PR #16, opened 2016-08-11
[Quicksort]Added shuffling of input

miczal:
This is safety measure against the worst case scenario - sorted input and O(n^2) complexity

### DIFF
--- a/sorts/quick_sort.py
+++ b/sorts/quick_sort.py
@@ -10,6 +10,12 @@ For manual testing run:
 python quick_sort.py
 """
 from __future__ import print_function
+from random import shuffle
+
+
+def sort(collection):
+    shuffle(collection)
+    return quick_sort(collection)
 
 
 def quick_sort(collection):
@@ -58,4 +64,4 @@ if __name__ == '__main__':
 
     user_input = input_function('Enter numbers separated by coma:\n')
     unsorted = [int(item) for item in user_input.split(',')]
-    print(quick_sort(unsorted))
+    print(sort(unsorted))


AnupKumarPanwar:
But Shuffle will add complexity in other cases

miczal:
I'm by no means mathematician, but shuffling of an input gives you a possibility to have average case complexity as often as theoretically possible in every case. It doesn't add to the overall average complexity - O(nlog n + n) is still O(nlog n). I've wrote this short script to let you know how important randomization of input is and how often people don't do it:

```
how_many = 100000
sorting_func = sort #this shuffles input
#sorting_func = quick_sort
tries = 10

accumulated_time = {'Sorted': 0,
                    'Reverse': 0,
                    'Random': 0,
                    'A lot of equal keys': 0,
                    'Same keys': 0}
for i in range(tries):
    data_sets = {'Sorted': [x for x in range(how_many)],
                 'Reverse': list(reversed([x for x in range(how_many)])),
                 'Random': [random.randint(0, how_many) for x in range(how_many)],
                 'A lot of equal keys': [x if x % 2 else 0 for x in range(how_many)],
                 'Same keys': [1] * how_many}
    for what, set in data_sets.items():
        start = time.clock()
        sorting_func(set)
        end = time.clock()
        accumulated_time[what] += (end - start)

print("Average time:")
for what, total_time in accumulated_time.items():
    print(what + ": " + str(total_time/tries))
```
Average time on my PC with shuffling of input:
```
Average time:
A lot of equal keys: 0.30944705735259076
Reverse: 0.5213236423849711
Random: 0.5092556249405173
Sorted: 0.5258773700452282
Same keys: 0.09532146027318696
```
In case of current implementation of `quick_sort` only in case of `Same keys` data and `Random` I didn't have a `RecursionError`
```
Average time:
Same keys: 0.016281707645539805
Random: 0.40859697925563115
```
Only after changing data sets length to **950 from 100000 (!!!)** I was able not to have an error.

See for yourself.

PR #19, opened 2016-08-16
Added Shell Sort

JakeBonek:
Added the Shell Sort algorithm and fixed the spelling in the other sorts.  I also added function documentation to heap_sort.py


JakeBonek:
I think that someone has been using tabs instead of spaces in heap_sort.py and that's why the travis ci build is failing.

H-Darji:
I'll fix it!

PR #64, opened 2017-01-20
Create randomnormalquicksort.py

prateekiiest:
This is program is implementing the quick sort algorithm for an array of numbers selected from a standard normal distribution and also the pivot chosen randomly between the starting element and the last element.
This program outputs the normal distribution graph of the array and also prints the number of comparisons.


dynamitechetan:
Please check the Travis CL build status

prateekiiest:
why is there an error showing no module called matplotlib.pyplot??

prateekiiest:
I did it in python 2.7. I am seeing the checks are being done with python 3+.
Is it the reason behind the error in build status??

PR #74, opened 2017-04-07
Script Output: The entire longest increasing subsequence instead of it's length

alaouimehdi1995:
The program now returns the entire longest increasing subsequence instead of returning only the length. Also: Code optimized and entirely explained in comments

### DIFF
--- a/dynamic_programming/longest_increasing_subsequence.py
+++ b/dynamic_programming/longest_increasing_subsequence.py
@@ -15,19 +15,24 @@ def longestSub(ARRAY): 			#This function is recursive
 		return ARRAY
 								#Else
 	PIVOT=ARRAY[0]
-	LONGEST_SUB=[]				#This array will contains the longest increasing sub array
-	for i in range(1,ARRAY_LENGTH):			#For each element from the array (except the pivot),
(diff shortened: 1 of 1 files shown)


dynamic_programming/longest_increasing_subsequence.py
```diff
@@ -1,12 +1,36 @@
...
-    return max(lis)
+'''
+Author  : Mehdi ALAOUI
+
+This is a pure Python implementation of Dynamic Programming solution to the longest increasing subsequence of a given sequence.
+
+The problem is  :
+Given an ARRAY, to find the longest and increasing sub ARRAY in that given ARRAY and return it.
+Example: [10, 22, 9, 33, 21, 50, 41, 60, 80] as input will return [10, 22, 33, 41, 60, 80] as output
+'''
```
ankit167: Please mention time and space complexity of the code

alaouimehdi1995: I just finished some optimizations on complexity, it was, on the worst case O(2^(n-1)), but now, I think that in the worst case, it's O(nlog(n)), if I'm wrong, please correct me.

PR #81, opened 2017-06-07
Added feature for negative number input

gupta2140:
Previously, the entry in tree was terminated by negative (which was the base condition for termination of entry loop). Now i've re-edited the code and have the user input any number of his choice. Added a feature that the user can terminate the entry in the tree at any point of time by just pressing 'N' or 'n' which is being checked at every input points. If a number is entered in place of 'N' or 'n', that number is entered into the node.

### DIFF
--- a/traverals/binary_tree_traversals.py
+++ b/traverals/binary_tree_traversals.py
@@ -13,37 +13,37 @@ class TreeNode:
 
 
 def build_tree():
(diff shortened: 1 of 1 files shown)


H-Darji:
@gupta2140 Before merging your request, Travis CI stated that all checks have passed. But after merge, it failed. Can you please check what is wrong?

PR #97, opened 2017-07-16
Added Stack implementation and some traditional Stack problems 

OmkarPathak:


0oo0:
Thanks 🎉.  We look forward to having more contributions from you.

PR #98, opened 2017-07-25
Update Breadth_First_Search.py

moun3imy:
Fixed the fact that all nodes except source are marked as visited twice


0oo0:
Thanks for the PR 🎉.

Issue #105, opened 2017-08-21
I'd like to contribute by adding the Trie data structure


JavonDavis:
I'm working on a Trie implementation and a simple example looking up a word in a dictionary. Thinking about using the Dictionary.txt file in the other folder.

dynamitechetan:
Cool. Please work on it.

JavonDavis:
Hey, @dynamitechetan I've sent in the Pull Request can you check it out?

PR #111, opened 2017-09-10
Trie implementation

JavonDavis:
Regarding #105 I've worked on a Trie implementation using a dict

### DIFF
--- a/data_structures/Trie/Trie.py
+++ b/data_structures/Trie/Trie.py
@@ -1,0 +1,21 @@
+'''
+A Trie/Prefix Tree is a kind of search tree used to provide quick lookup
+of words/patterns in a set of words. A basic Trie however has O(n^2) space complexity
+making it impractical in practice. It however provides O(max(search_string, length of longest word)) lookup
+time making it an optimal approach when space is not an issue.
+'''
+
+
+class TrieNode:
+    def __init__(self):
+        self.nodes = dict() # Mapping from char to TrieNode
+
+    def add_words(self, words: [str]):
+        for word in words:
+            self.add_word(word)
+
+    def add_word(self, word: str):
+        pass
+
+    def lookup_word(self, word: str) -> bool:
+        pass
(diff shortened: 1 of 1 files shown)


data_structures/Trie/Trie.py
```diff
@@ -0,0 +1,81 @@
+"""
+A Trie/Prefix Tree is a kind of search tree used to provide quick lookup
+of words/patterns in a set of words. A basic Trie however has O(n^2) space complexity
+making it impractical in practice. It however provides O(max(search_string, length of longest word)) lookup
+time making it an optimal approach when space is not an issue.
+
+This implementation assumes the character $ is not in any of the words. This character is used in the implementation
```
r0hit-gupta: Can you please find some way to achieve the desired result without setting a predefined character? An algorithm is supposed to consider all test cases in mind. Good job though!

data_structures/Trie/Trie.py
```diff
@@ -0,0 +1,81 @@
...
+        print(word, end=' ')
+
+    for key, value in node.nodes.items():
+        print_words(value, word + key)
+
+
+def test():
+    words = []
+    # Load words from text file into Trie
+    with open("../../other/Dictionary.txt", "r") as ins:
```
r0hit-gupta: Can you please switch this with a string sample? This would make it easier for the user to run and test the code. Thanks.

data_structures/Trie/Trie.py
JavonDavis: Thanks, no prob! I'll work on it

data_structures/Trie/Trie.py
JavonDavis: Is the sample I had in 9a52167 ok?

JavonDavis:
@sachinarora707 @r0hit-gupta I'm now using a boolean attribute instead of the $ and provided a list of strings instead of reading from the file

r0hit-gupta:
@JavonDavis great job. Please delete the empty` __init__.py` file and you are good to go.

PR #114, opened 2017-09-21
master

RiptideBo:
add bp neural network with 3 layers into Neural_Network File,


RiptideBo:
What does it mean by  'The Travis CI build failed'??

PR #115, opened 2017-09-22
add neuralnetwork_bp3.py

RiptideBo:
I add bp neural network with 3 layers into Neural Network File


RiptideBo:
I run the Travis CI test on the website, found some Error existing but not in my file. Thought maybe I could do some help to fix it

PR #116, opened 2017-09-22
master

RiptideBo:
I add convolution_neural_network.py

--Next 
I would improve this and add more comment or annotation about the detail
I would add a example file about the cnn model use based on MINST Data


RiptideBo:
Since I become a contributor in this project, does it mean that I could just clone this project to local. Then when I fix something, I could just push it??

RiptideBo:
Or I still have doing my job on my fork?  still have to stay sync with this project, pull request .etc.

Or I can make a new branch here?

Can anyone help with my confusion?

PR #122, opened 2017-09-28
Deleted similiar files

iharsh234:
closes #120.


sachinarora707:
@iharsh234 Can you check which one of the two copies of BFS (and DFS) has a better implementation? If both are same, then keep the one having proper comments and conventions.

PR #133, opened 2017-09-30
added counting sort

sgbasaraner:
hacktoberfest: I copied the comments from bubble_sort and left them as is (changed bubble_sort to counting_sort), and my counting sort implementation works with negative numbers too. if the comments will be a problem, notify me and I'd like to make the necessary changes.

### DIFF
--- a/sorts/counting_sort.py
+++ b/sorts/counting_sort.py
@@ -1,0 +1,57 @@
+"""
+This is pure python implementation of counting sort algorithm
+For doctests run following command:
+python -m doctest -v counting_sort.py
+or
+python3 -m doctest -v counting_sort.py
+For manual testing run:
+python counting_sort.py
+"""
+
+from __future__ import print_function
+
+
+def counting_sort(collection):
+    """Pure implementation of counting sort algorithm in Python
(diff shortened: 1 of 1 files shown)


andrealmeid:
Hello @sgbasaraner,
I read your code and I find some issues that you might be interest to notice. Two main features of counting sort is linear time complexity and stability. But, when you do this loop in lines 32-33
```python
for i in range(counting_arr_length):
    for j in range(length):
```
you are, in fact, running _j_ * _i_ times. This can easily achieves O(_n²_) time complexity, when _i_ >= _j_.

When you create the array, you do
```python
counting_arr_length = maximum + 1 + abs(minimum)
```
But we don't need all this positions, just `maximum + 1 - minimum` are enough to store all the values.

Finally, the stability. Stability in counting sort is an important feature because it's make counting sort useful to be used as radix sort subroutine. By the way you do, you lost the previous information about the original order of the elements.

I suggest check out some Wikipedia pages about sorting algorithms and _Introduction to Algorithms_ by Cormen, Leiserson, Rivest and Stein to find more about stability and time complexity. I made my own version, check out! (#165) Contact me if you have doubts or need more information :)

PR #172, opened 2017-10-19
path-1 forwarding ports with meterpreter API using namespaces with gui supported C# program. look at open ai.lnk file to see more.

SoftwareFanATIC:
PHP server side support. SQL  database querying. JAVA p2p connections along with python program for architecture support >> for linux not WINDOWS OR MACOSX


SoftwareFanATIC:
<XML? version="?
">

PR #180, opened 2017-10-20
fixed issue#171

mathildapurr:
algorithm now able to process string with characters other than brackets


sachinarora707:
Good piece of Code! 👍👍

PR #181, opened 2017-10-20
I request access to the files so I can contribute to the project!

DrBalar:


cclauss:
Why remove all travis tests??

Project Euler/Problem 16/sol1.py
```diff
@@ -0,0 +1,15 @@
...
+list_num = list(string_num)
+
+sum_of_num = 0
+
+print("2 ^",power,"=",num)
+
+for i in list_num:
+    sum_of_num += int(i)
+
+print("Sum of the digits are:",sum_of_num)
```
cclauss: https://docs.python.org/3/library/functions.html#sum

print("Sum of the digits are:", sum(int(i) for i in string_num))

ciphers/XOR_cipher.py
```diff
@@ -0,0 +1,239 @@
...
+
+		# precondition
+		assert (isinstance(key,int) and isinstance(content,str))
+
+		# testing for default arguments
+		if (key == 0):
+			if (self.__key == 0):
+				key = 1
+			else:
+				key = self.__key
```
cclauss: key = key or self.__key or 1

ciphers/XOR_cipher.py
```diff
@@ -0,0 +1,239 @@
...
+
+		# make sure key can be any size
+		while (key > 255):
+			key -= 255
+
+		# This will be returned
+		ans = []
+		resultNumber = 0
```
cclauss: resultNumber is assigned to but not used.

ciphers/XOR_cipher.py
```diff
@@ -0,0 +1,239 @@
...
+
+		for ch in content:
+			ans.append(chr(ord(ch) ^ key))
+
+		return ans
+
+
+	def encrypt_string(self,content, key = 0):
```
cclauss: return ''.join(self.encrypt(content, key))

ciphers/XOR_cipher.py
```diff
@@ -0,0 +1,239 @@
...
+			successful otherwise false
+			if key not passed the method uses the key by the constructor.
+			otherwise key = 1
+		"""
+
+		#precondition
+		assert (isinstance(file,str) and isinstance(key,int))
+
+		try:
+			fin = open(file,"r")
```
cclauss: use __with open()__ to avoid the need to close() manually and also to close() even if an exception is thrown.

ciphers/XOR_cipher.py
```diff
@@ -0,0 +1,239 @@
...
+			fout = open("encrypt.out","w+")
+
+			# actual encrypt-process
+			for line in fin:
+				fout.write(self.encrypt_string(line,key))
+
+			fin.close()
+			fout.close()
+
+		except:
```
cclauss: See PEP8 for reasons to avoid bare exceptions.

ciphers/XOR_cipher.py
christianbender: Thanks for your proposals

ciphers/XOR_cipher.py
christianbender: @cclauss 

Thanks for your proposals.I'm glad If I get some feedback.

greetings from Saarland

Christian

PR #225, opened 2017-11-29
Project Euler - Easier Solution To Problem 7

Richienb:
I've Made An Easier Solution To Problem 7
Upon Executing The File, You Will Be Asked `Enter The N'th Prime Number You Want To Get` And After You Specify It You Will Get The N'th Prime Number


Project Euler/Problem 07/sol2.py
```diff
@@ -0,0 +1,16 @@
+# By listing the first six prime numbers: 2, 3, 5, 7, 11, and 13, we can see that the 6th prime is 13. What is the Nth prime number?
+def isprime(number):
```
Richienb: Check If A Number Is A Prime Number

PR #232, opened 2017-12-28
Revert "Update bubble_sort.py"

sahilmalhotra24:


harshildarji:
@sahilmalhotra24 This one is conflicting with existing one!

PR #249, opened 2018-02-11
LU decomposition 

MindTraper:
Organised the Arithmetic analysis in the same folder and add a new function for the LU decomposition


dynamitechetan:
hi @MindTraper  please resolve the conflicts.

PR #286, opened 2018-04-14
Create Fibonacci.py

yesIamHasi:
This script contains python implementation of Fibonacci sequence in two forms.
1) Formula Implementation.
2) Iterative Implementation.


19ANDY:
#286 the code isn't correct

PR #287, opened 2018-04-14
some changes made with sqrt method 

19ANDY:


fibonacci.py
```diff
@@ -0,0 +1,36 @@
...
+
+        def iteration(self, num):
+            '''Return nth fibonacci number using iterative formula.'''
+            if num == 0: return 0
+            if num == 1: return 1
+            a = 1
+            b = 1
+            for i in range(num):
+                a,b = b, a+b
+                return b-a
```
christianbender: This line is a bug. You must indent properly the ```return```.
```python
for i in range(num):
    a,b = b, a+b
return b-a
```

fibonacci.py
```diff
@@ -0,0 +1,36 @@
...
+
+        def fibrange(self, start, stop, fast=True):
```
christianbender: Use list compreshion:  
```python
if not fast:
    fib_range = [self.iteration(i) for i in range(start,stop)]
    return fib_range
else:
    fib_range = [self.formula(i) for i in range(start,stop)]
    return fib_range
```

PR #295, opened 2018-04-16
Modernize Python 2 code to work for Python 3

### LINKED ISSUE #289: ProjectEuler -- Problem 1 -- solv2.py -- Error
For the Input ```1000``` I get ```233366.4```. The correct answer should be ```233168``` 
See [file](https://github.com/TheAlgorithms/Python/blob/master/Project%20Euler/Problem%2001/sol2.py)

bp274:
Fixes #289 which occurred because of difference in operation of `/` in Python 2 and Python 3 by replacing `/` with `//`.


christianbender:
@bp274  Thanks!!!

PR #300, opened 2018-05-19
Tabu Search

irokafetzaki:
I added Tabu Search for a Travelling Salesman Problem in the searches file. Analytical comments are available for each method. I also added tests for the methods of the Algorithm. Please let me know if you would like to accept my commit.

Thank you in advance.

### DIFF
--- a/searches/tabuTestData.txt
+++ b/searches/tabuTestData.txt
@@ -1,0 +1,10 @@
+a b 20
+a c 18
+a d 22
+a e 26
+b c 10
+b d 11
+b e 12
+c d 23
+c e 24
+d e 40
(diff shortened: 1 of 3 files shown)


harshildarji:
@irokafetzaki I've merged your commit, but just to let you know that in [tabu_search.py](https://github.com/irokafetzaki/Python/blob/35110b6e44cbe5b83878cd5d64a85466723a153a/searches/tabu_search.py), lines from [255-313](https://github.com/irokafetzaki/Python/blob/35110b6e44cbe5b83878cd5d64a85466723a153a/searches/tabu_search.py#L255-L313) are empty.

PR #302, opened 2018-05-22
Update merge_sort_fastest.py

yesIamHasi:
Thank you very much for pointing that out. I am really sorry I accidentally uploaded the wrong file last time. The changes you made didn't worked for some inputs. This one works fine.

### DIFF
--- a/sorts/merge_sort_fastest.py
+++ b/sorts/merge_sort_fastest.py
@@ -4,16 +4,51 @@ Takes an average of 0.6 microseconds to sort a list of length 1000 items.
 Best Case Scenario : O(n)
 Worst Case Scenario : O(n)
 '''
+# These modules are required for testing only.
+import random, copy
+import matplotlib.pyplot as plt
+
 def merge_sort(LIST):
+    '''
+    Returns sorted list.
+    '''
+    # Divide list into two parts
     start = []
     end = []
-    while len(LIST) > 1:
-        a = min(LIST)
-        b = max(LIST)
+    a = min(LIST)
+    b = max(LIST)
+    
+    while len(LIST)>1:
         start.append(a)
         end.append(b)
-        LIST.remove(a)
-        LIST.remove(b)
-    if LIST: start.append(LIST[0])
+        try:
+            LIST.remove(a)
+            LIST.remove(b)
+            a = min(LIST)
+            b = max(LIST)
+        except ValueError:
+            continue
     end.reverse()
-    return (start + end)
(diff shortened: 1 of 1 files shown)


harshildarji:
Hello @yesIamHasi, It would be really helpful if you could you please let use know about the inputs for which current code is not working as expected?

sorts/merge_sort_fastest.py
```diff
@@ -4,16 +4,51 @@
...
+    plt.show()
+    plt.plot(SORTED_LIST)
+    plt.xlabel('Merge Sorted Data')
+    plt.show()
+
+if __name__ == '__main__':
+    # Basic Test
+    LIST = [0,5,2,4,3,9,1,7,6,8]
+    print ('Unordered list {}'.format(LIST))
+    print ('Merge sorted list {}'.format(merge_sort(LIST)))
```
gabru-md: can we store the `merge_sort(LIST)` into some variable and then use it at multiple places?

sorts/merge_sort_fastest.py
```diff
@@ -4,16 +4,51 @@
 Best Case Scenario : O(n)
 Worst Case Scenario : O(n)
 '''
+# These modules are required for testing only.
+import random, copy
+import matplotlib.pyplot as plt
```
AnshulMalik: Umm, it's nice to see the output as graph, it really helps in visualising the data. But I think it's not required here. We can just test for the output.

sorts/merge_sort_fastest.py
```diff
@@ -4,16 +4,51 @@
...
+            continue
     end.reverse()
-    return (start + end)
+    return start + end
+
+def visualize(LIST, SORTED_LIST):
+    '''
+    Plots unsorted and sorted data.
+    '''
+    plt.plot(LIST)
```
christianbender: Changed the code like this:
```python
figure1 = plt.figure(1)
plt.plot(LIST)
plt.xlabel('Random Data')
figure1.show()
figure2 = plt.figure(2)
plt.plot(SORTED_LIST)
plt.xlabel('Merge Sorted Data')
figure2.show()

raw_input() # keeps the figures alive
```

With the code above it show up two figures parallel at the same time.

sorts/merge_sort_fastest.py
christianbender: @AnshulMalik 
good point! The function ```min``` and ```max``` have O(n). And this runtime is dominant.

PR #307, opened 2018-06-01
Updated caesar_cipher.py

PyDevthon:


christianbender:
@PyDevthon You must resolve your conflicts otherwise I can not merging

arjunmann73:
As this entire code is for educational purposes, it would be more user-friendly if comments were added to give a brief about what is happening in the program. 

I, being a student, have a few questions about the program:
1) It would make more sense if the function "encdec" was placed at the top of the code with the other functions but that wasn't the case in the code, is there a reason this was done or was it just random placement?
2) 
if __name__ == '__main__':
    import doctest
    doctest.testmod()

Could you please explain the above code snippet? Completely clueless about this part of the code.

3) Last but the most important, might sound like stupid question but please do help. If I am not wrong, the latest uploaded file contains 2 different source codes for the same program doing the same overall job, why is this the case? Why are their 2 different source codes given?

Thanks a lot.
Would love a reply

PR #308, opened 2018-06-01
Added data structure binary indexed tree. Runs fast update and fast q…

kaynaat007:
…uery operations on the array.

### DIFF
--- a/data_structures/BinaryIndexedTree/bit.py
+++ b/data_structures/BinaryIndexedTree/bit.py
@@ -1,0 +1,110 @@
+from __future__ import print_function
+
+__author__ = 'Nikhil Kumar'
+
+
+class BIT(object):
+    '''
+    Binary indexed tree is a simple data structure that is helpful in situations when we have lots
+    of queries which want to apply certain operation over a range of indexes in array. For example,
+    lets assume we have an array of length N. There can be two types of queries:
+        1. Change the value stored at an index i. (This is called a point update operation)
+        2. Find the sum of a prefix of length k. (This is called a range sum query)
+    In a traditional array, operation 1 would take O(1) time, while operation 2 would take O(N)
+    time in worst case. And if we have 1 million queries of type 2, we will end up wasting lot
+    of time.
+    To come out of such cases, we use binary indexed trees(BlsIT). ( segment trees is other solution )
+    I may run of space if i try to explain BIT here, but consider going to following links:
+
+    https://www.hackerearth.com/practice/data-structures/advanced-data-structures/fenwick-binary-indexed-trees/tutorial/
+    https://cs.stackexchange.com/questions/10538/bit-what-is-the-intuition-behind-a-binary-indexed-tree-and-how-was-it-thought-a
+
+    In general, it's a simple array where we store partial sums. Every index which is a power of 2,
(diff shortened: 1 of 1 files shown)


data_structures/BinaryIndexedTree/bit.py
```diff
@@ -0,0 +1,110 @@
...
+    The only reason we use BIT over segment tree becasue it takes less time to code it and
+    it takes less space in memory.
+    '''
+
+    def __init__(self, arr):
+        self.input = arr
+        self.length = len(self.input)
+        self.bit = [0] * (self.length + 1)  # we will use indexes from 1 ..N
+
+    def construct(self):
```
AnshulMalik: We probably don't need this method, it can be done in the constructor itself.

data_structures/BinaryIndexedTree/bit.py
```diff
@@ -0,0 +1,110 @@
...
+        for i in range(1, self.length + 1):
+            self.update(i, self.input[i - 1])
+
+    def update(self, index, val):
+        """
+        given and index and a value to update, update the element at the given index by value.
+        """
+
+        while index <= self.length:
+            self.bit[index] += val
```
AnshulMalik: Umm, the method name is update but, instead of update, it's adding the value to the index.
It'd be nice if we calculate the difference of previous and new value and add that to the index, then it'd match with the semantics.
We can have another method add for this purpose.

AnshulMalik:
And there is already a fenwick tree implementation: https://github.com/TheAlgorithms/Python/blob/master/data_structures/Binary%20Tree/FenwickTree.py

Do you think we should update that instead?

data_structures/BinaryIndexedTree/bit.py
kaynaat007: umm, didn't get you. Regarding the method name, all this method is doing is adding a value  `val` at index `index` as a result of which original value is updated by `val` amount. Hence the name of the method.   And since it's a fenwick tree, it's propagated till the root of the tree. All standard implementations name this method as  **update** only.

data_structures/BinaryIndexedTree/bit.py
kaynaat007: sure will move into init.  Also i will update existing fenwick tree in BinaryTree folder. I didn't see that before. I thought Binary Indexed Tree would have it's separate folder.  it should ? shouldn't it  ?.

data_structures/BinaryIndexedTree/bit.py
AnshulMalik: Yes, people have written this method in such a way. But the semantic attached with update is, to replace the existing value, addition and subtraction are different operations than update.
I don't know why people have implemented addition in update method. It is confusing when we read it first time.

data_structures/BinaryIndexedTree/bit.py
AnshulMalik: Yeah, BIT can have it's own directory, we need other's opinion here. @DravitLochan @harshildarji

data_structures/BinaryIndexedTree/bit.py
kaynaat007: let's take suggestion from other people too if unsure.  @DravitLochan, @harshildarji

PR #314, opened 2018-06-05
create readme for sorts

AnshulMalik:
The readme feels a bit all in one right now, we should split the docs up at different places, it'll be easier maintain as well as easy for people to find what they are looking for.

This is first step in the process, we probably need to have a dedicated readme for each algorithm, where we can put detailed stuff. (Lemme know your thoughts on this one)


AnshulMalik:
1. Yes, we should move them, but let's keep it for another PR.
2. I think folder names are self descriptive, but we can provide some info about a folder as.

poyea:
I'm closing this as this is resolved by other PRs. Please read our Contribution Guidelines for more details. Thank you for your pull request!🤩

PR #319, opened 2018-06-14
Included circularlinkedlist.py

yatesal:


data_structures/LinkedList/circularlinkedlist.py
```diff
@@ -0,0 +1,69 @@
+class Node:
+    def __init__(self,data):
+        self.data = data
+        self.next = None
+
```
christianbender: Please add a ```main``` function with a test case.

PR #326, opened 2018-07-08
subset_generation

Ambuj81:
generate all possible subset of size n of a given array of size r


dynamic_programming/subset_generation
```diff
@@ -0,0 +1,62 @@
+# python program to print all subset combination of n element in given set of r element .
+
```
christianbender: Your file must have the extension ```.py```

dynamic_programming/subset_generation
```diff
@@ -0,0 +1,62 @@
...
+#Current combination is ready to be printed,
+# print it
+ if(index== r):
+  for j in range(r):
+   print(data[j],end=" ")
+  print(" ")
+  return
+
+#  When no more elements are there to put in data[]
+
```
christianbender: Remove useless empty lines. You can write this:
```python

# comment
def foo():
   pass

# comment
def boo():
  pass
```

Above each comment you can write a empty line. For example:
```python
def combinationUtil(arr,n,r,index,data,i):
      
#Current combination is ready to be printed,
# print it
 if(index== r):
  for j in range(r):
   print(data[j],end=" ")
  print(" ")
  return

#  When no more elements are there to put in data[]
 if(i>=n):
  return

#current is included, put next at next
# location 
 data[index]=arr[i]
 combinationUtil(arr,n,r,index+1,data,i+1)
```

dynamic_programming/subset_generation
```diff
@@ -0,0 +1,62 @@
...
+ combinationUtil(arr,n,r,index,data,i+1)
+
+ 
+ # The main function that prints all combinations
+ #of size r in arr[] of size n. This function 
+ #mainly uses combinationUtil()
+ 
+ 
+ 
+def printcombination(arr,n,r):
```
christianbender: The parameter ```n``` is useless because you can use the function ```len(...)``` within this function.

dynamic_programming/subset_generation
```diff
@@ -0,0 +1,62 @@
...
+ 
+#Print all combination using temprary
+#array 'data[]'
+ combinationUtil(arr,n,r,0,data,0)
+
+ 
+# Driver function to check for above function
+arr = [10,20,30,40,50]
+
+r=3
```
christianbender: This variable ```r``` is useless because you can put the 3 right away in the function ```printcombination```

dynamic_programming/subset_generation
```diff
@@ -0,0 +1,62 @@
...
+ #of size r in arr[] of size n. This function 
+ #mainly uses combinationUtil()
+ 
+ 
+ 
+def printcombination(arr,n,r):
+
+# A temporary array to store all combination
+# one by one
+ data = list(range(r))
```
christianbender: Instead of this line write ```data = [0] * n```. Is more common.

dynamic_programming/subset_generation
```diff
@@ -0,0 +1,62 @@
...
+# print it
+ if(index== r):
+  for j in range(r):
+   print(data[j],end=" ")
+  print(" ")
+  return
+
+#  When no more elements are there to put in data[]
+
+ if(i>=n):
```
christianbender: Please put in spaces between operators like ```>= = + -``` etc.

dynamic_programming/subset_generation.py
```diff
@@ -15,48 +12,28 @@ def combinationUtil(arr,n,r,index,data,i):
...
 #  When no more elements are there to put in data[]
-
- if(i>=n):
+ if(i >= n):
   return
-
 #current is included, put next at next
 # location 
-  
  data[index]=arr[i]
```
christianbender: Please add spaces between all operators.

dynamic_programming/subset_generation.py
```diff
@@ -15,48 +12,28 @@ def combinationUtil(arr,n,r,index,data,i):
...
- 
- 
- 
 def printcombination(arr,n,r):
-
 # A temporary array to store all combination
 # one by one
- data = list(range(r))
- 
+ data = [0]*r
```
christianbender: Same things as above.

dynamic_programming/subset_generation.py
```diff
@@ -15,48 +12,28 @@ def combinationUtil(arr,n,r,index,data,i):
...
 #Print all combination using temprary
 #array 'data[]'
  combinationUtil(arr,n,r,0,data,0)
-
- 
 # Driver function to check for above function
 arr = [10,20,30,40,50]
-
-r=3
-n= len(arr)
```
christianbender: Same thing as above

Ambuj81:
pls review my changes

PR #331, opened 2018-07-22
Added code for memory function implementation

bharath-123:
Added a function which implements knapsack using memory functions. Also I updated the test cases to include test cases for the memory function implementation.

### DIFF
--- a/dynamic_programming/knapsack.py
+++ b/dynamic_programming/knapsack.py
@@ -28,7 +28,7 @@ def knapsack(W, wt, val, n):
 
     return dp[n][w]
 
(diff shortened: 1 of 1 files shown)


dynamic_programming/knapsack.py
```diff
@@ -12,3 +27,16 @@ def knapsack(W, wt, val, n):
                 dp[i][w] = dp[i-1][w]
 
     return dp[n][w]
+
+if name == '__main__':
```
christianbender: Write ```__name__```

bharath-123:
Resolved conflicts!

PR #332, opened 2018-07-24
multithreading and code explanation added

sliangooi:
I added multi threading on server and client program. Multiple clients might connect to a single server at the same time. I used threads to handle the server's multiple clients simultaneously to establish the connection. Besides I added some explanations of code in order to help contributors to understand the code.

### DIFF
--- a/simple-client-server/client.py
+++ b/simple-client-server/client.py
@@ -2,13 +2,20 @@
 
 import socket
 
+
 HOST, PORT = '127.0.0.1', 1400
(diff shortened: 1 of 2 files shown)


harshildarji:
Hello @sliangooi,
I agree to the @AnshulMalik's point. It would be great if you can make a different PR with the **multithreaded_http_server** as a different module (_or say script_) instead of changing the current simple one.
Thank you.

PR #335, opened 2018-08-13
Adding Simpler Solutions Project Euler

satanb4:
A simpler code has been added to project euler #16


AnshulMalik:
@Sayan97 Awesome, it'd really help if you add some comments telling what it does or simplify it a bit, so that it's easy to understand. :)

Sayan97:
Uses list comprehension to decrease the time complexity and speeds up the overall program.

Sayan97:
Variable declaration is not taking place. Works as iterator like in generators.

cclauss:
This solution seems incompatible with the other solution.
https://github.com/TheAlgorithms/Python/blob/26b2e60792f19a66dbd7e38f91a5b8febfdbc740/Project%20Euler/Problem%2016/sol1.py
 Could we have _prompts_ in the __input()__ calls so the user knows what is expected?  The current version would hang out automated testing.

PR #341, opened 2018-09-13
Minor update in the file Arrays.py

pandyamarut:
having output for the following code will help people get code early.


pandyamarut:
Successfully Merged the changes .
Thank you.

PR #343, opened 2018-09-13
Minor update in the file Arrays.py

pandyamarut:
having output for the following code will help people get code early.


AnshulMalik:
Not sure if It does

PR #350, opened 2018-09-25
Maths algo committed

FarhanKasmani:
## Basic Maths Algorithms
1. Euler Phi
2. Number of Divisors
3. Prime Factors
4. Segmented Sieve
5. Sieve of Eratosthenes
6. Sum of Divisors

### DIFF
--- a/Maths/EulersFunction.py
+++ b/Maths/EulersFunction.py
@@ -1,0 +1,9 @@
+def eulerPhi(n):
+    l = primeFactors(n)
+    l = set(l)
+    s = n
+    for x in l:
+        s *= (x - 1)/x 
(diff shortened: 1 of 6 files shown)


Maths/EulersFunction.py
```diff
@@ -0,0 +1,9 @@
+def eulerPhi(n):
+    l = primeFactors(n)
+    l = set(l)
+    s = n
+    for x in l:
+        s *= (x - 1)/x 
+    return s  
+
+print(eulerPhi(100))
```
ahmadrezash: You can define a "main" function whit a while(true) and use this function in this main function and get n from user.

Issue #315, opened 2018-06-05
Use consistent directories and file naming


AnshulMalik:
Currently, the directories and files don't follow a convention, some are `snake_case`, some `PascalCase`. We should stick to a convention.

elpaxoudis:
Greetings, I was going to start the same idea in a personal project but then I stumbled upon this one. So, I was thinking maybe I could contribute to this one instead.

I have never contributed to an open source project before. However, I could easily try this issue as my first. What do you think?

By the way I think that an all lower case convention is a suitable one if we consider them as [modules/packages](https://www.python.org/dev/peps/pep-0008/#package-and-module-names).

AnshulMalik:
Hey @elpaxoudis, Nice to have you here, you can definitely start with this one. And we can go ahead with snake case, as this is the standard.

elpaxoudis:
Nice! Thanks @AnshulMalik for letting me work on this. Just one final question. Should I change the source .py files into `snake_case` too?

AnshulMalik:
Yep @elpaxoudis

cclauss:
Fixed in #993 and to be enforced in our Travis CI testing with #1005

PR #362, opened 2018-09-30
Dirs files naming convention fix

elpaxoudis:
I hope I got everything right.
**Solves:** #315


AnshulMalik:
This is great work!

I found two files which are left somehow :P

arithmetic_analysis/Bisection.py
arithmetic_analysis/Intersection.py

elpaxoudis:
@AnshulMalik I think I got it now. Something strange was happening, the files in the directory were lowercase and when I changed them into uppercase again there was no change in git. Hence, I changed them to something else and then back to lowercase and I guess it worked, right?

PR #377, opened 2018-10-01
An updated get_initial_centroids function

prashant564:


AnshulMalik:
The earlier version is easier to understand @prashant564

prashant564:
Yes it was but the vectorized implementation is better... The lesser the
code the better. I would request u accept my pull request

AnshulMalik:
@prashant564 it's not always true `that lesser the code the better`. Our motive here is not to come up with the smallest and fastest code ever written, but to write algorithms in such a way that they perform well and are easier to understand for people. 
Hope you understand my point.

prashant564:
Yeah,  you are right.

PR #379, opened 2018-10-01
Add sorting unit tests

camilne:
This PR adds sorting unit tests under sorts/test. To run unit tests, run
> python -m unittest
It also helps to standardize some parts of sorting such as method signature and driver code for each file.


cclauss:
Seven file conflicts.

PR #404, opened 2018-10-03
Created a factorial algorith

RichaldoElias:


RichaldoElias:
Pentagonal Numbers

Maths/MagicSquareSeries.py
```diff
@@ -0,0 +1,2 @@
+def magicSquareSeries(n):
+    return n(n*n+1)/2
```
ashwek: this would cause _TypeError_ as it would interpret n as a function __n(__ _n*n+1_ __)__ _/2_

i think `n*(n*n+1)/2` is what you meant

PR #406, opened 2018-10-03
added Look and Say Series

RichaldoElias:


AnupKumarPanwar:
You've only defined the functions. Please write complete code with function calls.

PR #405, opened 2018-10-03
print() is a function in Python 3

cclauss:
@s-sanyal 

@TheAlgorithms Can we _please_ get Travis CI turned on so that these issues will be automatically detected before PRs are reviewed?

PR #407, opened 2018-10-03
Made the code Python 3 compatible

s-sanyal:


cclauss:
Not quite right...

$ __python2 -c "for x in 'abc': print (x),"__
```
a b c
```
$ __python3 -c "for x in 'abc': print (x),"__
```
a
b
c
```

See #405

cclauss:
Now it works great in Python 3 but it is now a syntax error in Python 2.  You need to add __from \_\_future\_\_ import print_function__ at the top of the file before any other imports.

harshildarji:
@cclauss @s-sanyal It still gives syntax error while executing with Python 2!

data_structures/Stacks/Stock-Span-Problem.py
```diff
@@ -37,7 +38,7 @@ def calculateSpan(price, S):
 # A utility function to print elements of array 
 def printArray(arr, n): 
     for i in range(0,n): 
-        print arr[i], 
+        print (arr[i],end =" ")
```
cclauss: Please delete the space...  __print (arr[i],end =" ")__ --> __print(arr[i], end=" ")__

__print()__ is now a function like all other functions so it should be formatted in the same way.

cclauss:
@ harshildarji Do you have `from __future__ import print_function` at the top of your file above all other imports?  What is the full text of the error that you are seeing?

harshildarji:
@cclauss The problem was in the explanation part at the very beginning of the script. But now it's solved, so I'm merging this PR now.

PR #418, opened 2018-10-04
Largets area of rectangle in histogram using stacks

ojasbansal:
Added a solution of a problem using stack as data structures.Problem finds the largest area of rectangle in histograms.


AnupKumarPanwar:
Aren't these 2 files same?

PR #425, opened 2018-10-05
Create sol5.py

alpylmz:
added a solve for the problem

### DIFF
--- a/Project Euler/Problem 01/sol5.py
+++ b/Project Euler/Problem 01/sol5.py
@@ -1,0 +1,8 @@
+a=3
+result=0
+while a=<1000:
+  if(a%3==0 and a%5==0):
+    result+=a
+  elif(a%15==0):
+    result-=a
+print(result)


shivg7706:
@alpylmz follow PEP8

cclauss:
@alpylmz Line 3 is a Python syntax error.  Could you please fix that and resubmit?

mrWiecek:
Shouldn't this solution be in project euler folder with all other solutions?

cclauss:
Does this solution actually solve the problem?

mrWiecek:
No it doesn't. With the syntax error you mentioned fixed it still creates infinite loop and it doesn't follow pep8. Now I see.

cclauss:
And will it solve the problem _even if_ the syntax error is fixed?

mrWiecek:
No it won't.

cclauss:
Should we delete this file?

mrWiecek:
Yes.

shuklalok:
Yes, delete this file. File path, logic and standards - all incorrect.

PR #426, opened 2018-10-05
Create sol4.py

alpylmz:
added a new solution to the problem


cclauss:
Change the name to __project_euler/problem_07/sol4.py__

PR #431, opened 2018-10-06
Remove pyc file

FannyVieira:
"By default, when executing Python code from files, the Python interpreter will automatically write a bytecode version of that file to disk, e.g. module.pyc.

These .pyc files should not be checked into your source code repositories.

Theoretically, this behavior is on by default, for performance reasons. Without these bytecode files present, Python would re-generate the bytecode every time the file is loaded. "
check more in [python documentation](https://docs.python-guide.org/writing/gotchas/#bytecode-pyc-files-everywhere)


piyush-kgp:
hey shouldn't this be taken care by gitignore (and it is indeed; in line 3)

PR #441, opened 2018-10-08
Naive Bayes Algorithm 

Himani2000:
the file is of jupyter notebook extension i.e ipynb . the naive bayes algorithm is implemented in the code


harshildarji:
Hey @Himani2000,
I must say you did really nice work here, but It would be really good if you can comment your code!
For example, you can explain what following line is doing:
```python
df.sl.apply(abc, args=(5,))
```
Thank you.

PR #447, opened 2018-10-11
Improve and Refactor the fibonnaciSeries.py (Recursion)

rafagarciac:
Wrap in the **__main__** structure and other improvements


rafagarciac:
@purveshmakode24  Take a look! :smile:

Maths/fibonacciSeries.py
```diff
@@ -1,16 +1,18 @@
...
 
-if limit <= 0:
-   print("Please enter a positive integer: ")
-else:
-   print(f"The first {limit} terms of the fibonacci series are as follows")
-   for i in range(limit):
-       print(recur_fibo(i))
+def main():
+    limit = int(input("How many terms to include in fibonacci series: "))
+    if isPositiveInteger:
```
harshildarji: I think you should write:
```python
if isPositiveInteger(limit):
```
Otherwise, It will not show the error message when provided a negative number.
Following is the output of your current script when you provide a negative number:
```python
How many terms to include in fibonacci series:  -7
The first -7 terms of the fibonacci series are as follows:
[]
```
Instead it should display something like:
```python
How many terms to include in fibonacci series:  -7
Please enter a positive integer.
```

PR #453, opened 2018-10-13
Create fibonacci2.py

Cabbaryus:


Maths/fibonacci2.py
```diff
@@ -0,0 +1,26 @@
+# Fibonacci Sequence Using Recursion
+
+
+def fibonacci(limit):
+    first_number = 1
```
harshildarji: Make it start from **0**.
[Why does the Fibonacci Series start with 0, 1?](https://math.stackexchange.com/questions/158699/why-does-the-fibonacci-series-start-with-0-1)

Maths/fibonacci2.py
```diff
@@ -0,0 +1,26 @@
...
+    second_number = 1
+
+    fibonaccci = [first_number , second_number ]
+
+    for i in range(limit):
+        first_number , second_number = second_number , first_number + second_number
+
+        fibonaccci.append(second_number)
+    print(fibonaccci)
```
harshildarji: Look at the output of your script:
```python
How many terms to include in fibonacci series: 5
The first 5 terms of the fibonacci series are as follows
[0, 1, 1, 2, 3, 5, 8]
```
Now, The length of the printed array should be **5**.

Maths/fibonacci2.py
```diff
@@ -0,0 +1,26 @@
+# Fibonacci Sequence Using Recursion
+
+
+def fibonacci(limit):
+    first_number = 1
+    second_number = 1
+
+    fibonaccci = [first_number , second_number ]
```
shivg7706: remove extra space

PR #471, opened 2018-10-17
Add problem 67 solution

Hyftar:
This adds the solution to the problem 67, which is exactly the same as the one from problem 18, but with a different triangle.


zhqu1148980644:
why dont't you compare a[i][j] and a[i-1][j+1]
```
for i in xrange(1, len(a)):
    for j in xrange(len(a[i])):
        if j != len(a[i - 1]):
            number2 = a[i - 1][j]
        else:
            number2 = 0
            
        if j+1 < (len(a[i])-1):
            number3 = a[i - 1][j + 1]
        else:
            number3 =0
            
        if j > 0:
            number1 = a[i - 1][j - 1]
        else:
            number1 = 0
        a[i][j] += max(number1, number2, number3)
```

Hyftar:
There's no need, you can only go diagonally two ways. This is clearer with the right formatting for the triangle.

zhqu1148980644:
But actually this two way could result to different answer.you can have a try.

Hyftar:
This algorithm is made to solve the problem at https://projecteuler.net/problem=67 and it provides a correct answer. That's why I cannot nor have to check for the upper right number.

zhqu1148980644:
you can try this two ways with about 7/8 rows of the raw txt file.You will get two different results, then you can see which is right by counting the maximum route sum visually.

zhqu1148980644:
what do you mean the right answer, that example's  maximum number 23? it's just coincidence.

zhqu1148980644:
sorry. I make a mistake. a number could only have two below and adjacent numbers

PR #481, opened 2018-10-18
smaller code

shawabhishek:


searches/binary_search.py
```diff
@@ -1,161 +1,14 @@
...
+        x = lower + (upper - lower) // 2
+        val = array[x]
+        if target == val:
+            return x
+        elif target > val:
+            if lower == x:   # these two are the actual lines
+                break        # you're looking for
+            lower = x
+        elif target < val:
+            upper = x
```
shivg7706: @shawabhishek Remeber a good code is that which have well-formatted comments, unit test, and must be understood by anyone may be non-programmer. I guess you get the point what I want to say.
Thanks

shawabhishek: I understand. thanks for guiding me.

PR #486, opened 2018-10-19
from IPython import get_ipython

cclauss:
@harshildarji  Help linters (and humans) understand where __get_ipython()__ comes from.

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./machine_learning/logistic_regression.py:19:1: F821 undefined name 'get_ipython'
get_ipython().run_line_magic('matplotlib', 'inline')
^
1    F821 undefined name 'get_ipython'
1
```

### DIFF
--- a/machine_learning/logistic_regression.py
+++ b/machine_learning/logistic_regression.py
@@ -16,6 +16,10 @@
 #importing all the required libraries
 import numpy as np
 import matplotlib.pyplot as plt
+try:
+     get_ipython
+except NameError:
+     from IPython import get_ipython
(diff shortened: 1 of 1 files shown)


harshildarji:
@cclauss To make `from IPython import get_ipython` work, we will have to run our script with **ipython** shell. Ordinary **python** shell will not work. So I think it is safe to comment out that line. User can remove the comment when they try to run with **ipython** or maybe in the **notebook**.

cclauss:
Alternative would be to try / catch ImportError but I think we are good as is...  Thanks for your guidance on getting this done.

PR #488, opened 2018-10-19
Clean up

RobbyB97:
Removed some content clearly not meant to be there, spell check, cleaned formatting, cited missing sources


RobbyB97:
You might want to add this :P I made sure every algorithm was cited and formatted them all nicely and there was some floating text and repeated things and text that was obviously not supposed to be there

PR #511, opened 2018-10-24
Create ExponentialSearch.py

niroshajayasundara:


harshildarji:
Direct copy from [Exponential Search - GeeksforGeeks](https://www.geeksforgeeks.org/exponential-search/)!

PR #527, opened 2018-10-26
Update PrimeCheck.py

rohitbabugaddeti:
I don't understand why range has got third argument since we can accomplish this functionality with just two arguments (we're taking the next value for range with step size 2 and however decrementing it later [in line 5]). So I think, changing it won't change the functionality of the algorithm. Hence I removed 4,5 lines from the code and modified the range iterator to give numbers from 2 to half of the number with step size 1

### DIFF
--- a/maths/PrimeCheck.py
+++ b/maths/PrimeCheck.py
@@ -1,8 +1,6 @@
 def primeCheck(number):
     prime = True
-    for i in range(2, int(number**(0.5)+1), 2):
-        if i != 2:
-            i = i - 1
+    for i in range(2, int(number**(0.5)+1)):
         if number % i == 0:
             prime = False
             break


Hyftar:
I think you've changed the logic here. The logic behind the original for loop it to iterate only on odd numbers and the number 2, while yours iterates on every single number instead. Of course, both algorithms give the same result, but your proposition is slightly less efficient.

rohitbabugaddeti:
> I think you've changed the logic here. The logic behind the original for loop is to iterate only on odd numbers and the number 2, while yours iterates on every single number instead. Of course, both algorithms give the same result, but your proposition is slightly less efficient.

Yeah , I changed it because I thought that the change improves the  code . But your answer looks promising :). Thanks for the comment.

PR #537, opened 2018-10-27
Change in caesar_cipher.py .......key not used correctly

lx1922:
I would like to propose the change I have made in caesar_cipher.py
The key was not being added or removed correctly..... so the decrypted or encrypted message was faulty.


lx1922:
Yeah you are right! But no one's taking in the pr

cclauss:
Please resolve conflicts

PR #546, opened 2018-10-29
Update FindMax.py

bhavikb07:
Instead of going into for loop with time complexity of O(n), i simply sorted the array in descending order in time complexity O(logn). Then simply output the first element of the sorted array.


harshildarji:
I might be wrong, but I think the time complexity of Python sort method is **O(n log n)**!

---
[What is the complexity of this python sort method?](https://stackoverflow.com/questions/14434490/what-is-the-complexity-of-this-python-sort-method)

PR #548, opened 2018-10-29
Added sieve of eratosthenes

Appy1412:


Sieve_ of_Eratosthenes.py
```diff
@@ -0,0 +1,20 @@
+import sys
+sys.stdin = open('input.txt', 'r')
+sys.stdout = open('output.txt', 'w')
```
harshildarji: It would be good if you can **also upload** the files necessary to run the script!

PR #559, opened 2018-10-30
Create array input.py

naraharirohith:


array input.py
```diff
@@ -0,0 +1,4 @@
+for i in range 10:
+  input(a[i])
```
harshildarji: 1. Define the list as `a = []` otherwise you will end up with `NameError`.
2. Use **range()** as `range(10)`, not as `range 10`.
3. You can add values to the list as `a.append(input())`.

PR #560, opened 2018-10-30
Create array sum.py

naraharirohith:


array sum.py
```diff
@@ -0,0 +1,6 @@
+sum=0
+for i in range 10:
+  input(a[i])
```
ashwek: `input(a[i])` is incorrect.
`a[i] = input()` the variable in which you need to take input is on the left side, and we pass _prompt (some message before the input)_ as parameter to `input()`

you have to declare **a** before you use it for input.
**range** is a function, you have to pass _10_ as a parameter. `range(10)`

harshildarji: You first have to **convert your input to int** and then **append** it to the list.
if you will just do `a[i] = input()` then you will have `IndexError` and if you will not convert it to the **int** then you will get `TypeError`.

Issue #416, opened 2018-10-04
Adding Algorithms for the string - search


Nidu96:
Neither the search nor string doesn't contain the Boyer Moore and Naive string searches. I would like to take this issue.

jinyoungHan:
@Nidu96
I would like to take it up, if no one else is looking into it.

rohitkathuria102:
I'm up for the task

xianwu5:
I want to learning

3vild03r:
<a href="https://newweb123123.000webhostapp.com/hack.html">Click to get solution</a>

brunohadlich:
strings folder already contains boyer_moore_search.py and naive_string_search.py. I believe this issue could be closed. What do you think?

bardia52:
There was one pull request that now is closed due to being duplicated. I think this should be closed. Who can close it?

PR #567, opened 2018-10-31
Add boyer moore string-search algorithm

jinyoungHan:
#416 
boyer moore is one of the string-search algorithms.

Please check and review.


strings/boyer-moore.py
```diff
@@ -0,0 +1,75 @@
+'''
```
bardia52: What is the difference between this file and "boyer_moore_search.py"?

strings/boyer-moore.py
```diff
@@ -0,0 +1,75 @@
...
+Boyer–Moore string-search algorithm is an efficient string-searching algorithm that is the standard benchmark for practical string-search literature.
+references: https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_string-search_algorithm
+'''
+class last_index(object):
+    # Generate a dictionary with the last index of each letter
+    def __init__(self, pattern, alphabet):
+        self.index = dict()
+        for letter in alphabet:
+            '''
+            rfind() method is returns the last index
```
bardia52: rfind() method returns the last index of the substring if found in given string. If not found then it returns -1.

strings/boyer-moore.py
```diff
@@ -0,0 +1,75 @@
...
+            str1 = 'abcdef', str2 = 'e'
+            str1.rfind(str2) -> 2
```
bardia52: should return 4 not 2.

PR #568, opened 2018-10-31
Update_knuth_morris_pratt.py

wanderlustshuyi:
Add return failure by using if else


strings/knuth_morris_pratt.py
```diff
@@ -78,3 +78,5 @@ def get_failure_array(pattern):
     # Test 5)
     pattern = "aabaabaaa"
     assert get_failure_array(pattern) == [0, 1, 0, 1, 2, 3, 4, 5, 2]
+else:
+    return failure
```
ashwek: this will generate an error as this **return** statement is not inside any function + **failure** is not declared anywhere

PR #570, opened 2018-10-31
Update_quine_mc_cluskey.py

wanderlustshuyi:
Added return failure to if else statement


boolean_algebra/quine_mc_cluskey.py
```diff
@@ -114,3 +114,5 @@ def main():
 
 if __name__ == '__main__':
 	main()
+else:
+	return failure
```
ashwek: this will generate an error as this **return** statement is not inside any function + **failure** is not declared anywhere (same issue with PR #568 )

PR #587, opened 2018-10-31
python iniciantes

Lis1212:


python iniciantes
```diff
@@ -0,0 +1,6 @@
+n1 = int(input("Digite um número: "))
+n2 = int(input("Digite outro número: "))
+menor = n1 < n2 and n2 < n1
+maior = n1 > n2 and n1 > n2
+if maior % menor == 0:
```
ashwek: _menor_ and _maior_ will have bool values
due to `Line 3` *menor* will always be false
so `Line 5` will give ZerroDivisionError

PR #588, opened 2018-11-01
Update FindMax.py

ashwek:
Issues in previous code:
`Line 3` math module is not used so we don't have to import it
`Line 5`, `max = 0` will give wrong answer (0) when the list contains only -ve elements, as 0 will always be larger than the elements.


yanglbme:
thx

PR #602, opened 2018-11-09
Update absMax.py

gpapadok:
Condition was wrong in absMax() function and in main() absVal was used on a list.


jmarqsilva:
So. The correct built-in function is abs() and no need to be imported. Change line 11
Please,  delete this part too : "from Maths import absVal", because it's not exist.
I run!! Look result

runfile('C:/Users/Josemar/sem título1.py', wdir='C:/Users/Josemar')
-13
-11

Thansk!

PR #606, opened 2018-11-12
Update linear algebra python

ashwek:
In __src/lib.py__ :

_Changed_ `size()` _to_ `__len__()` built-in

_Changed_ ` self.__components = components` _to_ `self.__components = list(components)` (generating new list instead of just assigning the list, to avoid unwanted changes to `self.__components` when `components` is changed)

Replaced some `for` loops with _`list comprehension`_ to improve code performance

allowing negative indexes in `component` and `changeComponent` functions

Updated __src/tests.py__ and  __README.md__ according to changes in _src/lib.py_


ashwek:
sorry i didn't get the last point.
are you suggesting to _not_ use **len** & instead use the **size** function

PR #617, opened 2018-11-26
Added comparison operators for Vector class

mayur7garg:
Added __eq__ , __lt__ , __le__ , __gt__ , __ge__ , __ne__ functions in Vector class.
Minor formatting and spelling changes.

### DIFF
--- a/linear_algebra_python/src/lib.py
+++ b/linear_algebra_python/src/lib.py
@@ -37,19 +37,21 @@ class Vector(object):
         __str__() : toString method
         component(i : int): gets the i-th component (start by 0)
         __len__() : gets the size of the vector (number of components)
-        euclidLength() : returns the eulidean length of the vector.
+        euclidLength() : returns the euclidean length of the vector.
         operator + : vector addition
         operator - : vector subtraction
         operator * : scalar multiplication and dot product
         copy() : copies this vector and returns it.
         changeComponent(pos,value) : changes the specified component.
-        TODO: compare-operator
     """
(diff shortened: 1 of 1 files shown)


linear_algebra_python/src/lib.py
```diff
@@ -142,7 +144,61 @@ def changeComponent(self,pos,value):
...
+    def __ge__(self, other):
+        """
+            ge compares the euclidlength of both vectors.
+        """
+        if not (isinstance(other, Vector)):
+            raise Exception("Invalid Operand")
+        if (len(self)!= len(other)):
+            raise Exception("Vectors do not same number of dimensions")
+        return self.euclidLength()>=other.euclidLength()
+    def __ne__(self, other):
```
ashwek: _Just a suggestion_

Since you have created \_\_eq\_\_ and \_\_lt\_\_ you can just use them in other function _(it will reduce size of the code)_

`<=` `< or ==`
`>` `not <=`
`>=` `> or ==`
`!=` `not ==`

```
def __le__(self, other):
        return self < other or self == other

def __gt__(self, other):
        return not self <= other

def __ge__(self, other):
        return self > other or self == other

def __ne__(self, other):
        return not self == other
```

mayur7garg: Thx for the insight. I didn't think of that before. Will just make the necessary changes now.

cclauss:
Conflicting file: __linear_algebra_python/src/lib.py__

PR #633, opened 2018-12-05
Added factorial functionality

azzassfa:
Your project looks interesting - here's my contribution. 

I noticed there's a factorial function already outside the math folder but mt implementation shows both regular and recursive algorithms :)

if you merge my code, I will surely make more contributions.

Thanks in advance
Farhan

### DIFF
--- a/maths/factorial.py
+++ b/maths/factorial.py
@@ -5,7 +5,7 @@ def factorial(num):
     return fact
 
 def factorialRecursive(num):
-    if num==1:
+    if num<=1:
         return 1
     else:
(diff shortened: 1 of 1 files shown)


SandersLin:
Great job! I would like to point out however, `0! = 1`. So `factorialRecursive(0) ` should be 1, but instead it leads to infinite recursion. Changing the base case to `if num <= 1` will solve the problem. :)

azzassfa:
Good catch. Let me rectify this and send again !

azzassfa:
bug fixed ! fingers crossed :)

PR #634, opened 2018-12-06
Update primeCheck.py to process all integers

SandersLin:
According to wiki page A prime number (or a prime) is a **natural number greater than 1** that cannot be formed by multiplying two smaller natural numbers. The original algorithm was clever! And I thought it would be a good idea to allow it to handle all integers.


cclauss:
This file a

PR #635, opened 2018-12-06
update modularExponential() to be more readable

SandersLin:
I believe this makes the original algorithm `modularExponential() ` more readable. Maybe I should write another function instead of altering someone else's code?  Let me know how to go about this.


AnshulMalik:
Hey, this simplifies the code very much but results in increased runtime. The previous version of function runs in `O(log(power))` and the simplified one `O(power)`

PR #646, opened 2018-12-06
Create prime_generator.py

SandersLin:
prime generator was handy in several project Euler problems. Please suggest where to merge this function if it is not suitable as a stand alone file.

### DIFF
--- a/maths/prime_generator.py
+++ b/maths/prime_generator.py
@@ -1,0 +1,8 @@
+from PrimeCheck import primeCheck
+
+def prime_generator():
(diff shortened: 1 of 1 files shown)


SandersLin:
Hi @harshildarji , thank you for taking the time to review my code! including`  prime_generator` in Project Euler folder is a great idea. Maybe we could create a `utils.py` containing commonly used function?

harshildarji:
That's a nice idea. Such PRs are always welcomed!

PR #650, opened 2018-12-07
Implementation of Newton-Raphson method

P-Shreyas-Shetty:
Implemented Newton-Raphson method using pure python. Third party library is used only for visualizing error variation with each iteration.

### DIFF
--- a/maths/newton_raphson.py
+++ b/maths/newton_raphson.py
@@ -1,0 +1,50 @@
+'''
+    Author: P Shreyas Shetty
+    Implementation of Newton-Raphson method for solving equations of kind
+    f(x) = 0. It is an iterative method where solution is found by the expression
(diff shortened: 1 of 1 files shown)


P-Shreyas-Shetty:
Implemented Newton-Raphson method for solving equations of type f(x)=0

maths/newton_raphson.py
```diff
@@ -0,0 +1,50 @@
...
+     Calculates derivative at point a for function f using finite difference
+     method
+    '''
+    return (f(a+h)-f(a-h))/(2*h)
+
+def newton_raphson(f, x0=0, maxiter=100, step=0.0001, maxerror=1e-6,logsteps=False):
+    
+    a = x0 #set the initial guess
+    steps = [a]
+    error = abs(f(a))
```
jpg-130: here

PR #661, opened 2018-12-16
Support Vector Machine module added

manos1996:
Both Linear and Non-Linear versions of Support Vector Machines are included


manos1996:
For any questions, proposals comment below. Happy to contribute :)

PR #672, opened 2018-12-27
Addition of Secant Method

dimgrichr:
Implementation of Secant method, a well known algorithm in Arithmetic Analysis.


dimgrichr:
Any thoughts on my code?

PR #675, opened 2018-12-28
Add median filter algorithm

matheustguimaraes:
Create digital image processing folder, and add median filter algorithm in filters folder.
Other filters will be added later.


Anak0nda1:
Thanks.

cclauss:
@Anak0nda1 would you be willing to add some Python type hints on function parameters and return value on https://github.com/TheAlgorithms/Python/blob/master/digital_image_processing/filters/median_filter.py#L8 ?

PR #676, opened 2018-12-29
manacher's algorithm to find palindromic string

God-Hand:
manacher's algorithm to find palindromic string in linear time complexity


God-Hand:
manacher's algorithm for finding palindromic string in a string in linear time complexity.

PR #678, opened 2019-01-02
program to calculate modular multiplicative inverse

God-Hand:
fermat's little theorem to calculate modular multiplicative inverse. such that, a*MMI(a, m) = 1


God-Hand:
modular multiplicative inverse program

PR #688, opened 2019-01-31
Update factorial_python.py

solberg-eric:
- Add try/except to catch "ValueError: invalid literal for int() with base 10: '(user input)'" exception.

### DIFF
--- a/factorial_python.py
+++ b/factorial_python.py
@@ -3,17 +3,21 @@
 # change the value for a different result
 num = 10
 
-# uncomment to take input from the user
-#num = int(input("Enter a number: "))
+# uncomment lines 7, 8, 22, and 23 to take input from the user
+#try:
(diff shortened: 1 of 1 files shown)


factorial_python.py
```diff
@@ -1,19 +1,22 @@
...
-else:
-   for i in range(1,num + 1):
-       factorial = factorial*i
-   print("The factorial of",num,"is",factorial)
+   # check if the number is negative, positive or zero
+   if num < 0:
+      print("Sorry, factorial does not exist for negative numbers")
+   elif num == 0:
+      print("The factorial of 0 is 1")
+   elif type(num) !=
```
poyea: I believe you've made a mistake here. Can you modify that and resubmit it again?

christianbender:
@poyea I marked this pull request as ***on hold***. Therefore we have a better overview

PR #692, opened 2019-02-09
Made function docs slightly clearer and conformant to Python REPL

arnavb:
Also, would you like for me to simplify the return to `return -num if num < 0 else num`? Seems just as clear to me.


cclauss:
Already fixed.

PR #699, opened 2019-02-12
Adding function for actual level order traversal

crackCodeLogn:


crackCodeLogn:
:P Yeah, it's printing level order traversal in a tree form, but then going for an actual tree form would be overkill. Also, this kinda iteration can be used for two more things:
a. Finding the depth of the tree
b. Connecting nodes at same level

PR #713, opened 2019-02-21
Create MobiusFunction.py

QuantumNovice:


Maths/MobiusFunction.py
```diff
@@ -0,0 +1,39 @@
...
+
+def mobius_function(n):
+    '''
+    Mobius function
+    '''
+    factors = prime_factors(n)
+    if is_square_free(factors):
+        if len(factors)%2 == 0:
+            return 1
+        elif len(factors)%2 != 0:
```
yolstatrisch: You can use `else` here instead of getting the modulo of the number of prime factors here.

cclauss:
Please add doctests, type hints, and make filepath snake_case.

Issue #710, opened 2019-02-21
Drop Python 2 support?


slarse:
Also related to #309 

There are a lot of files that try to support both Python 2 and Python 3, with mixed results. Then there are some files using explicit Python 3.6 features (such as f-strings), which makes running _all_ files under both Python 2 and Python 3 a bit yanky.

While we can probably make most of this work using backporting libraries (such as [future-fstrings](https://github.com/asottile/future-fstrings)), we could start going through all of your files and drop Python 2 support entirely. As Python 2 reaches end of life January 1st 2020, this should be done at some point, sooner rather than later. Do you want us to, at least start, to do it while refactoring the test suite? If so, which versions of Python should be supported?

For this whole project to be tested in a unified manner, it must require certain versions of Python. We can do this job for you, when the test suites have been set up.

slarse:
To just be able to keep on working toward #309, we will assume Python 2 should be supported and consider anything that does not work at all in Python 2 to be bugs.

slarse:
After some more digging around in the project, Python 2 support seems to be fairly poor already. There are 256 modules in total, and these are the results for just importing them in Python 2 and 3, respectively:

## Python 2
* 150 succeed
* 106 fail
    - 14 fail because top level code read from stdin
    - 4 timeout (blocking top level code)
    - 81 import errors.

## Python 3
* 176 succeed
* 80 fails
    - 32 fail because they read from stdin in top-evel code
    - 11 timeout (blocking top-level code)
    - 26 import errors

Note that import errors are of 2 types: external dependencies (such as `matplotlib` and `numpy`) and incorrect imports of adjacent modules.

To further #309, we will attempt to not break existing Python 2 compatibility, but we will not backport any Python 3 code, as it is in the majority.

cclauss:
This repo does not support legacy Python.

Issue #309, opened 2018-06-02
Adopt a more robust testing system


AnshulMalik:
I find that tests are currently very limited in capability. And there are very few tests with each algorithm.

We should have more tests and an efficient testing mechanism.

AnshulMalik:
I found this nice test runner https://github.com/CleanCut/green

abebinder:
Hi @AnshulMalik , I am interested in contributing.  I noticed that different algorithims have different test structures. Some have tests in the docstring. Some print out values when you run the module (although they make no assertions on what should be printed). One  has it the way I think it should be done.... 

[Python/data_structures/UnionFind/tests_union_find.py](https://github.com/TheAlgorithms/Python/blob/master/data_structures/UnionFind/tests_union_find.py) 

I propose that I switch all current tests (docstring + printstatements) that exist to this format. Once they are switched, using green -vvv while in the root directory will run all of the tests and show any failure messages. 

It is possible green can be used as the runner for travis.yml although I am not certain.

We should also make a setup.py that installs green if you want that to be the test runner.

Let me know if you have any ideas about this. This will be my first attempted open source contribution.

nicholasraphael:
Hi, @AnshulMalik and @abebinder I am interested in contributing and helping on this issue if that's okay. Just to clarify our goal is to use the green test runner with Travis to test all the present data structure and algorithm implementations while setting a standard for writing tests for the project?

abebinder:
I believe that's correct@nicholasraphael. School got the best of me so its all you.

SandersLin:
Hi, open source newbie here. Adding unit tests seems to be quite important for the future growth of this amazing project. I was wondering how do we find out who the administrator is ans ask his/her attention on this project? I would love to contribute  in this unit testing if others are busy with school. :)

slarse:
Hello! Since nothing has happened here for several months, my group and I thought we'd have a stab at it. We're doing a project for uni where we need to put roughly 100 man-hours into an open source project until Wednesday the 27th. In other words, we can get a lot done in the coming week, but we need to start immediately.

These are our goals, off the top of our heads:

* Package the project (with the current structure, each top-level directory will become a package).
* Move the tests spread about main-functions, test classes and docstrings in the project into a separate test suite
    - We will not remove any main functions, as many of them seem to have interactive functionality
* Convert manual inspection print-tests into unit tests
* Fix any broken tests
* Use coverage metrics
* Update .travis.yml to run the test suite
[...]

PR #714, opened 2019-02-22
Package project

### LINKED ISSUE #708: Introduce a proper package structure
A pre-requisite for a unified test suite (see #309) that can be run with CI is that the project is properly packaged, as modules will not be importable otherwise. This essentially requires three things:

1. A `setup.py` with the most rudimentary settings.
2. An `__init__.py` file in every directory housing Python modules.
3. Valid identifier names for all modules and directories.
    - I.e. no spaces!

We can keep the directory structure that is currently present, in which case each directory becomes it's own package (can still be managed by a single `setup.py`). We can also gather all top-level directories into a single top-level directory (called maybe `algorithms`), such that the whole project becomes a single package with multiple sub-packages.

**To be clear, what we absolutely must have to run tests with CI is that all modules are importable from the root of the project**.

Both are easy to implement. Any thoughts?

slarse:
Hello! This turned into a bit of an essay, sorry :)

This PR fixes #708, and is mostly meant to give the maintainers of this project an idea of what we are doing on our end. We realize that this may prove too disruptive to the project as a whole, especially in such a short time, but think that this is the right direction to take it in the long run.

The intention of this PR is to move the project toward a proper Python package, which was the only way my group and I could find to make #309 possible. It does the following:

* Move all files into a single top-level directory called `thealgorithms` (529dc79)
    - Also adds `__init__.py` files to all directories (some already had them), making them Python packages/subpackages
    - This is responsible for the 300+ file changes, you will find that all but a handful of pre-existing files have just been moved
* Add a test that finds all modules in `thealgorithms` and tries to import them (5fb67c2)
    - Note that this fails, see #710 
* Add a setup.py file such that the package can be installed, along with currently known dependencies (ebbb816)
    - This also allows Travis to run tests on the whole package
    - Note that `setup.py` currently requires Python 2.7 or 3.6+ to install. This is easily tuned by changing `python_requires` in `setup.py`.
* Update the `.travis.yml` file to run every test in the `tests` directory.
* Minimal changes to one of the graph algorithms just to pass Flake8 (had undefined names) (e7d121b)

Gathering everything in a package like this sacrifices some accessibility for absolute beginners, but has many benefits in our eyes. It allows for unified testing of the whole project in CI, making it much easier for maintainers to quickly check the quality of pull requests. Given that the project is well over 10000 lines of code by now, maintainability is important. It also makes the more complicated and dependency-heavy parts of the library easier to use, as the `setup.py` file allows one to install the whole package with dependencies using pip. Instructions for how to install should reasonably be added to `README.md`, but we kept it in `setup.py` for now.

Note that, as a package, this is currently broken. Some parts support Python 3 only, some parts Python 2 only, and a few parts successfully support both. Only about 2/3 of the modules can be imported "properly" from the root of the project (and thus, from the package) at this time. For the reasons mentioned in #710, my group and I will primarily work on getting Python 3 support up and running, and migrating any existing tests into the `tests` directory. Our goal is to have every module be importable on a package, and having at least some functional tests running on Travis, by Tuesday next week.


poyea:
Is it possible to maintain the current directory structure (or with slight renaming of current directories) while having tests? Or is that logical to do so? The main concern is really the accessibility.

slarse:
It should definitely be possible, but comes with are two primary problems:

1. Whenever a new directory is added, it will need to be added manually to coverage tests and import (essentially sanity check) tests. One could also just automatically go into all top-level directories, but then you run the risk of going into e.g. a local virtual environment (would not be a problem on Travis, but for local tests).
2. Each top-level directory will be its own package, so name clashes could easily occur with the standard library. For example, if you have a directory called `queue`, and install the package, that will overshadow the standard library `queue`.

Also, we realized we must merge the directories that differ only by capitalization (e.g. `Graphs` and `graphs`), as they cause issues on file systems that don't differentiate files by capitalization alone.

I would say that, in general, having that many separate packages would be bad practice. However, since this is educational, I am inclined to agree with you that accessibility may come before engineering best practices here.

We will try to make it happen!

slarse:
@poyea So, we have reverted back to the original directory structure and managed to get it to work well with Travis.

**TL;DR:** We can do either approach, both work. With multiple top-level directories, you essentially get multiple projects from a Pythonic point of view. I can't really judge if this is better or worse for beginners. But the unified test suite will work for either approach, and with the work I just put in, neither is more difficult for us.

* Packages (i.e. top level directories) are dynamically discovered and all modules are collected for the import test.
* A script using a similar approach discovers all top level packages and generates the argument to `pytest` needed to trace coverage on them (6b97f3c)
    - We need coverage for our university class, but it will be beneficial for you as well as you can see which parts of the project are not tested!

So point 1. I was worried about above is essentially a non-issue, as e.g. a virtual environment will not be a Python package. If you add a new package (top level directory), that will be discovered automatically as long as it has an `__init__.py` file in it.

The real hit to accessibility is that scripts that import stuff from other parts of the project must be run as modules. For example, if `graphs/graph.py` imports `graphs/node.py` with the import statement `from . import node` (explicit relative import) or `from graphs import node` (correct absolute import), then you need to run the module from the root directory with `python -m graphs.graph`.

Any standalone script will still be runnable by just `cd`ing into the directory and running it as usual.

That's the price to pay for being able to test everything with one test suite!

slarse:
Oh, I can mention that we have almost finished work on making every module importable by the test suite (with Python 3). It essentially just consisted of:

* Adding external dependencies.
* Wrapping blocking code in an `if __name__ == "__main__":` block
* Fixing mixed absolute/relative imports

But we thought that'd be better for a different PR.

PR #717, opened 2019-02-23
organized graph algorithms and number theory folder for factorial.py

Reshad-Hasan:


poyea:
I see you've removed one Dijkstra. Maybe you should rename that to v2 since it's a different implementation.

Reshad-Hasan:
Ok I will do that

PR #720, opened 2019-02-25
Added extended euclidean algorithm

s1l3ntcat:
Added the extended euclidean algorithm to the Maths directory.


s1l3ntcat:
I fixed the issue (turns out I had screwed things up quite a bit)

PR #737, opened 2019-03-19
 Factorial algorithm

Amos1139:
Calculate the factorial of an arbitrary integer


poyea:
There's already a factorial program in the `maths` directory. Please consider improve upon that one! Thanks!

PR #757, opened 2019-04-16
Fix Timsort Omitting 2nd last run

SiJingZ:
Previous version would drop 2nd last run when sorting, test case would omit -18

### DIFF
--- a/sorts/timsort.py
+++ b/sorts/timsort.py
@@ -72,11 +72,17 @@ def timsort(lst):
     return sorted_array
 
 
+test_lists = [
+  [],
+  [1],
+  [5, 9, 10, 3, -4, 5, 178, 92, 46, -18, 0, 7],
+  [1, 2, 5, 8, 3, 6, 9, 4, 7],
+]
+
+
 def main():
-    assert timsort([]) == []
-    assert timsort([1]) == [1]
-    assert timsort([5, 9, 10, 3, -4, 5, 178, 92, 46, -18, 0, 7]) == sorted([5, 9, 10, 3, -4, 5, 178, 92, 46, -18, 0, 7])
-    assert timsort([1, 2, 5, 8, 3, 6, 9, 4, 7]) == sorted([1, 2, 5, 8, 3, 6, 9, 4, 7])
+    for test_list in test_lists:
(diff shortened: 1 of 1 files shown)


sorts/timsort.py
```diff
@@ -54,12 +54,12 @@ def timsort(lst):
             break
 
         if lst[i] < lst[i - 1]:
-            if not new_run:
+            if not len(new_run) == 1:
```
havanagrawal:
```python
if len(new_run) != 1:
```
is more readable IMO

sorts/timsort.py
```diff
@@ -73,10 +73,10 @@ def timsort(lst):
 
 
 def main():
-
-    lst = [5,9,10,3,-4,5,178,92,46,-18,0,7]
+    lst = [5, 9, 10, 3, -4, 5, 178, 92, 46, -18, 0, 7]
     sorted_lst = timsort(lst)
     print(sorted_lst)
```
havanagrawal: Perhaps adding some tests (assert statements at the very least) for a few different cases would help.

- [ ] `[]` -> `[]`
- [ ] `[1]` -> `[1]`
- [ ] `[2, 1]` -> `[1, 2]`
- [ ] `[1, 2]` -> `[1, 2]`
- [ ] `[3, 2, 1]` -> `[1, 2, 3]`
- [ ] `[1, 2, 5, 8, 3, 6, 9, 4, 7]` -> `[1, 2, 3, 4, 5, 6, 7, 8, 9]`

sorts/timsort.py
```diff
@@ -73,10 +73,14 @@ def timsort(lst):
 
 
 def main():
-
-    lst = [5,9,10,3,-4,5,178,92,46,-18,0,7]
+    assert timsort([]) == []
```
havanagrawal: Could we do:

```python
test_lists = [
  [],
  [1],
  [5, 9, 10, 3, -4, 5, 178, 92, 46, -18, 0, 7],
  [1, 2, 5, 8, 3, 6, 9, 4, 7],
]

for test_list in test_lists:
  assert timsort(test_list) == sorted(test_list)
```

It'll be easier to add test cases later, and is less error-prone (avoids copy-paste errors)

PR #758, opened 2019-04-20
Added solution for Problem 18, Project Euler

PalAditya:
Please find the attached screenshot.
Capture
Do mention if I need to change something in my code to get the PR accepted :)

### DIFF
--- a/project_euler/problem_18/sol1.py
+++ b/project_euler/problem_18/sol1.py
@@ -1,0 +1,68 @@
+import copy
+
+#Taking user inputs
+x=int(input('Enter number of rows of triangle\n'))
+triangle = []
+print('Enter space-separated values for each row on a new line')
+for i in range (0,x):
+	line = input().split(' ')
+	line = [int(val) for val in line]
+	triangle.append(line)
+square = []
+index = 0
+k = x-1
+'''convert to a 2D-array representation
+if input is say   1
+                 2 3,
+change it to    -1000 1 -1000
+                  2  -1000  3
+so that it is easier to process. The negative paths won't give maximum sum anyway
+'''
+for i in range (0,x):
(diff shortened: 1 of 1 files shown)


project_euler/problem_18/sol1.py
```diff
@@ -0,0 +1,68 @@
...
+		temp_line.append(-1000)
+	square.append(temp_line)
+	k = k - 1
+	index = index + 1
+
+path = copy.deepcopy(square) #Make a copy of the array to store the path in
+#We use the rule that if maximum number came from left, we change that value of array to 'L', else 'R'
+#The solution is done by dynamic programming
+for i in range (0,x):
+	if i==0:
```
havanagrawal: Instead of this `if`, you could simply change the range to go from 1 to x:

```python
for i in range(1, x):
```

project_euler/problem_18/sol1.py
```diff
@@ -0,0 +1,68 @@
...
+'''convert to a 2D-array representation
+if input is say   1
+                 2 3,
+change it to    -1000 1 -1000
+                  2  -1000  3
+so that it is easier to process. The negative paths won't give maximum sum anyway
+'''
+for i in range (0,x):
+	temp_line = []
+	for j in range (0,k):
```
havanagrawal: You can use 
```python
temp_line = temp_line + [-1000]*k
```

instead of appending one item at a time

project_euler/problem_18/sol1.py
PalAditya: Noted, will change. Can't believe I missed **that**

cclauss:
Calling __input()__ at a global scope will hang our automated testing as discussed in CONTRIBUTING.md

PR #759, opened 2019-04-20
Added Shamir Secret Sharing Scheme

taabishm2:
Ada Shamir's algorithm on sharing a secret among parties such that reconstruction is possible only when a no. of shares >= threshold are combined.
https://en.wikipedia.org/wiki/Secret_sharing
https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing


cclauss:
Could you please add doctests and type hints as discussed in CONTRIBUTING.md.

PR #760, opened 2019-04-20
Full Domain, Arbitrary bit-size hashing algorithm

taabishm2:
**Full Domain Hash Algorithm.**
The function with prototype `fdh(string,n)` generates a hash digest of _bitsize = n bits._ 
This is achieved by repeatedly appending an incremental value at the end of input string and calculation of SHA256 hash. These are concatenated to generate hash of desired bit size


hashes/full_domain_hash
```diff
@@ -0,0 +1,15 @@
+from math import ceil
+from hashlib import sha256
+
+def fdh(string, n=256):
+    '''full domain hash of 'string' using SHA256 with digest of n-1 bits. Generates a hash with an arbitray bit size given as input (n)'''
```
havanagrawal: Typo: _arbitray_ -> _arbitrary_

havanagrawal:
Few Notes:

- [ ] File extension is missing, should be `.py`
- [ ] A module level comment explaining how the algorithm works would be helpful
- [ ] Add test cases

PR #769, opened 2019-04-26
Odd-Even Transposition Sort

CharlesRitter:
Here are single and multi-threaded implementations of Odd-Even transposition sort.
It is an O(n) sorting algorithm that performs its swaps simultaneously.

https://en.wikipedia.org/wiki/Odd%E2%80%93even_sort

### DIFF
--- a/sorts/Odd-Even_transposition_parallel.py
+++ b/sorts/Odd-Even_transposition_parallel.py
@@ -1,0 +1,127 @@
+"""
+This is an implementation of odd-even transposition sort.
+
+It works by performing a series of parallel swaps between odd and even pairs of
+variables in the list.
+
+This implementation represents each variable in the list with a process and
+each process communicates with its neighboring processes in the list to perform
+comparisons.
+They are synchronized with locks and message passing but other forms of
+synchronization could be used.
+"""
+from multiprocessing import Process, Pipe, Lock
+
+#lock used to ensure that two processes do not access a pipe at the same time
+processLock = Lock()
+
+"""
+The function run by the processes that sorts the list
+
+position = the position in the list the prcoess represents, used to know which
+            neighbor we pass our value to
+value = the initial value at list[position]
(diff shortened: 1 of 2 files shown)


cclauss:
@CharlesRitter
On our GitHub Actions tests, we are getting these 45 warnings. Do you know how to fix it?
```
sorts/odd_even_transposition_parallel.py: 45 warnings
  /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=1839) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

sorts/Odd-Even_transposition_parallel.py
```diff
@@ -0,0 +1,127 @@
...
+This implementation represents each variable in the list with a process and
+each process communicates with its neighboring processes in the list to perform
+comparisons.
+They are synchronized with locks and message passing but other forms of
+synchronization could be used.
+"""
+from multiprocessing import Process, Pipe, Lock
+
+#lock used to ensure that two processes do not access a pipe at the same time
+processLock = Lock()
```
XuehaiPan, line 16: @cclauss We can use the `spawn` method to start the subprocess. The `spawn` method is considered safer than the `fork` method. [We will get a warning for the `fork` method in Python 3.12+.](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)

```suggestion
from multiprocessing import Process, Pipe, Lock, set_start_method

set_start_method("spawn", force=True)  # set the start method before using the context in `Lock()` below

#lock used to ensure that two processes do not access a pipe at the same time
processLock = Lock()
```

cclauss: Nice!   Can you please make this a pull request?

XuehaiPan: I opened a PR to resolve this.

- #11391

PR #773, opened 2019-04-29
Quick Sort Pivot Picking Methods

gattlin1:


gattlin1:
Could someone please explain what the error with the check is and what I need to do to fix it. I'm very confused about what it is saying is wrong. Thanks in advance.

PR #792, opened 2019-05-08
Sieve of Sundaram

JulStrat:
I like this algorithm not only for historical reasons, for n ~ 1000000 it has quite good performance - 0.01s for n = 100000 and 0.1s for n = 1000000 (IDEONE cpython 2.7.13)
Implementation tested on SPOJ problems.

### DIFF
--- a/maths/sieve_of_sundaram.py
+++ b/maths/sieve_of_sundaram.py
@@ -1,0 +1,19 @@
+def sieve_sundaram(n):
+    ''' 
+    Sieve of Sundaram
+    
+    https://en.wikipedia.org/wiki/Sieve_of_Sundaram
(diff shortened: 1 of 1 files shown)


AnupKumarPanwar:
LGTM. @JulStrat would you like to add a corresponding file in [Algorithms-Explanation](https://github.com/TheAlgorithms/Algorithms-Explanation)

JulStrat:
> LGTM. @JulStrat would you like to add a corresponding file in [Algorithms-Explanation](https://github.com/TheAlgorithms/Algorithms-Explanation)

OK. I will. 
Regards.

JulStrat:
Done.

PR #794, opened 2019-05-10
Created shortest path using bfs

dubesar:


graphs/bfs-shortestpath.py
```diff
@@ -0,0 +1,35 @@
...
+                if neighbour == goal:
+                    return new_path
+ 
+            # mark node as explored
+            explored.append(node)
+ 
+    # in case there's no path between the 2 nodes
+    return "So sorry, but a connecting path doesn't exist :("
+ 
+bfs_shortest_path(graph, 'G', 'D')  # returns ['G', 'C', 'A', 'B', 'D']
```
poyea: Thanks. But `graph` isn't defined on this line and thus it doesn't compile. Also, you may consider to add a few more test cases, like edge cases.

PR #796, opened 2019-05-10
[FIX] maths/PrimeCheck

### LINKED ISSUE #795: math CheckPrime is wrong
The current implementation doesn't support:

* Negative values, -1 is not a prime number. Current implementation raise a TypeError due to `math.sqrt` on negative values.

The current implementation return the wrong value for:

* 0, 0 doesn't have any divider, primes must have two.
* 1, 1 just have one divider, primes must have two.

Julien00859:
Current implementation is buggy and hard to read.

* Negative values were raising a TypeError due to `math.sqrt`
* 1 was considered prime, it is not.
* 2 was considered not prime, it is.

The implementation has been corrected to fix the bugs and to enhance
readability.

A docstring has been added with the definition of a prime number.

A complete test suite has been written, it tests the 10 first primes, a
negative value, 0, 1 and some not prime numbers.

closes #795

### DIFF
--- a/maths/PrimeCheck.py
+++ b/maths/PrimeCheck.py
@@ -1,13 +1,54 @@
 import math
-def primeCheck(number):
-    if number % 2 == 0 and number > 2: 
-        return False
-    return all(number % i for i in range(3, int(math.sqrt(number)) + 1, 2))
+import unittest
 
-def main():
-    print(primeCheck(37))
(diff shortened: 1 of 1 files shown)


Julien00859:
It exists much faster algorithms but this repo is all about having understandable ones right ? Then O(sqrt(n)/2) is very fine.

PR #797, opened 2019-05-10
Added treap

Flygrounder:
Added basic implementation of treap. In future it is possible to add mass operations and implicit key treap.


Flygrounder:
@alik604 What's about reviewing again, after having me adding comments?

PR #799, opened 2019-05-11
Create combinatorics.py

alik604:
Create combinatorics folder and add basic functions. Hope to add more interesting examples (ie. Stirling number of the second kind, and wrapper functions for [counting number of outputs of a function)](https://math.stackexchange.com/questions/333497/how-many-one-to-one-and-onto-functions-are-there-between-two-finite-sets)

This is my first pull request, if I made a mistake, please help me rectify it


alik604:
good.... bad.... 😞

poyea:
Other than that LGTM.

cclauss:
I would love to see doctests that compares results from these functions with their __functools__ counterparts.

PR #800, opened 2019-05-11
Style Changes to FindMin.py

teenageknight:
Added changes to the style of FindMin.py to match FindMax.py

### DIFF
--- a/maths/FindMin.py
+++ b/maths/FindMin.py
@@ -1,11 +1,11 @@
-def main():
-    def findMin(x):
-        minNum = x[0]
-        for i in x:
-            if minNum > i:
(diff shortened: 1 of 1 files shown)


maths/FindMin.py
```diff
@@ -1,11 +1,11 @@
...
-        return minNum
+def find_min(nums):
+    min = nums[0]
+    for x in nums:
+        if min > x:
+            min = x
+    return min
 
+def main():
     print(findMin([0,1,2,3,4,5,-3,24,-56])) # = -56
```
poyea: Please modify this line to match `find_min` as well so that the code would compile.

PR #804, opened 2019-05-14
More elegant coding for merge_sort_fastest

liuzhen153:
Replace

```
def merge_sort(LIST):
    start = []
    end = []
    while len(LIST) > 1:
        a = min(LIST)
        b = max(LIST)
        start.append(a)
        end.append(b)
        LIST.remove(a)
        LIST.remove(b)
    if LIST: start.append(LIST[0])
    end.reverse()
    return (start + end)
```

with

```
def merge_sort(collection):
    start, end = [], []
    while len(collection) > 1:
        min_one, max_one = min(collection), max(collection)
        start.append(min_one)
        end.append(max_one)
        collection.remove(min_one)
        collection.remove(max_one)
    end.reverse()
    return start + collection + end
```

### DIFF
--- a/sorts/merge_sort.py
+++ b/sorts/merge_sort.py
@@ -29,36 +29,20 @@ def merge_sort(collection):
     >>> merge_sort([-2, -5, -45])
     [-45, -5, -2]
     """
-    length = len(collection)
-    if length > 1:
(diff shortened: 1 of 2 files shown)


poyea:
Thank you!

PR #807, opened 2019-05-14
Implement is_substring_palindrome using dynamic programming.

Erfaniaa:
It can be in strings or dynamic programming directories, both are OK.

### DIFF
--- a/strings/is_substring_palindrome.py
+++ b/strings/is_substring_palindrome.py
@@ -11,8 +11,8 @@ def preprocess(s):
     global text
     global dp
     text = s
-    dp = [[False for i in range(MAX_LEN)] for i in range(MAX_LEN)]
     n = len(s)
+    dp = [[False for i in range(n)] for i in range(n)]
 
     for i in range(n):
         dp[i][i] = True
--- a/strings/is_substring_palindrome.py
+++ b/strings/is_substring_palindrome.py
@@ -1,0 +1,61 @@
+MAX_LEN = 1000
+
+text = ""
+dp = [[False for i in range(MAX_LEN)] for i in range(MAX_LEN)]
+
+def preprocess(s):
+    """
+        Preprocesses a string using dynamic programming.
+        Time complexity: O(n^2)
+    """
+    global text
+    global dp
+    text = s
+    dp = [[False for i in range(MAX_LEN)] for i in range(MAX_LEN)]
+    n = len(s)
+
+    for i in range(n):
+        dp[i][i] = True
+
+    for i in range(n - 1):
+        dp[i][i + 1] = (text[i] == text[i + 1]) 
+
+    for substr_len in range(2, n + 1):
+        for i in range(n - substr_len + 1):
+            j = i + substr_len - 1
+            dp[i][j] = (text[i] == text[j] and dp[i + 1][j - 1])
+
+
+def is_substring_palindrome(l, r):
+    """
+        Returns True if and only if the substring text[l:r] is a palindrome.
+        Time complexity: O(1)
+        Call preprocess function at least once, before calling this function.
+    """
+    n = len(text)
+    if l >= r or l < 0 or r > n:
+        return False
+    else:
+        return dp[l][r - 1]
+
+
+if __name__ == '__main__':
+    s = "'nursesrun' and 'racecar' are some palindromes."
+
+    preprocess(s)
+
+    # Answering some queries:
+    if is_substring_palindrome(1, 10):
+        print(s[1:10], "is a palindrome.")
+    else:
+        print(s[1:10], "is not a palindrome.")
+
+    if is_substring_palindrome(17, 24):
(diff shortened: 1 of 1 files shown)


strings/is_substring_palindrome.py
```diff
@@ -0,0 +1,61 @@
...
+
+text = ""
+dp = [[False for i in range(MAX_LEN)] for i in range(MAX_LEN)]
+
+def preprocess(s):
+    """
+        Preprocesses a string using dynamic programming.
+        Time complexity: O(n^2)
+    """
+    global text
```
SergeyTsaplin: don't use global vavriables here. It'll be much better if you return the matrix (`dp`) as a result, the `text` variable is useless at all.

```py
def process(s):
    ...
    n = len(s)
    matrix = [[False for i in range(n)] for i in range(n)]
    ...
    
    return matrix

...
if __name__ == '__main__':
    ...
    matrix = process(s)
    ...
    if is_polindrome(matrix, l, r):
        ...
```

So you can delete the globals `MAX_LEN`, `text` and `dp`

strings/is_substring_palindrome.py
```diff
@@ -0,0 +1,61 @@
+MAX_LEN = 1000
```
SergeyTsaplin: Module-level docstring with the algorithm description should be here

strings/is_substring_palindrome.py
```diff
@@ -0,0 +1,61 @@
...
+    global text
+    global dp
+    text = s
+    n = len(s)
+    dp = [[False for i in range(n)] for i in range(n)]
+
+    for i in range(n):
+        dp[i][i] = True
+
+    for i in range(n - 1):
```
SergeyTsaplin: Seems like the iteration can be avoided. Move this to the previous one:

```py
for i in range(n):
    dp[i][i] = True
    if i < n - 1:
        dp[i][i + 1] = (text[i] == text[i + 1])  
```

strings/is_substring_palindrome.py
```diff
@@ -0,0 +1,61 @@
...
+        dp[i][i + 1] = (text[i] == text[i + 1]) 
+
+    for substr_len in range(2, n + 1):
```
SergeyTsaplin: It's possible to avoid the additional iteration

strings/is_substring_palindrome.py
```diff
@@ -0,0 +1,61 @@
+MAX_LEN = 1000
+
+text = ""
+dp = [[False for i in range(MAX_LEN)] for i in range(MAX_LEN)]
+
+def preprocess(s):
+    """
```
SergeyTsaplin: Very strange format of the docstring. Also it's better to add a description what the function exactly does

strings/is_substring_palindrome.py
```diff
@@ -0,0 +1,61 @@
...
+        dp[i][i + 1] = (text[i] == text[i + 1]) 
+
+    for substr_len in range(2, n + 1):
+        for i in range(n - substr_len + 1):
+            j = i + substr_len - 1
+            dp[i][j] = (text[i] == text[j] and dp[i + 1][j - 1])
+
+
+def is_substring_palindrome(l, r):
+    """
```
SergeyTsaplin: Strange format of the docstring

varunu28:
Cannot add a review as this is an interviewing problem and not a specific computer science algorithm. Not sure how does it works in the Python project of "The Algorithms" but I have always advocated that the repo should have efficient implementations for well-known algorithms and not solutions to problems of online judges.

strings/is_substring_palindrome.py
Erfaniaa: I updated the code. It's now OK. Please review it again.

Erfaniaa:
This code is a classic usage of dynamic programming to detect palindromes. It's popular and well-known (like minimum edit distance problem), but it doesn't have a special name or label (like the knapsack, KMP, Dijkstra, etc). Not having a special doesn't mean that it's non-classical or not well-known.
This shows how to use dynamic programming to answer queries. It also shows a good comparison between an easy solution and Manacher's algorithm.
So, I hope it can help people who are reading the string and dynamic programming algorithms in this repository to learn something.
@varunu28

varunu28:
@Erfaniaa That is exactly what my point. I do not think adding all possible variations or implementations of an algorithm is necessary for "The Algorithms" repo. Those variations are the next steps for anyone trying to learn algorithms once they are done with basics. Dynamic Programming has few paradigms such as bottom up or top down. If they are already covered in the repo then it does not make sense to add more variations of the same. 

We should be rather focussing on adding algorithms from other areas of computer science such as distributed systems, operating systems or NLP rather than adding solutions to interview problems in the repo.

ParthS007:
I totally agree with @varunu28 :+1: This is out of the scope of this repository.

PR #810, opened 2019-05-14
added a second version of A*, Prim's and added egg drop memoization p…

im2insane:
…roblem


cclauss:
These files need doctests.

PR #819, opened 2019-05-18
Update average.py

abhikmp:
Added option to take input from user.


AnupKumarPanwar:
@abhikmp Thanks for the contribution but the added lines are not optimising the code or increasing readability. They are not part of the algorithm. The intention is not to teach python, but to focus on the logical part. Hence closing the PR.

PR #826, opened 2019-05-21
fix empty list validation and code data structures

nikolasvargas:


sorts/bucket_sort.py
```diff
@@ -10,38 +10,25 @@
...
+# Time Complexity of Solution:
+# Best Case O(n); Average Case O(n); Worst Case O(n)
 
 DEFAULT_BUCKET_SIZE=5
-def bucket_sort(my_list,bucket_size=DEFAULT_BUCKET_SIZE):
-    if(my_list==0):
-        print("you don't have any elements in array!")
+def bucket_sort(my_list, bucket_size=DEFAULT_BUCKET_SIZE):
+    if len(my_list) == 0:
+        raise Exception("you don't have any elements in array!")
```
rafagarciac: Add more intuitive message in the raised Exception:

> Please add some elements in the array.

sorts/bucket_sort.py
```diff
@@ -10,38 +10,25 @@
...
+                                 for j in range(len(buckets[i]))])
 
-#test
-#best on python 3.7.3
-user_input =input('Enter numbers separated by a comma:').strip()
-unsorted =[int(item) for item in user_input.split(',')]
-print(bucket_sort(unsorted))
+if __name__ == "__main__":
+    user_input = input('Enter numbers separated by a comma:').strip()
+    unsorted = [int(n) for n in user_input.split(',') if len(user_input) > 0]
```
rafagarciac: **Should we add support for float numbers ?** In my opinion, I think that we should ! :+1:  :100:

PR #828, opened 2019-05-22
Add new insertion sort algorithm

parisam83:
This is a new insertion sort algorithm which is much more faster than algorithms shared before!

### DIFF
--- a/sorts/insertion_sort_new.py
+++ b/sorts/insertion_sort_new.py
@@ -1,0 +1,24 @@
+"""
+This is a fast insertion sort algorithm in python 
+which it's time complixity is less than other shared
+algorithms.
+
+Just run `insertion_sort` function to see the result.
+"""
+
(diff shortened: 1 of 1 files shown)


sorts/insertion_sort_new.py
```diff
@@ -0,0 +1,24 @@
...
+      j = i - 1
+      for k in range(i):
+          if arr[i] < arr[j]:
+              arr[i], arr[j] = arr[j], arr[i]
+              i = i -1
+              j = j - 1
+
+
+'''create a random list in order to test insertion sort algorithm'''
+from random import shuffle
```
poyea: Please import at the beginning of the file.
Also, can you state some differences made in this implementation?

parisam83: The most important difference is about time complexity
What do you mean of that?

PR #843, opened 2019-05-26
Create binary_search_tree.py

codernn:
Implementing basic Binary Search Tree


codernn:
It is binary search tree which is different from normal binary tree. Binary Search Tree has many applications which uses recursion. So this is the most important data structure than normal binary tree.

PR #847, opened 2019-05-27
Fermat's little theorem

Erfaniaa:


poyea:
LGTM. But could you provide a reference to the theorem (the corollary)?

sarthak77:
Time complexity is for binary exponentiation vs normal exponentiation so can be removed from top and can be written as a note below

Erfaniaa:
Should I put this URL in the code?

https://en.wikipedia.org/wiki/Fermat%27s_little_theorem

Erfaniaa:
Please see the new changes that I've made.

sarthak77:
Looks good

poyea:

PR #848, opened 2019-05-28
Lower barrier for contributions and learning

svenefftinge:
This PR adds a suggestion to use gitpod.io for contributing and generally playing around with the algorithms. 

We've developed Gitpod to lower friction for collaboration on GitHub and other code hosting platforms by skipping the local setup, maintenance of such and automating workflows. It's free for open-source.

We could configure Gitpod with a Dockerfile so it preinstalls certain tools or python packages, but it looked like most algorithms worked out of the box so I skipped additional configuration.


poyea:
LGTM. @AnupKumarPanwar What do you think?

PR #867, opened 2019-06-02
Correct the wrong iterative DFS implementation

ZJUGuoShuai:
The old iterative DFS is wrong. The order it explores the graph is not in a DFS way. I think my code is correct.

Reference:
- <https://www.techiedelight.com/depth-first-search/>
- <https://www.geeksforgeeks.org/iterative-depth-first-traversal/>


poyea:

PR #869, opened 2019-06-02
Add solutions to Project Euler problem 56, 58, 63

nehamsoni:


poyea:
Thanks. Only 3 files added? I'll review it later.

project_euler/problem_56/sol1.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import print_function
```
poyea: We don't need this anymore. As stated in our contribution guide, Python means Python 3.x+.

project_euler/problem_56/sol1.py
```diff
@@ -0,0 +1,18 @@
...
+
+try:
+	xrange		#Python 2
+except NameError:
+	xrange = range	#Python 3
```
poyea: Just use `range()`, for the same reason.

project_euler/problem_56/sol1.py
cclauss: https://docs.python.org/3/howto/pyporting.html#prevent-compatibility-regressions

This directive changes how Python 2 operates.  It makes it a syntax error in Python 2 to add a legacy __print__ statement.  This prevents Python 2 developers from breaking Python 3 compatibility.

poyea: Right. If we consider any forward compatibility, this should be here. It doesn't matter a lot as long as it's on the first line.

PR #872, opened 2019-06-04
Removed Graphs and move prim.py to graphs

NielXu:
It is weird that the graphs folder does not show up when I open the project locally, therefore, I modified it directly on GitHub.


VipinJain1:
Great, thanks.

PR #873, opened 2019-06-04
circular_linked_list.py

TheIncredibleHarsh:
A circular linked list where the last node points to the first node. I don't know why it is missing and I think it should be here. Helps in traversing the list in circular order and we don't have to start traversing from head every time


TheIncredibleHarsh:
Why is the LGTM check failing? It gives error on some other file, I'm sorry but I have never worked with it before

cclauss:
LGTM tests are failing because of the Python syntax error in #425

PR #880, opened 2019-06-06
Fix the Python Syntax Error that is causing LGTM to fail

cclauss:
Fixes #425

The __elif__ in the current solution will __never__ be executed.

### DIFF
--- a/Project Euler/Problem 01/sol5.py
+++ b/Project Euler/Problem 01/sol5.py
@@ -1,3 +1,7 @@
+#!/usr/bin/env python3
+
 """Find the sum of all the multiples of 3 or 5 below 1000."""
+
+# https://projecteuler.net/problem=1
 
 print(sum(i if 0 in (i % 3, i % 5) else 0 for i in range(1000)))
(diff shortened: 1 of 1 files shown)


StephenGemin:
I'm seeing the same thing for a pull request I submitted yesterday.  I made the update in my branch, but it won't it still fails the LGTM check.

Elif statement is in Project Euler/Problem 01/sol5.py

cclauss:
The folks at LGTM tried to explain it to me but I do not understand their explanation.  A real test runner would be preferable to LGTM.

https://discuss.lgtm.com/t/tests-are-not-automatically-run-when-lgtm-yml-is-modified/1446/5

shuklalok:
The solution, too, is incorrect. Problem statement is "Find the sum of all the multiples of 3 or 5 below 1000."

cclauss:
```python
print(sum(i if 0 in (i % 3, i % 5) else 0 for i in range(1000)))
233168
```

poyea:
Thanks again, this is done by #882. I've pushed this to that PR.

PR #885, opened 2019-06-08
added project euler problem 56,58,63,81,82,83

nehamsoni:
done all the changes according to guidelines
sorry for inconvenience i am little new to github 😉😁


nehamsoni:
pls suggest the changes if needed 🙄

cclauss:
Is this the replacement of #869?  If so, please edit the title of this PR to be the same as #869 so that the title of the PR tells us what it is trying to achieve.  🙏

cclauss:
One thing that would be really helpful with these Euler solutions would be a comment that states the problem to be solved at the top of each file.

nehamsoni:
> One thing that would be really helpful with these Euler solutions would be a comment at the top of each file that states the problem to be solved.

do you mean at top of each file i should specify
`# problem number <81> project Euler`

cclauss:
https://projecteuler.net/problem=1

nehamsoni:
done anything else...!!🙂

nehamsoni:
added files for problem 82 and 83 aswell🙃

project_euler/problem_81/sol1.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+
+matrix=[]
+fi=open('input.txt','r')
+for line in fi:                         #.   matrix1 is used to calculate
+    line=line.split(',')                #.   d{} which counts repeated node's frequency
+    line=list(map(int, line))
+    matrix.append(line)
+fi.close()
```
cclauss: Replace seven lines with two:
```python
with open('input.txt') as in_file:
    matrix = [[int(x) for x in line.split(',')] for line in in_file]
```

nehamsoni: done.

project_euler/problem_81/sol1.py
```diff
@@ -36,6 +36,7 @@ def dijkstra(initial,graph):
 
 with open('input.txt') as in_file:
     matrix = [[int(x) for x in line.split(',')] for line in in_file]
+in_file.close()
```
cclauss: __with open() as__ removes the need to call __.close()__.

Also one line was deleted in these changes that needs to be re-added.

nehamsoni: oh yess i hope its good now...😉

cclauss: Nice!  Thanks.

PR #886, opened 2019-06-08
Remove the Python Syntax Error that is causing our build to fail

cclauss:
As discussed at https://github.com/TheAlgorithms/Python/pull/425#issuecomment-497402836 this file contains syntax, logic, style, and naming errors that are causing our LGTM build process to fail.  Line 3 is a Python syntax error, the logic does not solve the problem, and the file is in the wrong directory.

@StephenGemin, @shuklalok, @vikaskyadav, @poyea, @mrWiecek your reviews please.

### DIFF
--- a/Project Euler/Problem 01/sol5.py
+++ b/Project Euler/Problem 01/sol5.py
@@ -1,8 +1,0 @@
-a=3
-result=0
-while a=<1000:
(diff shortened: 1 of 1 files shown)


poyea:
This is done by #882. Thanks.

PR #888, opened 2019-06-08
Enable automated testing with CircleCI

cclauss:


AnupKumarPanwar:
@cclauss I recommend using Travis CI only and avoiding multiple CI tools.

PR #892, opened 2019-06-09
Remove punctuation and control characters

StephenGemin:
Removes punctuation from string.punctuation.  The dash is replaces with a whitespace to preserve hyphenated words.  There is an option to remove control characters as well.

I've used this many times when searching for word frequencies, etc.  Not sure if it belongs with all these algorithms, but it's an "algorithm" to remove punctuation :D

To test, look at the tests folder and run test_punctuation.py


strings/parse_punctuation.py
```diff
@@ -0,0 +1,39 @@
+# parse_punctuation.py
+
+import codecs
+import string
+
+
+def remove_punc(path, remove_ctrl_char=False):
```
cclauss: This looks really cool!

Please add type hints on this principle function: __def remove_punc(path: str, remove_ctrl_char: bool=False)-> str:__

Issue #837, opened 2019-05-25
Matrix multiplication fails for matrix that are not Square matrix


sandeepsj:
Currently The Matrix Multiplication function can multiply only square matrices
current code is
`def multiply(matrix_a, matrix_b):
    matrix_c = []
    n = len(matrix_a)
    for i in range(n):
        list_1 = []
        for j in range(n):
            val = 0
            for k in range(n):
                val = val + matrix_a[i][k] * matrix_b[k][j]
            list_1.append(val)
        matrix_c.append(list_1)
    return matrix_c
`
~~~~~~~        ~~~~~~~
This could be changed to
 def multiply(matrix_a, matrix_b):
    matrix_c = []
    n = len(matrix_a)       #no of rows in Mat_a
    m = len(matrix_a[0])    #no of cols in Mat_a
    l = len(matrix_b[0])    #no of cols in Mat_b
    for i in range(n):
        list_1 = []
        for j in range(l):
            val = 0
            for k in range(m):
                val = val + matrix_a[i][k] * matrix_b[k][j]
            list_1.append(val)
        matrix_c.append(list_1)
    return matrix_c

paupaulaz:
Hey, I would love to take this issue and solve it as my first contribution :)

cclauss:
Is this fixed?  Can this issue be closed?

PR #898, opened 2019-06-11
Fixed error in matrix multiplication that didn't multiple non-square …

dhruvsharma78:
…matrices. Also added checks for invalid dimensions

The original code for matrix multiplication did not support non square matrices (reference to #837 )

The function also did not check if the 2 matrices can even be multiplied. 

The error has been fixed, and the check has been added


matrix/matrix_multiplication_addition.py
```diff
@@ -15,12 +15,19 @@ def scalarMultiply(matrix , n):
 
 def multiply(matrix_a, matrix_b):
     matrix_c = []
```
StephenGemin: with numpy arrays, no need to define matrix_c at the beginning of the function.  Also, since I can't access the beginning of this module would need to add "import numpy as np".

Below is a suggestion to reduce the number of code lines.  Maybe I'm missing the point of this module, but using numpy would be a lot more efficient.  I know the rest of the doc has no docstring, but I figured I would add one in the suggestion.

You can the entire function with just 8 lines of code.

```suggestion
    """
    :param matrix_a: matrix as a list or array, 1st matrix that defines #columns for matrix rule [a1 x a2]
    :type matrix_a: list of int or list of float
    :param matrix_b: matrix as a list or array, 2nd matrix that defines #rows for matrix rule [b1 x b2]
    :type matrix_b: list of int or list of float
    :return: numpy array [a1 x b2]
    """
    matrix_a = np.array(matrix_a)
    matrix_b = np.array(matrix_b)
    
    try:
        matrix_c = np.dot(matrix_a, matrix_b)
        return matrix_c
    except ValueError:
        raise
```

dhruvsharma78: This definitely works. In fact, a lot of stuff in this project can be done using in-built/library functions, but I think the idea of this project is to show how the core algorithms work.

Using ```numpy.dot``` would definitely make it work in lesser lines of code, and possibly even be faster, but that level of abstraction would hide what exactly is happening.

I may be wrong though, and if what we're going for is smaller code and speed, your suggestion would work (Although I think ```numpy.matmul``` would be a better choice)

StephenGemin: That makes sense.  You're right on both points ☺.

I never heard of numpy.matmul, now I know.
https://www.numpy.org/devdocs/reference/generated/numpy.dot.html

TLDR: If both a and b are 2-D arrays, it is matrix multiplication, but using matmul or a @ b is preferred.

Appreciate the feedback!

Issue #840, opened 2019-05-26
New Algorithms required in Machine Learning


siddharthjain1611:
Need to add polynomial regression, SVM, SVR, and other important algorithms.

StephenGemin:
There will be multiple updates to follow.  I'd like to pool any regression analysis into a separate module.  It doesn't make sense to put these under machine learning as they are used outside of machine learning too.

anand372:
is this still open i would like to  be assigned for implementing and posting of SVM's?

cclauss:
You do not need to wait for assignment...  Just look though https://github.com/TheAlgorithms/Python/blob/master/DIRECTORY.md and if you see a hole then submit a pull request to fill it.

cclauss:
@siddharthjain1611 What still needs to be done, or can this issue be closed?

QuantumNovice:
#1609

siddharthjain1611:
Yes we can close it.

PR #901, opened 2019-06-14
least squares poly regression

StephenGemin:
First of multiple submissions for Issue #840 

Regression class
1. least squares regression of nth order
3. output polynomial coefficients
2. plot result


StephenGemin:
For the commits after the force-push


**Within the matrix folder:**
1. Removed init.py
2. Added pytests for matrix operations as it was difficult to see if things were working correctly just running it through print statements on the main file. There were edge cases that the algorithms did not account for.

**Made the following changes to matrix_operation.py**
1. added matrix subtraction to matrix_operation.py
2. added matrix size checks for addition and subtraction as there were previously no checks
3. fixed typo in matrix multiplication loop that was in Pull Request TheAlgorithms#898 on TheAlgorithms/Python
4. PEP8 styling changes

PR #906, opened 2019-06-17
add a generic heap

SaurabhGoyal:
- Adds a heap which is generic and can be made Min or max using just a key function.
- Also easier to read and understand because of modularity of functions.
- Also adds `update_item` and `delete_item` functionalities.


data_structures/Heap/heap_generic.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def extract_top(self):
+        """Returns top item tuple (Calculated value, item) from heap and removes it as well if present"""
+        top_item_tuple = self.get_top()
+        if top_item_tuple:
+            self.delete_item(top_item_tuple[0])
+        return top_item_tuple
+
+
+# # Usage
```
cclauss: Can we replace the usage with [__doctests__](https://docs.python.org/3/library/doctest.html) so that we can run __python3 -m doctest -v data_structures/Heap/heap_generic.py__ to understand if the code is working as expected?

PR #914, opened 2019-06-23
better implementation for midpoint

wuminbin:


poyea:
Can you explain the method, with certain examples?

wuminbin:
@poyea 
In File "arithmetic_analysis/bisection.py", I make sure that the type of midpoint is always float by divided by float, while the origin code can't work well when using Python2.

In both File("arithmetic_analysis/bisection.py" and "searches/binary_search.py"), I change the algorithm of midpoint from "(a+b) / 2" to "a+(b-a)/2" in case the result of a plus b is to large.

PR #916, opened 2019-06-26
[WIP] Adding poly regression

StephenGemin:
For Issue #840

Regression class
--least squares regression of nth order
--output polynomial coefficients
--plot result

I'm re-entering this because I made a royal mess of #901.  Now I'm just entering changes for the poly regression and keeping other changes separate.

### DIFF
--- a/regression/regr.py
+++ b/regression/regr.py
@@ -1,0 +1,119 @@
+"""
+    References
+    ------------
+    https://en.wikipedia.org/wiki/Polynomial_regression
+"""
+
+import numpy as np
+from scipy import linalg
+import matplotlib.pyplot as plt
+from matplotlib import style
+style.use('seaborn')
+
+# from matrix import matrix_operation as mat_op
+
+
+class Regression:
+    """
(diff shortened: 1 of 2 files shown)


regression/regr.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def the_algorithm_ls_reg(self, order=1):
+        """
+        :param order: nth order polynomial
+        :type order: int
+        :return: ndarray
+
+
+        _____________________________________________
+        Still under development
```
AnupKumarPanwar: If the algorithm is under progress, please mention that in the commit title by prefixing  **[WIP]**  to the commit message or apply the corresponding tag.

StephenGemin: @AnupKumarPanwar  made the change, thanks for the tip.  This will probably stay in development as it depends on another PR #925.

cclauss:
Is this unblocked now that #925 has landed?

StephenGemin:
Yes it is, thank you!

PR #925, opened 2019-06-27
Add error & test checks for matrix_operations.py

StephenGemin:
1. Included error checks in matrix operation. There were some cases where the functions would not work correctly.

2. PEP8 changes to matrix_operations.py.

3. Added test cases for matrix operations to improve error checking.


StephenGemin:
Would need this reviewed/approved in order to implement #916

matrix/tests/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    mat_ops: tests for matrix operations
```
cclauss: Please consider adding a carriage return at the end of this file.  It is a good idea to have a carriage return as the last line of your Python file but my real interest is beyond that.  This change will kick off a new Travis CI test and pytest is now a part of that testing and I am interested to see Travis finds, runs, and passes your tests.

PR #930, opened 2019-06-29
adding factorial

jpg-130:
calculating factorial using memoization added

### DIFF
--- a/dynamic_programming/factorial.py
+++ b/dynamic_programming/factorial.py
@@ -1,0 +1,19 @@
+#Factorial of a number using memoization
+def factorial(num):
+    if num<0:
+        return -1
+    if result[num]!=-1:
+        return result[num]
+    else:
+        result[num]=num*factorial(num-1)
+        #uncomment the following to see how recalculations are avoided
+        #print(result)
+        return result[num]
+
+#factorial of num
+result=[-1]*10
+result[0]=result[1]=1
+print(factorial(5))
+#uncomment the following to see how recalculations are avoided
+# print(factorial(3))
+# print(factorial(7))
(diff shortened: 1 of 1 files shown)


cclauss:
Could you please add a doctest or two?

jpg-130:
It is showing that the travis-ci build failed, however, it works well on the local machine. What changes should I incorporate?

cclauss:
https://travis-ci.org/TheAlgorithms/Python/builds/558476077#L705
Think like a compiler would.  Is the variable spelled correctly?  Is it in scope?

cclauss:
Need another hint?

jpg-130:
@cclauss I'll try.

dynamic_programming/factorial.py
```diff
@@ -0,0 +1,35 @@
...
+##result[0]=result[1]=1
+##print(factorial(5))
+# print(factorial(3))
+# print(factorial(7))
+
+
+if __name__ == "__main__":
+    import doctest
+    result=[-1]*10
+    result[0]=result[1]=1
```
cclauss: Hint: Move lines 33 and 34 ___inside___ the __factorial()__ function.

jpg-130: Ok but if I do that the values are recomputed every time the function is called i.e. I don't want to recompute 3! if 7! is already computed.

cclauss: Understood.  Then put those two lines at the __top__ of the file at _global scope_ __before the function and the __main()__

jpg-130: Thank You!

PR #931, opened 2019-06-30
adding RSA using user defined functions

jpg-130:
User-defined functions are used instead of inbuilt functions.

### DIFF
--- a/ciphers/rsa.py
+++ b/ciphers/rsa.py
@@ -1,0 +1,60 @@
+def multiplicative_inverse(a, m) : 
+    a = a % m; 
+    for x in range(1, m) : 
+        if ((a * x) % m == 1) : 
+            return x 
+    return 1
+
+def convert_list_to_string(l):
+    l = map(str, l)
+    string = ""
+
+    return string.join(l)
+p,q=11,13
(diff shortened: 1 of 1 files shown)


ciphers/rsa.py
```diff
@@ -0,0 +1,60 @@
+def multiplicative_inverse(a, m) :
```
cclauss: Please add one or more [__doctests__](https://docs.python.org/3/library/doctest.html) for this function.

ciphers/rsa.py
```diff
@@ -0,0 +1,60 @@
...
+    for x in range(1, m) : 
+        if ((a * x) % m == 1) : 
+            return x 
+    return 1
+
+def convert_list_to_string(l):
+    l = map(str, l)
+    string = ""
+
+    return string.join(l)
```
cclauss: Let's simplify to a one liner: __return "".join(str(item) for item in l)__ and then please add a doctest.

Please indent the remaining code underneath __if \_\_name\_\_ == "\_\_main\_\_": so that [our automated testing](https://travis-ci.org/TheAlgorithms/Python/builds/561354685#L608) does not run all of this code when it executes the pytests.

PR #940, opened 2019-07-02
Create Outliers

AugustofCravo:
Added algorithm that prints outliers and strong outliers from a list.

### DIFF
--- a/maths/Outliers
+++ b/maths/Outliers
@@ -1,0 +1,85 @@
+import statistics as st
+
+def OutliersUp(list):
+    """
+    Prints the upper outilier of a list.
+    Ex.: list = [1,2,3,4,5,30]
+    Upper Outilier = 30
+    """
+    print ("Upper Outlier: ")
+    for i in list:
+        if i > ThirdQuartile(list)+1.5*InterQuartileAmplitude(list):
+            print (i)
+
+def OutliersDown(list):
+    """
+    Prints the lower outilier of a list.
+    Ex.: list = [20,100,200,300,400]
+    Lower Outilier = 20
+    """
+    print ("Lower Outlier: ")
+    for i in list:
+        if i < FirstQuartile(list)-1.5*InterQuartileAmplitude(list):
+            print (i)
+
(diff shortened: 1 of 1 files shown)


maths/Outliers
```diff
@@ -0,0 +1,85 @@
...
+def OutliersUp(list):
+    """
+    Prints the upper outilier of a list.
+    Ex.: list = [1,2,3,4,5,30]
+    Upper Outilier = 30
+    """
+    print ("Upper Outlier: ")
+    for i in list:
+        if i > ThirdQuartile(list)+1.5*InterQuartileAmplitude(list):
+            print (i)
```
cclauss: Let me rewrite this function with a few optional suggestions:
```python
from typing import List


def OutliersUp(numbers: List[int]) -> None:
    """
    Prints the upper outilier of a list.
    >>> OutliersUp([1, 2, 3, 4, 5, 30])
    Upper Outlier: 30
    """
    threshold = ThirdQuartile(numbers) + 1.5 * InterQuartileAmplitude(numbers)
    for i in numbers:
        if i > threshold:
            print("Upper Outlier: {}".format(i))
```
This provides Python typehints to say what kind of function parameters we expect and what this function will return (nothing).  We avoid using __list__ as a variable name because there is already a Python builtin using that name.  Added two [__doctest__](https://docs.python.org/3/library/doctest.html) lines at the end of the function's docstring so that our [automated testing](https://github.com/TheAlgorithms/Python/blob/master/.travis.yml#L14) can then run the first line and raise an error if the result does not match the second line.  For both performance and readability reasons, we only calculate the __threshold__ only once instead of multiple times.  We lose the space between __print__ and __(__ because __print()__ is now a function like all other Python functions.  And we use __str.format()__ to print out our results on just one line.

cclauss:
Please change the filename to lowercase with a __.py__ extension: __outliers.py__

cclauss:
@AugustofCravo Can you please work on getting the tests to pass so that we can merge?

PR #942, opened 2019-07-03
Create converte_binario_decimal.py

Dhandarah:
creates a binary number conversion function for decimals.

### DIFF
--- a/converte_binario_decimal.py
+++ b/converte_binario_decimal.py
@@ -1,0 +1,17 @@
+def converterb_d(n):
+    decimal = 0
+    n = str(n)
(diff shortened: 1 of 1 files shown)


PatOnTheBack:
@Dhandarah Would it be better to create this program in English since it is a more widespread language?

The text could be changed from:

- "Digite o número binário" to "Enter a binary number"
- "O número binário {} corresponde a {} em decimal" to "The binary number {} corresponds to {} in decimal"
- "converte_binario_decimal.py" to "binary_to_decimal.py"

The variable `tam` might also need an English translation.

PR #961, opened 2019-07-05
Generate all subsequences using backtracking

Erfaniaa:


Erfaniaa:
> Hi, why there is no files changed?
> image

@siriak 
There was a mistake. I added a new file, please check it.

PR #884, opened 2019-06-07
Directory.md

jaspalsb:
Directory that was suggested in #860

PR #963, opened 2019-07-06
Travis CI: Build Directory.md

cclauss:
As suggested at https://github.com/TheAlgorithms/Python/pull/884#issuecomment-508688850

@poyea Please consider a better name for this script like __build_directory_md.py__ or __scripts/build_directory_md.py__.  Putting a ~ in a filename on Linux/macOS is not a good idea because ~ is the home directory.


cclauss:
Closed in favor of #964

PR #965, opened 2019-07-06
Fix bracktrack time complexity

Erfaniaa:
@poyea @siriak please review and merge this one.


Erfaniaa:
@AnupKumarPanwar

backtracking/all_permutations.py
```diff
@@ -2,7 +2,7 @@
 	In this problem, we want to determine all possible permutations
 	of the given sequence. We use backtracking to solve this problem.
 
-	Time complexity: O(n!),
+	Time complexity: O(n! * n),
```
siriak: Why the complexity is `O(n! * n)`?

Erfaniaa: At least, we print the permutations. Each permutation size is n.

AnupKumarPanwar: Fixing a character at the 0th index, recursion will run n! times. And this will be repeated for every character of the string. So O(n! * n). Makes sense?

Erfaniaa: For the last character, I mean the (n-1)th one, we iterate through numbers from 0 to n-1.
So for each n! states, our for loop takes O(n) time.

siriak: I'd agree on this. But what about generating subsequences? We don't have any loops there.

Erfaniaa:
> LGTM
@AnupKumarPanwar 
Please don't merge it now.
This commit contains two changes.
I'm not sure about the time complexity of all_subsequences. Is it O(2^n) or O(2^n * n)? What's your idea?

backtracking/all_permutations.py
Erfaniaa: I guess you're right.
T(n) = 2 * T(n - 1) + 1, so T(n) equals 2^n - 1.
But I think it's better to write O(2^n + n), because we fill the answer list.
And one another thing: how can I commit only one file changes?

siriak: Anyway, O(2^n + n) = O(2^n)

backtracking/all_subsequences.py
```diff
@@ -2,7 +2,7 @@
 	In this problem, we want to determine all possible subsequences
 	of the given sequence. We use backtracking to solve this problem.
 
-	Time complexity: O(2^n),
+	Time complexity: O(2^n * n),
```
siriak:
```suggestion
	Time complexity: O(2^n),
```

AnupKumarPanwar: Right this should be O(2^n)

fty48jj19:
e  eu ja disse para vc nao me obrig f toma um atitude sobre vc vc esta
enpondo algo que nao quero dessa maneira vc eu e um contribuite vou espulsa
vc coloca outro no seu lugar  estou te avisando ,

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Livre
de vírus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>.
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Em sáb, 6 de jul de 2019 às 12:35, Anup Kumar Panwar <
notifications@github.com> escreveu:

PR #968, opened 2019-07-07
Simplied password_generator.py

qckzr:
- Added main() function
- Simplified password_generator
- Formatted file according to pep 8


other/password_generator.py
```diff
@@ -2,34 +2,45 @@
...
-print('[ If you are thinking of using this passsword, You better save it. ]')
+def password_generator(max_length):
+    letters = [letter for letter in string.ascii_letters]
+    digits = [digit for digit in string.digits]
+    symbols = [symbol for symbol in string.punctuation]
+    chars = letters + digits + symbols
+    random.shuffle(chars)
+    password = ''.join(random.choice(chars)
+                       for x in range(max_length))
+    return password
```
cclauss: Really cool ideas here!  You sparked my imagination over my morning coffee.  Thanks.

I approve this PR in its current form _but_ what about...
```python
from random import choice
from string import ascii_letters, digits, punctuation

def password_generator(length=8):
    '''
    >>> len(password_generator())
    8
    >>> len(password_generator(length=16))
    16
    >>> len(password_generator(257))
    257
    >>> len(password_generator(length=0))
    0
    >>> len(password_generator(-1))
    0
    '''
    chars = tuple(ascii_letters) + tuple(digits) + tuple(punctuation)
    return ''.join(choice(chars) for x in range(length))
```
Some of the logic behind these suggestions:
* Change __max_length__ to just __length__ because it _completely_ determines the length of the output
* Give __length__ a sensible default value so that users can just fire-and-forget.  Perhaps this value should be higher
* Add doctests so that we can test the code by running __python3 -m doctest -v other/password_generator.py__
* tuples should be faster to create then lists and should take up slightly less memory
* We only need to randomize once.  __random.choice()__ should be a lighter weight operation than __random.shuffle()__
* __password_generator()__ should generate a password so there is no need to create a variable named __password__ only to get rid of it on the following line.

other/password_generator.py
```diff
@@ -2,34 +2,45 @@
...
 
 
 def random_characters(ctbi, i):
-  pass  # Put your code here...
+    pass  # Put your code here...
+
+
+def main():
+    max_length = int(
+        input('Please indicate the max length of your password: '))
```
cclauss: Deal gracefully with leading or trailing whitespace in user input by adding a __.strip()__ call to the end of __.input()__:
```python
input('Please indicate the max length of your password: ').strip()
```

fty48jj19:
e  por favor cada um que uma coisa  a diverssa codigos  a pessao aqui nao
tem modernidade a pessoa antigas nao posso diferecia nada eu jogo os
codigo  nao posso esta seguido nada nao  saiba tem gente que lega mas tem
outras q fica na entençao de me prejudica  com codigos suftware  e outro
codigo que pode me prejudica  entenda isso por favor  nem todos posso ficar
seguindo nao ,

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Livre
de vírus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>.
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Em dom, 7 de jul de 2019 às 12:20, cclauss <notifications@github.com>
escreveu:

PR #971, opened 2019-07-08
convolve and sobel

lighttxu:
add img_convolve algorithm and sobel_filter algorithm


AnupKumarPanwar:
I guess gaussian_filter.py already exists.

Issue #959, opened 2019-07-05
timsort


xu561865:
When I used [2, 4, 5, 3, 1] to test, it just returned [1, 2, 4, 5]

FrogBattle:
Hello,

After some play with the code I reached the following conclusion:
The 3 is left out in the following if clause : 
`        if lst[i] < lst[i - 1]:
            if not new_run:
                runs.append([lst[i - 1]])
                new_run.append(lst[i])
            else:
                runs.append(new_run)
                new_run = []
`
Lines 56-62.
The current index i is not assigned to "new_run" before the same variable is set to empty list, thus the 3 is lost. It is supposed to be added to the list on the next loop iteration, although this is stopped by the first check done here : 
`
        if i == length - 1:
            new_run.append(lst[i])
            runs.append(new_run)
            break
`
Lines 51-54.

If you put the if clause with the break statement last, the program may add some elements to the new_run list twice. This happens as new_run initializes to [lst[0]] and it may happen that it adds some elements twice to different runs.

In general, if you rely on the next loop iteration to add the previous element into a new_run list, it must do it explicitly ( if new_run==[]), which would mean another conditional. Right now, you will not lose an element if and only if the next iterated element is also smaller than the one before him : 
`
        if lst[i] < lst[i - 1]:
            if not new_run:
                runs.append([lst[i - 1]])
                new_run.append(lst[i])
            else:
                runs.append(new_run)
                new_run = []
        else:
            new_run.append(lst[i])
`
Lines 56-64.

Hope this helps.
Alex

cclauss:
Can this issue be closed?

sadeghipakijoo:
thanks...

PR #972, opened 2019-07-08
Update tim_sort.py

FrogBattle:
Update tim_sort.py

The previous algorithm was skipping numbers, according to issue #959, and my own tests.
The version I am applying uses a while loop, which works correctly and is easier to compute, as there is no break statement.
This implementation only builds up increasing runs, without making decreasing ones, which are later inverted.

### DIFF
--- a/sorts/tim_sort.py
+++ b/sorts/tim_sort.py
@@ -42,36 +42,26 @@ def merge(left, right):
 
 
 def tim_sort(lst):
-    runs, sorted_runs = [], []
-    length = len(lst)
-    new_run = [lst[0]]
-    sorted_array = []
-
-    for i in range(1, length):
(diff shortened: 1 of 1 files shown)


cclauss:
How about adding some doctests?
```python
def tim_sort(lst):
    """
    >>> tim_sort("Python")
    ['P', 'h', 'n', 'o', 't', 'y']
    >>> tim_sort((1.1, 1, 0, -1, -1.1))
    [-1.1, -1, 0, 1, 1.1]
    >>> tim_sort(list(reversed(list(range(7)))))
    [0, 1, 2, 3, 4, 5, 6]
    >>> tim_sort([3, 2, 1]) == insertion_sort([3, 2, 1])
    True
    """
```

Issue #973, opened 2019-07-08
Adding Flake8 F4 Tests to find Import Errors


PatOnTheBack:
@PatOnTheBack if you add [__F4__](https://flake8.pycqa.org/en/latest/user/error-codes.html) to the flake8 tests in __.travis.yml__, you will find other import-related issues to fix.  Thanks for your attention to these details.

_Originally posted by @cclauss in https://github.com/TheAlgorithms/Python/pull/920#issuecomment-509011067_

Like cclauss said, adding Flake8 F4 to the Travis file can help us find import errors that reduce code qulaity.

cclauss:
https://github.com/TheAlgorithms/Python/blob/master/.travis.yml#L8

PR #974, opened 2019-07-08
Add Flake8 F4 Tests to .travis.yml

PatOnTheBack:
F4 tests check for import errors. Implements issue #973


cclauss:
Another approach is to do smaller PRs...  Like a PR that just adds F401 to __.travis.yml__ and fixes the three instances of that.  Another PR that adds F403 and fixes the two instance of that.  Small moves.

cclauss:
Should we progress this one or close it?

PatOnTheBack:
I think we should progress. Checking for import errors can be helpful in the future, not just now.

PR #977, opened 2019-07-08
added decimal to hexadecimal conversion

jasper256:
I didn't update the directory because I noticed that the other conversion algorithms were not listed

### DIFF
--- a/conversions/decimal_to_hexadecimal.py
+++ b/conversions/decimal_to_hexadecimal.py
@@ -1,0 +1,44 @@
+""" Convert Base 10 (Decimal) Values to Hexadecimal Representations """
+
+# set decimal value for each hexadecimal digit
+values = {
+    0:'0',
+    1:'1',
+    2:'2',
+    3:'3',
+    4:'4',
+    5:'5',
+    6:'6',
+    7:'7',
+    8:'8',
+    9:'9',
+    10:'a',
+    11:'b',
+    12:'c',
+    13:'d',
+    14:'e',
+    15:'f'
+}
+
+def decimal_to_hexadecimal(decimal):
+    """ take decimal value, return hexadecimal representation as str """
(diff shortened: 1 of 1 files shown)


PatOnTheBack:
If I add this line `print("999098 in hexadecimal is", decimal_to_hexadecimal(999098))` to the `main()` function, I get an error. Does your program not support large numbers?

```
('5 in hexadecimal is', '5')
('15 in hexadecimal is', 'f')
('37 in hexadecimal is', '25')
('255 in hexadecimal is', 'ff')
('4096 in hexadecimal is', '100')
  File "decimal_to_hex.py", line 45, in <module>
    main()
  File "decimal_to_hex.py", line 42, in main
    print("999098 in hexadecimal is", decimal_to_hexadecimal(999098))
  File "decimal_to_hex.py", line 30, in decimal_to_hexadecimal
    hexadecimal = values[remainder] + hexadecimal
KeyError: 235
```

jasper256:
Thank you for bringing this to my attention. After combing through the algorithm again I believe I have found the error that would cause the key to go out of bounds as the number of digits necessary increase. I will fix this as soon as I get home.

jasper256:
The program has now been revised.  I used 999,098 as a test and checked the results with a calculator, which came back as successful.

jasper256:
I honestly don't know much about automated testing, but it's something I'd be happy to look into in the future. I'm currently about to head to the airport for a trip I'm taking, and I won't be able to sit down in front of a computer for a couple weeks. I won't even have internet access through my phone in a few days, so if anyone posts comments just know that I'll get to them when I can in about two weeks.

cclauss:
Cool.  Let's land this one now and you can pick up tests when you return.  Read thru the doctest page on your flight because it is simple but really helpful to have tests on your code.  Safe travels.

PR #981, opened 2019-07-09
Reduced size of notebook file by one third

Blues1998:


project_euler/problem_13/sol2.py
```diff
@@ -0,0 +1,5 @@
+sum = 0
+with open("num.txt",'r') as f:
+    for line in f:
+        sum += int(line)
```
cclauss: Please consider reimplementing as a list comprehension that uses the Python builtin [__sum()__](https://docs.python.org/3/library/functions.html#sum).

cclauss: Python builtin [__divmod__](https://docs.python.org/3/library/functions.html#divmod)

cclauss: Python builtin [__divmod__](https://docs.python.org/3/library/functions.html#divmod)

project_euler/problem_16/sol2.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import print_function
+n = 2**1000
+r = 0
+while n:
+    r, n = r + n % 10, n // 10
```
cclauss: Python builtin [__divmod()__](https://docs.python.org/3/library/functions.html#divmod)

PR #991, opened 2019-07-10
add canny edge detection algorithm and modify sobel_filter

lighttxu:
add canny edge detection algorithm and modify sobel_filter


digital_image_processing/edge_detection/canny.py
```diff
@@ -0,0 +1,103 @@
...
+    Non-maximum suppression. If the edge strength of the current pixel is the largest compared to the other pixels 
+    in the mask with the same direction, the value will be preserved. Otherwise, the value will be suppressed. 
+    """
+    for row in range(1, image_row - 1):
+        for col in range(1, image_col - 1):
+            direction = gradient_direction[row, col]
+
+            if (0 <= direction < 22.5) \
+                    or (15 * PI / 8 <= direction <= 2 * PI) \
+                    or (7 * PI / 8 <= direction <= 9 * PI / 8):
```
cclauss: Can we avoid the backslashes (see PEP8) with something like:
```python
            if (
                0 <= direction < 22.5
                or 15 * PI / 8 <= direction <= 2 * PI
                or 7 * PI / 8 <= direction <= 9 * PI / 8
            ):
```

digital_image_processing/edge_detection/canny.py
```diff
@@ -0,0 +1,103 @@
...
+            if dst[row, col] == weak:
+                if dst[row, col + 1] == 255 \
+                        or dst[row, col - 1] == 255 \
+                        or dst[row - 1, col] == 255 \
+                        or dst[row + 1, col] == 255 \
+                        or dst[row - 1, col - 1] == 255 \
+                        or dst[row + 1, col - 1] == 255 \
+                        or dst[row - 1, col + 1] == 255 \
+                        or dst[row + 1, col + 1] == 255:
+                    dst[row, col] = strong
```
cclauss: Can we avoid the backslashes with something like:
```python
                if 255 in (
                    dst[row, col + 1],
                    dst[row, col - 1],
                    dst[row - 1, col],
                    dst[row + 1, col],
                    dst[row - 1, col - 1],
                    dst[row + 1, col - 1],
                    dst[row - 1, col + 1],
                    dst[row + 1, col + 1],
                ):
```

digital_image_processing/edge_detection/canny.py
lighttxu: Yeah. It looks clear.

digital_image_processing/edge_detection/canny.py
lighttxu: Wow, concise and clear!

PR #995, opened 2019-07-10
Update rod_cutting.py

SandersLin:
A hopefully clearer implementation without dependence on global variables.

### DIFF
--- a/dynamic_programming/rod_cutting.py
+++ b/dynamic_programming/rod_cutting.py
@@ -1,27 +1,19 @@
-### SOLUTION ###
-"""
-Optimal Substructure: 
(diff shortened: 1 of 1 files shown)


cclauss:
Nice!  Please consider adding a doctest.

SandersLin:
Thanks! Added doctests.

cclauss:
Doctests for length = -1, length = 3.2, prices = []

SandersLin:
Modified doctests and type hints. 
I wonder when doing input validation is too much though.  For example, when the type hint and documentation clearly states  prices should be List[int], but some bloke passed in list of strings, the author shouldn't be responsible.

PR #996, opened 2019-07-10
Update 3n+1.py

SandersLin:
Made variable names more meaningful and removed nested functions.

### DIFF
--- a/maths/3n+1.py
+++ b/maths/3n+1.py
@@ -1,19 +1,19 @@
-def main():
-    def n31(a):# a = initial number
-        c = 0
-        l = [a]
-        while a != 1:
-            if a % 2 == 0:#if even divide it by 2
-                a = a // 2
-            elif a % 2 == 1:#if odd 3n+1
-                a = 3*a +1
-            c += 1#counter
-            l += [a]
+def n31(a):# a = initial number
+    counter = 0
+    path = [a]
+    while a != 1:
+        if a % 2 == 0:
+            a = a // 2
+        else:
+            a = 3*a +1
+        counter += 1
+        path += [a]
+    return path , counter
 
-        return l , c
-    print(n31(43))
-    print(n31(98)[0][-1])# = a
-    print("It took {0} steps.".format(n31(13)[1]))#optional finish
+def main():
+    num = 43
+    path , length = n31(num)
+    print("The Collatz sequence of {0} took {1} steps. \nPath: {2}".format(num,length, path))
 
 if __name__ == '__main__':
     main()
--- a/maths/3n+1.py
+++ b/maths/3n+1.py
@@ -1,4 +1,11 @@
-def n31(a):# a = initial number
+def n31(a):
+    """
+    Returns Collatz sequence of number a
+    
+    >>> n31(4)
+    ([4,2,1],3)
+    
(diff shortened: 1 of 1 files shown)


maths/3n+1.py
```diff
@@ -1,19 +1,19 @@
...
-        c = 0
-        l = [a]
-        while a != 1:
-            if a % 2 == 0:#if even divide it by 2
-                a = a // 2
-            elif a % 2 == 1:#if odd 3n+1
-                a = 3*a +1
-            c += 1#counter
-            l += [a]
+def n31(a):# a = initial number
```
cclauss: Readability:  Please add two spaces between the code and the comment.

SandersLin: Sorry, which comment are you referring to ? `# a = initial number` ? Maybe we could do without the comment and choose a better variable name instead of `a`  ;)

cclauss: The PEP8 rule applies to any line in Python that has both code and a comment.

Here is something slick for you to try:
```python
def n31(initial_number):
    """
    >>> n31(43)  # doctest: +ELLIPSIS
    ([43, ...], 29)
    """
    <the rest of the code goes here>
```
And then on the command line do __python3 -m doctest -v maths/3n+1.py__.  This is a doctest and our __.travis.yml__ file guides our automated testing to test your function on each commit.

cclauss:
What should happen on a negative number? a floating point number?

```python
def n31(initial_number):
    """
    >>> n31(43)  # doctest: +ELLIPSIS
    ([43, ...], 29)
    """
```

SandersLin:
Good catch! Added argument validation and type hints.

maths/3n+1.py
```diff
@@ -15,10 +21,10 @@ def n31(a):
             a = 3*a +1
         counter += 1
         path += [a]
-    return path , counter
+    return path, counter + 1
```
cclauss: Would it also work to __return path, len(path)__ ?

maths/3n+1.py
```diff
@@ -1,11 +1,17 @@
-def n31(a):
+from typing import Tuple, List
+
+def n31(a: int) -> Tuple[List[int], int]: 
     """
-    Returns Collatz sequence of number a
-    
+    Returns Collatz sequence of a number
```
cclauss: This comment says that we are only returning the sequence but we are returning the sequence and its length.

maths/3n+1.py
SandersLin: Yes, that is more elegant, thanks!

maths/3n+1.py
SandersLin: updated docstring

maths/3n+1.py
```diff
@@ -1,19 +1,30 @@
...
+            a = a // 2
+        else:
+            a = 3*a +1
+        path += [a]
+    return path, len(path)
+
+def main():
+    num = 4
+    path , length = n31(num)
+    print("The Collatz sequence of {0} took {1} steps. \nPath: {2}".format(num,length, path))
```
cclauss: You can drop the 0, 1, and 2 here and drop the 0 in the __raise__ statements above.

PR #998, opened 2019-07-10
Update average_median.py

SandersLin:
added doctest, fixed `TypeError: list indices must be integers or slices, not float` error due to` number/2` producing `float` type as index.


cclauss:
Just move the doctests to the __end__ of the docstring

PR #1004, opened 2019-07-11
Update abs_Min.py

SandersLin:
from Maths.abs import absVal does not work.


cclauss:
Can you please rename the file to be all lowercase?

AnupKumarPanwar:
@SandersLin Please resolve the conflicts

maths/abs_min.py
```diff
@@ -1,4 +1,4 @@
-from Maths.abs import absVal
+from abs import absVal
```
AnupKumarPanwar: absVal doesn't exist in abs.py

SandersLin: This should resolve the conflicts. Thanks

PR #1007, opened 2019-07-12
Adding log_likelihood

rishu2403:
Calculating the Log-Likelihood
The log-likelihood can be viewed as a sum over all the training data.

### DIFF
--- a/machine_learning/logistic_regression.py
+++ b/machine_learning/logistic_regression.py
@@ -27,6 +27,11 @@ from sklearn import datasets
 def sigmoid_function(z):
     return 1 / (1 + np.exp(-z))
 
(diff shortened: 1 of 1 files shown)


machine_learning/logistic_regression.py
```diff
@@ -55,6 +62,12 @@ def logistic_reg(
         J = cost_function(h, y)
 
         iterations += 1  # update iterations
+        
+         if step % 10000 == 0:
```
cclauss: The Travis CI tests fail because there is one space too many here.  Python is picky about indentation.

poyea:
This PR seems to be outdated. Thank you.

PR #1025, opened 2019-07-15
Fix error message and format with python/black

cclauss:
@SandersLin Your review please?

### DIFF
--- a/dynamic_programming/climbing_stairs.py
+++ b/dynamic_programming/climbing_stairs.py
@@ -1,27 +1,42 @@
+#!/usr/bin/env python3
+
+
 def climb_stairs(n: int) -> int:
-  """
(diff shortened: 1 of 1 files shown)


dynamic_programming/climbing_stairs.py
```diff
@@ -1,27 +1,42 @@
+#!/usr/bin/env python3
```
poyea: LGTM. The shebang line can be removed btw. Thank you!

cclauss: On Macs and Linux the shebang line enables an effective shorthand.  It’s presence allows us to execute the script with __./climb*__  so why do we want to type more that that?

poyea: I know. Alright, perhaps we could make a note on this in the guidelines later.

PR #1029, opened 2019-07-17
Added Burrows-Wheeler transform algorithm.

brunohadlich:
I added a docstring with parameters and return explanation but I'm not sure what is the right way to write it, I looked at google for a reference and only found PEP 257 that does not define exactly how these things should be written in a strict way.

I also found this [stack overflow question](https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format) that talks about PEP 287 and how reStructuredText (reST) format is used by [Sphinx](http://www.sphinx-doc.org/en/master/) to generate documentation and seems to be a good approach.

I think it would be nice to adopt a style to standardize the project documentation. What you guys think?


compression/burrows_wheeler.py
```diff
@@ -0,0 +1,165 @@
...
+    >>> all_rotations("panamabanana")
+    ['panamabanana', 'anamabananap', 'namabananapa', 'amabananapan',\
+ 'mabananapana', 'abananapanam', 'bananapanama', 'ananapanamab',\
+ 'nanapanamaba', 'anapanamaban', 'napanamabana', 'apanamabanan']
+    >>> all_rotations(5)
+    Traceback (most recent call last):
+        ...
+    TypeError: The parameter string type must be str.
+    """
+    if not (type(string) is str):
```
cclauss: PEP8 says
> Object type comparisons should always use __isinstance()__ instead of comparing types directly.

[__flake8 E721__](https://lintlyci.github.io/Flake8Rules) will find these automatically.

compression/burrows_wheeler.py
```diff
@@ -0,0 +1,165 @@
...
+
+def all_rotations(string):
+    """
+    :param str string: The string that will be rotated len(string) times.
+    :return: A list with len(string) rotations of the parameter string.
+    :rtype: list[str]
+    :raises TypeError: If the string parameter type is not str.
+    Examples:
+
+    >>> all_rotations("^BANANA|")
```
cclauss: `>>> all_rotations("^BANANA|")  # doctest: +NORMALIZE_WHITESPACE`

This allows you to lose the backslashes and indent the way your want because PEP8 says:
> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

compression/burrows_wheeler.py
```diff
@@ -0,0 +1,165 @@
...
+    >>> bwt_transform(4)
+    Traceback (most recent call last):
+        ...
+    TypeError: The parameter string type must be str.
+    >>> bwt_transform('')
+    Traceback (most recent call last):
+        ...
+    ValueError: The parameter string must not be empty.
+    """
+    if not (type(string) is str):
```
cclauss: isinstance()

compression/burrows_wheeler.py
```diff
@@ -0,0 +1,165 @@
...
+    ordered_rotations = [""] * len(bwt_string)
+    for x in range(len(bwt_string)):
+        for i in range(len(bwt_string)):
+            ordered_rotations[i] = bwt_string[i] + ordered_rotations[i]
+        ordered_rotations.sort()
+    return ordered_rotations[idx_original_string]
+
+
+if __name__ == "__main__":
+    string = input("Provide a string that I will generate its BWT transform: ")
```
cclauss: Consider doing __input().strip()__ to remove leading and/or trailing whitespace from user input.

compression/burrows_wheeler.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+    original_string = reverse_bwt(
+        result["bwt_string"], result["idx_original_string"]
+    )
+    print(
+        (
+            "Reversing Burrows Wheeler tranform for entry '{}' we get original"
+            " string '{}'"
+        ).format(result["bwt_string"], original_string)
+    )
```
cclauss: When the line wraping gets to be too much, I usually write:
```python
fmt = Reversing Burrows Wheeler tranform for entry '{}' we get original string '{}'"
print(fmt.format(result["bwt_string"], original_string))
```
But do what feels right to you...  Awesome work here!

brunohadlich:
@cclauss my last commit takes into account your last comments. Thanks they were really helpful. If everything is ok I think it can be merged.

PR #1030, opened 2019-07-17
created a playfair cipher

rishu2403:
Cryptographic playfair cipher using python


harshildarji:
Hey @rishu2403,
We already have an implementation of Playfair cipher [here](https://github.com/TheAlgorithms/Python/blob/master/ciphers/playfair_cipher.py).
Please take a look at it.
Thank you.

PR #1033, opened 2019-07-17
Fix Fisher-Yates shuffle

obelisk0114:
The old version of Fisher-Yates shuffle is wrong.

    def FYshuffle(LIST):
        for i in range(len(LIST)):
            a = random.randint(0, len(LIST)-1)
            b = random.randint(0, len(LIST)-1)
            LIST[a], LIST[b] = LIST[b], LIST[a]
        return LIST

If the probability that the second element goes to second last position is 1/n, it's the ideal case.
However, it's not 1/n in this case.

    The probability is P(choose the second element) * P(choose the second last element)
    = 1/n * 1/n * Sigma(others)
    But Sigma(others) < n in this case.

Correct me if I am wrong.


cclauss:
Are we OK if __a == b__?  I think so but it is kind of a noop and might effect the Probability calculation.

obelisk0114:
> Are we OK if **a == b**? I think so but it is kind of a noop and might effect the Probability calculation.

It's OK if **a == b**. The ideal probability 1/n means every element has equal probability to go to every position (including current position).

QuantumNovice:
I agree @obelisk0114. 
This is the correct algorithm.
`
def FYshuffle(LIST):
    for a in range(len(LIST)):
        b = random.randint(0, len(LIST)-1)
        LIST[a], LIST[b] = LIST[b], LIST[a]
    return LIST
`

poyea:
Migrate to #1186, closing.

Issue #1024, opened 2019-07-15
Fix the LGTM issues


cclauss:
The LGTM service is run by Travis CI on all pull requests and it currently reports:
```
2 Errors
12 Warnings
10 Recommendations
```
https://lgtm.com/projects/g/TheAlgorithms/Python/alerts/?mode=list

Please try to fix ___just a few___ of these issues in a pull request.

For speed of review, please do not try to do them all.  Where possible add doctests as well.

cclauss:
@epicalyx, @harshildarji, @rishu2403 Could one of you please consider fixes to unused variables in logistic_regression?  They might be false positives or they might be worth fixing.

AmritK10:
I would like to work on the issue of unused variables in logistic_regression.

cclauss:
Go for it!

denis-trofimov:
Gonna handle an error in root/matrix/matrix_operation.py

denis-trofimov:
> hello can someone help me solve this :

Hi @tchatec I guess you should create new issue. It is the wrong place to ask such question.

PR #1034, opened 2019-07-18
Fixes for issue "Fix the LGTM issues #1024"

brunohadlich:


project_euler/problem_03/sol1.py
```diff
@@ -28,18 +28,22 @@ def isprime(no):
...
-    
+
     >>> solution(13195)
     29
     >>> solution(10)
     5
     >>> solution(17)
     17
+    >>> solution(0)
+    -1
```
cclauss: Please consider adding a negative integer, a floating point number, an empty list, and a string.

cclauss:
Related to #1024

project_euler/problem_03/sol1.py
brunohadlich: My last commit adds code to handle such situations as well as doctests.

cclauss:
Test failure

PR #1035, opened 2019-07-18
Simplify the pytest command by using --ignore options

cclauss:
Let's see where our doctests are still broken... 12 files must currently be ignored.

Switch pytest from a whitelist to a blacklist...

@poyea I would be very interested to get your review on this PR so that we can make sure that we progress thru these remaining 12 files and so that new pull requests do not add new issues for running doctests.

We are now have [> 125 doctest tests](https://travis-ci.org/TheAlgorithms/Python/builds/561236433#L750) that run on each pull request.  I would like us to be able to do that with just __pytest . --doctest-modules__


cclauss:
Done elsewhere.

PR #1047, opened 2019-07-19
scripts/build_directory_md.py: Add more ignores and ignores_exts

cclauss:
Also ignores all paths that start with a . (dot).  The output (with blank lines removed) is in ~#1046~ #1072.


cclauss:
Closing in favor to #1076

PR #1051, opened 2019-07-20
Adds Gaussian Function in maths section

QuantumNovice:


cclauss:
This is cool!  In this PR, please focus only on adding the Gaussian function.  Please drop the __.md__ files that are not directly tied to the Gaussian function.

In general, we have the autogenerated __DIRECTORY.md__ for users to find algorithms so if the __.md__ files do not contain explanations, we would rather not have them.

Awesome that you have added doctests but how about a few more? Both good parameters and bad parameters: Large numbers, negative numbers, strings, lists.

QuantumNovice:
> 
> 
> This is cool! In this PR, please focus only on adding the Gaussian function. Please drop the **.md** files that are not directly tied to the Gaussian function.
> 
> In general, we have the autogenerated **DIRECTORY.md** for users to find algorithms so if the **.md** files do not contain explanations, we would rather not have them.
> 
> Awesome that you have added doctests but how about a few more? Both good parameters and bad parameters: Large numbers, negative numbers, strings, lists.
Thanks for consideration. On it.

cclauss:
All submissions will be tested with [__mypy__](http://www.mypy-lang.org) so we encourage to add [__Python type hints__](https://docs.python.org/3/library/typing.html) where it makes sense to do so.

cclauss:
OPTIONAL: Extra credit if you run your code through [__python/black__](https://github.com/python/black) for readability.

OPTIONAL: You can also try to minimize parens:
```python
    return 1 / sqrt(2 * pi * sigma ** 2) * exp(-(x - mu) ** 2 / 2 * sigma ** 2)
```

cclauss:
Lowercase directory and Python filenames please:
* __Maths/Gaussian.py__ --> __maths/gaussian.py__

cclauss:
Output did not match...
* https://travis-ci.org/TheAlgorithms/Python/builds/561368477#L681

QuantumNovice:
The test failed because in Python 3.7.3 
The exception message is 'Result too large'
`OverflowError: (34, 'Result too large')`
where as in python 3.7.1 the exception string is 
`OverflowError: (34, 'Numerical result out of range')`
My fork is too old. I am refreshing my updating my fork to the latest commit. I'll take care of that.

cclauss:
```
OverflowError: (34, ...)
```

QuantumNovice:
Fixed the error with `+IGNORE_EXCEPTION_DETAIL`
[New Pull](https://github.com/TheAlgorithms/Python/pull/1054)

PR #1060, opened 2019-07-21
Code to change contrast

QuantumNovice:


digital_image_processing/change_contrast.py
```diff
@@ -0,0 +1,35 @@
+"""
+Changing contrast with PIL
+
+This algorithm is used in
+https://noivce.pythonanywhere.com/ python web app.
```
cclauss: Please provide a URL to the code repository as well.  We need to ensure that the code was published under a license similar to [__this repo's license__](https://github.com/TheAlgorithms/Python/blob/master/LICENSE.md).

QuantumNovice: I wrote that. 
This is a hello world app I wrote with Flask https://noivce.pythonanywhere.com/ .

cclauss: Good.  Spelling: "novice"

digital_image_processing/change_contrast.py
```diff
@@ -0,0 +1,35 @@
...
+
+python/black: True
+flake8 : True
+"""
+
+from PIL import Image
+
+
+def change_contrast(img, level: int):
```
cclauss: def change_contrast(img: Image, level: int) -> ???:

digital_image_processing/change_contrast.py
QuantumNovice: I am using free version. So novice isn't available.

digital_image_processing/change_contrast.py
```diff
@@ -0,0 +1,35 @@
...
+        return 128 + factor * (c - 128)
+
+    return img.point(contrast)
+
+
+if __name__ == "__main__":
+    # Load image
+    img = Image.open("image_data/lena.jpg")
+    # Change contrast to 170
+    cont_img = change_contrast(img, 170.0)
```
cclauss: Type hint above says __int__ but we are passing a __float__.  Which should it be?

digital_image_processing/change_contrast.py
QuantumNovice: On it.

digital_image_processing/change_contrast.py
```diff
@@ -0,0 +1,35 @@
...
+from PIL import Image
+
+
+def change_contrast(img, level: int):
+    """
+    Function to change contrast
+    """
+    factor = (259 * (level + 255)) / (255 * (259 - level))
+
+    def contrast(c):
```
cclauss: type hints here?

digital_image_processing/change_contrast.py
QuantumNovice: Corrected.

cclauss:
If __float__ works instead of __int__, then I am cool with that because it would deliver more fine-grained control but go with whatever works best for your algorithm.

cclauss:
👍 Cool to see that mypy is working...
https://travis-ci.org/TheAlgorithms/Python/builds/561654753#L563

digital_image_processing/change_contrast.py
```diff
@@ -17,7 +17,7 @@ def change_contrast(img: Image, level: int) -> Image:
     """
     factor = (259 * (level + 255)) / (255 * (259 - level))
 
-    def contrast(c: int) -> int:
+    def contrast(c: int) -> float:
```
cclauss: Should __c__ be an __int__ or a __float__?

Please add "ready for review" when you are satisfied (and the Travis tests pass).

QuantumNovice: `c` should be `int` it is actually individual color value of a pixel.
i.e
`[(c1, c2, c3), (c1, c2, c3)]`

digital_image_processing/change_contrast.py
```diff
@@ -0,0 +1,35 @@
...
+https://noivce.pythonanywhere.com/ python web app.
+
+python/black: True
+flake8 : True
+"""
+
+from PIL import Image
+
+
+def change_contrast(img: Image, level: int) -> Image:
```
cclauss: Last question: Should __level__ be an int or a float?

QuantumNovice: I did this keeping C++ in mind but in Python `float` is faster than `int`.

digital_image_processing/change_contrast.py
```diff
@@ -0,0 +1,35 @@
...
+        every bit.
+        """
+        return 128 + factor * (c - 128)
+
+    return img.point(contrast)
+
+
+if __name__ == "__main__":
+    # Load image
+    img = Image.open("image_data/lena.jpg")
```
cclauss: A [slightly better approach](https://pillow.readthedocs.io/en/stable/reference/open_files.html#image-lifecycle) would be: __with Image.open() as img:__ and then indent the remaining lines.  That would close the image file when the _with_ block ends.  Leaving files open is not a good idea especially image files which can take up a lot of memory.

digital_image_processing/change_contrast.py
QuantumNovice: Changed level to `float`

digital_image_processing/change_contrast.py
QuantumNovice: Done

QuantumNovice:
What would you suggest for writing doc-tests for image processing algorithms?

cclauss:
I see a lot of repos use https://pypi.org/project/selenium to compare images on webpages but I have not done it myself.

PR #1061, opened 2019-07-21
Create  feature_extraction/bag_of_words.py

QuantumNovice:
Travis build was successful.
https://travis-ci.org/QuantumNovice/Python-1/builds/561687140


feature_extraction/bag_of_words.py
```diff
@@ -0,0 +1,39 @@
+"""
+Refrences:
+https://en.wikipedia.org/wiki/Bag-of-words_model
+@author QuantumNovice
+
+flake8: ignore
+bag_of_words.py:21:80: E501 line too long (96 > 79 characters)
+bag_of_words.py:24:80: E501 line too long (87 > 79 characters)
```
cclauss: Please do a [doctest normalize whitespace](https://github.com/TheAlgorithms/Python/blob/61fec83242bb00766141ffa2571414bbba68ab6e/maths/prime_factors.py#L21) and then wrap the long lines.

feature_extraction/bag_of_words.py
```diff
@@ -0,0 +1,39 @@
...
+    Counts the repitition of words in data.
+    >>> data = ['John likes to watch movies. Mary likes movies too']
+    >>> bagOfwords(data)
+    {'John': 1, 'likes': 2, 'to': 1, 'watch': 1, 'movies.': 1, 'Mary': 1, 'movies': 1, 'too': 1}
+    >>> data = ['John also likes to watch football games.']
+    >>> bagOfwords(data)
+    {'John': 1, 'also': 1, 'likes': 1, 'to': 1, 'watch': 1, 'football': 1, 'games.': 1}
+    """
+    counter: Dict[str, int] = {}
+    for line in data:
```
cclauss: Please simplify with https://docs.python.org/3/library/collections.html#collections.Counter

feature_extraction/bag_of_words.py
```diff
@@ -0,0 +1,39 @@
+"""
```
cclauss: I am reluctant to create a new directory unless we know of a lot of algorythms that we want to add to it.  Would this fit in one of the directories that already exists?

feature_extraction/bag_of_words.py
QuantumNovice: On it. Is now resolved in next commit.

feature_extraction/bag_of_words.py
QuantumNovice: Feature extraction algorithms are
There are a lot of feature extraction algorithm. They usually go with machine learning.
Would `machine_learning/feature_extraction` be acceptable?

cclauss:
Unfortunately i do not see how this is an improvement over what is built into the Python Standard Library.  As the caller to your function, I would rather receive a Counter than the dict returned.  See the Counter examples and see if there is some new value in this contribution.

QuantumNovice:
> 
> 
> Unfortunately i do not see how this is an improvement over what is built into the Python Standard Library. As the caller to your function, I would rather receive a Counter than the dict returned. See the Counter examples and see if there is some new value in this contribution.

I see that too. It's no improvement. Thanks. I'll close the pull.

PR #1071, opened 2019-07-25
added automated doctest to decimal_to_hexadecimal.py in conversions

jasper256:


cclauss:
Please add tests for a negative number (-256) a floating point number (16.16) and a hex number (0xfffff).  Thanks.

jasper256:
Will do @cclauss. I might add an assert statement to weed out floats that don't equal an integer.  I'll try to get it done tomorrow if I have time

cclauss:
I would be OK with a test that fails...
```
>>> decimal_to_hexadecimal(16.16)  # doctest: +ELLIPSIS
        ...
        KeyError: ...
```

Also to consider...

The line [__decimal, remainder = divmod(decimal, 16)__](https://docs.python.org/3/library/functions.html#divmod) would allow you to eliminate three other lines in this script.

jasper256:
@cclauss thank you for your suggestions.  I believe that I may have seen divmod before but I honestly forgot it existed.  Anyway, I think you will also find that with more extensive test cases and comments on said cases the exact capabilities of the program are more clear now.

cclauss:
My sense is that the output could be clearer a __0xffff__ ranther than just __ffff__.  Your thoughts?

conversions/decimal_to_hexadecimal.py
```diff
@@ -21,23 +21,60 @@
 }
 
 def decimal_to_hexadecimal(decimal):
-    """ take decimal value, return hexadecimal representation as str """
+    """
+        take integer decimal value, return hexadecimal representation as str 
+        >>> decimal_to_hexadecimal(5)
+        '5'
```
cclauss: I believe that you can drop the quotes if you want.

conversions/decimal_to_hexadecimal.py
```diff
@@ -21,23 +21,60 @@
...
+        >>> decimal_to_hexadecimal(999098)
+        'f3eba'
+        >>> # negatives work too
+        >>> decimal_to_hexadecimal(-256)
+        '-100'
+        >>> # floats are acceptable if equivalent to an int
+        >>> decimal_to_hexadecimal(17.0)
+        '11'
+        >>> # other floats will error
+        >>> decimal_to_hexadecimal(16.16)
```
cclauss: Let's simplify to:
```python
 >>> decimal_to_hexadecimal(16.16)  # doctest: +ELLIPSIS
...
AssertionError
```

conversions/decimal_to_hexadecimal.py
jasper256: When I drop the quotes, the doctest errors.  This is my first time using doctest so I do not know if there is a workaround.

conversions/decimal_to_hexadecimal.py
jasper256: Got it.

jasper256:
> My sense is that the output could be clearer as **0xffff** ranther than just **ffff**. Your thoughts?

As a novice programmer I trust your thoughts. I had never seen this notation before, but I looked it up and it seems commonly accepted so I added it.

conversions/decimal_to_hexadecimal.py
```diff
@@ -22,44 +22,34 @@
...
         >>> decimal_to_hexadecimal(4096)
-        '1000'
+        '0x1000'
         >>> decimal_to_hexadecimal(999098)
-        'f3eba'
+        '0xf3eba'
         >>> # negatives work too
         >>> decimal_to_hexadecimal(-256)
-        '-100'
+        '0x-100'
```
cclauss: Please add a test:
```
>>> decimal_to_hexadecimal(-256) == hex(-256)
True
```

jasper256: Done.  This also made me realize that I was supposed to put '-' before '0x', which I fixed

jasper256:
The pleasure was all mine.  Every time you suggested a feature, I had to look it up and learn how to implement it, which is helping me improve as a programmer.  This was my first time contributing to an open source project, and I look forward to working on more in the future!

cclauss:
Awesome!  Working on open source can be very rewarding when done with an attitude of continuous learning and collaboration.  You are off to a nice start here.  Adding support for floats is still possible...https://docs.python.org/3/library/stdtypes.html#float.hex  ;-)

PR #1073, opened 2019-07-25
Enable pytest to run on machine_learning/perceptron.py 

cclauss:
Indent global code under `if __name__ == '__main__':` so __pytest__ does not get hung up on this file.


AnshulMalik:
Travis is failing for some weird reason.

PR #1076, opened 2019-07-26
Rewrite build_directory_md.py

cclauss:
Rewrite __scripts/build_directory_md.py__ for ease of maintenance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,5 @@ script:
       --ignore=machine_learning/random_forest_classification/random_forest_classification.py
       --ignore=machine_learning/random_forest_regression/random_forest_regression.py
 after_success:
-  - python scripts/build_directory_md.py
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -19,5 +19,5 @@ script:
       --ignore=machine_learning/random_forest_classification/random_forest_classification.py
       --ignore=machine_learning/random_forest_regression/random_forest_regression.py
 after_success:
-  - python scripts/build_directory_md.py
+  - scripts/build_directory_md.py > DIRECTORY.md
```
poyea: Does this line mean travis will write to Directory on successful builds?

cclauss: Yes but... It is only writing in its local directory.  This does not commit those changes back into the repo.  I could investigate how to do the latter on __branch==master__ if that would be of interest.

PR #1080, opened 2019-07-26
Add if __name__ == '__main__': to machine_learning/perceptron.py

cclauss:
A second attempt to achieve #1073 which got into a bad git state.

Also run [__psf/black__](https://github.com/psf/black) on the code.


cclauss:
Fixed elsewhere.

PR #1084, opened 2019-07-27
Update decimal_to_binary.py

Corruption13:
Added the ability to convert fractional numbers and floats to binary

### DIFF
--- a/conversions/decimal_to_binary.py
+++ b/conversions/decimal_to_binary.py
@@ -1,25 +1,43 @@
-"""Convert a Decimal Number to a Binary Number."""
+"""Convert ANY Decimal Number to a Binary Number."""
 
+def dtbconverter(num):
+    """ Function inputs a float value and returns a list as output """
+    decimal_accuracy = 7
 
-def decimal_to_binary(num):
-    """Convert a Decimal Number to a Binary Number."""
-    binary = []
-    while num > 0:
-        binary.insert(0, num % 2)
-        num >>= 1
-    return "".join(str(e) for e in binary)
+    # The part before decimal point in List<string> format
+    whole = []
+    # The part after decimal point in List<string> format
+    fractional = ['.']
 
+    # Extract fractional number part of decimal
+    mantissa = round(num % 1, decimal_accuracy)
+    # Extract whole number part of decimal.
+    w_num = int(num)
 
-def main():
-    """Print binary equivelents of decimal numbers."""
-    print("\n2 in binary is:")
-    print(decimal_to_binary(2))  # = 10
-    print("\n7 in binary is:")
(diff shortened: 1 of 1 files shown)


cclauss:
Please add som doctests for our automated testing.

conversions/decimal_to_binary.py
```diff
@@ -1,25 +1,43 @@
-"""Convert a Decimal Number to a Binary Number."""
+"""Convert ANY Decimal Number to a Binary Number."""
 
+def dtbconverter(num):
+    """ Function inputs a float value and returns a list as output """
+    decimal_accuracy = 7
```
cclauss: Please consider changing this to __def decimal_to_binary(num: float, decimal_accuracy: int = 7) -> float:__  The current function name is more descriptive, the type hints will help callers and mypy and allowing callers to mess with __decimal_accuracy__ has more value than hardcoding it.

Also, please consider adding some __doctests__ with  ints, floats, zero, negative numbers, and for error checking, an empty list and a string.

conversions/decimal_to_binary.py
```diff
@@ -1,25 +1,43 @@
...
+        i = i + 1
+
+    # Loop to find binary of whole number part.
+    while w_num != 0:
+        whole.append(w_num % 2)
+        w_num = w_num // 2
+    whole.reverse()
+
+    binary = whole + fractional
+    return "".join(binary)
```
cclauss: Callers will appreciate if we return a float, not a string.  __return float(whole + fractional)__

conversions/decimal_to_binary.py
```diff
@@ -1,25 +1,43 @@
...
-    print(decimal_to_binary(7))  # = 111
-    print("\n35 in binary is:")
-    print(decimal_to_binary(35))  # = 100011
-    print("\n")
+    i = 0     # Counter
+
+    # Loop to find binary of mantissa part
+    while i < decimal_accuracy:
+        mantissa = mantissa * 2
+        mantissa = int(mantissa // 1)
```
cclauss: // is integer division so the __int()__ is redundant.

conversions/decimal_to_binary.py
```diff
@@ -1,25 +1,43 @@
...
+        fractional.append(str(mantissa))
+        # Extacting mantissa
+        mantissa = round(mantissa % 1, decimal_accuracy)
+        if mantissa == 0:
+            break  # Removes trailing zeros.
+        i = i + 1
```
cclauss: i += 1

cclauss:
Works as designed.  The current functionality mirrors the Python builtin [__bin()__](https://docs.python.org/3/library/functions.html#bin) which also does not accept floating point values.

PR #1089, opened 2019-07-31
project_euler/problem_10

itsvinayak:
edit new solution using sieve-of-eratosthenes to solve problem 10 of project Euler

### DIFF
--- a/project_euler/problem_10/sol3.py
+++ b/project_euler/problem_10/sol3.py
@@ -1,20 +1,64 @@
-#https://projecteuler.net/problem=10
-#using Sieve_of_Eratosthenes
+"""
+https://projecteuler.net/problem=10
+
+Problem Statement:
+The sum of the primes below 10 is 2 + 3 + 5 + 7 = 17.
+
+Find the sum of all the primes below two million.
+
+using Sieve_of_Eratosthenes : 
+
+The sieve of Eratosthenes is one of the most efficient ways to find all primes smaller than n when n is smaller than 10 million
+"""
+
+try:
+    raw_input  # Python 2
+except NameError:
+    raw_input = input  # Python 3
+
+try:
+    xrange  # Python 2
+except NameError:
+    xrange = range  # Python 3
+
 
 def prime_sum(n):
-    l=[0 for i in range(n+1)]
+    """Returns the sum of all the primes below n.
+    
 
-    l[0]=1
-    l[1]=1
-    for i in range(2,int(n**0.5)+1):
-        if l[i]==0:
-            for j in range(i * i, n+1, i):
-                    l[j]=1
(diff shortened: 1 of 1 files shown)


project_euler/problem_10/sol3.py
```diff
@@ -1,20 +1,64 @@
...
+
+try:
+    raw_input  # Python 2
+except NameError:
+    raw_input = input  # Python 3
+
+try:
+    xrange  # Python 2
+except NameError:
+    xrange = range  # Python 3
```
cclauss: Sorry to be a pain but...  Let's simplify and just use __input()__ and __range()__ because this repo no longer supports legacy Python.

project_euler/problem_10/sol3.py
```diff
@@ -1,20 +1,64 @@
...
-    print(s)
+    return(s)
 
 if __name__ == '__main__':
-    prime_sum(2000000)
+    
+    #import doctest
+    #doctest.testmod()
+    
+    print(prime_sum(int(raw_input())))
```
cclauss: __print(prime_sum(int(input().strip())))__ will ensure that your code will gracefully deal with leading and/or trailing whitespace in user input.

project_euler/problem_10/sol3.py
```diff
@@ -1,20 +1,64 @@
...
+        >>> prime_sum(5000)
+        1548136
+        >>> prime_sum(10000)
+        5736396
+        >>> prime_sum(7)
+        10
+
+    """
+    
+    list = [0 for i in xrange(n+1)]
```
cclauss: __list__ should not be used as a variable name in Python because there is __list__ is a _builtin_ that we should avoid _shadowing_.  https://docs.python.org/3/library/functions.html

project_euler/problem_10/sol3.py
```diff
@@ -1,20 +1,64 @@
...
+        >>> prime_sum(2000000)
+        142913828922
+        >>> prime_sum(1000)
+        76127
+        >>> prime_sum(5000)
+        1548136
+        >>> prime_sum(10000)
+        5736396
+        >>> prime_sum(7)
+        10
```
cclauss: Tests for negative number, floating point number, string?  This will ensure that your code is robust and predictable.

itsvinayak: it's not for a negative number, floating-point number, string. it only works for positive numbers,
it a question base problem if anyone uses it, should only enter positive numbers.

thank you,

cclauss:
Thanks for the contribution...

An optimization was possible around:
```python
    s = 0
    for i in range(n):
        if list_[i] == 0:
            s += i
    return s
# could be rewritten as
    return sum(i for i in range(n) if list_[i] == 0) 
```

PR #1092, opened 2019-08-02
improve image convolve algorithm and add hog feature

lighttxu:
Implementation of image convolve is improved to support non-square filter.
And I have added hog feature extraction


cclauss:
Thanks for these modifications.  Would you be willing to add some type hints, doctests, or other pytests?

cclauss:
Sorry...  I meant real Python type hints: https://docs.python.org/3/library/typing.html
https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html
https://realpython.com/python-type-checking/

lighttxu:
Hmm... got it.  I will fix them later.  
Thanks for the advice.

PR #1094, opened 2019-08-02
Fixed SyntaxError in matrix_operation.py (#1024)

AnomNomidas:


cclauss:
I do not see the syntax error in the current code...
```
>>> shape = "abcd"
>>> (f"operands could not be broadcast together with shape "
>>>  f"({shape[0]}, {shape[1]}), ({shape[2]}, {shape[3]})")
'operands could not be broadcast together with shape (a, b), (c, d)'
```

littlechrome:
I think it is false positive caused by the use of f-string.
I tried running currect code in python 3.6 and no problem occurred
https://discuss.lgtm.com/t/python-false-positive-for-f-strings/1874

PR #1101, opened 2019-08-06
print() is a function just like every other function

cclauss:
No space between __print__ and __(__.


harshildarji:
@cclauss Although I have reviewed this, don't merge it before #1102 otherwise it may create conflicts for the file [quadratic_equations_complex_numbers.py](https://github.com/TheAlgorithms/Python/blob/c5f04044252bd742f0ca46d044467f48764f9cda/maths/quadratic_equations_complex_numbers.py).

Issue #989, opened 2019-07-09
pytest --doctest-modules data_structures fails


cclauss:
Missing __if \_\_name\_\_ == "\_\_main\_\_":__ ?

cclauss:
pytest --doctest-modules data_structures
============================================================================== test session starts ==============================================================================
platform darwin -- Python 3.7.4, pytest-4.4.1, py-1.8.0, pluggy-0.9.0
rootdir: /Users/cclauss/Python/itinerant_futurizer/Python
collected 0 items / 5 errors

==================================================================================== ERRORS =====================================================================================
____________________________________________________________ ERROR collecting data_structures/hashing/double_hash.py ____________________________________________________________
data_structures/hashing/double_hash.py:3: in <module>
    from .hash_table import HashTable
E   ImportError: attempted relative import with no known parent package
____________________________________________________ ERROR collecting data_structures/hashing/hash_table_with_linked_list.py ____________________________________________________
data_structures/hashing/hash_table_with_linked_list.py:1: in <module>
    from .hash_table import HashTable
E   ImportError: attempted relative import with no known parent package
_________________________________________________________ ERROR collecting data_structures/hashing/quadratic_probing.py _________________________________________________________
data_structures/hashing/quadratic_probing.py:3: in <module>
    from .hash_table import HashTable
E   ImportError: attempted relative import with no known parent package
________________________________________________________ ERROR collecting data_structures/stacks/balanced_parentheses.py ________________________________________________________
data_structures/stacks/balanced_parentheses.py:3: in <module>
    from stack import Stack
E   ModuleNotFoundError: No module named 'stack'
____________________________________________________ ERROR collecting data_structures/stacks/infix_to_postfix_conversion.py _____________________________________________________
data_structures/stacks/infix_to_postfix_conversion.py:5: in <module>
    from .Stack import Stack
E   ModuleNotFoundError: No module named 'stacks.Stack'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 5 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================================================================ 5 error in 0.43 seconds ============================================================================

QuantumNovice:
[...]

PR #1108, opened 2019-08-06
Fixed #575.1 Travis. Extension of the issue #989

QuantumNovice:
The error was caused by providing relative path to stack.py script.


cclauss:
This PR has a useless title.  The  title should tell the reader what problem(s) are being solved without looking up other PRs and Issues.  Links to other items are better in the body than the title because in the body they can be clicked on.

cclauss:
Not required.  The current code works.

PR #1111, opened 2019-08-07
Digital image processing doctests

mrvnmchm:
Added pytests to digital_image_processing. Also, potential typo found in `digital_image_processing/filters/median_filter.py line 18`

Found typo during testing.
image

Error resolved after fix
image

PR #1117, opened 2019-08-08
Add digital image processing tests

mrvnmchm:
All previous corrections from #1111 kept


cclauss:
@mrvnmchm There is at least one of these tests that really slows down our Travis CI tests.  It seems to talk several seconds.  Would you be willing to investigate to see which test is slowing us down and if there is a way to accelerate it?

cclauss:
[Last release date was 2011](https://pypi.org/project/OL-GeoIP/1.2.4/#history)

PR #1119, opened 2019-08-09
Update tim_sort.py -> update merge

jmanh128:
The merge algorithm was incredibly slow. In testing, the one I am submitting about 3 times quicker than the recursive one.

### DIFF
--- a/sorts/tim_sort.py
+++ b/sorts/tim_sort.py
@@ -25,16 +25,17 @@ def insertion_sort(lst):
 
 
 def merge(left, right):
-    if not left:
-        return right
-
-    if not right:
-        return left
-
-    if left[0] < right[0]:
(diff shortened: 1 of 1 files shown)


cclauss:
Please add doctests lilt timsort() has.

It would be awesome if the __main()__ function did timeit tests of the various sort functions vs. builtin sort() and sorted().

https://docs.python.org/3/library/timeit.html

jmanh128:
I am not quite sure what exactly you're asking. Are you asking me to re-implement the old algorithm and add the timeit() to compare the two? Or is it something else. Thank you

cclauss:
Not need to bring back the old algorithm.  Two requests...
1. Add doctests to merge sort like the doctests in timsort()
2. Add timeit benchmarks so we can see relative performance of Tim vs Merge vs the two builtins.

PR #1128, opened 2019-08-13
Added sudoku solving program in backtracking algorithms

adithbharadwaj:
I have added sudoku.py which is a program to solve the famous sudoku puzzle in python. Please review it and suggest improvements/comments, if any.

### DIFF
--- a/backtracking/sudoku.py
+++ b/backtracking/sudoku.py
@@ -13,6 +13,17 @@
 
 '''
 
+# assigning initial values to the grid 
+initial_grid = [[3,0,6,5,0,8,4,0,0], 
+        		[5,2,0,0,0,0,0,0,0], 
+        		[0,8,7,0,0,0,0,3,1], 
+        		[0,0,3,0,1,0,0,8,0], 
+        		[9,0,0,8,6,3,0,0,5], 
+        		[0,5,0,0,9,0,6,0,0], 
+        		[1,3,0,0,0,0,2,5,0], 
+        		[0,0,0,0,0,0,0,7,4], 
+        		[0,0,5,2,0,6,3,0,0]] 
+
 def is_safe(grid, row, column, n):
 	'''
 	This function checks the grid to see if each row,
@@ -23,16 +34,16 @@ def is_safe(grid, row, column, n):
 	'''
 
 	for i in range(9):
-		if(grid[row][i] == n):
+		if grid[row][i] == n:
 			return False
 		
 	for i in range(9):
-		if(grid[i][column] == n):
+		if grid[i][column] == n:
 			return False		
 	
 	for i in range(3):
 		for j in range(3):
-			if(grid[(row - row%3) + i][(column - column%3) + j] == n):
+			if grid[(row - row%3) + i][(column - column%3) + j] == n:
 				return False
 
 	return True			
@@ -45,10 +56,10 @@ def is_completed(grid):
 	and There is no repeating number in any column, row or 3x3 subgrid.
 	
 	'''
-	
-	for i in range(9):
-		for j in range(9):
-			if(grid[i][j] == 0):
+
+	for row in grid:
+		for cell in row:
+			if cell == 0:
 				return False
 
 	return True
@@ -63,7 +74,7 @@ def find_empty_location(grid):
 
 	for i in range(9):
 		for j in range(9):
-			if(grid[i][j] == 0):
+			if grid[i][j] == 0:
 				return i, j	
 
 
@@ -75,17 +86,17 @@ def sudoku(grid):
 	
 	'''
 
-	if(is_completed(grid)):
+	if is_completed(grid):
 		return True
 
 	row, column = find_empty_location(grid)	
 
 	for digit in range(1, 10):	
-		if(is_safe(grid, row, column, digit)):
+		if is_safe(grid, row, column, digit):
 			grid[row][column] = digit
 			
-			if(sudoku(grid)):
-				return True
+			if sudoku(grid):
+				return grid
 			
 			grid[row][column] = 0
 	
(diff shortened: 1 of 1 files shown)


backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+	and There is no repeating number in any column, row or 3x3 subgrid.
+	
+	'''
+	
+	for i in range(9):
+		for j in range(9):
+			if(grid[i][j] == 0):
+				return False
+
+	return True
```
cclauss: OPTIONAL: Challenges:
1. Try to do this without hardcoding 9 and without using __range()__ like: __for row in grid: for cell in row:__
2.Try making the body of this function a one-liner using nested list comprehensions.

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+	'''
+	This function checks the grid to see if each row,
+	column, and the 3x3 subgrids contain the digit 'n'.
+	It returns False if it is not 'safe' (a duplicate digit
+	is found) else returns True if it is 'safe'
+	
+	'''
+
+	for i in range(9):
+		if(grid[row][i] == n):
```
cclauss: Lose unneeded parens... __if(grid[row][i] == n):__ --> __if grid[row][i] == n:__  # because Python != JavaScript ;-)

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+			return False
+		
+	for i in range(9):
+		if(grid[i][column] == n):
```
cclauss: Parens.

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+def find_empty_location(grid):
+	'''
+	This function finds an empty location so that we can assign a number 
+	for that particular row and column.
+
+	'''
+
+	for i in range(9):
+		for j in range(9):
+			if(grid[i][j] == 0):
```
cclauss: Parens

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+	
+	'''
+
+	if(is_completed(grid)):
+		return True
+
+	row, column = find_empty_location(grid)	
+
+	for digit in range(1, 10):	
+		if(is_safe(grid, row, column, digit)):
```
cclauss: Parens

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+			grid[row][column] = digit
+			
+			if(sudoku(grid)):
```
cclauss: Parens

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+
+	for i in range(9):
+		for j in range(9):
+			print(grid[i][j], end = ' ')
+		print()
+
+
+if __name__ == '__main__':
+	# assigning initial values to the grid 
+	grid = [[3,0,6,5,0,8,4,0,0],
```
cclauss: Put this grid at the top of the file so that we can use it in automated testing.

backtracking/sudoku.py
```diff
@@ -0,0 +1,127 @@
...
+			if(grid[i][j] == 0):
+				return i, j	
+
+
+def sudoku(grid):
+	'''
+	Takes a partially filled-in grid and attempts to assign values to 
+	all unassigned locations in such a way to meet the requirements 
+	for Sudoku solution (non-duplication across rows, columns, and boxes) 
+
```
cclauss: Let's have the __sudoku()__ function not change the incoming grid and let's have it return a solution or None.  The first statement in the body of the function should be to make a separate copy: __grid = grid[:]__.  This allows the caller to compare their grid against the solution returned by the function.

Returning a solution allows us to build [doctests](https://docs.python.org/3/library/doctest.html) for the function.  These can be run as part of our automated tests

backtracking/sudoku.py
adithbharadwaj: Sure. will do. You mean make it a global variable?

backtracking/sudoku.py
adithbharadwaj: sure! will remove all unnecessary parens :)

backtracking/sudoku.py
cclauss: Yes please.

cclauss: Perhaps one grid that has a solution and a second grid that does not so that we can test that the function does the right thing in both cases.

backtracking/sudoku.py
adithbharadwaj: Yeah sure. A separate copy of 'grid' makes sense. Will make these changes and submit a pull request soon. Thanks a lot for your suggestions. :)

backtracking/sudoku.py
cclauss: If you want to _automate_ that kind of code cleanup, then consider using the Python Software Foundation's [__black__](https://github.com/psf/black).

cclauss:
Just FYI: Look at the sudoku files in __pytudes__ to get other implementation ideas an for some really difficult puzzles https://github.com/norvig

backtracking/sudoku.py
```diff
@@ -75,17 +86,17 @@ def sudoku(grid):
 	
 	'''
 
-	if(is_completed(grid)):
+	if is_completed(grid):
 		return True
```
cclauss: __return grid__

cclauss:
Please fix line 90

backtracking/sudoku.py
```diff
@@ -0,0 +1,144 @@
...
+	
+	'''
+
+	for i in range(9):
+		if grid[row][i] == n:
+			return False
+		
+	for i in range(9):
+		if grid[i][column] == n:
+			return False
```
cclauss: OPTIONAL: Would it be possible to compress lines 46-52 onto just three lines?

cclauss:
Are we ready to merge?

adithbharadwaj:
Yeah we are ready to merge. I have made all the changes you suggested and also added a grid with no solution as well.

adithbharadwaj:
Thanks a lot, sir! I am a beginner to open source and would love to contribute more to the repository. Any guidance or suggestions are much appreciated. Thanks again :)

cclauss:
https://github.com/TheAlgorithms/Python/blob/master/backtracking/sudoku.py#L97

To run these doctest locally, do __python3 -m doctest -v backtracking/sudoku.py__

https://github.com/TheAlgorithms/Python/issues/created_by/cclauss might be good ways to get more involved.

PR #1131, opened 2019-08-13
Solution for Problem Euler 56

shuklalok:
Solution for problem Euler 56

### DIFF
--- a/project_euler/problem_56/sol1.py
+++ b/project_euler/problem_56/sol1.py
@@ -1,15 +1,27 @@
-def maximum_digital_sum(a,b):
-    '''
-    Considering natural numbers of the form, a**b, where a, b < 100,
-    what is the maximum digital sum?
-    :param a:
-    :param b:
-    :return:
-    '''
+import typing
+
+
+def maximum_digital_sum(a,b) -> int:
+    """
+        Considering natural numbers of the form, a**b, where a, b < 100,
+        what is the maximum digital sum?
+        :param a:
+        :param b:
+        :return:
+        >>> maximum_digital_sum(10,10)
+        45
+
+        >>> maximum_digital_sum(100,100)
+        972
+
+        >>> maximum_digital_sum(100,200)
+        1872
+    """
+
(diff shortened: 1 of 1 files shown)


project_euler/problem_56/sol1.py
```diff
@@ -0,0 +1,15 @@
+def maximum_digital_sum(a,b):
```
cclauss: Please consider adding Python [__type hints__](https://docs.python.org/3/library/typing.html) like: __def maximum_digital_sum(a: int, b: int) -> int:__ so that __mypy__  can do type checking in our automated testing.

Also, please consider adding Python [__doctests__](https://docs.python.org/3/library/doctest.html) so that __pytest__ can run those tests in our automated testing.

https://github.com/TheAlgorithms/Python/blob/master/.travis.yml

project_euler/problem_56/sol1.py
```diff
@@ -1,15 +1,27 @@
...
-    Considering natural numbers of the form, a**b, where a, b < 100,
-    what is the maximum digital sum?
-    :param a:
-    :param b:
-    :return:
-    '''
+import typing
+
+
+def maximum_digital_sum(a,b) -> int:
```
cclauss: Can __a__ and __b__ only be __ints__ or could they be __floats__ as well?  Same question on the __return__ value.

project_euler/problem_56/sol1.py
```diff
@@ -1,15 +1,27 @@
-def maximum_digital_sum(a,b):
-    '''
-    Considering natural numbers of the form, a**b, where a, b < 100,
-    what is the maximum digital sum?
-    :param a:
-    :param b:
-    :return:
-    '''
+import typing
```
cclauss: This does not need to be imported if it is not being used.

project_euler/problem_56/sol1.py
shuklalok: **a** and **b** can be only **int**.

cclauss: (a: int,b: int) -> int:

PR #1132, opened 2019-08-13
Fixed bug doubly_linked_list.py

amadionix:
If I were to use deleteHead with a list containing only one node I would get an error when trying to access "previous" of a None object

### DIFF
--- a/data_structures/linked_list/doubly_linked_list.py
+++ b/data_structures/linked_list/doubly_linked_list.py
@@ -24,16 +24,17 @@ class LinkedList:           #making main class named linked list
     def deleteHead(self):
         temp = self.head
         self.head = self.head.next                   # oldHead <--> 2ndElement(head) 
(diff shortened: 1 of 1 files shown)


cclauss:
Can you please add automated (doctest) test that fail on the current code and pass on the new code so that we and prove the change work and we can make sure that future changes do not break our tests?

amadionix:
I thought this repo was test-free, since I didn't find any tests anywhere inside it: am I missing something?
I am actually working on rewriting a similar library, along with good test jobs. But so far it changes a bit from this one here, like namings and so on. I might pull request some blocks of that once they are complete (for example the linked list one)

cclauss:
Yes... Missing something... https://travis-ci.org/TheAlgorithms/Python/jobs/571383525#L575

amadionix:
Great. Thank you for telling me. I will try to find the time to do what you asked.

PR #1138, opened 2019-08-17
added doctests for compare_string and is_for_table

pathak-deep15:
>>>compare_string('0010','0110')
       '0_10'
>>> is_for_table('__1','011',2)
	True
The above doctests were added

### DIFF
--- a/boolean_algebra/quine_mc_cluskey.py
+++ b/boolean_algebra/quine_mc_cluskey.py
@@ -2,6 +2,9 @@ def compare_string(string1, string2):
 	"""
 	>>> compare_string('0010','0110')
 	'0_10'
+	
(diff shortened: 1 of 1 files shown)


cclauss:
This is awesome and it works as expected but instead of putting all the tests into a single comment, I would encourage you to move the doctests under their appropriate functions.  The reasons for this is code reuse and readability.  Let's say that I read this file and I would like to copy the algorithm __compare_string()__ and paste it into my own project.  It would be good if the testcase moved along with the function.  The readability dimension comes in when I am reading through the code, I see the definition of the function and immediately after I see a sample call which helps me to know if I properly understand the intention and effects of calling that function.

Great work and thanks for your contribution.

PR #1140, opened 2019-08-19
Fixing lgtm warning in interpolation search per #1024

qckzr:


qckzr:
Fixing one lgtm warning #1024

cclauss:
I do not think this LGTM issue is worth fixing.  We do not want to turn on __debug__ just to placate this linter.

PR #1141, opened 2019-08-19
Fixing lgtm issue in basic graphs

qckzr:


graphs/basic_graphs.py
```diff
@@ -141,7 +141,7 @@ def dijk(G, s):
 from collections import deque
 
 
-def topo(G, ind=None, Q=[1]):
+def topo(G, ind=None, Q):
```
cclauss: The Travis CI tests are failing.  A proper workaround would be:
```python
def topo(G, ind=None, Q=None):
    if Q is None:
        Q = [1]
```

PR #1142, opened 2019-08-19
ENH: refactored longest common subsequence, also fixed a bug with the sequence returned

maxwell-aladago:


dynamic_programming/longest_common_subsequence.py
```diff
@@ -1,37 +1,83 @@
...
 """
 from __future__ import print_function
 
-try:
-    xrange          # Python 2
-except NameError:
-    xrange = range  # Python 3
 
-def lcs_dp(x, y):
+def lcs_dp(x: str, y: str):
```
cclauss: Should we have a more intuitive function name?

I would just return the substring because the caller can easily do __len(substring)__ if that information is really needed.

maxwell-aladago:
Yes, I actually tried changing but decided not to because you initially
approved it.

PR #1147, opened 2019-08-23
Merge pull request #1 from TheAlgorithms/master

jiveshs98:
Updating repo


cclauss:
Why does this keep happening?  Why not just clone or resync the repo?

jiveshs98:
Actually, I wanted to know if there is a way to resync the repo without using git bash. I have tried, but in vain. Please tell me a solution.

cclauss:
The first three commands:
https://devguide.python.org/gitbootcamp/#syncing-with-upstream

jiveshs98:
Thank you!!

PR #1148, opened 2019-08-24
docs: remove FUNDING.yml and use default file

yanglbme:
See https://github.com/TheAlgorithms/.github


cclauss:
Why is this PR a good idea?

yanglbme:
@cclauss Using default files for @TheAlgorithms helps us avoid doing some repetitive work in every repo. Also, we can develop a unified specification for the organization.

And that's why I think it's good :)

yanglbme:
@poyea By default, the <kbd>Sponsor</kbd> button will not show up in every repository.  

The owner of @TheAlgorithms or creator of the repository have to manually check the option `Sponsorships` in Setting/Features to show it up.

image

PR #1161, opened 2019-08-30
Update DIRECTORY

jai-dewani:
Index more problems from sections like
* Machine Learning
* Project Euler
* Scripts


jai-dewani:
Just saw the URL problem, will fix it

scripts/build_directory_md.py
```diff
@@ -36,7 +36,7 @@ def print_directory_md(top_dir: str = ".") -> None:
         if filepath != old_path:
             old_path = print_path(old_path, filepath)
         indent = (filepath.count(os.sep) + 1) if filepath else 0
-        url = "/".join((URL_BASE, filepath, filename)).replace(" ", "%20")
+        url = "/".join((URL_BASE, filepath.split(os.sep)[1], filename)).replace(" ", "%20")
```
cclauss: These are URLs which ___always___ use slashes.  __os.sep__ on Windows is a backslash which is not a good idea in a URL.

jai-dewani: I used os.sep to remove the back slashes in filepath
Before that filepath were something like "\\\\folder" and that's why bad URLs were generated.

PR #1162, opened 2019-08-31
Created Sherman Morrison method

McDic:
I created method to apply Sherman-Morrison formula to Matrix.

Sherman-Morrison formula is important to reduce time complexity on calculating inverse matrix on some special cases. This formula is also used to improve BFGS update performance. For deeper stuffs please look into this documents:

https://en.wikipedia.org/wiki/Sherman%E2%80%93Morrison_formula
https://en.wikipedia.org/wiki/Broyden%E2%80%93Fletcher%E2%80%93Goldfarb%E2%80%93Shanno_algorithm
https://en.wikipedia.org/wiki/Quasi-Newton_method

### DIFF
--- a/matrix/sherman_morrison.py
+++ b/matrix/sherman_morrison.py
@@ -1,4 +1,8 @@
 class Matrix:
+    """
+    <class Matrix>
+    Matrix structure.
+    """
 
     def __init__(self, row: int, column: int, default_value: float = 0):
         """
--- a/matrix/sherman_morrison.py
+++ b/matrix/sherman_morrison.py
@@ -37,7 +37,13 @@ class Matrix:
         string_format_identifier = "%%%ds" % (max_element_length,)
 
         # Make string and return
-        s += "\n".join("[" + ", ".join(string_format_identifier % (obj,) for obj in row_vector) + "]" for row_vector in self.array)
+        def single_line(row_vector): 
+            nonlocal string_format_identifier
+            line = "["
+            line += ", ".join(string_format_identifier % (obj,) for obj in row_vector)
+            line += "]"
+            return line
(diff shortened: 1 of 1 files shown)


McDic:
@cclauss  I added doctests.

By the way, I'm sorry, what do you mean for list comprehension? Do you want some method to iterate over all elements in matrix as single 1x1 element or row/column vector?

matrix/sherman_morrison.py
```diff
@@ -30,39 +37,73 @@ def __str__(self):
         string_format_identifier = "%%%ds" % (max_element_length,)
 
         # Make string and return
-        for row_vector in self.array:
-            s += "["
-            s += ", ".join(string_format_identifier % (obj,) for obj in row_vector)
-            s += "]\n"
+        s += "\n".join("[" + ", ".join(string_format_identifier % (obj,) for obj in row_vector) + "]" for row_vector in self.array)
```
cclauss: This line can not be read on GitHub without left-right scrolling.  Please wrap it so this is no longer the case.

matrix/sherman_morrison.py
```diff
@@ -30,39 +37,73 @@ def __str__(self):
...
+        >>> a+b
+        Matrix consist of 2 rows and 1 columns
+        [-1]
+        [-1]
         """
 
         # Validation
-        assert(self.row == another.row and self.column == another.column)
         assert(isinstance(another, Matrix))
+        assert(self.row == another.row and self.column == another.column)
```
cclauss: assert(0)

assert(False)

Lose the parens ()

McDic:
If you have more things to fix then please write comments. Thank you.

McDic:
@cclauss Is there anything needed to change?

PR #1164, opened 2019-08-31
solution to problem 551 from project euler

b63:
Hello,
This would be my first contribution, so if there is anything I could have done better, I would greatly appreciate any feedback.
Thanks!

### DIFF
--- a/project_euler/problem_551/sol1.py
+++ b/project_euler/problem_551/sol1.py
@@ -23,6 +23,9 @@ def next_term(a_i, k, i, n):
     smallest term for which c > 10^k when the terms are written in the form:
             a(i) = b * 10^k + c
 
+    For any a(i), if digitsum(b) and c have the same value, the difference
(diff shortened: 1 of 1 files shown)


project_euler/problem_551/sol1.py
```diff
@@ -0,0 +1,192 @@
...
+
+    while i < n:
+        i += 1
+        addend = ds_c + ds_b
+        diff += addend
+        ds_c = 0
+        for j in range(k):
+            s = a_i[j] + addend
+            a_i[j] = s % 10
+            addend = s // 10
```
cclauss: Combine 131 and 132 into a single line with _builtin_ __divmod()__:
__addend, a_i[j] = divmod(s, 10)

b63: nice! also changed it in a few more places

b63:
My bad, I changed some of the less descriptive names and also added some comments to aid readability.

PR #1166, opened 2019-09-02
Add radix2 FFT

KirilBangachev:
Created a dynamic implementation of the radix - 2 Fast Fourier Transform for fast polynomial multiplication.

Reference: https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm#The_radix-2_DIT_case


maths/radix2_fft.py
```diff
@@ -0,0 +1,202 @@
...
+            inverseC.pop()
+        return inverseC
+
+    # Overwrite __str__ for print(); Shows A, B and A*B
+    def __str__(self):
+        A = "A = "
+        B = "B = "
+        C = "A*B = "
+        for i in range(self.len_A):
+            A += str(self.polyA[i]) + "*x^" + str(i) + " + "
```
cclauss: A += f"{self.polyA[i]}*x{i} + "

KirilBangachev: I updated the `__str__ ` method using f prefix for clarity and list comprehension+String.join() for performance.

maths/radix2_fft.py
```diff
@@ -174,20 +174,25 @@ def __multiply(self):
...
-            A += str(self.polyA[i]) + "*x^" + str(i) + " + "
-        for i in range(self.len_B):
-            B += str(self.polyB[i]) + "*x^" + str(i) + " + "
-        for i in range(self.len_B + self.len_A - 1):
-            C += (
-                str(self.product[i])
-                + "*x^"
-                + str(i)
-                + " + "
-            )
```
cclauss: A = "A = " + ".join(f"{item}*x^{i}" for i, item in enumerate(self.polyA))

maths/radix2_fft.py
```diff
@@ -174,20 +174,25 @@ def __multiply(self):
...
+            ]
+        )
+        C = "A*B = " + " + ".join(
+            [
+                f"{self.product[i]}*x^{i}"
+                for i in range(len(self.product))
+            ]
+        )
+
         return A + "\n \n" + B + "\n \n" + C
```
cclauss: return "\n\n".join((A, B, C))

maths/radix2_fft.py
KirilBangachev: I switched it to `self.polyA[:self.len_A]` instead as `self.polyA` has some leading zeros necessary for the FFT.

PR #1169, opened 2019-09-04
Fully refactored the rod cutting module. 

maxwell-aladago:


dynamic_programming/rod_cutting.py
```diff
@@ -1,57 +1,193 @@
-from typing import List
+"""
+This module provides two implementations for the rod-cutting problem:
+1. A naive recursive implementation which has an exponential runtime
+2. Two dynamic programming implementations which have quadratic runtime
 
-def rod_cutting(prices: List[int],length: int) -> int:
```
cclauss: A __List[int]__ is more strict than a __list__ of anything.

maxwell-aladago: Making a list[int] is too restrictive since the prices can be floats as well. 

If values which do not have arithmetic operations are provided, then a ValueError will be thrown.

maxwell-aladago: > Making a `list[int]` is too restrictive since the prices can be `floats` as well.
> You can also have mixed prices like `[1, 1.2, 3]`.
> 
> If values which do not work well with `addition` are provided, then a ValueError will be thrown. That's something a user should handle, I think.
> 
> Edit: Also, the return type can be `float` or `int` depending on the prices.
> 
> @cclauss

@cclauss, do you insist that I enforced the values of the prices to be integers?

cclauss: Nope.  I am fine with support both int and float

PR #1184, opened 2019-09-17
Removed break in bubble_sort.py

PRITI1999:
Removed the unnecessary usage of break and restored single entry single exit mechanism.

### DIFF
--- a/sorts/bubble_sort.py
+++ b/sorts/bubble_sort.py
@@ -24,7 +24,7 @@ def bubble_sort(collection):
     length = len(collection)
     swapped = True
     i = 0
-    while swapped == True and i < n: # Stop iteration if the collection is sorted.
+    while swapped == True and i < length: # Stop iteration if the collection is sorted.
         swapped = False 
         for j in range(length-1-i):
             if collection[j] > collection[j+1]:
--- a/sorts/bubble_sort.py
+++ b/sorts/bubble_sort.py
@@ -22,14 +22,15 @@ def bubble_sort(collection):
     True
     """
     length = len(collection)
-    for i in range(length-1):
-        swapped = False
+    sorted = False
+    i = 0
+    while sorted == False and i < n: # Stop iteration if the collection is sorted.
+        sorted = True 
         for j in range(length-1-i):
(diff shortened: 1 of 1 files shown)


sorts/bubble_sort.py
```diff
@@ -22,14 +22,15 @@ def bubble_sort(collection):
     True
     """
     length = len(collection)
-    for i in range(length-1):
-        swapped = False
+    sorted = False
```
cclauss: There is really nothing wrong with using __break__.  It is part of the language.

__sorted__ is a builtin in Python so shadowing it with a variable name is a bad idea.

sorts/bubble_sort.py
```diff
@@ -22,14 +22,15 @@ def bubble_sort(collection):
...
+    while sorted == False and i < n: # Stop iteration if the collection is sorted.
+        sorted = True 
         for j in range(length-1-i):
             if collection[j] > collection[j+1]:
-                swapped = True
+                sorted = False
                 collection[j], collection[j+1] = collection[j+1], collection[j]
-        if not swapped:
-            break  # Stop iteration if the collection is sorted.
+         i += 1
```
cclauss: Is there an indentation error (even multiple of four) on this line?

sorts/bubble_sort.py
```diff
@@ -22,14 +22,15 @@ def bubble_sort(collection):
     True
     """
     length = len(collection)
-    for i in range(length-1):
-        swapped = False
+    swapped = True
+    i = 0
+    while swapped == True and i < n: # Stop iteration if the collection is sorted.
```
poyea: What is `n`? Do you mean `length`?

cclauss: https://travis-ci.org/TheAlgorithms/Python/builds/586149305#L600

PRITI1999: Thanks. Replaced n by length

cclauss:
What is the advantage of implicating this PR?  It has more lines, more tests, and more complexity than the current code but I do not understand what problem that this change is solving.

PR #1185, opened 2019-09-18
Create beaufort-cipher.py

egrzeszczak:


ciphers/beaufort-cipher.py
```diff
@@ -0,0 +1,57 @@
...
+
+def filterInput(string):
+    """Filters input from characters that aren't in the alphabet"""
+    for n in range(len(string)):
+        if not string[n] in alphabet:
+            string = string.replace(string[n], alphabet[0])
+    return string
+
+
+def assignKeyMask(input, key):
```
cclauss: __input__ is a builtin function in Python 3 so it is not a good idea to shadow it by reusing that as a variable name.

egrzeszczak: Changed. Now passing an length in instead of full text. Name changed

PR #1188, opened 2019-09-20
Fixed bug and added test case that catches the bug

amadionix:


cclauss:
Tests are not being found by pytest
https://travis-ci.org/TheAlgorithms/Python/builds/587396593#L626-L628

PR #1190, opened 2019-09-20
Fix lgtm error display #1024

denis-trofimov:
#1024 @cclauss 
$ flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude=venv
0

### DIFF
--- a/data_structures/heap/binomial_heap.py
+++ b/data_structures/heap/binomial_heap.py
@@ -307,7 +307,7 @@ class BinomialHeap:
         # No right subtree corner case
         # The structure of the tree implies that this should be the bottom root
         # and there is at least one other root
-        if self.min_node.right == None:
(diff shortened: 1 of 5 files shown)


cclauss:
LGTM is wrong about __fix: Syntax Error lgtm display in matrix/matrix_operation.py.__

The original code is correct and is nota syntax error.

denis-trofimov:
I agree, lgtm is false positive.
Later I found warning can be turn off by comment.

So I will later revert that.

denis-trofimov:
> LGTM is wrong about **fix: Syntax Error lgtm display in matrix/matrix_operation.py.**
> 
> The original code is correct and is nota syntax error.

I filed an issue https://github.com/Semmle/ql/issues/1990

PR #1192, opened 2019-09-21
Update decision_tree.py

FourthNight:
Error on line 128. Test case cannot be both input and label.


cclauss:
Would you be willing to add some test cases to appear in
https://travis-ci.org/TheAlgorithms/Python/builds/587785967#L643

PR #1194, opened 2019-09-22
Add disjoint set

luoheng23:


data_structures/disjoint_set/disjoint_set.py
```diff
@@ -0,0 +1,62 @@
...
+    make x as a set.
+    """
+    # rank is the distance from x to its' parent
+    # root's rank is 0
+    x.rank = 0
+    # p is x's parent
+    x.p = x
+
+
+def unionSet(x, y):
```
cclauss: __union_set()__ would be more Pythonic.

data_structures/disjoint_set/disjoint_set.py
```diff
@@ -0,0 +1,62 @@
+class node:
+    def __init__(self, data):
+        self.data = data
+
+
+def makeSet(x):
```
cclauss: __make_set()__ would be more Pythonic.

data_structures/disjoint_set/disjoint_set.py
```diff
@@ -0,0 +1,62 @@
...
+    x, y = findSet(x), findSet(y)
+    if x.rank > y.rank:
+        y.p = x
+    else:
+        x.p = y
+        if x.rank == y.rank:
+            y.rank += 1
+
+
+def findSet(x):
```
cclauss: __find_set()__ would be more Pythonic.

data_structures/disjoint_set/disjoint_set.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    return the parent of x
+    """
+    if x != x.p:
+        x.p = findSet(x.p)
+    return x.p
+
+
+def test_disjoint_set():
```
cclauss: Please add a doctest here so that our automated tests will pick this up.
```python
def test_disjoint_set():
    """
    >>> test_disjoint_set()
    """

File: data_structures/disjoint_set/disjoint_set.py [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+    assert findSet(vertex[3]) == findSet(vertex[4])
+    assert findSet(vertex[4]) == findSet(vertex[5])
+
+
+def main():
+    test_disjoint_set()
+
+
+if __name__ == "__main__":
+    main()
```
cclauss: Let’s simplify.  Replace line 62 with line 58 and then delete lines 55-58.

data_structures/disjoint_set/disjoint_set.py
```diff
@@ -0,0 +1,62 @@
+class node:
```
cclauss: Classes in Python are Title case so __Node__, not __node__ please.

luoheng23:
> This is supercool! Congratulations and thanks for sharing.
> 
> A nice addition would be a function like:
> **def as_python_set(node: Node) -> set:**
> 
> And an addition test, **as_python_set(vertex[0]).is_disjoint(as_python_set(vertex[3]))**
> https://docs.python.org/3.7/library/stdtypes.html#frozenset.isdisjoint

Thank you for your advice. I have updated the code.  
But I have problem with addtional function. The function **as_python_set** is hard to implement. Disjoint set is a tree structure, each node only has a pointer to its parent, so it's hard to get the whole set using just one node. The test has the same function with **assert find_set(vertex[2]) != find_set(vertex[3])**, for each node has only one pointer.

cclauss:
Put this towards the bottom of your file:
```python
def as_python_set(node: Node) -> set:
    python_set = set()
    while True:
        python_set.add(node.data)
        if node == node.p:
            # assert node.rank == 0, f"{python_set} root node has zero rank: {node.rank}"
            return python_set
        node = node.p


if __name__ == "__main__":
    test_disjoint_set()
    vertex = [Node(i) for i in range(6)]
    for v in vertex:
        make_set(v)
    union_set(vertex[0], vertex[1])
    union_set(vertex[1], vertex[2])
    union_set(vertex[3], vertex[4])
    union_set(vertex[4], vertex[5])
    for node in vertex:
        print(node.data, as_python_set(node))
```
This might help to see issues to fix. Ideally, we would like __as_python_set(vertex[2])__ to be __{0, 1, 2}__.

data_structures/disjoint_set/disjoint_set.py
```diff
@@ -0,0 +1,67 @@
...
+
+def make_set(x):
+    """
+    make x as a set.
+    """
+    # rank is the distance from x to its' parent
+    # root's rank is 0
+    x.rank = 0
+    # p is x's parent
+    x.p = x
```
cclauss: Let's call this __x.parent__ instead of __x.p__ then the comment above can be removed.  Single character variable names are old school.

cclauss:
Perhaps we are not creating __parents__ with the current code but are creating __children__ instead.  ;-)

My sense is that __rank__ is not really helping us and could safely be dropped.

Goal: Build a Python __set__ from either the root node of a union_set or the leaf node of a union_set.

luoheng23:
> Perhaps we are not creating **parents** with the current code but are creating **children** instead. ;-)
> 
> My sense is that **rank** is not really helping us and could safely be dropped.
> 
> Goal: Build a Python **set** from either the root node of a union_set or the leaf node of a union_set.

I don't think we need a **node-to-set** method.
**disjoint set** is to quickly determine if two nodes are in the same set. The idea is  choosing a root, and all nodes point to it.
For example,  for **a, b, c, d**. Final **disjoint set** is not like  **a -> b -> c -> d**, but like **a -> d, b -> d, c -> d, d -> d**.  Then **find_set** can quickly return **d** in just one operation. So we can quickly know whether they are in same set. 
So it's impossible to get whole set through a node, and this is not purpose of this data structure.
**Rank** is an optimize method for **disjoint set**, it can make the tree more flat, then the time complexity will be reduced. [refercence here](https://en.wikipedia.org/wiki/Disjoint-set_data_structure), with such method, **disjoint set** can be faster than normal **set**.

cclauss:
Sorry for being so dense.  You are correct.  It works perfectly.  Here is how I tested it.
```python
def find_python_set(node: Node) -> set:
    """
    Return a Python Standard Library set that contains i.
    """
    sets = ({0, 1, 2}, {3, 4, 5})
    for s in sets:
        if node.data in s:
            return s
    raise ValueError(f"{i} is not in {sets}")


def test_disjoint_set():
    """
    >>> test_disjoint_set()
    """
    vertex = [Node(i) for i in range(6)]
    for v in vertex:
        make_set(v)

    union_set(vertex[0], vertex[1])
    union_set(vertex[1], vertex[2])
    union_set(vertex[3], vertex[4])
    union_set(vertex[3], vertex[5])

    for node0 in vertex:
        for node1 in vertex:
            if find_python_set(node0).isdisjoint(find_python_set(node1)):
                assert find_set(node0) != find_set(node1)
            else:
                assert find_set(node0) == find_set(node1)
```

Please put a link to the Wikipedia article in a comment at the top of the file and then we can land this.  Thanks for your patience.

luoheng23:
> Sorry for being so dense. You are correct. It works perfectly. Here is how I tested it.
> 
> ```python
> def find_python_set(node: Node) -> set:
>     """
>     Return a Python Standard Library set that contains i.
>     """
>     sets = ({0, 1, 2}, {3, 4, 5})
>     for s in sets:
>         if node.data in s:
>             return s
>     raise ValueError(f"{node.data} is not in {sets}")
> 
> 
> def test_disjoint_set():
>     """
>     >>> test_disjoint_set()
>     """
>     vertex = [Node(i) for i in range(6)]
>     for v in vertex:
>         make_set(v)
> 
>     union_set(vertex[0], vertex[1])
>     union_set(vertex[1], vertex[2])
>     union_set(vertex[3], vertex[4])
>     union_set(vertex[3], vertex[5])
> 
>     for node0 in vertex:
>         for node1 in vertex:
>             if find_python_set(node0).isdisjoint(find_python_set(node1)):
>                 assert find_set(node0) != find_set(node1)
>             else:
>                 assert find_set(node0) == find_set(node1)
> ```
> 
> Please put a link to the Wikipedia article in a comment at the top of the file and then we can land this. Thanks for your patience.

Thank you for your reply.
Your test is better, I have updated the test and reference.

PR #1197, opened 2019-09-23
Create stooge_sort

Charitoc:


sorts/stooge_sort
```diff
@@ -0,0 +1,23 @@
+def stoogesort(arr, l, h):
```
cclauss: Please be kind to your reader and use better variable names than __l__ and __h__.  Would it be possible to remove l and h and pass in only __arr__?

sorts/stooge_sort
```diff
@@ -0,0 +1,23 @@
...
+    if l >= h:
+        return
+  
+    # If first element is smaller than the last then swap them
+    if arr[l]>arr[h]:
+        t = arr[l]
+        arr[l] = arr[h]
+        arr[h] = t
```
cclauss: A more Pythonic swap is a one liner: __arr[l], arr[h] = arr[h], arr[l]

sorts/stooge_sort
```diff
@@ -0,0 +1,23 @@
...
+  
+    # If there are more than 2 elements in the array
+    if h-l+1 > 2: 
+        t = (int)((h-l+1)/3)
```
cclauss: __t = (h-l+1) // 3__  # // is integer division

sorts/stooge_sort
cclauss: Please insert a doctest of the form:
```python
def stoogesort(arr, l: int, h: int) -> None:
    """
    >>> arr = list(reversed(range(0, 11, 2)))
    >>> stoogesort(arr, 0, len(arr) - 1)
    >>> print(arr)
    [0, 2, 4, 6, 8, 10]
    """
```
and then locally run __python3 -m doctest -v stooge_sort.py__ to ensure that test passes.

PR #1199, opened 2019-09-24
Create Linear_Time_Sort.py

viraatdas:


sorts/Linear_Time_Sort.py
```diff
@@ -0,0 +1,63 @@
...
+    # Copy the output array to arr[], so that  
+    # arr[] now contains sorted numbers according  
+    # to current digit  
+    for i in range(n):  
+        arr[i] = output[i]  
+  
+# The main function to that sorts arr[] of  
+# size n using Radix Sort  
+def sort(arr, n) : 
+
```
cclauss: Let’s move line 53 (n = len(arr)) to here.  The caller only needs to give us an array and we can get its length ourselves.

Please add doctests to both functions.

sorts/Linear_Time_Sort.py
```diff
@@ -0,0 +1,63 @@
+# Python3 the implementation to sort an  
+# array of size n 
+  
+# A function to do counting sort of arr[]  
+# according to the digit represented by exp.  
+def countSort(arr, n, exp):
```
cclauss: Python functions use the form __count_sort()__ with no capital letters.

cclauss:
Failing tests need to be fixed https://travis-ci.org/TheAlgorithms/Python/builds/589212548#L618

PR #1200, opened 2019-09-25
Deque

paulo9428:
At Double ended queue, added "pop", "popleft", "remove" method


paulo9428:
Sorry, i just wanna complement double ended queue methods of this file. I think this code has no ingenuous points. So i agree with the pr the previous writter @cclauss push newly. And i also totally agree with your opinion. Thank you for your great advice. It was my first event of pushing pr. It leaves me great motive to comtribute on open source. I am satisfied just with that.

PR #1202, opened 2019-09-25
Min head with decrease key functionality

Raj1998:
I have added MinHeap data structure with decrease key function in O(logn) time. I have also added sample usage at the bottom. 

Please let me know if there are any changes.

### DIFF
--- a/data_structures/heap/min_heap.py
+++ b/data_structures/heap/min_heap.py
@@ -1,0 +1,140 @@
+# Min head data structure
+# with decrease key functionality - in O(log(n)) time
+
+
+class MinHeap:
+    def __init__(self, array):
+        self.idx_of_element = {}
+        self.heap = self.buildHeap(array)
+
+    def getParentIdx(self, idx):
+        return (idx - 1) // 2
+
+    def getLeftChildIdx(self, idx):
+        return idx * 2 + 1
+
+    def getRightChildIdx(self, idx):
+        return idx * 2 + 2
+
+    def buildHeap(self, array):
+        lastIdx = len(array) - 1
+        startFrom = self.getParentIdx(lastIdx)
+
+        for idx, i in enumerate(array):
+            self.idx_of_element[i] = idx
+
+        for i in range(startFrom, -1, -1):
+            self.siftDown(i, array)
+        return array
+
+    # this is min-heapify method
+    def siftDown(self, idx, array):
+        while True:
+            l = self.getLeftChildIdx(idx)
+            r = self.getRightChildIdx(idx)
+
+            smallest = idx
+            if l < len(array) and array[l] < array[idx]:
+                smallest = l
+            if r < len(array) and array[r] < array[smallest]:
+                smallest = r
+
+            if smallest != idx:
+                array[idx], array[smallest] = array[smallest], array[idx]
+                self.idx_of_element[array[idx]], self.idx_of_element[
+                    array[smallest]
+                ] = (
+                    self.idx_of_element[array[smallest]],
+                    self.idx_of_element[array[idx]],
+                )
+                idx = smallest
+            else:
+                break
+
+    def siftUp(self, idx):
+        p = self.getParentIdx(idx)
+        while p >= 0 and self.heap[p] > self.heap[idx]:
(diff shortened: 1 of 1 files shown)


data_structures/heap/min_heap.py
```diff
@@ -0,0 +1,140 @@
...
+
+        for idx, i in enumerate(array):
+            self.idx_of_element[i] = idx
+
+        for i in range(startFrom, -1, -1):
+            self.siftDown(i, array)
+        return array
+
+    # this is min-heapify method
+    def siftDown(self, idx, array):
```
cclauss: sift_down()

cclauss:
Please add at least one https://docs.python.org/3/library/doctest.html

data_structures/heap/min_heap.py
Raj1998: Ok, I'll change names.

Raj1998: Shall I change all method names to `sift_down()` convention or only this method?

Raj1998: I have changed the names and also added doctest

cclauss: In Python the convention is that functions and methods use_lowercase_names() while classes UseUppercaseNames().

Raj1998: Changed

data_structures/heap/min_heap.py
```diff
@@ -0,0 +1,154 @@
+# Min head data structure
+# with decrease key functionality - in O(log(n)) time
+
+
+class Node:
+    def __init__(self, val, name):
```
cclauss: It will be more intuitive to change the order of the arguments to be __name, value__ so it is like a Python dictionary __key, value__

data_structures/heap/min_heap.py
```diff
@@ -0,0 +1,154 @@
...
+        self.val = val
+        self.name = name
+
+    def __str__(self):
+        return self.name
```
cclauss: __return f"{self.\_\_class\_\_.\_\_name\_\_}({self.name}, {self.value})"__
The Python convention is that \_\_str\_\_() should print out exactly what it would take to build a new instance.

data_structures/heap/min_heap.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class MinHeap:
+    """
+    >>> r = Node(-1, "R")
+    >>> b = Node(6, "B")
+    >>> a = Node(3, "A")
+    >>> x = Node(1, "X")
+    >>> e = Node(4, "E")
+    >>> myMinHeap = MinHeap([r, b, a, x, e])
```
cclauss:
```
>>> print(b)
Node('b', 6)
```

data_structures/heap/min_heap.py
```diff
@@ -0,0 +1,154 @@
...
+    >>> myMinHeap.decrease_key(b, -17)
+    """
```
cclauss:
```
>>> print(b)
Node('b', -17)
```

Raj1998: I have added this and btw it it is 
`>>> print(b)
Node(B, -17)` 
without quotes around B

data_structures/heap/min_heap.py
```diff
@@ -0,0 +1,166 @@
...
+    >>> print(myMinHeap.get_value("B"))
+    -17
+    """
+
+    def __init__(self, array):
+        self.idx_of_element = {}
+        self.heap_dict = {}
+        self.heap = self.build_heap(array)
+
+    def get_parent_idx(self, idx):
```
cclauss: You could implement the ``__getitem__()__ method here that could enable __my_heap["B"]__ functionality by using __self.idx_of_element__ to use the key to find the node and return the value.

https://docs.python.org/3/reference/datamodel.html#object.__getitem__

data_structures/heap/min_heap.py
cclauss: Did this test fail?

data_structures/heap/min_heap.py
Raj1998: Added

PR #1205, opened 2019-09-25
Fix the build typo: fn --> fn1

cclauss:
I messed up and broke the build.  Can I please get a rapid review?

PR #1206, opened 2019-09-26
Adding stooge sort

Charitoc:


cclauss:
The failing Travis CI builds are not caused by this PR but are instead caused by my mistake in #1205.

sorts/stooge_sort.py
```diff
@@ -0,0 +1,37 @@
...
+def stooge(arr, i, h):
+
+    """
+
+>>> arr = [2, 4, 5, 3, 1]
+>>> stoogesort(arr)
+>>> print(arr)
+[1, 2, 3, 4, 5]
+
+"""
```
cclauss: The doctest for __stooge_sort()__ should be in the __stooge_sort()__ function, not in the __stooge()__ function.

Let's move lines 6, 8, 9, 10, 11, and 13 up to just below line 1.  The blank lines 7 and 12 can be deleted.  It should look like:
```python
def stooge_sort(arr):
    """
    >>> arr = [2, 4, 5, 3, 1]
    >>> stoogesort(arr)
    >>> print(arr)
    [1, 2, 3, 4, 5]
    """
    stooge(arr, 0, len(arr) - 1)


def stooge(arr, i, h):
```

Charitoc: Done

cclauss:
While waiting, you should be able to test this locally with __python3 -m doctest -v sorts/stooge_sort.py__

Charitoc:
It is ready, could you do the review?

Charitoc:
Thank you!

cclauss:
On thing to think about in the future is the difference between __list.sort()__ and __sorted(list)__ in Python.  The first is an in-place sort (like this stooge_sort()) with the second returns a new sorted list without modifying the original list.  Both are useful but I tend to like the second better.

https://docs.python.org/3/howto/sorting.html

Charitoc:
Ok, thank you for your time.

PR #1220, opened 2019-10-01
Update cryptomath_module.py

Hocnonsense:
u2 and v2 are useless
why not remove it?


cclauss:
Conflicts with existing ciphers/cryptomath_module.py and no doctests or type hints.

PR #1226, opened 2019-10-01
Improve prim.py

dunderbruno:
- Updated file to suit PEP8
- Running example
- Improves indexes management


cclauss:
Please consider adding some doctests to the functions.

https://docs.python.org/3/library/doctest.html

PR #1227, opened 2019-10-01
Create support_vector_regression.py

panwalas:
Adding new type of algorithm namely- support vector regression


cclauss:
The global code needs to be wrapped in `if __name__ == "__main__":` as discussed in CONTRIBUTING.md.  If this is not done then all of this code is run by pytest even though there are no tests in this submission.

PR #1233, opened 2019-10-01
Update rsa_cipher.py

Hocnonsense:
change module name &
sys.exit -> raise Exception


cclauss:
Nice!  Please raise a more specific exception instead of a general exception to help users of that algorithm.

cclauss:
Like a __ValueError()__

PR #1236, opened 2019-10-01
Diophantine Equation

OddExtension5:


OddExtension5:
okay sure!

cclauss:
1. Awesome work!  Thanks massively.
2. Are there any other functions that could benefit from doctests?
3. Instead of creating new directories (__hashing/number_theory__) can we fit these files this into data_structures/hashes) or should we create a new (data_structures/number_theory) directory.  We want to avoid creating too many directories at the root and would like to reduce the nesting of directories.  Your advise on categorization would be a big help.

OddExtension5:
1. Thanks!
2. Yes, but other functions are already created somewhere like extended_gcd, gcd, etc., and they have their doctest. So, according to me putting more doctest into my code is not required.
3.  No, we have to make a separate directory for number theory because hashing and number theory are different from each other.  But we can reduce the nesting of directories by making one directory named Bitcoin or Blockchain as this field required knowledge of Hashing, Number Theory, Cryptography, Abstract(Modern) Algebra (Group Theory, Finite Fields) and Private&Public Key&Signature Algorithms. 

P.S.  If you made separate directory then it will be easier for me or others adding algorithms related to Blockchain and users can easily follow our directory. With that said In the future, I will add more algorithms related to bitcoin, for example, Schnorr Digital Signature Algorithm, Finite Field, Elliptic Curve Digital Signature Algorithm(ECDSA).

OddExtension5:
Closing in favour of #1248

PR #1237, opened 2019-10-01
Update bubble_sort.py

madforpython:
Check this code too.I think its optimised and ShortnSweet for beginners


cclauss:
In what way is it better than the current code?

PR #1239, opened 2019-10-01
Update greater_common_divisor.py

madforpython:
Check This code.It's beginner friendly.

### DIFF
--- a/maths/greater_common_divisor.py
+++ b/maths/greater_common_divisor.py
@@ -1,25 +1,25 @@
-"""
-Greater Common Divisor.
+# Python program to find the G.C.D of two input number
 
-Wikipedia reference: https://en.wikipedia.org/wiki/Greatest_common_divisor
-"""
+# define a function
+def computeGCD(x, y):
 
+# choose the smaller number
+    if x > y:
+        smaller = y
+    else:
+        smaller = x
+    for i in range(1, smaller+1):
+        if((x % i == 0) and (y % i == 0)):
+            gcd = i
+            
+    return gcd
 
-def gcd(a, b):
-    """Calculate Greater Common Divisor (GCD)."""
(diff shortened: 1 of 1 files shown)


maths/greater_common_divisor.py
```diff
@@ -1,25 +1,25 @@
-"""
-Greater Common Divisor.
+# Python program to find the G.C.D of two input number
 
-Wikipedia reference: https://en.wikipedia.org/wiki/Greatest_common_divisor
```
cclauss: Why should we take this away??

maths/greater_common_divisor.py
```diff
@@ -1,25 +1,25 @@
...
-"""
+# define a function
+def computeGCD(x, y):
```
cclauss: What about __greatest_common_denominator()__??

maths/greater_common_divisor.py
```diff
@@ -1,25 +1,25 @@
...
 
+# choose the smaller number
+    if x > y:
+        smaller = y
+    else:
+        smaller = x
```
cclauss: __smaller = min(x, y)__

maths/greater_common_divisor.py
madforpython: Sharing My Experience.When i was at my initial Learning stage i used to face several difficulties undersatnding code which contains complexities.So i prefer keeping codes ShortnSweet.

maths/greater_common_divisor.py
madforpython: It's Greatest Common Divisor Code

maths/greater_common_divisor.py
madforpython: as I mentioned above.I like to keep codes beginner friendly.

PR #1240, opened 2019-10-01
added beadsort.py

PedroDCN:


sorts/bead_sort.py
```diff
@@ -0,0 +1,42 @@
+
+from itertools import zip_longest
```
thescriptninja: Hey @PedroDCN. Apparently the build is failing as you have imported `zip_longest` and haven't used it. Please remove it.

PR #1242, opened 2019-10-01
Levenshtein

dunderbruno:
Another implementation of Levenshtein distance.


cclauss:
Why?  How is it better than the one that we already have?

PR #1243, opened 2019-10-01
Pythagoras

dunderbruno:
Distance between two points in space.


cclauss:
Please consider adding some doctests https://docs.python.org/3/library/doctest.html

PR #1255, opened 2019-10-02
Modified caesar_cipher.py to work correctly

DanDevX:
- Fixed encrypt function, now it returns upper and lower case letters instead of any character
- Fixed encrypt and decrypt function, now they can handle spaces correctly
- Changed key range from 1 - 95 to 0 - 25 (after 25 it just repeats itself e.g. 1 and 27 will give you the same result)
- Removed main() from choice == '3' as it was called twice after the bruteforce function


cclauss:
Closing.  Conflicting files an no doctests.

PR #1256, opened 2019-10-02
Added bogobogosort, ran psf/black on all files

WilliamHYZhang:
Hi,

I added code for bogobogosort, and also ran psf/black on all files.

Hope this helps!


cclauss:
It has GOT to be two separate pull requests. It is too difficult to closely read 30 files.

WilliamHYZhang:
Alright, will do.

PR #1257, opened 2019-10-02
Add problem 32 solution

Hyftar:
This PR adds the solution to the Project Euler's Problem 32

### DIFF
--- a/project_euler/problem_32/solution.py
+++ b/project_euler/problem_32/solution.py
@@ -1,0 +1,62 @@
+"""
+We shall say that an n-digit number is pandigital if it makes use of all the
+digits 1 to n exactly once; for example, the 5-digit number, 15234, is 1 through
+5 pandigital.
+
+The product 7254 is unusual, as the identity, 39 × 186 = 7254, containing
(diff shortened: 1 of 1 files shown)


project_euler/problem_32/sol1.py
```diff
@@ -0,0 +1,56 @@
+"""
+We shall say that an n-digit number is pandigital if it makes use of all the digits 1 to n exactly once; for example, the 5-digit number, 15234, is 1 through 5 pandigital.
```
cclauss: Wrap to 88 max

project_euler/problem_32/sol1.py
```diff
@@ -0,0 +1,56 @@
...
+    45228
+    """
+    return sum(
+        set(
+            map(
+                lambda pandigital: int(''.join(pandigital[5:9])),
+                filter(
+                    lambda combination: isCombinationValid(combination),
+                    itertools.permutations('123456789')
+                )
```
cclauss: Comprehension(s) instead?

PR #1260, opened 2019-10-03
Add problem 18 solution

Hyftar:
This adds the solution to the problem 18 which is basically a copy-paste from the solution of the problem 67 except with a different triangle.


project_euler/problem_18/sol1.py
```diff
@@ -0,0 +1,64 @@
+"""
+By starting at the top of the triangle below and moving to adjacent numbers on the row below, the maximum total from top to bottom is 23.
```
cclauss: Reading this comment on GitHub requires left/right scrolling which reduces readability.  Please wrap all lines at 88 characters max.

project_euler/problem_18/sol1.py
```diff
@@ -0,0 +1,64 @@
...
+    1074
+    """
+    script_dir = os.path.dirname(os.path.realpath(__file__))
+    triangle = os.path.join(script_dir, 'triangle.txt')
+
+    with open(triangle, 'r') as f:
+        triangle = f.readlines()
+
+    a = map(lambda x: x.rstrip('\r\n').split(' '), triangle)
+    a = list(map(lambda x: list(map(lambda y: int(y), x)), a))
```
cclauss: Please try to rewrite these as list comprehensions as discussed in
https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md

Hyftar:
Uncertain about what broke the Travis-CI build. Could be a false negative?

cclauss:
Rename your file to something other than sol1.py.

PR #1261, opened 2019-10-03
Add problem 23 solution

Hyftar:
This PR adds the solution to the Project Euler's problem 23


project_euler/problem_23/sol1.py
```diff
@@ -0,0 +1,40 @@
+"""
+A perfect number is a number for which the sum of its proper divisors is exactly equal to the number. For example, the sum of the proper divisors of 28 would be 1 + 2 + 4 + 7 + 14 = 28, which means that 28 is a perfect number.
```
cclauss: Please wrap these lines at 88 characters max for readability.

PR #1275, opened 2019-10-04
sparse table range type queries

chaudhary19:
Minimum, Maximum and Sum Range Query over Sparse Table


Erfaniaa:
Please review this pull request. @poyea @siriak

PR #1277, opened 2019-10-05
psf/black code formatting

WilliamHYZhang:
Ran psf/black code formatting on all files, compliant with PEP8 standard.


cclauss:
Black and PEP8 do not agree on line length but 88 seems more reasonable for modern code than 79.

PR #1283, opened 2019-10-06
Add Iterative Quick Sort

Pranjal1751:


cozek:
Thank you for your contribution.
Kindly fix your file name so that it passes the automated tests.
'file name.py' should be 'file_name.py'. There should not be any capital letters or spaces in the filename. More about this in the CONTRIBUTING.md.
Also, please add some doctests to your functions. You can take reference from the existing quick sort contributions.

PR #1285, opened 2019-10-06
Create segment_tree_min_range_query

adi0311:
Segment Tree implementation for the minimum in a particular range and update of an element queries. I have used Python dictionary to create the tree because in that case, we do not need to define a Python list of size twice that of the original array.


cclauss:
Automated testing is not finding the tests
https://travis-ci.org/TheAlgorithms/Python/builds/594222307#L343

PR #1292, opened 2019-10-07
less code

shellhub:


maths/abs.py
```diff
@@ -10,11 +10,7 @@ def abs_val(num):
     >>abs_val(0)
     0
     """
-    if num < 0:
-        return -num
-
-    # Returns if number is not < 0
-    return num
+    return -num if num < 0 else num;
```
cclauss: Please remove the semicolon!  This is Python, not JavaScript!

shellhub: updated, Thanks

PR #1314, opened 2019-10-08
Added Bozo Sort

ammar0211:
Added a bozo sort algorithm.

### DIFF
--- a/sorts/bozo_sort.py
+++ b/sorts/bozo_sort.py
@@ -1,0 +1,27 @@
+"""
+Bozo sort consists out of checking if the input sequence is sorted and if not swapping randomly two elements.
+This is repeated until eventually the sequence is sorted.
+Takes in an array of numbers and provides a sorted array.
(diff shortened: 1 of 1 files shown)


sorts/bozo_sort.py
```diff
@@ -0,0 +1,27 @@
...
+        if array[i]>array[i+1]:
+            return False
+    return True
+def bozo_sort(array):
+    while not sort_check(array):
+        i,j=random.randint(0,len(array)-1),random.randint(0,len(array)-1)
+        array[i],array[j]=array[j],array[i]
+    return array
+
+n=int(input('Enter Size of Array: '))
```
cozek: taking input from stdin is not recommended. Please read CONTRIBUTING.md
Perhaps you can just use a small list as input instead, for demonstrating this.

Is it possible to use fix the random seed for adding doctests to your functions?

cclauss:
Guessing randomly is not an algorithm.

Issue #1317, opened 2019-10-08
add python program that surfs the web.


archu5:

Kunal614:
It is showing merging is block

cclauss:
Travis CI tests need to be passing.  See CONTRIBUTING.md for details.

Kunal614:
didn't get , can you help.

mohanish2504:
Hi,

I am a beginner in open source contribution, I will like to contribute! please consider the following reply.
There are 3 Libraries in python already which can help in surfing the web:
 1.Selenium Web Driver
 2.Scrappy
 3.Beautiful soup
It will still great if you can provide some explanations on kind of surfing you want to perform.

onlinejudge95:
@mohanish2504 we currently have some `web crawling` programs in [web_programming](https://github.com/TheAlgorithms/Python/tree/master/web_programming) directory.

If you are interested, do check them and raise a PR if you want to add any new program or propose a new feature in the currently listed ones.

For more information about how to contribute to this repo please read the [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md)

PR #1319, opened 2019-10-08
solves #1317

archu5:


cozek:
Hi @archu5. I think you have misunderstood what network flow is.
Please read [this](https://en.wikipedia.org/wiki/Network_flow_problem).
Thank you for your time.

PR #1320, opened 2019-10-08
Update doubly_linked_list.py

higgsboson1209:
Prints the list contents in reverse


data_structures/linked_list/doubly_linked_list.py
```diff
@@ -68,6 +68,12 @@ def display(self):  # Prints contents of the list
             current.displayLink()
             current = current.next
         print()
+    def displayreverse(self): #prints contents of the list in reverse
```
cclauss: Please add an underscore like __display_reverse()__ and at least one doctest.

PR #1325, opened 2019-10-09
Added Armstrong Number

rajat1812:
I have added Armstrong number in the math module


rajat1812:
Sir I will try to fix the code tonight.

PR #1329, opened 2019-10-09
Added Self Organizing Map

rajat1812:
I have created a simple SOM for the  fraud detection


cozek:
Thank you for your contribution.
Some observations:
1. You did not run flake as asked in the CONTRIBUTION.md. Please read it. This is failing the automated test.
2. You have used a non-standard module and dataset. Is the dataset free for use/distribution? Given the non-generalization of the use cases of SOM, I think it would be better if you could implement it on something such as the iris dataset instead. [Read this](https://github.com/scikit-learn/scikit-learn/issues/2892).
3. Implementing it from scratch on the iris dataset would be much more educational and suitable for this repo.

PR #1332, opened 2019-10-09
Modified longest_common_subsequence.py for successful doctests

jpg-130:


cozek:
Doctests will still be successful without that.
try `python3 -m doctest file.py -v`

PR #1333, opened 2019-10-09
Adding doctests for sum_of_subset.py

jpg-130:


dynamic_programming/sum_of_subset.py
```diff
@@ -22,14 +27,9 @@ def isSumSubset(arr, arrLen, requiredSum):
     # uncomment to print the subset
     # for i in range(arrLen+1):
     #     print(subset[i])
+    print(subset[arrLen][requiredSum])
```
cozek: Unnecessary . Try this.
take test.py as
```
def returns_true():
    """
    >>> returns_true()
    True
    """
    return True
```

then
`python3 -m doctest test.py -v`

PR #1345, opened 2019-10-11
SieveOfEratosthenes.py

swatiprajapati08:
prime numbers less than N in O(N)


cozek:
Hi @swatiprajapati08. It says the code is contributed by Rituraj Jain on line 74. But your name is different. I am just confused.
Plus there is already a sieve in this repository. How is yours better or different?
Plus, have you read CONTRIBUTION.md?

PR #1351, opened 2019-10-12
update non-recursive factorial

yuriimchg:
- wrapped the algorithm with function;
- add docstring;
- renamed variables / constants.


cclauss:
https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings does not encourage overindenting docstrings.

PR #1369, opened 2019-10-15
Update dbscan.py

MittalS211:
Corrected error that variable was defined multiple times.


cclauss:
The existing usage is correct.

PR #1370, opened 2019-10-15
Update polynom-for-points.py

MittalS211:
Fixed infinite loop in line 70 by commenting line 71


cclauss:
Could you please add one or more doctests that fail because there is an infinite loop.

PR #1371, opened 2019-10-15
Update interpolation_search.py

MittalS211:
Corrected unreachable code at line 129


cclauss:
Let’s leave debug turned off.  The l’inter is just being too picky in this case.

PR #1372, opened 2019-10-15
Added addition function

Dheemanth-gowda:


cclauss:
Nice but this function does not really help in this file and it does not return the sum value to its caller.

PR #1374, opened 2019-10-15
Added Pytests for Decission Tree mean_squared_error method

hash84:
Modified the mean_squared_error to be a static method

Created the Test_Decision_Tree class
Consists of two methods
1. **helper_mean_squared_error_test**: This method calculates the mean squared error manually without using numpy. Instead, a for loop is used for the same.
2. **test_one_mean_squared_error**: This method considers a simple test case and compares the results by the helper function and the original mean_squared_error method of Decision_Tree class. This is done using asert
keyword.

Execution:

PyTest installation
```
pip3 install pytest 
```
OR 
```
pip install pytest
```

Test function execution
```
pytest decision_tree.py
```

### DIFF
--- a/machine_learning/decision_tree.py
+++ b/machine_learning/decision_tree.py
@@ -15,6 +15,7 @@ class Decision_Tree:
         self.min_leaf_size = min_leaf_size
         self.prediction = None
 
+    @staticmethod
     def mean_squared_error(self, labels, prediction):
         """
         mean_squared_error:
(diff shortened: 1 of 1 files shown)


hash84:
@cclauss I have done the required changes and the pytest are being executed successfully in Travis CI

hash84:
Thanks a lot this is my first pull request and first to be merged!

PR #1377, opened 2019-10-16
Update Gaussian.py

agrdivyam99:
Gaussian Function corrected


cclauss:
Please add doctests that fail on the current implementation but pass on the proposed implementation.

PR #1389, opened 2019-10-18
Python program that surfs  3 site at a time

Kunal614:
add input in the compiling time  like --  python3 project1.py (man)


project1.py
```diff
@@ -0,0 +1,21 @@
+from bs4 import BeautifulSoup
+import requests
+#from urllib.request import urlopen
+import webbrowser
+import sys
+from fake_useragent import UserAgent
```
cclauss: https://travis-ci.com/TheAlgorithms/Python/builds/132468677#L334

Kunal614: Please explain I don't understand what to do and how to do.

Kunal614: I did't get error

cclauss: Comment out the line that it is complaining about and then see if the script still runs.  flake8 is saying that the import is not needed.

Kunal614:
it is needed otherwise program will work for 2 - 3 times then stop working.

project1.py
Kunal614: > Comment out the line that it is complaining about and then see if the script still runs. flake8 is saying that the import is not needed.

commented

PR #1390, opened 2019-10-18
Update g_topological_sort.py

amadionix:
I had a version which I think is better than the one which was here. It also handles the case in which there are cycles in the graph, giving as output an empty list (= impossible scheduling). It also has a graph class which I think it's nice to have there. I think it can be improved, I was actually curious if somebody had a better version of it.


cclauss:
No doctests?  No typehints?  See CONTRIBUTING.md for details.

Please provide a https://docs.python.org/3.8/library/timeit.html test that demonstrates that this is faster than the original or a testcase to prove that this is more correct than the original.

PR #1393, opened 2019-10-18
added runge-kutta

infrontoftheforest:


maths/runge_kutta.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def runge_kutta(f, y0, x0, h, x_end):
+    """
+    Calculate the numeric solution at each step to the ODE f(x, y) using RK4
+
+    https://en.wikipedia.org/wiki/Runge-Kutta_methods
+
+    Arguments:
+    f -- The ode as a function of x and y
```
cclauss: ode_func instead of f?
stepsize instead of h?

poyea: It's fine. The math-like notation is used.

PR #1394, opened 2019-10-18
added explicit euler's method

infrontoftheforest:


cclauss:
The single character variable name are so old school.  Is there anything that could be done to make them more readable?

infrontoftheforest:
I've updated the variable names

cclauss:
Nice!

PR #1396, opened 2019-10-18
Create dinic.py

danielx285:
Dinic's algorithm for maximum flow

### DIFF
--- a/graphs/dinic.py
+++ b/graphs/dinic.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python2.7
 INF = float("inf")
 
 class Dinic:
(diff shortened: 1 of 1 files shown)


graphs/dinic.py
```diff
@@ -0,0 +1,54 @@
...
+        self.lvl = [0] * n
+        self.ptr = [0] * n
+        self.q = [0] * n
+        self.adj = [[] for _ in range(n)]
+
+    def add_edge(self, a, b, c, rcap=0):
+        self.adj[a].append([b, len(self.adj[b]), c, 0])
+        self.adj[b].append([a, len(self.adj[a]) - 1, rcap, 0])
+
+    def dfs(self, v, t, f):
```
cclauss: Please make API understandable to the first time user.  In of `dfs(self, v, t, f)`, something more like `determine_fastest_surfer(self, viscosity, timber, finances)`as discussed in CONTRIBUTING.md.  Also, these methods/functions need doctests.

graphs/dinic.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python2.7
```
cclauss: Please remove line 1.  This repo no longer supports legacy Python.

graphs/dinic.py
danielx285: Ok, I will change the file.

cclauss:
https://docs.python.org/3/library/doctest.html is what is needed on key functions/methods.

Can we have a better name for `rcap` ?

PR #1401, opened 2019-10-19
The time complexity of every algorithms make its value 

SinghSujitkumar:
A timer has been added which calculates the time of execution of a sort , i.e. bubble sort.

### DIFF
--- a/sorts/bubble_sort.py
+++ b/sorts/bubble_sort.py
@@ -1,5 +1,5 @@
 import time
-start = time.process_time()
+start = 0
 def bubble_sort(collection):
     """Pure implementation of bubble sort algorithm in Python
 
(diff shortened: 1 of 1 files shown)


sorts/bubble_sort.py
```diff
@@ -37,3 +39,4 @@ def bubble_sort(collection):
     user_input = input("Enter numbers separated by a comma:").strip()
     unsorted = [int(item) for item in user_input.split(",")]
     print(*bubble_sort(unsorted), sep=",")
```
cclauss: We are also counting the time that the user enters the data.  Let’s only count sort time...

Please move line 2 to be just above this line and also indent line 42.

SinghSujitkumar: Okay . Thanks a lot for the obersvation. Will do the changes

PR #1407, opened 2019-10-20
i add some thing in README just for try

yichung-cheng:


cclauss:
If it was content that would help contributors then I would merge it.

PR #1410, opened 2019-10-20
Factorial of a number.

Aayush-Vats:


cclauss:
You might need to set some limits on the input...
```python
# Python program to find the factorial of a given number
def factorial(n):
    return 1 if n in (0, 1) else n * factorial(n - 1)


for num in (1, 5, 100, 500, 1000, 1.1, -1, 10000):
    print(f"Factorial of {num} is {factorial(num)}")
```

maths/Factorial
```diff
@@ -0,0 +1,12 @@
+# Python 3 program to find  
+# factorial of given number 
+def factorial(n): 
+      
+    # single line to find factorial 
+    return 1 if (n==1 or n==0) else n * factorial(n - 1);
```
onlinejudge95: This code is functional for relatively small values of `n`.

Try finding factorial of `1000`

PR #1416, opened 2019-10-21
added solution 7 for problem_01

suadjelili:
I have added solution 7 for problem 01.


project_euler/problem_01/sol7.py
```diff
@@ -0,0 +1,32 @@
+"""
```
cclauss: Please edit the filename to be `problem_07/sol7.py` so that it is in the appropriate directory.

project_euler/problem_02/sol5.py
```diff
@@ -0,0 +1,46 @@
+"""
```
cclauss: Please edit the filename to be `problem_05/sol5.py` so that it is in the appropriate directory.

PR #1419, opened 2019-10-22
Project Euler problems 06, 20

chttrjeankr:
Added new solutions for problem_06 and problem_20 for project_euler

### DIFF
--- a/project_euler/problem_20/sol3.py
+++ b/project_euler/problem_20/sol3.py
@@ -1,0 +1,39 @@
+"""
+n! means n × (n − 1) × ... × 3 × 2 × 1
+
+For example, 10! = 10 × 9 × ... × 3 × 2 × 1 = 3628800,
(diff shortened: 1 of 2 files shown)


project_euler/problem_20/sol3.py
```diff
@@ -0,0 +1,39 @@
...
+    >>> solution(3)
+    6
+    >>> solution(2)
+    2
+    >>> solution(1)
+    1
+    >>> solution(0)
+    1
+    """
+    return sum(map(int, str(factorial(n))))
```
cclauss: sum(int(x) for x in str(factorial(0))) as discussed in CONTRIBUTING.md

chttrjeankr: There already exists such an implementation using list comprehension in [sol2.py](https://github.com/TheAlgorithms/Python/blob/906c985de3b0156ac50a1bd10d0c803f3439cf4e/project_euler/problem_20/sol2.py#L29)
`return sum([int(x) for x in str(factorial(n))])`

Should I then remove my solution considering it redundant?

PR #1420, opened 2019-10-22
Add algorithm to rotate images

matheustguimaraes:


digital_image_processing/rotation/rotation.py
```diff
@@ -0,0 +1,50 @@
...
+        get_rotation(gray_img, pts1, pts2, img_rows, img_cols),
+        get_rotation(gray_img, pts2, pts3, img_rows, img_cols),
+        get_rotation(gray_img, pts2, pts4, img_rows, img_cols),
+    ]
+
+    # plot different image rotations
+    fig = plt.figure(1)
+    titles = ["Original", "Rotation 1", "Rotation 2", "Rotation 3"]
+    for i in range(0, len(images)):
+        plt.subplot(2, 2, i + 1), plt.imshow(images[i], "gray")
```
cclauss: for i, image in enumerate(images):
    plt.subplot(2, 2, i + 1), plt.imshow(image, "gray")

digital_image_processing/rotation/rotation.py
```diff
@@ -0,0 +1,50 @@
+from matplotlib import pyplot as plt
+from numpy import float32
+import cv2
+
+
+def get_rotation(img, pt1, pt2, rows, cols):
```
cclauss: Type hints?  https://docs.python.org/3/library/typing.html

digital_image_processing/rotation/rotation.py
matheustguimaraes: Added enumerate in loop.

digital_image_processing/rotation/rotation.py
matheustguimaraes: Added type hints in get_rotation(). Thanks for the help!

PR #1421, opened 2019-10-22
psf/black code formatting

chttrjeankr:
ran `black .` on directory `\Python`


chttrjeankr:
> In the future, please do not include reformatting in the same PR with code changes.... It makes your reviewer go cross-eyed.

will surely keep that mind :sweat_smile:

PR #1428, opened 2019-10-24
Added binary exponentiaion with respect to modulo

Anzoteh96:


maths/miller_rabin.py
```diff
@@ -0,0 +1,46 @@
+from binary_exp_mod import bin_exp_mod
```
cclauss: https://travis-ci.com/TheAlgorithms/Python/builds/133302996#L347

Try adding a . like `from .binary_exp_mod import bin_exp_mod`

maths/miller_rabin.py
```diff
@@ -0,0 +1,46 @@
...
+import random
+
+# This is a probabilistic check to test primality, useful for big numbers!
+def is_prime(n, prec = 1000):
+  if (n < 2):
```
cclauss: Please add one or more doctests here as discussed in CONTRIBUTING.md.

maths/miller_rabin.py
```diff
@@ -0,0 +1,46 @@
...
+if __name__ == '__main__':
+  n = int(input("Enter bound : ").strip())
+  primes = []
+  for i in range(n + 1):
+    if is_prime(i):
+      primes.append(i)
+  print("Here's the list of primes")
+  for p in primes:
+    print(p, end=" ")
+  print()
```
cclauss:
```python
if __name__ == '__main__':
    n = abs(int(input("Enter bound : ").strip()))
    print("Here's the list of primes:")
    print(", ".join(str(i) for i in range(n + 1) if is_prime(i)))
```

maths/miller_rabin.py
Anzoteh96: Added a test on it

maths/miller_rabin.py
Anzoteh96: Did the change

cclauss:
This file is not being tested https://travis-ci.com/TheAlgorithms/Python/builds/133415919#L417

PR #1448, opened 2019-10-25
Add gaussian_elimination.py for solving linear systems

isaacgdo:
This code implements the Gaussian elimination method for solving linear systems.

- The gaussian elimination function scales the matrix representing the linear system into a triangular matrix.

- The retroactive resolution function, in turn, uses the triangular matrix to find each of the solution's variables from bottom to top.


isaacgdo:
Thank you, @cclauss. I intend to submit more codes soon.

PR #1452, opened 2019-10-25
updated factorial_python

percy07:
Optimized the factorial code.

### DIFF
--- a/graphs/dfs.py
+++ b/graphs/dfs.py
@@ -1,44 +1,60 @@
-"""pseudo-code"""
+# Python program to print DFS traversal for complete graph 
+from collections import defaultdict 
 
-"""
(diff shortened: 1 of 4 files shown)


cclauss:
Sorry but I am going to close this one.  Please do not be discouraged and please contribute again after reading CONTRIBUTING.md/

This PR touches too many files that are not related to each other.  Three separate PRs would be much easier to review.  The goal of this repo is to demonstrate algorithms and the logic of how they work and how they can be optimized to go faster and/or take less memory.  We all know that `max()` is a Python `builtin` and that `math.factorial()` is readily available in the Python Standard Library but many files in this repo reconstruct those functions from scratch to demonstrate the logic and trade-offs that go into implementing those algorithms.

PR #1453, opened 2019-10-25
Adding deterministic miller rabin primality test

bizzfitch:
There's a miller-rabin primality test in the cipher folder, but I figured a deterministic one for "small" numbers would be interesting. Deterministic (meaning True is prime) for n < ~ 3.32 e 24, probabilistic for larger. With Miller Rabin, a False is always composite- no false negatives.

See https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test for details.

### DIFF
--- a/maths/miller_rabin.py
+++ b/maths/miller_rabin.py
@@ -1,0 +1,133 @@
+"""Created by Nathan Damon, @bizzfitch on github"""
+
+
+def miller_rabin(n, allow_probable=False):
+    """Deterministic Miller-Rabin algorithm for primes ~< 3.32e24.
+
+    Uses numerical analysis results to return whether or not the passed number
+    is prime. If the passed number is above the upper limit, and
+    allow_probable is True, then a return value of True indicates that n is
+    probably prime. This test does not allow False negatives- a return value
+    of False is ALWAYS composite.
+
+    Parameters
+    ----------
+    n : int
+        The integer to be tested. Since we usually care if a number is prime,
+        n < 2 returns False instead of raising a ValueError.
+    allow_probable: bool, default False
+        Whether or not to test n above the upper bound of the deterministic test.
+
+    Raises
+    ------
+    ValueError
+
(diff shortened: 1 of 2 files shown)


cclauss:
The tests are not being discovered by pytest
https://travis-ci.com/TheAlgorithms/Python/builds/133548451#L425

Why not put both miller-rabin implementations in the same directory with different names to ease side-by-side study?

bizzfitch:
Any idea what I did wrong with the tests? I put them under test_miller_rabin, which runs under main. 

That's a good point. My thought process was that since this test is deterministic for most computable numbers, it can be used as a general fast primality test. I also don't have any additional implementation that made me think of putting it in the cipher folder. If you think it's best, I'm happy to move it, because I'm not convinced it belongs in maths :)

cclauss:
I am not sure on the test thing.  It has been bugging me for a while now.  I am no expert on [pytest discovery rules](http://doc.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery) but perhaps @codecalec could give us guidance on why is works in [his PRs].

cclauss:
Ahh...  Perhaps the test_function() needs to be in a file with the name of the form `test_***.py`.

maths/miller_rabin.py
```diff
@@ -0,0 +1,133 @@
...
+            continue
+        # if pr is False, then the above loop never evaluated to true,
+        # and the n MUST be composite
+        return False
+    return True
+
+
+def test_miller_rabin():
+    """Testing a nontrivial (ends in 1, 3, 7, 9) composite
+    and a prime in each range.
```
cclauss: Since we run doctests. you can trick `pytest` into running these tests just by adding `>>> test_miller_rabin()` inside this docstring.

codecalec: This seems to be what is in other methods in the maths folder.

PR #1454, opened 2019-10-25
Added Isotonic regression script in Python (Machine Learning)

RitwickGhosh:


cclauss:
CONTRIBUTING.md is quite clear that PRs should not modify DIRECTORY.md.  Our automated tests fail and we do not support plagiarism.   https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/_isotonic.pyx

PR #1458, opened 2019-10-25
Update odd_even_transposition_parallel.py

samarthsehgal97:


sorts/odd_even_transposition_parallel.py
```diff
@@ -144,7 +144,7 @@ def main():
     print("Initial List")
     print(*arr)
 
-    list = OddEvenTransposition(arr)
+
```
cclauss: Can you please do two things?
1. Convert 140-143 into a one liner using a list comprehension.
2. Put line 147 back in and replace `list` with `arr`.

PR #1459, opened 2019-10-25
Update a_star.py

samarthsehgal97:


cclauss:
Why should this change be made?  Reviews please from:
@WilliamHYZhang
@vysor
@Reshad-Hasan
@AlexDvorak

AlexDvorak:
I understand why, because it is not set anywhere else. But it should either be set somewhere in the main loop, or remove the reference to it in the main while loop.

PR #1463, opened 2019-10-25
Update linear_search.py

percy07:
linear search updated.

### DIFF
--- a/searches/linear_search.py
+++ b/searches/linear_search.py
@@ -31,10 +31,12 @@ def linear_search(sequence, target):
     >>> linear_search([0, 5, 7, 10, 15], 6)
 
     """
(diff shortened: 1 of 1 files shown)


searches/linear_search.py
```diff
@@ -31,12 +31,14 @@ def linear_search(sequence, target):
     >>> linear_search([0, 5, 7, 10, 15], 6)
 
     """
-    for index, item in enumerate(sequence):
-        if item == target:
-            return index
+    
+    for i in range(len(sequence)):
+        if sequence(i) ==target:
```
chttrjeankr: if sequence[i] == target:

cclauss:
https://docs.python.org/3/library/functions.html#enumerate

PR #1464, opened 2019-10-25
circular linkedlist program

Dhyani01:
circular linked list program is added !

### DIFF
--- a/data_structures/linked_list/circular_linked_list.py
+++ b/data_structures/linked_list/circular_linked_list.py
@@ -1,0 +1,93 @@
+class Node:
+    def __init__(self, data):
+        self.data = data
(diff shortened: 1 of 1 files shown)


data_structures/linked_list/circular_linked_list.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # This function is only for empty list
+    def addToEmpty(self, data):
+
+        if (self.last != None):
+            return self.last
+
+        # Creating the newnode temp
+        temp = Node(data)
+        self.last = temp
```
cclauss: Why create a temporary variable that we never use again?

onlinejudge95:
Do we need this

Issue #1368, opened 2019-10-15
decimal_to_octal return


afeiguo:
the return code is wrong for big number, e.g.:

`print("{0:g}".format(float(1000000001.0)))`

the return is **1e+09**, not **'1000000001'**

afeiguo:
> the return code is wrong for big number, e.g.:
> 
> print("{0:g}".format(float(1000000001.0)))
> 
> the return is 1e+09, not '1000000001'

coding should change:   return "{0:g}".format(float(octal))           =>
                return  str(int(octal))

blaxii:
You know you could justt do 
`def decimal_to_octal(num):
    """Convert a Decimal Number to an Octal Number."""
    calcul = ""
    while num > 0:
        reste = num % 8
        num = num // 8
        calcul = str(reste) + calcul
    return int(calcul)`

blaxii:
If you want you can do:

> def decimal_to_octal(num):

    """Convert a Decimal Number to an Octal Number."""
    calcul = ""
    while num > 0:
        rest = num % 8
        num = num // 8
        calcul = str(rest) + calcul
    return int(calcul)

As I see your code is too complex for the program you want to do.

cclauss:
num, rest = divmod(num, 8)

onlinejudge95:
The issue has been fixed here https://github.com/TheAlgorithms/Python/pull/1482. Hence closing the issue.

PR #1482, opened 2019-10-28
Fixed issue #1368

Jawpral:
The return value for large numbers is now not in scientific notation.

### DIFF
--- a/conversions/decimal_to_octal.py
+++ b/conversions/decimal_to_octal.py
@@ -6,8 +6,14 @@ import math
 # https://github.com/TheAlgorithms/Javascript/blob/master/Conversions/DecimalToOctal.js
 
 
-def decimal_to_octal(num):
-    """Convert a Decimal Number to an Octal Number."""
+def decimal_to_octal(num: int) -> str:
+    
+    """Convert a Decimal Number to an Octal Number.
+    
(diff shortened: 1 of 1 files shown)


conversions/decimal_to_octal.py
```diff
@@ -16,7 +16,7 @@ def decimal_to_octal(num):
         counter += 1
```
cclauss: Let’s add some doctests because this function is not quite right in a few ways...  Please change the top of the function to:
```python
def decimal_to_octal(num: int) -> str:

    """Convert a Decimal Number to an Octal Number.
    
    >>> for i in (0, 2, 8, 64, 65, 216, 255, 256, 512):
    ...     decimal_to_octal(i) == oct(i)
    True
    """
```
Then change line 19 to something like `return f"0o{int(octal)}"` and then do `python3 -m doctest -v conversions/decimal_to_octal.py` to make sure those tests pass.

Jawpral: The test case fails, but when I do it manually there is no error as such

Jawpral: I changed the code as suggested nevertheless

cclauss:
https://travis-ci.com/TheAlgorithms/Python/builds/133787349#L551

PR #1486, opened 2019-10-28
Added a brute force program for Transposition cipher

LawJarp-A:
Cracking the Transposition cipher with brute force

### DIFF
--- a/ciphers/brute_force_transposition_cipher.py
+++ b/ciphers/brute_force_transposition_cipher.py
@@ -1,0 +1,29 @@
+import math
+
+def decryptMessage(key, message):
+
+    numCols = math.ceil(len(message) / key)
+    numRows = key
+    numShadedBoxes = (numCols * numRows) - len(message)
+    plainText = [""] * numCols
+    col = 0
+    row = 0
+
+    for symbol in message:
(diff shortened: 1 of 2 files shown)


onlinejudge95:
@Jawpral are you working on it?

ciphers/brute_force_transposition_cipher.py
```diff
@@ -0,0 +1,29 @@
+import math
+
+def decryptMessage(key, message):
```
onlinejudge95: follow consistent conventions for naming a function, either use **snake_case** or use **camelCase**, since the repo is `Python` you should only be using **snake_case**

conversions/decimal_to_octal.py
```diff
@@ -16,7 +16,7 @@ def decimal_to_octal(num):
         counter += 1
         num = math.floor(num / 8)  # basically /= 8 without remainder if any
         # This formatting removes trailing '.0' from `octal`.
-    return "{0:g}".format(float(octal))
+    return int(octal)
```
onlinejudge95: You know right, we want to convert a **decimal** to an **octal**

PR #1501, opened 2019-10-28
rename and add doctest 

shellhub:
@cclauss

### DIFF
--- a/maths/find_max.py
+++ b/maths/find_max.py
@@ -10,11 +10,11 @@ def find_max(nums):
     True
     True
     """
-    max = nums[0]
+    max_num = nums[0]
     for x in nums:
-        if x > max:
-            max = x
-    return max
+        if x > max_num:
+            max_num = x
+    return max_num
 
 
 def main():
--- a/maths/find_min.py
+++ b/maths/find_min.py
@@ -1,17 +1,25 @@
-"""Find Minimum Number in a List."""
+def find_min(nums):
+    """
+    Find Minimum Number in a List
+    :param nums: contains elements
+    :return: max number in list
+
+    >>> for nums in ([3, 2, 1], [-3, -2, -1], [3, -3, 0], [3.0, 3.1, 2.9]):
+    ...     find_min(nums) == min(nums)
+    True
+    True
+    True
+    True
+    """
+    min_num = nums[0]
+    for num in nums:
+        if min_num > num:
+            min_num = num
+    return min_num
 
 
 def main():
-    """Find Minimum Number in a List."""
-
-    def find_min(x):
-        min_num = x[0]
-        for i in x:
-            if min_num > i:
-                min_num = i
-        return min_num
-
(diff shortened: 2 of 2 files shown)


maths/find_max.py
```diff
@@ -1,16 +1,21 @@
-# NguyenU
```
cclauss: This file is out of sync with `master`.  There are already doctests on this file:
https://github.com/TheAlgorithms/Python/blob/master/maths/find_max.py#L6

shellhub: @cclauss Please review again!

cclauss:
Please write a https://docs.python.org/3/library/timeit.html that proves that the new implementation is faster than the current implementation.  My bet is that the function and slice overhead means that it is not.

shellhub:
> Please write a https://docs.python.org/3/library/timeit.html that proves that the new implementation is faster than the current implementation. My bet is that the function call and slice overhead means that it is not.

Did you mean below code implementation is slower ?
``` python
    max_num = nums[0]
    for i in range(1, len(nums)):
        if nums[i] > max_num:
            max_num = nums[i]
```

cclauss:
Not sure but that is my _suspicion_ but a timeit test can tell us for sure.

shellhub:
After I tested using 
``` python
def find_max(nums):
    max_num = nums[0]
    for i in range(1, len(nums)):
        if nums[i] > max_num:
            max_num = nums[i]
    return max_num
print(timeit.timeit("find_max([2, 4, 9, 7, 19, 94, 5])", setup="from __main__ import find_max", number=10000))
time: 0.022231400000000002
```

``` python
def find_max(nums):
    max = nums[0]
    for x in nums:
        if x > max:
            max = x
    return max
print(timeit.timeit("find_max([2, 4, 9, 7, 19, 94, 5])", setup="from __main__ import find_max", number=10000))
time: 0.007430200000000005
```

From the result, You are right. Thanks

cclauss:
This is awesome work.  I think we should fill this file with as many different implementations of __find_max()__ as we can think of with timeit tests for each one in an effort to find that fastest pure Python implementation.

How does `print(timeit.timeit("max([2, 4, 9, 7, 19, 94, 5])", number=10000))` perform using the Python builtin?

shellhub:
> print(timeit.timeit("max([2, 4, 9, 7, 19, 94, 5])", number=10000))

`0.004652999999999997` , Very fast

maths/find_min.py
```diff
@@ -1,17 +1,21 @@
-"""Find Minimum Number in a List."""
+def find_min(nums):
+    """
+    Find Minimum Number in a List
+    :param nums: contains elements
+    :return: max number in list
 
+    >>> find_min([1, 3, 5, 7, 9]) == 1
```
cclauss: Let's use the same tests data as find_max.py uses.

shellhub: Updated. Please review again

PR #1503, opened 2019-10-28
Improve doctest and comment for maximum sub-array problem

poyea:
More examples and comment description for `max_sub_array.py`.

### DIFF
--- a/dynamic_programming/max_sub_array.py
+++ b/dynamic_programming/max_sub_array.py
@@ -2,9 +2,6 @@
 author : Mayank Kumar Jha (mk9440)
 """
 from typing import List
-import time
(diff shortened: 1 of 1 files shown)


dynamic_programming/max_sub_array.py
```diff
@@ -64,6 +71,9 @@ def max_sub_array(nums: List[int]) -> int:
 
 
 if __name__ == "__main__":
+    """
+    A random simulation of this algorithm.
+    """
```
cclauss: Let's move the `matplotlib` import down here because it is only needed in the `__main__` section and that third party library won't be imported when this file is used as a library of functions.

PR #1509, opened 2019-10-29
Update palindrome.py

percy07:
Add Doctests.

### DIFF
--- a/other/palindrome.py
+++ b/other/palindrome.py
@@ -1,5 +1,25 @@
 # Program to find whether given string is palindrome or not
 def is_palindrome(str):
+    """
+    This function return True is given string is Palindrom other wise return False.
+
+    >>> is_palindrome('MALAYALAM')
+    True
+    >>> is_palindrome('String')
+    False
+    >>> is_palindrome('rotor')
+    True
+    >>> is_palindrome('level')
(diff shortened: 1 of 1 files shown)


other/palindrome.py
```diff
@@ -1,5 +1,25 @@
 # Program to find whether given string is palindrome or not
 def is_palindrome(str):
```
cclauss: `str` is a builtin in Python so this usage "shadows" that builtin which is not a good idea.  Can you please change all occurrences of `str` with just plain `s`?

Also, please change the function signature by adding Python type hints: `def is_palindrome(s: str) -> bool:` and add similar type hints to the function `recursive_palindrome()`.

Let's add another function:
```python
def slice_palindrome(s: str) -> bool:
    return s == s[::-1]
```

other/palindrome.py
```diff
@@ -1,5 +1,25 @@
...
+    >>> is_palindrome('level')
+    True
+    >>> is_palindrome('A')
+    True
+    >>> is_palindrome('BB')
+    True
+    >>> is_palindrome('ABC')
+    False
+    """
+
```
cclauss: Please remove one of these two blank lines.

PR #1512, opened 2019-10-29
Update requirements.txt

percy07:
add two packages.
1. collections
2. itertools


cclauss:
These modules are already in the Python Standard Library.

PR #1517, opened 2019-10-29
Update basic_maths.py

percy07:
Add Doctests.

### DIFF
--- a/maths/basic_maths.py
+++ b/maths/basic_maths.py
@@ -2,64 +2,56 @@
 import math
 
 
-def prime_factors(n):
+def prime_factors(n: int) -> list:
     """Find Prime Factors.
-    >>> prime_factors(10)
-    [2,5]
+    >>> prime_factors(100)
+    [2, 2, 5, 5]
     """
     pf = []
     while n % 2 == 0:
         pf.append(2)
         n = int(n / 2)
-
     for i in range(3, int(math.sqrt(n)) + 1, 2):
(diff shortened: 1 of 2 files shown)


maths/basic_maths.py
```diff
@@ -41,7 +47,10 @@ def number_of_divisors(n):
 
 
 def sum_of_divisors(n):
-    """Calculate Sum of Divisors."""
+    """Calculate Sum of Divisors.
+    >>> sum_of_divisors(72)
+    195
```
cclauss: 195.0

maths/basic_maths.py
```diff
@@ -3,7 +3,10 @@
 
 
 def prime_factors(n):
-    """Find Prime Factors."""
+    """Find Prime Factors.
+    >>> prime_factors(10)
+    [2,5]
```
cclauss: [2, 5]  # doctests care about whitespace.

cclauss:
Please add type hints on all functions:
* `def prime_factors(n: int) -> list:`
* `def number_of_divisors(n: int) -> int:`
* `def sum_of_divisors(n: int) -> float:`  # or should it return an int?
* `def euler_phi(n: int) -> float:`  # or should it return an int?

cclauss:
It would be much cooler if your could make sure that these doctests pass on your own computer _before_ submitting the pull request: `python3 -m doctest -v maths/basic_maths.py`

PR #1523, opened 2019-10-30
Update quick_select.py

percy07:
Add Doctests.

### DIFF
--- a/searches/quick_select.py
+++ b/searches/quick_select.py
@@ -26,6 +26,16 @@ def _partition(data, pivot):
 
 
 def quickSelect(list, k):
+    """
+    >>> quickSelect([2,4,5,7,899,54,32], 5)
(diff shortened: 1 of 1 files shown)


searches/quick_select.py
```diff
@@ -26,6 +26,16 @@ def _partition(data, pivot):
 
 
 def quickSelect(list, k):
```
cclauss: Please change the function signature to be `def quick_select(items: list, index: int):`  This does the following...
1. Use a Pythonic function name instead of a function name from JavaScript.
2. Change the parameter __list__ to be __items__ because __list__ is a Python builtin data type so it is not a suitable variable name.
3. Add typehints for the list and the int
4. Change the parameter __k__ to be __index__ because that name is more self-documenting.

PR #1524, opened 2019-10-30
Fractional knapsack

ptarun:


cclauss:
Please look at the changes that have I made.  Please replace the URLs if you have better sources.  The doctest results surprise me because the `capacity` parameter is largely being ignored.  Are the doctest results what was expected?

poyea:

cclauss:
This is landed but...
> The doctest results surprise me because the capacity parameter is largely being ignored. Are the doctest results what was expected?_

PR #1527, opened 2019-10-30
Add log function

MarkMoretto:
I didn't see a file covering the log function.  This one takes a base argument with the default being 10.

The code was run through black, flake8, and doctest per contribution guidelines.

### DIFF
--- a/maths/log_function.py
+++ b/maths/log_function.py
@@ -1,0 +1,37 @@
+"""Logarithm function"""
+
+
+def log(x, base=10):
+    """
+    This function returns the log value of a given number.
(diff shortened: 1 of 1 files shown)


cclauss:
Please add tests for: log(-1), log(0), log(0.1), log(5), log(10_000)
The output should match `math.log()` in all cases.

```python
import math
print([math.log(x) for x in (0.1, 1, 5, 10, 100, 1_000, 10_000)])
```

MarkMoretto:
Hi and thanks!

Are we allowed to use unittest?  I don't know of another way to test an exception being raised or thrown.

PR #1530, opened 2019-10-31
Create Perfect Number

percy07:
A number is a perfect number if is equal to sum of its proper divisors, that is, sum of its positive divisors excluding the number itself.


cclauss:
Closing because:
* Filename does not end in `.py`
* Filename contains spaces
* Filename contains UPPERCASE characters
* no URL to source describing the algorithm
* no doctests
* no type hints
* code not formatted with psf/black

PR #1534, opened 2019-10-31
perfect square

shellhub:
* add perfect square number

### DIFF
--- a/maths/perfect_square.py
+++ b/maths/perfect_square.py
@@ -1,7 +1,7 @@
 import math
 
 
-def perfect_square(num):
+def perfect_square(num: int) -> bool:
     """
     Check if a number is perfect square number or not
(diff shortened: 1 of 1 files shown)


shellhub:
@cclauss Please review!

maths/perfect_square.py
```diff
@@ -0,0 +1,27 @@
+import math
+
+
+def perfect_square(num):
```
cclauss: https://docs.python.org/3/library/typing.html

shellhub: Like below code?
``` python
def perfect_square(num: int) -> bool:
```

cclauss: Perfect!

shellhub: updated! Please review again!

PR #1536, opened 2019-10-31
recursive quick sort

metehaansever:


sorts/recursive-quick-sort.py
```diff
@@ -0,0 +1,6 @@
+def quick_sort(l):
+    if len(l) <= 1:
+        return l
+    else:
+        return quick_sort([e for e in l[1:] if e <= l[0]]) + [l[0]] +\
+            quick_sort([e for e in l[1:] if e > l[0]])
```
cclauss:
```python
        return (quick_sort([e for e in l[1:] if e <= l[0]]) + [l[0]] +
                quick_sort([e for e in l[1:] if e > l[0]]))
```

cclauss: PEP8 recommends avoiding \ in Python code.

PR #1537, opened 2019-10-31
add max sum contigous subsequence

taufikalgi:


dynamic_programming/max_sum_contigous_subsequence.py
```diff
@@ -0,0 +1,17 @@
+def maxSubArraySum(nums):
+  if not nums: return 0
```
cclauss: Please add a doctest as discussed in __CONTRIBUTING.md__.

PR #1542, opened 2019-10-31
Travis CI: Write & print DIRECTORY.md on one line

mrvnmchm:
I have been successful at updating the DIRECTORY.md file. However, after multiple tests and failures I am unable to achieve pushing back to the PR that initiated the build. Apparently this is due to Travis CI not including secure environment variables in PR builds, see travis-ci/travis-ci#1500


.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: python
 python: 3.7
 cache: pip
-before_install: pip install --upgrade pip setuptools
+before_install: 
+  - pip install --upgrade pip setuptools
```
cclauss: Why this change?  How is it different than the current code?

.travis.yml
```diff
@@ -11,5 +12,4 @@ script:
   - mypy --ignore-missing-imports .
   - pytest . --doctest-modules
 after_success:
-  - scripts/build_directory_md.py > DIRECTORY.md
-  - cat DIRECTORY.md
+  - scripts/build_directory_md.py 2>&1 | tee DIRECTORY.md
```
cclauss: Why this change?  How is it different than the current code?

.travis.yml
mrvnmchm: Formatting leftovers during previous attempts of testing builds. I'll revert this.

.travis.yml
mrvnmchm: This change eliminates a line and does both operations. Printing the output, and writing it to the file, see [tee](https://en.wikipedia.org/wiki/Tee_(command))

chttrjeankr:
I see the Travis CI build didn't pass. Isn't that a problem?

cclauss:
I made a whitespace change to @mrvnmchm’s PR before landing it which confuses Travis CI but all is well.  https://travis-ci.com/TheAlgorithms/Python/builds/134508235#L170

PR #1544, opened 2019-11-01
Fix bellman_ford.py

JakobMusik:


cclauss:
There is no context to evaluate if the original is correct or this change is correct or both are wrong.  There are no doctests and no sample data and no URL to a description of the algorithm and some sample data.

Should the reader look at https://www.geeksforgeeks.org/bellman-ford-algorithm-dp-23/ or at https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm or what??

JakobMusik:
> There is no context to evaluate if the original is correct or this change is correct or both are wrong. There are no doctests and no sample data and no URL to a description of the algorithm and some sample data.
> 
> Should the reader look at https://www.geeksforgeeks.org/bellman-ford-algorithm-dp-23/ or at https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm or what??

In wikipedia, it says 'for each edge (u, v) with weight w in edges:' . To some extent, i think the code should be changed to fit this.

Issue #1460, opened 2019-10-25
DIRECTORY.md not being updated by TravisCI


chttrjeankr:
- [x] .travis.yml isn't updating DIRECTORY.md automatically as it should
- [x] scripts/build_directory_md.py needs can have some minor changes too.  #1461

cclauss:
If you can make Travis CI write back to the incomming pull request, that would be cool.  Writing back to `master` might put Travis into an infinite loop.  @jneeven is doing some magic in lgeiger/black-action#2 but that is based on GitHub Actions, not Travis CI.

mrvnmchm:
A fix for the DIRECTORY.md, can't get Travis CI to push back to the initiating PR. #1542

cclauss:
https://github.com/TheAlgorithms/Python/blob/master/.github/workflows/autoblack.yml Is writing changes back to a PR.  GitHub Actions Beta ends in two weeks and then it will go live,

mrvnmchm:
Nice! That looks very close to the same changes I was making to the travis.yml. When the GH Actions is released I'll make the updates. Thanks @cclauss

cclauss:
See if they will let you into the beta... https://github.com/features/actions

poyea:
@cclauss @mrvnmchm GitHub Actions is now available for all public repositories.

cclauss:
#1569

mrvnmchm:
Should I make any changes to #1549, or do I just need another peer review?

cclauss:
https://github.com/TheAlgorithms/Python/pull/1570/checks?check_run_id=303560970#step:4:550

cclauss:
Writing back to a PR is NOT allowed in GitHub Actions.  See the other workflow to see how to use __push__ instead.

chttrjeankr:
Can this issue now be considered resolved then?

cclauss:
This should remain open until #1572 lands.

PR #1546, opened 2019-11-01
Add autoDIR Action

mrvnmchm:
This Github action updates the DIRECTORY.md file and pushes back to GH, if it is necessary. FIles updated are from black.py and build_directory_md.py. Not sure if I need to adjust the GH auth.


mrvnmchm:
Added for Issue #1460

cclauss:
You need to rebase.  It is _much_ easier to review if each PR touches only one file.

mrvnmchm:
I can do that. Should I cancel this PR and resubmit after rebase?

cclauss:
> Should I cancel this PR?

If the PR touches more than one file then yes.

.github/workflows/autoDIR.yml
```diff
@@ -0,0 +1,29 @@
+name: autoDIR
```
cclauss: # The objective of this GitHub Action is to add a new DIRECTORY.md file to a pull request if needed.
name: directory_writer

.github/workflows/autoDIR.yml
```diff
@@ -0,0 +1,29 @@
...
+      matrix:
+        python-version: [3.7]
+    steps:
+
+      - uses: actions/checkout@v1
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Update DIRECTORY
```
cclauss: name: Update DIRECTORY.md

.github/workflows/autoDIR.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: |
+          scripts/build_directory_md.py 2>&1 | tee DIRECTORY.md
+          git config --global user.name 'autoDIR'
```
cclauss: directory_writer

.github/workflows/autoDIR.yml
```diff
@@ -0,0 +1,29 @@
...
+          git config --global user.email 'mrvnmchm@users.noreply.github.com'
+          git remote set-url origin https://x-access-token:${{ secrets.gh_token }}@github.com/$GITHUB_REPOSITORY
+          git checkout $GITHUB_HEAD_REF
+          git diff-files --quiet
```
cclauss: I believe that if this command fails then the entire PR is marked with a failing test ❌ which we do not want.

On line 28, use `fixup:` instead of `Update:` because when a merge happens git will autosquish that commit into the previous commit.

Put a `\n` at the end of the file so that GitHub does not put 🚫 at the end of the file.

.github/workflows/autoDIR.yml
mrvnmchm: I'll make this change for the upcoming PR

.github/workflows/autoDIR.yml
mrvnmchm: Will make this change in upcoming PR

.github/workflows/autoDIR.yml
mrvnmchm: Will make this change in upcoming PR

.github/workflows/autoDIR.yml
mrvnmchm: That's correct @cclauss, the `git diff-files --quiet` commands exit code is the same as for git diff-index (0 == no differences; 1 == differences). In this case, if failure, it will continue with the push. Would you have any suggestion for evaluating if the DIRECTORY.md file has changed?

I will make the changes to line 28, and add the newline at the end in upcoming PR

cclauss:
$ __python3__
```python
>>> import hashlib
>>> def get_file_hexdigest(filepath: str="DIRECTORY.md") -> str:
...     with open(filepath, "rb") as in_file:
...         return hashlib.md5(in_file.read()).hexdigest()
...
>>> get_file_hexdigest("DIRECTORY.md")
'b1c7c4f112361d85016fa4d50cd681e9'
```

PR #1547, opened 2019-11-01
Update gaussian_filter.py

frifich:
changed double for loop to use itertools 's product. This way range is called only twice.


cclauss:
Is the new implementation faster than the current one?  Does it use less memory?  How do we know?

PR #1548, opened 2019-11-01
Update gaussian_filter.py

frifich:
Changed embedded for loops with product. This way range(dst_height) is called only once, instead of being called $dst_height times.

### DIFF
--- a/digital_image_processing/filters/gaussian_filter.py
+++ b/digital_image_processing/filters/gaussian_filter.py
@@ -22,7 +22,7 @@ def gaussian_filter(image, k_size, sigma):
     # im2col, turn the k_size*k_size pixels into a row and np.vstack all rows
     image_array = zeros((dst_height * dst_width, k_size * k_size))
     row = 0
-    for i, j in product(range(dst_height), range(dst_height)): 
+    for i, j in product(range(dst_height), range(dst_width)): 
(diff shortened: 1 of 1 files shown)


digital_image_processing/filters/gaussian_filter.py
```diff
@@ -21,8 +22,7 @@ def gaussian_filter(image, k_size, sigma):
     # im2col, turn the k_size*k_size pixels into a row and np.vstack all rows
     image_array = zeros((dst_height * dst_width, k_size * k_size))
     row = 0
-    for i in range(0, dst_height):
-        for j in range(0, dst_width):
+    for i, j in product(range(dst_height), range(dst_height)):
```
cclauss: The previous implementation used both width and height so it would work for images that are not square.  The new implementation only uses height so it assumes that all images are square.

cclauss: Would https://docs.scipy.org/doc/numpy/reference/generated/numpy.prod.html be even faster?

frifich: https://docs.scipy.org/doc/numpy/reference/generated/numpy.prod.html  is the internal product, it doesn't have the same functionality as itertools.product

PR #1549, opened 2019-11-02
directory_writer (#1)

mrvnmchm:
* directory_writer

* fixup: Format Python code with psf/black

* fixup: DIRECTORY.md


mrvnmchm:
add directory_writer GH action for #1460

PR #1561, opened 2019-11-12
Python Program to check a palindrome string

SKAUL05:


cclauss:
I think we have enough already https://github.com/TheAlgorithms/Python/search?q=palindrome

PR #1564, opened 2019-11-12
Panagram Script Added

SKAUL05:


strings/check_panagram.py
```diff
@@ -0,0 +1,24 @@
+# Created by sarathkaul on 12/11/19
+
+
+def check_panagram(input_str: str) -> None:
```
cclauss: Please remove all printing from our function so it is an algorithm that could be dropped into a larger program...
```python
def check_panagram(input_str: str="The quick brown fox jumps over the lazy dog") -> bool:
```
All that printing should happen outside the function (algorithm).

Please add a test:
```
>>> check_panagram("The quick brown fox jumps over the lazy dog")
False
```

strings/check_panagram.py
```diff
@@ -0,0 +1,24 @@
...
+    >>> check_panagram("The quick brown fox jumps over the lazy dog")
+    The quick brown fox jumps over the lazy dog is a Panagram String
+    >>> check_panagram("My name is Unknown")
+    My name is Unknown is not a Panagram String
+    """
+    frequency = set()
+
+    for alpha in input_str:
+        if alpha != " ":
+            frequency.add(alpha.lower())
```
cclauss: frequency.add(alpha.replace(" ", "").lower())  # and then remove line 15

strings/check_panagram.py
```diff
@@ -0,0 +1,24 @@
...
+    if len(frequency) == 26:
+        print(f"{input_str} is a Panagram String")
+    else:
+        print(f"{input_str} is not a Panagram String")
```
cclauss: print(f"{input_str} is {'not ' if len(frequency) == 26 else ''} a panagram string")

strings/check_panagram.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+if __name__ == "main":
+    check_panagram("INPUT STRING")
```
cclauss: Put the printing down here.

cclauss:
Perhaps https://docs.python.org/3/library/string.html#string.ascii_lowercase should be part of your solution.

cclauss:
Please add this test.
```
    >>> check_panagram("The quick brown fox jumps over the la_y dog")
    False
```

SKAUL05:
Added

PR #1565, opened 2019-11-14
Added Remove duplicate script and updated requirements.txt

SKAUL05:


requirements.txt
```diff
@@ -14,3 +14,4 @@ scikit-fuzzy
 sklearn
 sympy
 tensorflow
+ordered_set
```
cclauss: Looks like it breaks the alphabetical ordering on requirements.txt.

Why do we need to add this module?

strings/remove_duplicates.py
```diff
@@ -0,0 +1,22 @@
...
+    >>> remove_duplicates("Python is great and Java is also great")
+    'Python is great and Java also'
+    """
+    sen_list = sentence.split(" ")
+    check = OrderedSet()
+
+    for a_word in sen_list:
+        check.add(a_word)
+
+    return " ".join(check)
```
cclauss: return " ".join(sorted(check))  # and lose the dependency.

SKAUL05: That seems correct

cclauss:
git rm strings/check_panagram.py~HEAD

PR #1568, opened 2019-11-14
fix implementation errors.

Jiang-zzz:
I revised my implementation and found out that I have miss a inner loop for t.  x and y should be recalculated everytime when t is divisble by 2. I have also included a more readble source for this algorithm.


cclauss:
The doctests give the same results as previously?

Jiang-zzz:
> The doctests give the same results as previously?

Yes, as previous version can factor the number, it's just less efficient.

PR #1572, opened 2019-11-14
 ||  true

cclauss:
Fixes: #1460

@chttrjeankr


chttrjeankr:
Issue #1460 finally fixed then :clap:

PR #1574, opened 2019-11-16
Adding circular_queue.py

Himan10:
Circular FIFO Queue implementation using Python list with fixed range.
Time complexity 
1. Enqueue(put) - O(1)
2. Dequeue(get) - O(1)
data_structures/queue/circular_queue.py

### DIFF
--- a/data_structures/queue/circular_queue.py
+++ b/data_structures/queue/circular_queue.py
@@ -68,7 +68,9 @@ class CircularQueue:
         index
         >>> cq = CircularQueue(5)
         >>> cq.dequeue()
-        False
+        Traceback (most recent call last):
+           ...
+        Exception: UNDERFLOW
         >>> cq.enqueue("A")
         >>> cq.enqueue("B")
         >>> cq.dequeue()
--- a/data_structures/queue/circular_queue.py
+++ b/data_structures/queue/circular_queue.py
@@ -15,7 +15,8 @@ class CircularQueue:
         >>> cq = CircularQueue(5)
         >>> len(cq)
         0
-        >>> cq.enqueue("A")
+        >>> cq.enqueue("A")  # doctest: +ELLIPSIS
+        <circular_queue.CircularQueue object at ...
         >>> len(cq)
         1
         """
@@ -26,7 +27,8 @@ class CircularQueue:
         >>> cq = CircularQueue(5)
         >>> cq.is_empty()
         True
-        >>> cq.enqueue("A")
+        >>> cq.enqueue("A")  # doctest: +ELLIPSIS
+        <circular_queue.CircularQueue object at ...
         >>> cq.is_empty()
         False
         """
@@ -37,7 +39,8 @@ class CircularQueue:
         >>> cq = CircularQueue(5)
         >>> cq.first()
         False
-        >>> cq.enqueue("A")
+        >>> cq.enqueue("A")  # doctest: +ELLIPSIS
+        <circular_queue.CircularQueue object at ...
         >>> cq.first()
         'A'
         """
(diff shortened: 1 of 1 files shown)


data_structures/queue/circular_queue.py
```diff
@@ -0,0 +1,41 @@
...
+        return self.size == 0
+
+    def first(self):
+        if self.is_empty():
+            return False
+        return self.array[self.front]
+
+    def enqueue(self, data):
+        if self.size >= self.n:
+            raise "QUEUE IS FULL"
```
cclauss: Raising a string might have worked in Python 2 but it does not in Python 3.

Himan10: Thank you, sir,  Yeah sure I'll remove it but i guess instead of raising a string, returning True or False might be a better option

cclauss: Please consider `raise Exception("Queue is full")`

cclauss: Raising an exception is more Pythonic.

cclauss:
Tests...
```python
# Implementation of Circular Queue (using Python lists)

class CircularQueue:
    """ Circular queue with fixed capacity """

    def __init__(self, n: int):
        self.n = n
        self.array = [None] * self.n
        self.front = 0  # index of the first element
        self.rear = 0
        self.size = 0

    def __len__(self) -> int:
        """
        >>> cq = CircularQueue(5)
        >>> len(cq)
        0
        >>> cq.enqueue("A")
        >>> len(cq)
        1
        """
        return self.size

    def is_empty(self) -> bool:
        """
        >>> cq = CircularQueue(5)
        >>> cq.is_empty()
        True
        >>> cq.enqueue("A")
        >>> len(cq)
        False
        """
        return self.size == 0

    def first(self):
        """
        >>> cq = CircularQueue(5)
        >>> cq.first()
        False
        >>> cq.enqueue("A")
        >>> cq.first()
        'A'
        """
        return False if self.is_empty() else self.array[self.front]

    def enqueue(self, data):
        """
        This function insert an element in the queue using self.rear value as an index
        >>> cq = CircularQueue(5)
        >>> cq.enqueue("A")
        >>> (cq.size, cq.first(), cq.last())
        1, 'A', 'A'
        >>> cq.enqueue("A")
        >>> (cq.size, cq.first(), cq.last())
        2, 'A', 'B'
        """
        if self.size >= self.n:
            return False

        self.array[self.rear] = data
        self.rear = (self.rear+1)%self.n
        self.size += 1
        return self

    def dequeue(self):
        """
        This function removes an element from the queue using on self.front value as an
        index
        >>> cq = CircularQueue(5)
        >>> cq.dequeue()
        False
        >>> cq.enqueue("A")
        >>> cq.enqueue("B")
        >>> cq.dequeue()
        'B'
        >>> (cq.size, cq.first(), cq.last())
        1, 'A', 'A'
        >>> cq.dequeue()
        False
        """
        if self.size == 0:
            return False

        temp = self.array[self.front]
        self.array[self.front] = None
        self.front = (self.front + 1)%self.n
        self.size -= 1
        return temp
```

Himan10:
@cclauss well thanks sir for the test, actually it's my first time contributing in some one's Repo. also i wasn't familiar with doctest but thanks for your suggestions and all the things.

cclauss:
Perhaps you should adopt the same get()/put() interface as a Python non-blocking [queue.Queue()](https://docs.python.org/3/library/queue.html)...
>>> from queue import Queue
>>> Queue().get(block=False)  # This will raise queue.Empty

Himan10:
yeah right, I already checked that queue module I just want to implement the queue's dequeue(get) method for O(1) complexity
BTW, can you tell me why there are problems still remains in my code? 
at first, Building by Travis Cl was perfectly done but after adding some doctest in every function, Travis Cl failed?

Himan10:
> doctest: Catch "Exception: UNDERFLOW"
okay okay seems like, I need to remove Exception from there and add return statements

cclauss:
Raising the exception is right and very Pythonic.  It is what the Python Standard Library __queue.Queue()__ does.  I was just trying to modify the doctest to accept the Exception being raised.

cclauss:
The problem that we have now is that __enqueue()/put()__ is returning a value.  I do not believe that that is expected.

Himan10:
> The problem that we have now is that **enqueue()/put()** is returning a value. I do not believe that that is expected.

So what if we don't return anything from the enqueue() function? maybe why I was returning self so that I can do some operations like this
cq.enqueue(10).dequeue().enqueue(20)

cclauss:
Ahhh... A "fluent" interface...  OK.  That makes sense.  Try this...
```python
>>> cq.enqueue("A")  # doctest: +ELLIPSIS
<circular_queue.CircularQueue object at ...
```

Himan10:
Thank you for approving it, I really appreciate your editions and yeah I learned about some new stuff last night surely going to follow these things when writing python code. :smiley:

PR #1575, opened 2019-11-17
Improve bellman_ford.py

poyea:
* Fix out-of-range error on line 41
* Add a type hint


cclauss:
```python
def printDist(dist, V):
    print("\nVertex Distance")
    distances = ('INF' if d == float('inf') else d for d in dist)
    print("\t".join(f"{i}\t{d}" for i, d in enumerate(distances)))
```

PR #1576, opened 2019-11-17
Word Occurence Script Added

SKAUL05:
This PR request is related to an algorithm that counts the occurrences of all the words in a sentence.


cclauss:
Like https://docs.python.org/3/library/collections.html#collections.Counter

PR #1578, opened 2019-11-17
Add pytest-cov

sirex:
Also added coverage report in .travis.yml file.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+omit =
+    .env/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,4 @@
 [report]
-sort = Miss
+sort = Cover
 omit =
     .env/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,4 @@
 [report]
+sort = Miss
 omit =
     .env/*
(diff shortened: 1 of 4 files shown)


cclauss:
Nice!!  Thanks!  Is there a way to get a __sorted__ list showing those files with no tests or with little test coverage?

sirex:
> Nice!! Thanks! Is there a way to get a **sorted** list showing those files with no tests or with little test coverage?

By default it is sorted by file name, bet here are other options:

> sort (string, default “Name”): Sort the text report by the named column. Allowed values are “Name”, “Stmts”, “Miss”, “Branch”, “BrPart”, or “Cover”.

Changed sorting to `Miss` (by missing statements).

cclauss:
Can we do a reverse sort on cover?  That would put the 5% near the top of the list and the 95% near the bottom.  This is super cool.  Thanks for doing it!

sirex:
> Can we do a reverse sort on cover? That would put the 5% test coverage near the top of the list and the 95% near the bottom. This is super cool. Thanks for doing it!

I'm not sure, but I guess reverse is not supported. Documentation lists these sort options:

- Name
- Stmts
- Miss
- Branch
- BrPart
- Cover

PR #1581, opened 2019-11-18
Reverse Words

SKAUL05:


strings/reverse_words.py
```diff
@@ -0,0 +1,21 @@
+# Created by sarathkaul on 18/11/19
+
+
+def reverse_words(input_str: str) -> str:
+    """
+    Reverses words in a given string
+    >>> sentence = "I love Python"
+    >>> reverse_words(sentence) == " ".join(sentence.split()[::-1])
```
cclauss: Let's also keep the other test to help the reader understand...
```python
>>> reverse_words(sentence)
'Python love I'
```

SKAUL05: Okay Cool

strings/reverse_words.py
```diff
@@ -0,0 +1,24 @@
...
+    True
+    >>> sentence = "I love Python"
```
cclauss: __sentence__ is already defined on line 7 so I think we can remove line 10.

strings/reverse_words.py
```diff
@@ -0,0 +1,23 @@
...
+    >>> reverse_words(sentence)
+    'Python love I'
+    """
+    input_str = input_str.split(" ")
+    new_str = list()
+
+    for a_word in input_str:
+        new_str.insert(0, a_word)
+
+    return " ".join(new_str)
```
chttrjeankr: I just got the idea that `return " ".join(input_str[::-1])` would have probably worked the same way.

cclauss: Line 8

chttrjeankr: I noticed it being used as a doctest, and NOT the actual function. :sweat_smile: Hence, the comment.

PR #1590, opened 2019-11-23
added current stock price

vansh1999:
Added "current stock price", web scrapper to fetch current stock prices using python and beautiful soup


cclauss:

PR #1593, opened 2019-11-26
Update singly_linked_list.py

vikasit12:
printing current.data rather than node address in __repr__ for a more readable print statement


cclauss:
eval(repr(c)) == c

The output of `__repr__()` _should look like a valid Python expression that could be used to recreate an object with the same value_.

https://docs.python.org/3.4/reference/datamodel.html#object.__repr__

PR #1611, opened 2019-12-04
Pythonize longest_increasing_subsequence_o(nlogn).py

rirze:
Use Pythonic expressions and standard lib functions


cclauss:
Can you please run the current algorithm and the proposed algorithm thru https://docs.python.org/3.8/library/timeit.html or similar to prove the performance gains of the proposed changes?

rirze:
Well, the proposed changes are not performance-driven...

It's to clean up the code and make it easier to understand. 
I will provide a runtime comparison either way I guess.

PR #1615, opened 2019-12-06
Improved readability

GeorgeChambi:


divide_and_conquer/convex_hull.py
```diff
@@ -88,7 +88,7 @@ def __le__(self, other):
         return False
 
     def __repr__(self):
-        return "({}, {})".format(self.x, self.y)
+        return f"({self.x}, {self.y})"
```
cclauss: __return f"Point({self.x}, {self.y})"__

__repr__() should _return a string that would yield an object with the same value when passed to eval()_
https://docs.python.org/3/library/functions.html#repr

graphs/dijkstra_algorithm.py
```diff
@@ -139,9 +139,9 @@ def dijkstra(self, src):
         self.show_distances(src)
 
     def show_distances(self, src):
-        print("Distance from node: {}".format(src))
+        print(f"Distance from node: {src}")
         for u in range(self.num_nodes):
-            print("Node {} has distance: {}".format(u, self.dist[u]))
+            print("Node {u} has distance: {self.dist[u]}")
```
cclauss: Need an __f__.

machine_learning/cancel_data.csv
```diff
@@ -0,0 +1,569 @@
+842302,M,17.99,10.38,122.8,1001,0.1184,0.2776,0.3001,0.1471,0.2419,0.07871,1.095,0.9053,8.589,153.4,0.006399,0.04904,0.05373,0.01587,0.03003,0.006193,25.38,17.33,184.6,2019,0.1622,0.6656,0.7119,0.2654,0.4601,0.1189
```
cclauss: Whatup with this file!?!

GeorgeChambi: ah sorry, one of the files made it when I was testing

PR #1634, opened 2019-12-14
Bug Fixed in newton_raphson_method.py

rohitjoshi21:
Without using math function,function like sin and exp cannot be evaluated.
Adding PRECISION variable explicitly helps user to play with precision.


cclauss:

cclauss:
__from math import sin, cos__ # will work

__from math import *__ # will not

rohitjoshi21:
Then we should mention each function like log,exp

rohitjoshi21:
Isn't there any other way ?

cclauss:
from math import *  # noqa: F821

rohitjoshi21:
Its still showing the error? Why?

arithmetic_analysis/newton_raphson_method.py
```diff
@@ -4,10 +4,10 @@
 # quickly find a good approximation for the root of a real-valued function
 from sympy import diff
 from decimal import Decimal
-from math import *
+from math import sin,cos,tan,log,exp
 
 
-PRECISION = 10 ** -10
+precision = 10 ** -10
```
cclauss: All UPPERCASE is correct for a constant in Python.

cclauss: https://www.python.org/dev/peps/pep-0008/#constants

rohitjoshi21:
What's the mistake now ?

cclauss:
Don’t import these functions unless your code is going to use them.
https://travis-ci.com/TheAlgorithms/Python/builds/141069414#L356

rohitjoshi21:
These functions are being used by the program. But the function name is not explicitly mentioned in the program code. What should I do then ?

arithmetic_analysis/newton_raphson_method.py
```diff
@@ -4,31 +4,35 @@
...
 from sympy import diff
 from decimal import Decimal
+from math import sin,log,exp
+
+
+PRECISION = 10 ** -10
 
 
 def NewtonRaphson(func, a):
     """ Finds root from the point 'a' onwards by Newton-Raphson method """
```
cclauss: How about some doctests...

"""
>>> NewtonRaphson("sin(x)", 2))
0
>>> NewtonRaphson("x**2 - 5*x +2", 0.4))
0
>>> NewtonRaphson("x**2 - 5", 0.1))
0
>>> NewtonRaphson("log(x)- 1", 2))
0
"""
     # Exponential Roots
     print("exp(x) - 1 = 0", NewtonRaphson("exp(x) - 1", 0))
     print("The root of exp(x) - 1 = 0 is", NewtonRaphson("exp(x) - 1", 0))
"""

arithmetic_analysis/newton_raphson_method.py
```diff
@@ -4,31 +4,45 @@
 # quickly find a good approximation for the root of a real-valued function
 from sympy import diff
 from decimal import Decimal
+from math import *
```
onlinejudge95:
```
>>> import this
The Zen of Python, by Tim Peters
...
Explicit is better than implicit.
...
```
Prefer
```
import math

math.ceil()
```

arithmetic_analysis/newton_raphson_method.py
```diff
@@ -4,31 +4,45 @@
...
+
+
+PRECISION = 10 ** -10
 
 
 def NewtonRaphson(func, a):
```
onlinejudge95: Prefer **snake_case** for **function** name. Use **CamelCase** for **class** names.

arithmetic_analysis/newton_raphson_method.py
```diff
@@ -4,31 +4,45 @@
...
+    >>> NewtonRaphson("log(x)- 1", 2))
+    2.718281828458937968804105873
+    """
+    
+    x = a
     while True:
-        c = Decimal(a) - (Decimal(eval(func)) / Decimal(eval(str(diff(func)))))
 
-        a = c
+        x = Decimal(x) - (Decimal(eval(func)) / Decimal(eval(str(diff(func)))))
```
onlinejudge95: `Decimal` class is present in `decimal` module, import that if you want to use it.

[Source](https://docs.python.org/3.8/library/decimal.html#decimal.Decimal)

PR #1642, opened 2019-12-24
Added to maths and strings

coleman2246:
Added from python math library combinations,gamma and radian methods. From the python string library split,upper, and lower were added.

### DIFF
--- a/maths/gamma.py
+++ b/maths/gamma.py
@@ -1,0 +1,32 @@
+from math import factorial
+
+def gamma(num):
+    '''
+    >>> gamma(1)
+    1
+
+    >>> gamma(0)
+    Traceback (most recent call last):
+        ...
+    ValueError:  math domain error
+    
+    >>> gamma(10)
+    362880
+    
+    >>> gamma(-199)
+    Traceback (most recent call last):
+        ...
+    ValueError:  math domain error
+    '''
+
+
+    if num < 1:
+        raise ValueError(" math domain error")
+    else:
+        return factorial(num-1)
+
+
+if __name__ == "__main__":
+    from doctest import testmod
+    testmod()
+    
--- a/maths/gamma.py
+++ b/maths/gamma.py
@@ -1,32 +1,60 @@
-from math import factorial
+import math
+from scipy.integrate import quad
+from numpy import inf
 
-def gamma(num):
-    '''
-    >>> gamma(1)
-    1
+
+def gamma(num: float) -> float:
+    """
+    https://en.wikipedia.org/wiki/Gamma_function
+    In mathematics, the gamma function is one commonly 
+    used extension of the factorial function to complex numbers. 
+    The gamma function is defined for all complex numbers except the non-positive integers
+
+
+    >>> gamma(-1)
+    Traceback (most recent call last):
+        ...
+    ValueError: math domain error
+
+    
 
     >>> gamma(0)
     Traceback (most recent call last):
         ...
-    ValueError:  math domain error
-    
-    >>> gamma(10)
-    362880
-    
-    >>> gamma(-199)
+    ValueError: math domain error
+
+
+    >>> gamma(9)
+    40320.0
+
+    >>> from math import gamma as math_gamma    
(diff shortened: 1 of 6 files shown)


strings/upper.py
```diff
@@ -0,0 +1,32 @@
...
+    
+    for char in word:
+        # converting to ascii value int value and checking to see if char is a lower letter
+        # if it is a capital letter it is getting shift by 32 which makes it a capital case letter
+        if ord(char) >= 97 and ord(char) <= 122:
+            upper_string += chr(ord(char) -32)
+        else:
+            upper_string += char
+    
+    return upper_string
```
cclauss: This can be a one liner plus add type hints...
```python
def upper(word: str) -> str:
    return "".join(chr(ord(char) - 32) if 97 <= ord(char) <= 122 else char)
```

strings/upper.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+
+if __name__ == "__main__":
+    from doctest import testmod
+    testmod()
+
```
cclauss: Python files should end in one and only one __\n__ otherwise the GitHub file viewer adds a 🚫 at the end of the file.

strings/split.py
```diff
@@ -0,0 +1,32 @@
+
+def split(string,seperator = " "):
```
cclauss: Type hints: __def split(string: str, seperator: str=" ") -> list:__

strings/lower.py
```diff
@@ -0,0 +1,37 @@
...
+
+    for char in word:
+        # converting to ascii value int value and checking to see if char is a capital letter
+        # if it is a capital letter it is getting shift by 32 which makes it a lower case letter
+        if ord(char) >= 65 and ord(char) <= 90:
+            lower_string += chr(ord(char) + 32)
+        else:
+            lower_string += char
+
+    return lower_string
```
cclauss: This can be a one liner plus add type hints...
```python
def lower(word: str) -> str:
    return "".join(chr(ord(char) + 32) if 65 <= ord(char) <= 90 else char)
```

maths/radians.py
```diff
@@ -0,0 +1,24 @@
...
+'''
+def radians(degree):
+
+    """
+    >>> radians(180)
+    3.141592653589793
+    >>> radians(92)
+    1.6057029118347832
+    >>> radians(274)
+    4.782202150464463
```
cclauss: When replicating a Python Standard Library function, please try to do side-by-side tests...

Add tests:
```
>>> from math import radians as math_radians
>>> all(radians(i) == math_radians(i) for i in range(-2, 361))
True
```
Test should always try negative numbers and zeros.  Testing for floating point numbers is also good idea.

maths/radians.py
```diff
@@ -0,0 +1,24 @@
+from math import pi
+'''
+
+Coverts the given angle from degrees to radians
+https://en.wikipedia.org/wiki/Radian
+'''
+def radians(degree):
```
cclauss: __def radians(degree: int) -> float:__  # Please see CONTRIBUTING.md

maths/gamma.py
```diff
@@ -0,0 +1,32 @@
...
+    ValueError:  math domain error
+    
+    >>> gamma(10)
+    362880
+    
+    >>> gamma(-199)
+    Traceback (most recent call last):
+        ...
+    ValueError:  math domain error
+    '''
```
cclauss: Can we test these results against https://docs.python.org/3/library/math.html#math.gamma especially with floating point and negative numbers.

coleman2246: Added . The function had to be re-written as the old version would only work with positives integers greater than one.

maths/radians.py
coleman2246: Added

maths/radians.py
coleman2246: Test was added

strings/upper.py
coleman2246: Change was made

strings/upper.py
coleman2246: Formatting issue was fixed

strings/lower.py
coleman2246: Change has been added

strings/split.py
coleman2246: Type hint has been added

PR #1647, opened 2019-12-26
Richest neighbours algorithm

bluedistro:
Implemented richest neighbors algorithm


cclauss:
Backslash n “\n” in a string is perfectly acceptable in Python.  It is backslashes _outside_ of strings that are frowned upon.

PR #1648, opened 2019-12-28
Added Strassen divide and conquer algorithm to multiply matrices.

AlexLeka98:


cclauss:
Thanks for this contribution.  Please read CONTRIBUTING.md and add doctests and type hints as discussed there and format your code with psf/black as discussed in that doc.

PR #1651, opened 2019-12-29
Fix floor division

kylepw:
Noticed floor division that calculates the parent index becomes inaccurate with bigger numbers (index will differ based on odd vs even number of elements), so I added this measure.

### DIFF
--- a/data_structures/heap/min_heap.py
+++ b/data_structures/heap/min_heap.py
@@ -40,7 +40,10 @@ class MinHeap:
         return self.get_value(key)
 
     def get_parent_idx(self, idx):
-        return (idx - 1) // 2
+        if len(self.heap_dict) % 2 == 0:
+            return idx // 2
+        else:
+            return (idx // 2) - 1
 
     def get_left_child_idx(self, idx):
         return idx * 2 + 1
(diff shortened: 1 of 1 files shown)


cclauss:
Nice!  Please add one or more doctests to this method that would fail on the current implementation but would pass your suggested implementation.

data_structures/heap/min_heap.py
```diff
@@ -40,10 +40,10 @@ def __getitem__(self, key):
         return self.get_value(key)
 
     def get_parent_idx(self, idx):
-        if len(self.heap_dict) % 2 == 0:
-            return idx // 2
+        if (len(self.heap_dict) - 1) % 2 == 0:
```
cclauss: if len(self.heap-dict) % 2:

kylepw: I was running this min-heap implementation on a Dijkstra's shortest distance problem in a separate module with a lot of vertex data and was getting sometimes correct but other times wrong results until I tweaked the parent index function and some other calculations. Not sure how I'd put that in a simple doctest here. This seems to run fine with small inputs, so I'm going to leave it be. Thanks for taking the time to review!

PR #1652, opened 2020-01-01
Create deque_doubly.py

Himan10:
Implementing Deque ADT using Doubly Linked List...
Every operation of Deque using DLL runs in constant time


cclauss:
No doctests or type hints as discussed in CONTRIBUTING.md

Himan10:
> No doctests or type hints as discussed in CONTRIBUTING.md

Do i need to provide doctests for private class too ?/

cclauss:
> Do i need to provide doctests for private class too?

Only if we want to make sure that it works the way that we think it should.

Himan10:
> > Do i need to provide doctests for private class too?
> 
> Only if we want to make sure that it works the way that we think it should.

okay I'll.. but all the doctests which i put in my code are working on my machine. but why travis CI isn't accepting it?

data_structures/linked_list/deque_doubly.py
```diff
@@ -0,0 +1,144 @@
...
+    
+    ### DEqueu Remove Operations (At the front, At the end) ###
+    
+    def remove_first(self):
+        """ removal from the front
+        >>> d = LinkedDeque()
+        >>> d.remove_first()
+        Traceback (most recent call last):
+           ...
+        Exception: List is empty
```
cclauss: This should be the same exception and a similar error message as __[].pop()__ --> IndexError: pop from empty list

So something like __IndexError: remove_first from empty list__

data_structures/linked_list/deque_doubly.py
```diff
@@ -0,0 +1,144 @@
...
+        >>> d.add_first('A') # doctest: +ELLIPSIS
+        <deque_doubly.LinkedDeque object at ...
+        >>> d.remove_first()
+        'A'
```
cclauss: Please add a test to make sure that d is empty.

data_structures/linked_list/deque_doubly.py
```diff
@@ -0,0 +1,144 @@
...
+            raise Exception('List is empty')
+        return self._delete(self._header._next)
+    
+    def remove_last(self):
+        """ removal in the end
+        >>> d = LinkedDeque()
+        >>> d.remove_last()
+        Traceback (most recent call last):
+           ...
+        Exception: List is empty
```
cclauss: This should be the same exception and a similar error message as __[].pop()__ --> IndexError: pop from empty list

So something like __IndexError: remove_last from empty list__

data_structures/linked_list/deque_doubly.py
```diff
@@ -0,0 +1,144 @@
...
+        >>> d.add_first('A') # doctest: +ELLIPSIS
+        <deque_doubly.LinkedDeque object at ...
+        >>> d.remove_last()
+        'A'
```
cclauss: Please add a test to make sure that d is empty.

Himan10:
@cclauss sir, why it's not been approved?

data_structures/linked_list/deque_doubly.py
Himan10: Yeah, i've done something like ... 
```
>>> d = LinkedDeque()
>>> d.is_empty()
True
>>> d.add_first('A') # doctest: +ELLIPSIS
<deque_doubly.LinkedDeque object at ...
>>> d.remove_last()
 'A'
>>> d.is_empty()
True
```
is it okay sir?

cclauss:
```
<deque_doubly.LinkedDeque object at ...
—>
<linked_list.deque_doubly.LinkedDeque object at ...
```

Himan10: I don't know why you added those two lines after <deque_doubly ... 
are you running this from the linked_list directory?

PR #1653, opened 2020-01-03
Typo?

logan-lieou:
newNod -> newNode


cclauss:
Instead, can we go newNod -> new_node

This would be in alignment with the Python variable naming discussed in CONTRIBUTING.md

PR #1660, opened 2020-01-06
enhance swapping code in link

lanzhiwang:
enhance swapping code in link

### DIFF
--- a/data_structures/linked_list/swap_nodes.py
+++ b/data_structures/linked_list/swap_nodes.py
@@ -11,8 +11,9 @@ class Linkedlist:
     def print_list(self):
         temp = self.head
         while temp is not None:
-            print(temp.data)
+            print(temp.data, end=' ')
             temp = temp.next
+        print()
 
     # adding nodes
     def push(self, new_data):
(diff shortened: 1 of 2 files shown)


data_structures/linked_list/swap_nodes.py
```diff
@@ -53,6 +53,22 @@ def swapNodes(self, d1, d2):
...
+    def swapNodes2(self, d1, d2):
+        if d1 == d2:
+            return
+        else:
+            D1 = self.head
+            while D1 is not None and D1.data != d1:
+                D1 = D1.next
+
+            D2 = self.head
+            while D2 is not None and D2.data != d2:
```
onlinejudge95: Checking for `None` should be done like `if D1` or `while D1`

data_structures/linked_list/swap_nodes.py
```diff
@@ -53,6 +53,22 @@ def swapNodes(self, d1, d2):
             D1.next = D2.next
             D2.next = temp
 
+    def swapNodes2(self, d1, d2):
```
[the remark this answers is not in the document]
cclauss: swap_nodes_2() would be the proper Python name.  See CONTRIBUTING.md.  Also please add type hints to the function signature.

data_structures/linked_list/swap_nodes.py
```diff
@@ -11,8 +11,9 @@ def __init__(self):
     def print_list(self):
         temp = self.head
         while temp is not None:
-            print(temp.data)
+            print(temp.data, end=' ')
             temp = temp.next
+        print()
 
     # adding nodes
     def push(self, new_data):
```
onlinejudge95: Needs type hints and more pythonic condition checking

cclauss: Do we want to limit the data type?  Currently this can be used with many different data types and I consider that a feature, not a bug.

data_structures/linked_list/swap_nodes.py
```diff
@@ -21,72 +22,32 @@ def push(self, new_data):
...
                 return
 
-            D1.data, D2.data = D2.data, D1.data
+            node_1.data, node_2.data = node_2.data, node_1.data
 
-# swapping code ends here
 
 
 if __name__ == "__main__":
     list = Linkedlist()
```
cclauss: __list__ is the name of a Python builtin so it should not be used as a variable name.

data_structures/linked_list/swap_nodes.py
onlinejudge95: Then let's be consistent by using `typing.Any`

data_structures/linked_list/swap_nodes.py
onlinejudge95: Also according to [PEP8](https://www.python.org/dev/peps/pep-0008/#class-names) should the class names not be `LinkedList` instead of `Linkedlist`

PR #1675, opened 2020-01-11
Add a program to evaluate a string in prefix notation (Polish Notation)

M0r13n:
We already have a script for evaluating the postfix notation. So I added another for prefix notation.


data_structures/stacks/prefix_evaluation.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # push operand to stack
+        if is_operand(c):
+            stack.append(int(c))
+
+        else:
+            # pop values from stack can calculate the result
+            # push the result onto the stack again
+            o1 = stack.pop()
+            o2 = stack.pop()
```
onlinejudge95: What if there is only a single input in the string
`evaluate("9")`
A better error message would be useful.

M0r13n: I think the current behaviour is mathematically correct. Just like in infix notation x = 9 is a valid expression. At least I think so. So `evaluate("9")` should output "9" as it does.

onlinejudge95: Early morning sleep made me skip the entire logic :p

cclauss: ☕️

PR #1699, opened 2020-01-18
Counting word in a sentence using dictionary comprehension

ttuanho:
### **Describe your change:**

* [x] Add an algorithm?  No.
* [x] Fix a bug or typo in an existing algorithm? Improve the existing implementation with more succint code
* [x] Documentation change? No need.

### **Checklist:**

### DIFF
--- a/strings/word_occurence2.py
+++ b/strings/word_occurence2.py
@@ -1,0 +1,22 @@
+# Developed by Tuan H. on 18/1/20
+
+"""
+Counting word in a sentence using dictionary comprehension
+"""
+
+
+def word_occurence(sentence: str) -> dict:
+    occurence = {}
+    words = sentence.split(' ')
+    unique_words = list(set(words))
+    
+    # Creating a dictionary containing count of each word
+    occurence = {w:words.count(w) for w in unique_words}
+    
+    return occurence
+
+
+if __name__ == "__main__":
+    input_sentence = input('Input: ')
+    for word, count in word_occurence(input_sentence).items():
+        print(f"{word}: {count}")
--- a/strings/word_occurence2.py
+++ b/strings/word_occurence2.py
@@ -8,8 +8,8 @@ from collections import defaultdict
 def word_occurence(sentence: str) -> dict:
     """
     >>> from collections import Counter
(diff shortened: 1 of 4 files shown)


ttuanho:
Still have no idea why I [failed the codespell check](https://github.com/TheAlgorithms/Python/pull/1699/checks?check_run_id=396785578). I appreciate if someone can explain a bit more.

Thanks.

strings/word_occurence2.py
```diff
@@ -8,8 +8,8 @@
 def word_occurence(sentence: str) -> dict:
     """
     >>> from collections import Counter
-    >>> all(word_occurence(s) == Counter(s.split()) for s in
-    ...     ("a b a b c a", "I love to love Python ❤️ 💕 ❤️", ""))
+    >>> all(word_occurence(s) == Counter(s.split(' ')) for s in
+    ...     ("a b a b c a", "I love to love Python ❤️ 💕 ❤️", "", " ; ;"))
```
cclauss: The code puts out the wrong output when the input is an empty string.

strings/word_occurence2.py
```diff
@@ -8,8 +8,8 @@
 def word_occurence(sentence: str) -> dict:
     """
     >>> from collections import Counter
-    >>> all(word_occurence(s) == Counter(s.split()) for s in
-    ...     ("a b a b c a", "I love to love Python ❤️ 💕 ❤️", ""))
+    >>> all(word_occurence(s) == Counter(s.split(' ')) for s in
```
cclauss: What happens when the input string has two spaces in a row? `"a a  a b  b"`  --  I think we are better off using __s.split()__ than __s.split(" ")__

ttuanho: I think line 17 `unique_words = list(set(words))` helps us to solve this issue.

strings/word_occurence2.py
ttuanho: Then I should change line 16 to `.split()`. Otherwise, it does count "empty" char as a string tho:
```
python3 -i word_occurence2.py 
Input: 
: 1
>>> word_occurence('')
{'': 1}
```
Thanks for the suggestions.

PR #1707, opened 2020-01-22
Create function_composition.py

QuantumNovice:
### Added function `Compose` class that facilitates function composition in a way that lead to optics.
Makes function composition possible in a mathematical way.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


QuantumNovice:
This works for any function/object. No need for mypy static type hints.

cclauss:
In theory, would this be possible for functions that took two parameters like x,y or lat, lon?

QuantumNovice:
> 
> 
> In theory, would this be possible for functions that took two parameters like x,y or lat, lon?
Yes, int theory although I am not entirely sure how I would do it here. I think eventually those functions are reduced to a single variable functions via partial differentiation.

cclauss:
*args, **kwargs would allow us to pass an arbitrary number of params but...

QuantumNovice:
> 
> 
> *args, **kwargs would allow us to pass an arbitrary number of params but...
In case of multivariate functions this notation becomes ambiguous because if
f = a(x,y)
g = b(x,y)
then `( f * g)(x)` could be any of `f(g(x,y), y)` , `f(x, g(x,y)`

https://en.wikipedia.org/wiki/Function_composition

cclauss:
If the call is `( f * g)(*args)` then
```
    answer = f(g(*args), args[-1])
    if f(args[0], g(*args)) != answer:
        raise Exception()
    return answer
```
That _might_ work `if len(args) in (1, 2)`

PR #1726, opened 2020-02-03
Improve collatz_sequence algorithm

TheSuperNoob:
- Add more doc tests and type checking to make sure only natural
  numbers are used

- Simplified the algorithm slightly
	This new version is also between 10-15% faster for really
	long sequences

`python -m timeit -s 'from maths.collatz_sequence import collatz_sequence' 'collatz_sequence(989345275647)'` (1348 steps)

Old algorithm: 

> 1000 loops, best of 5: 369 usec per loop

New algorithm: 

> 1000 loops, best of 5: 313 usec per loop

### DIFF
--- a/maths/collatz_sequence.py
+++ b/maths/collatz_sequence.py
@@ -1,19 +1,33 @@
-def collatz_sequence(n):
+from typing import List
+
+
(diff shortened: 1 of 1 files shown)


onlinejudge95:
Good game

PR #1727, opened 2020-02-04
Second most repeated character in a sequence

SKAUL05:
This `PR` is about adding an algorithm which would find the second most repeated character in a sequence

### DIFF
--- a/other/nth_repeated.py
+++ b/other/nth_repeated.py
@@ -3,7 +3,7 @@
 from collections import Counter
 
 
-def nth_repeated(sequence: list, nth_repeated: int) -> str:
+def count_repeated(sequence: list, nth_repeated: int) -> str:
     """
     >>> sequence = ['Algorithms','Algorithm','Python','Python','The','Python','The']
     >>> nth_repeated = 2
@@ -35,4 +35,4 @@ if __name__ == "__main__":
         "Python",
     ]
     nth_repeated = 2
-    print(nth_repeated(input_sequence, nth_repeated))
+    print(count_repeated(input_sequence, nth_repeated))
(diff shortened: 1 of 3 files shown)


SKAUL05:
> How about generalizing the algorithm to find **kth** most repeated character in a sequence

@onlinejudge95 Changed to support nth repeated

other/nth_repeated.py
```diff
@@ -0,0 +1,36 @@
+# Created by sarathkaul on 04/02/20
+# Script to print second most repeated character in sequence
+from collections import Counter
+
+
+def second_repeated(sequence: list, nth_repeated: int) -> str:
```
onlinejudge95: Change the method name`

other/nth_repeated.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    >>> sequence = ['Algorithms','Algorithm','Python','Python','The','Python','The']
+    >>> nth_repeated = 2
```
onlinejudge95: What if `nth_repeated` is < 1.

SKAUL05:
@onlinejudge95 Changes done

cclauss:
The algorithm is a one liner...
```python
def count_repeated(sequence: list, nth_repeated: int) -> str:
        return "" if nth_repeated < 1 else Counter(sequence).most_common()[nth_repeated - 1][0]
```

cclauss:
Sorry @ SKAUL05 but closing as this has now become a demonstration of how to use the Python Standard Library rather than an algorithm.  Please do not be discouraged because this has been a learning experience.  I hope you will make other submissions of algorithms in the near future.

PR #1755, opened 2020-02-16
Added False Position algorithm

MatteoRaso:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/arithmetic_analysis/false_position.py
+++ b/arithmetic_analysis/false_position.py
@@ -5,7 +5,7 @@ Created on Sat Feb  8 19:25:41 2020
 """
 
 
-def false_position(f, lower_bound: float, upper_bound: float, error: float):
+def main(f, lower_bound: float, upper_bound: float, error: float):
     """An implementation of the false position method used for root-finding.
     The algorithm is extremely similar to the secant method in that it uses
     the x-intercept of the line connecting f(lower_bound) and f(upper_bound) 
--- a/arithmetic_analysis/false_position.py
+++ b/arithmetic_analysis/false_position.py
@@ -1,0 +1,57 @@
+"""
+Created on Sat Feb  8 19:25:41 2020
+
+@author: MatteoRaso
+"""
+
+
+def false_position(f, lower_bound: float, upper_bound: float, error: float):
+    """An implementation of the false position method used for root-finding.
+    The algorithm is extremely similar to the secant method in that it uses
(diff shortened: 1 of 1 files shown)


arithmetic_analysis/false_position.py
```diff
@@ -0,0 +1,57 @@
...
+    INPUTS:
+    f: The function which who's root we want to find.
+    lower_bound: A value at which f(lower_bound) < 0
+    upper_bound: A value at which f(upper_bound) > 0
+    error: The maximum tolerated error
+    
+    OUTPUT:
+    The approximated root bounded by lower_bound and upper_bound
+    """
+    if f(lower_bound) > 0:
```
onlinejudge95: For edge case like these, how about raising an exception

arithmetic_analysis/false_position.py
```diff
@@ -0,0 +1,57 @@
...
+    
+    INPUTS:
+    f: The function which who's root we want to find.
+    lower_bound: A value at which f(lower_bound) < 0
+    upper_bound: A value at which f(upper_bound) > 0
+    error: The maximum tolerated error
+    
+    OUTPUT:
+    The approximated root bounded by lower_bound and upper_bound
+    """
```
onlinejudge95: Some doctests would help

arithmetic_analysis/false_position.py
MatteoRaso: I'm not sure if I understand what you mean. The program would still run without that check, but it wouldn't be guaranteed to converge.

cclauss:
```python
if f(lower_bound) > 0:
    raise ValueError("f(lower_bound) must be less than zero")

Comment (MatteoRaso) [reply arithmetic_analysis/false_position.py]: Thank you for the clarification.

File: arithmetic_analysis/false_position.py [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+                + upper_bound
+            )
+            # Narrowing the interval.
+            if f(root) > 0:
+                upper_bound = root
+
+            else:
+                lower_bound = root
+
+        return root
```
TheSuperNoob: You should round this to the same number of decimals as your error tolerance as the last few digits will not be accurate. 

Also consider using pythons decimal module for arbitrary precision if you want the error tolerance to be arbitrarily small.

PR #1757, opened 2020-02-16
Created geodesy section with one algorithm

eightysixth:
### **Describe your change:**
Created a geodesy section with one formula implemented. [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) calculates the great circle distance between two points in a sphere.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/geodesy/haversine_distance.py
+++ b/geodesy/haversine_distance.py
@@ -1,4 +1,4 @@
-from math import asin, atan, cos, sin, sqrt, tan, radians
+from math import asin, atan, cos, radians, sin, sqrt, tan
 
 
 def haversine_distance(lat1: float, lon1: float, lat2: float, lon2: float) -> float:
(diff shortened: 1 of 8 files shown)


geodesy/haversine_distance.py
```diff
@@ -0,0 +1,49 @@
...
+def haversine_distance(lat1, lon1, lat2, lon2):
+    """
+    Calculate great circle distance between two points in a sphere,
+    given longitudes and latitudes. 
+    (https://en.wikipedia.org/wiki/Haversine_formula)
+
+    Args:
+        lat1, lon1: latitude and longitude of coordinate 1
+        lat2, lon2: latitude and longitude of coordinate 2
+        returnAngle: Toggle to return distance or angle
```
cclauss: Line 13 can be removed?

eightysixth: Oops, yes it can be removed.

geodesy/haversine_distance.py
```diff
@@ -0,0 +1,49 @@
+from math import asin, atan, cos, sin, sqrt, tan, pow, radians
+
+
+def haversine_distance(lat1, lon1, lat2, lon2):
```
cclauss: def haversine_distance(lat1: float, lon1: float, lat2: float, lon2: float) -> float:

geodesy/haversine_distance.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Args:
+        lat1, lon1: latitude and longitude of coordinate 1
+        lat2, lon2: latitude and longitude of coordinate 2
+        
+    Returns: 
+        geographical distance between two points in metres
+
+    >>> int(haversine_distance(37.774856, -122.424227, 37.864742, -119.537521)) # From SF to Yosemite
+    254352
```
cclauss: What is the unit of measure in great circle calculations?  Google maps says 164 miles or 264 kilometers but that is road distance, not as-the-crow-flies distance.  Helping readers connect what they are learning with what they already know will improve the value of this submission.

geodesy/haversine_distance.py
```diff
@@ -0,0 +1,48 @@
...
+    # Equation https://en.wikipedia.org/wiki/Haversine_formula#Formulation
+    flattening = (AXIS_A - AXIS_B) / AXIS_A
+    phi_1 = atan((1 - flattening) * tan(radians(lat1)))
+    phi_2 = atan((1 - flattening) * tan(radians(lat2)))
+    lambda_1 = radians(lon1)
+    lambda_2 = radians(lon2)
+
+    # Equation
+    sin_sq_phi = pow(sin((phi_2 - phi_1) / 2), 2)
+    sin_sq_lambda = pow(sin((lambda_2 - lambda_1) / 2), 2)
```
TheSuperNoob: Using `math.pow()` is very slow compared to pythons `**` notation for powers.

eightysixth: What about sqrt vs ** .5?

TheSuperNoob: According to [this stackoverflow thread](https://stackoverflow.com/questions/327002/which-is-faster-in-python-x-5-or-math-sqrtx) you should use `math.sqrt`.

TheSuperNoob: And as a side note, we can simplify the above lines even more: `a * a` is a lot faster than `a ** 2`.

So you can change the above lines to `(sin(phi_2 - phi_1) / 2) * (sin(phi_2 - phi_1) / 2)` etc

geodesy/haversine_distance.py
eightysixth: I'm planning on adding two more algorithms which improve approximation so haversine is a good starter method. Perhaps, a readme file can follow with comparisons and detailed explanations. For now:

We know that the globe is "sort of" spherical, so a path between two points 
    isn't exactly a straight line. We need to account for the Earth's curvature 
    when calculating distance from point A to B. This effect is negligible for 
    small distances but adds up as distance increases. The Haversine method treats
    the earth as a sphere which allows us to "project" the two points A and B 
    onto the surface of that sphere and approximate the spherical distance between
    them. Since the Earth is not a perfect sphere, other methods which model the
    Earth's ellipsoidal nature are more accurate but a quick and modifiable
    computation like Haversine can be handy for shorter range distances.

geodesy/haversine_distance.py
eightysixth: Did that and made it a little more readable

geodesy/haversine_distance.py
cclauss: What is the unit of measure?  Is it meters, kilometres, miles, lightyears?

eightysixth: Line 15 - metres.. Clarified that in the latest commit as well.

 unit of measure meant is irrelevant for explanation.

PR #1763, opened 2020-02-18
Implemented geodesy - Lambert's ellipsoidal distance

eightysixth:
### **Describe your change:**
Addition of a more accurate algorithm to calculate distances between two points on the Earth's surface. This method treats the Earth as an oblate ellipsoid.

Tests can be cross-referenced [here](https://geographiclib.sourceforge.io/scripts/geod-google.html) - This is a much more precise algorithm than Lambert's but can be used for comparison.
Insert lat1, lon1 lat2, lon2 into the "Inverse" field and "Calculate" to verify tests. s12 value is the geodesic

* [x ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [x ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).

### DIFF
--- a/geodesy/lamberts_ellipsoidal_distance.py
+++ b/geodesy/lamberts_ellipsoidal_distance.py
@@ -45,7 +45,7 @@ def lamberts_ellipsoidal_distance(
     # Distance in metres(m)
     AXIS_A = 6378137.0
     AXIS_B = 6356752.314245
-    RADIUS = 6378137
+    EQUATORIAL_RADIUS = 6378137
 
     # Equation Parameters
     # https://en.wikipedia.org/wiki/Geographical_distance#Lambert's_formula_for_long_lines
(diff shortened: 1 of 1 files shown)


geodesy/lamberts_ellipsoidal_distance.py
```diff
@@ -0,0 +1,83 @@
...
+    '4,138,992 meters'
+    >>> f"{lamberts_ellipsoidal_distance(*SAN_FRANCISCO, *VENICE):0,.0f} meters"
+    '9,737,326 meters'
+    """
+
+    # CONSTANTS per WGS84 https://en.wikipedia.org/wiki/World_Geodetic_System
+    # Distance in metres(m)
+    AXIS_A = 6378137.0
+    AXIS_B = 6356752.314245
+    RADIUS = 6378137
```
TheSuperNoob: Are both `AXIS_A` and `RADIUS` needed? According to the Wikipedia article in the comments the equatorial radius = 6378137 which is the same value you have for both `AXIS_A` and `RADIUS`

I would remove one and rename to `EQUATORIAL_RADIUS` for a more descriptive/accurate name.

poyea: I think that's fine. Axis A and B are used in one place, and the radius is used in another calculation.

PR #1782, opened 2020-02-22
add example to estimate area under line using montecarlo

singlav:
### An implementation of the Monte Carlo method to find area under
       y = x where x lies between 0 to 1
    1. Let x be a uniformly distributed random variable between 0 and 1
    2. expected value of x = integration of x from 0 to 1
    3. Finding expected value of x:
        a. Repeatedly draw x from uniform distribution
        b. expected value = average of those values
    4. Actual value = 1/2
    5. Print the estimated, actual and error value

### **Checklist:**


singlav:
@cclauss @onlinejudge95  please review. Thanks.

maths/montecarlo.py
```diff
@@ -37,6 +37,29 @@ def circle(x: float, y: float):
     print("The total error is ", abs(pi - pi_estimate))
 
 
+def area_under_line_estimator(iterations: int):
```
cclauss: Given that we want this to be an algorithm that we could drop into a larger program, please streamline the area_under_line_estimator() function to return a floating point value.  Also, please move all calls to the __print()__ function outside of area_under_line_estimator().

singlav: Thanks. I'll do that.

singlav:
Thanks for the suggestion. I made the change.

TheSuperNoob:
Could we generalize this more instead of locking it for just the identity function f(x) = x ? Maybe the function can take any function as input as well a lower bound and upper bound (instead of between 0 and 1) to estimate the definite integral of any function f.

maths/montecarlo.py
```diff
@@ -49,7 +50,7 @@ def area_under_line_estimator(iterations: int) -> float:
     5. Returns estimated value
     """
     
-    estimate = sum(uniform(0,1) for i in range(iterations)) / iterations
+    estimate = mean(uniform(0,1) for i in range(iterations))
     
     return estimate
```
cclauss: We can just `return mean(uniform(0,1) for i in range(iterations))` because the function name already tells us that this is an estimate.

singlav:
Thanks for the suggestion. Would it be fine if we do that incrementally in subsequent commits?

maths/montecarlo.py
cclauss: Let’s change i to _ to signal that we are not using this value.

singlav: Thanks. will make the change.

cclauss:
Sure.

maths/montecarlo.py
onlinejudge95: Also, i think `uniform(0,1)` would fail **black** checks

singlav:
@cclauss Thanks for making the necessary changes. Does it look fine now ?

singlav:
@cclauss Thanks for making the necessary changes. Does it look fine now ?

cclauss:
Nice!!  Thanks for your patience and persistence!

singlav:
Thanks for the help. I am new to Github and its my first contribution :)

PR #1795, opened 2020-03-13
add relu function

inoue0426:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/maths/relu.py
+++ b/maths/relu.py
@@ -1,0 +1,39 @@
+"""
+This script demonstrates the implementation of the ReLU function.
+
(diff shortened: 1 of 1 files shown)


cclauss:
Do we really need numpy?  Given that it is so trivial, perhaps two versions of the function...
1. `relu()` that does not use numpy
2. `relu_numpy()` the current function that uses numpy

Perhaps a `timeit` test or similar to benchmark the two versions against each other.

harshildarji:
@cclauss I agree with the idea of `timeit`, but I think, for now,  the use of `numpy` is best here.

PR #1803, opened 2020-03-15
Fix rehashing function will not call insert_data function

ehdgua01:
### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/data_structures/hashing/hash_table.py
+++ b/data_structures/hashing/hash_table.py
@@ -62,7 +62,7 @@ class HashTable:
         self.size_table = next_prime(self.size_table, factor=2)
         self._keys.clear()
         self.values = [None] * self.size_table  # hell's pointers D: don't DRY ;/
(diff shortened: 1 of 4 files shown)


data_structures/hashing/hash_table.py
```diff
@@ -62,7 +62,7 @@ def rehashing(self):
         self.size_table = next_prime(self.size_table, factor=2)
         self._keys.clear()
         self.values = [None] * self.size_table  # hell's pointers D: don't DRY ;/
-        map(self.insert_data, survivor_values)
+        list(map(self.insert_data, survivor_values))
```
cclauss: Good catch!  But... `self.insert_data()` returns no values so instead of allocating a list that we do not need of values that are all guaranteed to be `None`, let's just do:
```suggestion
        for value in survivor_values:
            self.insert_data(value)
```

PR #1813, opened 2020-03-26
Updated to linear time complex solution.

itsvaibhav01:
Extra detail added for two different possible solutions, one Naive and another using stacks. 
Updated the algorithm to O(n) time complexity from before which was O(N^2).

### DIFF
--- a/data_structures/stacks/next_greater_element.py
+++ b/data_structures/stacks/next_greater_element.py
@@ -7,17 +7,29 @@ def printNGE(arr):
     13 -- 21
     21 -- -1
     3 -- -1
+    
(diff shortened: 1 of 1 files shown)


cclauss:
This is cool but let's do is slightly differently...
* [ ] Keep _both_ algorithms in the same file as two separate functions:
    * [ ] `next_greatest_element()` and
    * [ ] `next_greatest_element_slow()`
* [ ] Both should return the result instead of printing it
* [ ] Both should have the same function signature
* [ ] Both the same doctests.
* [ ] We should benchmark them with [`timeit`](https://docs.python.org/3.8/library/timeit.html) or similar.

PR #1822, opened 2020-03-31
Added new Algorithm to find middle element of Linked List

meSajied:
Please read CONTRIBUTING.md
> 1 files contain uppercase characters:
> data_structures/linked_list/MiddleElementOfLinkedList.py


cclauss:

cclauss:
Cool!  Now let’s encapsulate the algorithm into a function so that we could drop it into a larger program.  How about a function like:
```python
def find_middle_node(list_head: Node) -> Node:
    """
    Given the head_node of a linked list, return the middle node.
    >>> Node("e", Node("d", Node("c", Node("b", Node("a", None))))).data
    c
    """
    # your implementation goes here...
    return middle_node
```

meSajied:
I changed the algorithm so it can take input..

cclauss:
No tests... https://travis-ci.com/github/TheAlgorithms/Python/builds/159585571#L795-L797

meSajied:
why that happend

cclauss:
> Why that happens?

Add doctests as discussed at https://github.com/TheAlgorithms/Python/pull/1822#issuecomment-607034537

cclauss:
Our testing infrastructure is having a bad day https://www.traviscistatus.com

While waiting, please try testing on your computer:
`python3 -m doctest -v middle_element_of_linked_list.py`

cclauss:
The doctests need to be on a function or method like __middle_element()__, not on __main__().

data_structures/linked_list/middle_element_of_linked_list.py
```diff
@@ -13,6 +13,16 @@ def push(self, new_data:int) -> int:
...
     def middle_element(self) -> int:
+        '''
+        >>> link = LinkedList()
+        >>> link.push(4)
+        >>> link.push(3)
+        >>> link.push(8)
+        >>> link.push(7)
+        >>> link.push(9)
+        >>> link.push(11)
+        >>> link.middle_element()
```
cclauss, line 24: After this line you need to put the value (without the >>>) that you expect link.middle_element() to return.  This is the test.  If `link.middle_element()` returns that value then the test will pass.

cclauss: There is no `return` statement in `LinkedList.push()` so it does not currently return anything.

data_structures/linked_list/middle_element_of_linked_list.py
```diff
@@ -0,0 +1,56 @@
...
+        12
+        >>> link.push(17)
+        17
+        >>> link.push(7)
+        7
+        >>> link.push(3)
+        3
+        >>> link.middle_element()
+        10
+        >>>
```
cclauss, line 39: Please add the following lines an then I thing we are ready to merge.
```suggestion
        >>> link.push(20)
        20
        >>> link.push(-20)
        -20
        >>> LinkedList.middle_element()
        12
```

data_structures/linked_list/middle_element_of_linked_list.py
```diff
@@ -0,0 +1,60 @@
...
+    def push(self, new_data:int) -> int:
+        new_node = Node(new_data) 
+        new_node.next = self.head 
+        self.head = new_node
+        return self.head.data
+
+    def middle_element(self) -> int:
+        '''
+        >>> link = LinkedList()
+        >>> link.push(5)
```
cclauss, line 19: One more situation that we should include in our tests...  What happens when we call middle_element() on an __empty__ LinkedList?

```suggestion
        >>> link = LinkedList()
        >>> link.middle_element()
        Traceback (most recent call last):
        ...
        AttributeError: 'NoneType' object has no attribute 'data'
        >>> link.push(5)
        >>> link.push(5)
```

PR #1857, opened 2020-04-12
Rename math/add.py to maths/add.py

cclauss:
In the USA we say math but is the UK-influenced countries, they say maths.


poyea:
mypy raises an error on line 19:
```
maths/add.py:19: error: No overload variant of "sum" matches argument types "int", "int"
maths/add.py:19: note: Possible overload variant:
maths/add.py:19: note:     def [_T, _S] sum(__iterable, Iterable[_T], start: _S) -> Union[_T, _S]
maths/add.py:19: note:     <1 more non-matching overload not shown>
```

cclauss:
Good catch.  Thanks!!

PR #1872, opened 2020-04-16
Initialize set with source in DFS

SandersLin:
### **Describe your change:**
checking if adjacent vertex is in explored set before pushing on to stack. The main difference between DFS and BFS is DFS uses a stack and BFS uses a FIFO queue. 

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


poyea:
I'll further fix this file.

graphs/dfs.py
```diff
@@ -1,33 +1,34 @@
...
-    while stack:
-        v = (
-            stack.pop()
-        )  # one difference from BFS is to pop last element here instead of first one
 
-        if v in explored:
-            continue
-
-        explored.add(v)
+def dfs(graph: Dict, start: str) -> Set[int]:
```
cclauss, line 18:
```suggestion
def depth_first_search(graph: Dict, start: str) -> Set[int]:
```

graphs/dfs.py
```diff
@@ -1,33 +1,34 @@
...
+    """Depth First Search on Graph
 
+       :param graph: directed graph in dictionary format
+       :param vertex: starting vectex as a string
+       :returns: the trace of the search
```
cclauss, line 23: Please add a doctest.

poyea: I'm not sure how to write doctest for this one. The function returns different outputs on different runs.

cclauss: Could we do:
```
>>> all(x in output for x in graph)
True
# and/or
>>> all(x in output for x in graph.values())
True
```

cclauss: If not then perhaps add a comment explaining that doctests would be difficult.

poyea: That works!

PR #1888, opened 2020-04-17
fixed error in  factorial.py

SandersLin:
### **Describe your change:**
`factorial(11)` returns` index out of range`. To implement memoization an auxiliary function is required.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


dynamic_programming/factorial.py
```diff
@@ -12,26 +10,26 @@ def factorial(num):
...
 
     if num < 0:
         return "Number should not be negative."
+
+    return factorial_aux(num, result)
+
+
+def factorial_aux(num, result):
+
+    if num == 0 or num == 1:
```
cclauss, line 23:
```suggestion
    if num in (0, 1):
```

swapnanildutta: This makes more sense.

pawarashish564:
we can use `math.factorial` if required .

dynamic_programming/factorial.py
```diff
@@ -1,37 +1,37 @@
 # Factorial of a number using memoization
-result = [-1] * 10
-result[0] = result[1] = 1
 
 
 def factorial(num):
```
cclauss, line 4:
```suggestion
def factorial(num: int) -> int:
```

dynamic_programming/factorial.py
```diff
@@ -1,37 +1,37 @@
...
     """
+    result = [-1] * (num + 1)
 
     if num < 0:
         return "Number should not be negative."
+
+    return factorial_aux(num, result)
+
+
+def factorial_aux(num, result):
```
cclauss, line 23: Type hints?  Doctests?

dynamic_programming/factorial.py
```diff
@@ -1,37 +1,37 @@
...
+    if num == 0 or num == 1:
+        return 1
+
     if result[num] != -1:
         return result[num]
     else:
-        result[num] = num * factorial(num - 1)
-        # uncomment the following to see how recalculations are avoided
-        # print(result)
+        result[num] = num * factorial_aux(num - 1, result)
```
fhlasek, line 31: Technically, this is more of a "[memoization](https://stackoverflow.com/questions/6184869/what-is-the-difference-between-memoization-and-dynamic-programming#:~:text=Both%20Memoization%20and%20Dynamic%20Programming,solving%20the%20problem%20bottom%2Dup.)" approach. While being perfectly correct, the more appropriate algorithm for this section may look as follows 
```suggestion
    result[0] = 1
    for i in xrange(num):
        result.append((i + 1) * result[i])
    return result[num]
```

cclauss: xrange() was removed from Python on 1/1/2020.

The canonical way to do memoization in Python is with the [__lru_cache__](https://docs.python.org/3/library/functools.html#functools.lru_cache) function decorator.

PR #1921, opened 2020-04-30
Graph coloring

CSenshi:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/backtracking/coloring.py
+++ b/backtracking/coloring.py
@@ -28,12 +28,9 @@ def valid_coloring(
     >>> valid_coloring(neighbours, colored_vertices, color)
     False
     """
-    for i in range(len(neighbours)):
-        # Constraint Failed
-        if neighbours[i] == 1 and colored_vertices[i] == color:
-            return False
-    # All neighbours satisfy constraint
-    return True
+    # Does any neighbour not satisfy the constraints
+    return not any(neighbour == 1 and colored_vertices[i] == color
+                   for i, neighbour in enumerate(neighbours))
 
 
 def util_color(
--- a/backtracking/coloring.py
+++ b/backtracking/coloring.py
@@ -86,9 +86,7 @@ def util_color(
     return False
 
 
-def color(
-    graph: List[List[int]], max_colors: int
-) -> Tuple[bool, Union[List[int], None]]:
+def color(graph: List[List[int]], max_colors: int) -> Tuple[bool, List[int]]:
     """ 
     Wrapper function to call subroutine called util_color
     which will either return True or False.
@@ -113,4 +111,4 @@ def color(
     if util_color(graph, max_colors, colored_vertices, 0):
         return True, colored_vertices
 
-    return False, None
+    return False, []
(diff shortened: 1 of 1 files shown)


backtracking/coloring.py
```diff
@@ -0,0 +1,118 @@
...
+    >>> color = 2
+    >>> valid_coloring(neighbours, colored_vertices, color)
+    False
+    """
+    for i in range(len(neighbours)):
+        # Constraint Failed
+        if neighbours[i] == 1 and colored_vertices[i] == color:
+            return False
+    # All neighbours satisfy constraint
+    return True
```
cclauss, line 36:
```suggestion
    # Does any neighbour not satisfy the constraints
    return not any(neighbour == 1 and colored_vertices[i] == color
                   for i, neighbour in enumerate(neighbours))
```

backtracking/coloring.py
```diff
@@ -0,0 +1,118 @@
...
+    >>> util_color(graph, max_colors, colored_vertices, index)
+    False
+    """
+
+    # Base Case
+    if index == len(graph):
+        return True
+
+    # Recursive Step
+    for i in range(0, max_colors):
```
cclauss, line 79:
```suggestion
    for i in range(max_colors):
```

backtracking/coloring.py
```diff
@@ -0,0 +1,118 @@
...
+            if util_color(graph, max_colors, colored_vertices, index + 1):
+                return True
+            # Backtrack
+            colored_vertices[index] = -1
+    return False
+
+
+def color(
+    graph: List[List[int]], max_colors: int
+) -> Tuple[bool, Union[List[int], None]]:
```
cclauss, line 93: Isn't `typing.Optional[X]` the way to do `Union[X, None]`

Why return `None` when we could just return an empty list?

backtracking/coloring.py
```diff
@@ -0,0 +1,118 @@
+from typing import List, Tuple, Union
+
+"""
```
cclauss, line 3: Please move this docstring above the imports.

backtracking/coloring.py
CSenshi: Thanks for responding! 
What do you think is better? Returning empty list of ints or leaving as it is and change to `typing.Optional[X]`

cclauss: Empty list.  We consistently return one data type (a list).

cclauss: Do we need to return two values?  Is there a situation where the return value is True but the list is empty or False but the list is not empty?

Some doctests on `color()` would be a welcome addition.

CSenshi: There is only one case, where graph is empty list algorithm return (True, []), but it is edge case and probably no one will use it. We can change it to return one data type and when no coloring is found return empty list. Do you think it's better?

cclauss: I thing it is much easier for the caller if we only return the list.

CSenshi: changed :+1:

backtracking/coloring.py
CSenshi: done :+1:

PR #1932, opened 2020-05-03
update crawl_google_results.py

oldcryptogeek:
### **Describe your change:**
Add conditional statement to handle Google Map results in the browser.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [N/A] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [N/A] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [N/A] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [N/A] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

### DIFF
--- a/web_programming/crawl_google_results.py
+++ b/web_programming/crawl_google_results.py
@@ -19,7 +19,7 @@ if __name__ == "__main__":
 
     print(len(links))
     for link in links:
-        if link.text == "Maps": 
-            webbrowser.open(f"{link.get('href')}")
-        else: 
+        if link.text == "Maps":
+            webbrowser.open(link.get('href'))
+        else:
             webbrowser.open(f"http://google.com{link.get('href')}")
--- a/web_programming/crawl_google_results.py
+++ b/web_programming/crawl_google_results.py
@@ -19,5 +19,7 @@ if __name__ == "__main__":
 
     print(len(links))
     for link in links:
-        while (link.text == "Maps"): webbrowser.open(f"{link.get('href')}")[:0]
-        else: webbrowser.open(f"http://google.com{link.get('href')}")
+        if link.text == "Maps": 
+            webbrowser.open(f"{link.get('href')}")
+        else: 
+            webbrowser.open(f"http://google.com{link.get('href')}")


web_programming/crawl_google_results.py
```diff
@@ -19,4 +19,5 @@
 
     print(len(links))
     for link in links:
-        webbrowser.open(f"http://google.com{link.get('href')}")
+        while (link.text == "Maps"): webbrowser.open(f"{link.get('href')}")[:0]
```
cclauss, line 22: So many things confuse me about this single line of code!
1. What search term would show us why this change is useful?
2. Why use `while` instead of `if`?  Won’t that send us in an infinite loop?
3. Why put the parentheses around `link.text == "Maps"`?  psf/black will remove these.
4. Why put the conditional and the statement on the same line?  psf/black will put them on separate lines.
5. Why create an f-string that contains no {}?
6. `webbrowser.open()` will always return `None` so why try to slice `None`?
7. `>>> 'abc'[:0]` is the empty string ('').  Why try to create an empty slice and then not use it?

Please explain.

oldcryptogeek: Thanks for your feedback.  Your criticisms are well founded.

oldcryptogeek: 1. The search term used was "bitcoin"
2. You are correct, I misused the while statement.
3. You are correct.
4. You are correct.
5. Each statement has f strings.
6. To solve the while problem I created for myself, inadvertently.
7. You are correct.

web_programming/crawl_google_results.py
```diff
@@ -19,4 +19,7 @@
 
     print(len(links))
     for link in links:
-        webbrowser.open(f"http://google.com{link.get('href')}")
+        if link.text == "Maps": 
+            webbrowser.open(f"{link.get('href')}")
+        else:
```
cclauss, line 24:
```suggestion
        if link.text == "Maps":
            webbrowser.open(link.get('href'))
        else:
```

oldcryptogeek: Of course.. Thank you.

PR #1938, opened 2020-05-03
Change_method_name

### LINKED ISSUE #1926: Same name for an attribute and a function
Hi, I'm new to programming and I'm not sure if it's a problem, but the code(
Python/data_structures/queue/queue_on_list.py) have the same name for an attribute and a function.
```
class Queue:
    def __init__(self):
        self.entries = []
        self.length = 0
        self.front = 0

    def front(self):
        return self.entries[0]
```
When executed it gives me the error:
TypeError: 'int' object is not callable

mrcloma:
Fixes #1926 
Change method name because it's the same name as an attribute.

### DIFF
--- a/data_structures/queue/queue_on_list.py
+++ b/data_structures/queue/queue_on_list.py
@@ -43,7 +43,7 @@ class Queue:
     """Enqueues {@code item}
     @return item at front of self.entries"""
 
-    def front(self):
+    def first(self):
         return self.entries[0]
 
     """Returns the length of this.entries"""


cclauss:
Why does this pull request have a `diff` that changes every single line in the file?  Is it line ending?

mrcloma:
Maybe I made a mistake but at first only line 46 was changed.

mrcloma:
Yes, its the line endings probably.

cclauss:
Fixed in #1943

PR #1974, opened 2020-05-11
Update linear_search.py

Himanshu-77:
Comment modified in line 17 as Sorting not necessary in Linear Search

### DIFF
--- a/searches/linear_search.py
+++ b/searches/linear_search.py
@@ -14,7 +14,7 @@ python linear_search.py
 def linear_search(sequence, target):
     """Pure implementation of linear search algorithm in Python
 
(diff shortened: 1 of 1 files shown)


searches/linear_search.py
```diff
@@ -14,7 +14,7 @@
 def linear_search(sequence, target):
     """Pure implementation of linear search algorithm in Python
 
-    :param sequence: some sorted collection with comparable items
+    :param sequence: some collection with comparable items (as sorted items not required in Linear Search)
```
cclauss, line 17:
```suggestion
    :param sequence: a collection with comparable items (as sorted items not required in Linear Search)
```

Himanshu-77: Now I have updated it.

PR #1982, opened 2020-05-14
Added strand sort

mateuszz0000:
### **Describe your change:**
I didn't find strand sort (https://en.wikipedia.org/wiki/Strand_sort) implemenation so here it is.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/sorts/strand_sort.py
+++ b/sorts/strand_sort.py
@@ -1,0 +1,60 @@
+from functools import partial
+import operator
+
+
+def strand_sort(arr: list, solution: list, _operator: callable):
+    """
+    Strand sort implementation
+    source: https://en.wikipedia.org/wiki/Strand_sort
+
+    :param arr: Unordered input list
+    :param solution: Ordered output list
+    :param _operator: Operator lt/gt to specify ascent/descent version
+
+    Examples:
+    >>> solution_1 = []
+    >>> strand_sort([4, 2, 5, 3, 0, 1], solution_1, operator.gt)
+    >>> solution_1
+    [0, 1, 2, 3, 4, 5]
+
+    >>> solution_2 = []
+    >>> strand_sort([4, 2, 5, 3, 0, 1], solution_2, operator.lt)
+    >>> solution_2
+    [5, 4, 3, 2, 1, 0]
+    """
+    if not arr:
+        return
+
+    sublist = [arr.pop(0)]
+    for i, item in enumerate(arr):
+        if _operator(item, sublist[-1]):
+            sublist.append(item)
+            arr.pop(i)
+
+    #  merging sublist into solution list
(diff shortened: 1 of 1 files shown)


sorts/strand_sort.py
```diff
@@ -0,0 +1,60 @@
+from functools import partial
+import operator
+
+
+def strand_sort(arr: list, solution: list, _operator: callable):
```
cclauss, line 5: This sort would be easier to use (and test) if it returned a list instead of requiring the caller to allocate and pass in a `solution` list.  It would also be cool if `strand_sort()` was defined to work like the Python builtin [`sorted()`](https://docs.python.org/3/library/functions.html#sorted) which has an optional `reverse` parameter that defaults  to `False`.  We don't need to support the `key` parameter of `sorted()` but the `reverse` would make a lot of sense.

sorts/strand_sort.py
```diff
@@ -0,0 +1,60 @@
...
+                    solution.insert(i, item)
+                    break
+            else:
+                solution.append(item)
+
+    strand_sort(arr, solution, _operator)
+
+
+strand_asc = partial(strand_sort, _operator=operator.gt)
+strand_desc = partial(strand_sort, _operator=operator.lt)
```
cclauss, line 51: I'm not a fan.  Let's lose `partial` and define `strand_sort()` to work like the Python builtin [`sorted()`](https://docs.python.org/3/library/functions.html#sorted) which has an optional `reverse` parameter that defaults  to `False`.

sorts/strand_sort.py
mateuszz0000: I am agreed. One question because this implementation needs recursive passing solution array because algorithm depends on deleting items from input source. Should I  avoid recursion?

sorts/strand_sort.py
mateuszz0000: Sure! Just wondering: is`partial` bad practice?

sorts/strand_sort.py
cclauss:
```python
def strand_sort(arr: list, reverse: bool=False, solution: list=None):
    solution = solution or []
    your code goes here...
```
This will allow the caller to just pass in the `arr` but also allows the function to call itself recursively.  Best of both worlds.

sorts/strand_sort.py
cclauss: Partial has its place.  I tend to use it rarely.  The real point is that sort algorithms in Python should try to have similar calling conventions to the builtin `sorted()` so developers find them intuitive (Pythonic) to use.

mateuszz0000:
@cclauss should I refactor something more?

PR #1989, opened 2020-05-16
Bbp formula

ken437:
### **Describe your change:**
Adds an algorithm that approximates pi using the Bailey-Borwein-Plouffe formula, described at https://en.wikipedia.org/wiki/Bailey%E2%80%93Borwein%E2%80%93Plouffe_formula

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/maths/bbp_formula.py
+++ b/maths/bbp_formula.py
@@ -1,0 +1,47 @@
+def bbp_pi_approximation(num_iterations: int = 1000) -> float:
+    """
+    Uses a popular pi-approximating algorithm known as the 
+    Bailey-Borwein-Plouffe (BBP) formula to approximate pi.
+    Wikipedia page:
+    https://en.wikipedia.org/wiki/Bailey%E2%80%93Borwein%E2%80%93Plouffe_formula
+    @param num_iterations: nonnegative integer representing the number 
+    of terms in the formula to sum (more iterations leads to more accuracy)
+    @return: the approximation of pi given by the formula
+    
+    >>> bbp_pi_approximation(-10)
+    Traceback (most recent call last):
+      ...
+    ValueError: Please input a nonnegative integer
+    >>> bbp_pi_approximation(1.7)
+    Traceback (most recent call last):
+      ...
+    ValueError: Please input a nonnegative integer
+    >>> "%.8f" % bbp_pi_approximation()
+    '3.14159265'
+    >>> "%.8f" % bbp_pi_approximation(1000)
+    '3.14159265'
+    >>> "%.8f" % bbp_pi_approximation(0)
+    '3.13333333'
+    >>> "%.8f" % bbp_pi_approximation(1)
+    '3.14142247'
+    """
+    if (not isinstance(num_iterations, int)) or num_iterations < 0:
+        raise ValueError("Please input a nonnegative integer")
+    pi_approx = 0.0
+    exponential_part = 1.0
+    # num_iterations + 1 because sum notation includes the last term
+    for index in range(num_iterations + 1):
+        pi_approx += exponential_part * (
+            4 / (8 * index + 1)
+            - 2 / (8 * index + 4)
+            - 1 / (8 * index + 5)
+            - 1 / (8 * index + 6)
+        )
+        exponential_part /= 16
+    return pi_approx
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


ken437:
I didn't intend to include aliquot_sum.py in the pull request but it was added for some reason. Is there a way to make my pull request only include the addition of bbp_formula.py?

TheSuperNoob:
This is pretty cool! One issue with this implementation however: 

Currently it is limited to pythons standard double precision calculation. Which means regardless of how many iterations you supply you function with it will not get any more precise than that. To get around this we can use pythons builtin module `Decimal` which allows for arbitrary precision. 

One note about the formula itself: 

This is mainly used for finding the n-th number of pi without needing to calculate any of the previous digits! Which means we can efficiently calculate the 1 billionth digit of pi without needing to know the digits leading up to that one. So maybe another PR can be opened using with this same file but implements a new function called `nth_digit_of_pi()` or something similar to more fully explore the power of this formula.

maths/bbp_formula.py
```diff
@@ -0,0 +1,47 @@
...
+    '3.13333333'
+    >>> "%.8f" % bbp_pi_approximation(1)
+    '3.14142247'
+    """
+    if (not isinstance(num_iterations, int)) or num_iterations < 0:
+        raise ValueError("Please input a nonnegative integer")
+    pi_approx = 0.0
+    exponential_part = 1.0
+    # num_iterations + 1 because sum notation includes the last term
+    for index in range(num_iterations + 1):
```
TheSuperNoob, line 33: I'm not sure the + 1 part actually is needed. By definition this is an infinite sum which just gets more and more accurate with each iteration, so anytime you stop is technically the last iteration :) 
And it might fit better to how you define `num_iterations` as well. As an example, default value for `num_iterations` is 1000, and starting with 0 and ending with 999 gives you a total of 1000 iterations.

Very minor detail though and shouldn't affect the final result.

cclauss:
> I didn't intend to include aliquot_sum.py

Rebase this branch on master.

maths/bbp_formula.py
ken437: I tried to design it so that num_iterations represented the number on top of the sigma expression in the formula. Now that I think about it, though, it might have been better to name the variable something like upper_limit_of_summation rather than num_iterations

ken437:
> This is pretty cool! One issue with this implementation however:
> 
> Currently it is limited to Python’s standard double precision calculation. Which means regardless of how many iterations you supply you function with it will not get any more precise than that. To get around this we can use Python’s builtin module `Decimal` which allows for arbitrary precision.
> 
> One note about the formula itself:
> 
> This is mainly used for finding the n-th number of pi without needing to calculate any of the previous digits! Which means we can efficiently calculate the 1 billionth digit of pi without needing to know the digits leading up to that one. So maybe another PR can be opened using with this same file but implements a new function called `nth_digit_of_pi()` or something similar to more fully explore the power of this formula.

The first paragraph definitely sounds like a good idea. Maybe I could even let the user input the number of digits after the decimal point to round to (e.g. if they input "2" they get 3.14). As for the second paragraph, I could definitely do that in another pull request; I think it would be cool if our repository had an efficient general pi algorithm in one file and a digit extraction algorithm in another file.

TheSuperNoob:
> The first paragraph definitely sounds like a good idea. Maybe I could even let the user input the number of digits after the decimal point to round to (e.g. if they input "2" they get 3.14).

Yes you can do this! From what i can tell each iteration of this formula gives 1 digit of pi so theres not a whole lot you need to change, only update the calculations to be of type `Decimal` where the precision is equal to the number of iterations you do. :) 

> I think it would be cool if our repository had an efficient general pi algorithm in one file and a digit extraction algorithm in another file.

I have already in the past implemented [Chudnovsky's algorithm](https://en.wikipedia.org/wiki/Chudnovsky_algorithm) which is as far as im aware the most efficiant method of calculating many digits of pi. [chudnovsky_algorithm.py](https://github.com/TheAlgorithms/Python/blob/master/maths/chudnovsky_algorithm.py) 
this yields around 14 digits of pi each iteration.

TheSuperNoob:
I assume this revision is abandoned in favor of #1996?

PR #1995, opened 2020-05-18
some trigonometric algorithm

zhbi98:
### **Describe your change:**

Add an algorithm
Added maths.py file in the math directory, internally implemented some trigonometric algorithm, such as sqrt, pow, sin, cos, tan, factorial, power.

### DIFF
--- a/maths/maths.py
+++ b/maths/maths.py
@@ -1,0 +1,138 @@
+
+# package maths
+# Implement some 
+# trigonometric algorithm
+
+Pi = 3.14159265358979323846
+
+
+def _abs(value):
+    # >>> _abs(-10)
+    # 10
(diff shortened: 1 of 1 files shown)


maths/maths.py
```diff
@@ -0,0 +1,138 @@
...
+    # >>> _abs(-10)
+    # 10
+    # >>> _abs(10)
+    # 10
+    if value < 0:
+        value = -value
+    else:
+        value = value
+
+    return value
```
cclauss, line 19: return -value if value < 0 else value

zhbi98: I want to push this part of the code to this algorithm warehouse, what should I do

PR #2000, opened 2020-05-18
some trigonometric algorithm

zhbi98:
### **Describe your change:** 
Created a new file maths.py
Internally implemented some trigonometric functions, root root, factorial, power algorithm

### DIFF
--- a/maths/maths.py
+++ b/maths/maths.py
@@ -1,0 +1,138 @@
+
+# package maths
+# Implement some 
+# trigonometric algorithm
+
+Pi = 3.14159265358979323846
+
+
+def _abs(value):
+    # >>> _abs(-10)
+    # 10
+    # >>> _abs(10)
+    # 10
+    if value < 0:
+        value = -value
+    else:
+        value = value
+
+    return value
+
+
+def _power(value, n):
+    # >>> _power(-10, 2)
+    # 100
+    # >>> _power(10, 2)
+    # 100
+    result = 1
+
+    if n == 0:
+        result = 1
+    else:
+        for i in range(n):
+            result = result * value
+
+    return result
+
+
+# newton iteration method
+def _sqrt(value):
+    if value < 0:
+        print("_sqrt: Value must be greater than or equal to 0")
+    else:
+        x = value
+        t = 0
+        # X(k+1) = 1/2 * (X(k) + vale/X(k))
+        while _abs(x - t) > 1e-15:
+            t = x
+            x = 0.5 * (x + (value / x))
+    
+    # >>> _sqrt(9)
+    # 3
+    # >>> _sqrt(25)
+    # 5
+    return x
+
+
+def _sin(value):
+    value = value * Pi / 180
+    t = value
+    x = 0
+    n = 1
+
+    while _abs(t) > 1e-15:
+        x = x + t
+        n = n + 1
+        t = -t * value * value / (2 * n - 1) / (2 * n - 2)
+
+    if x > 0 and x < 1e-15:
+        x = 0
+
+    # >>> _sin(90)
+    # 1
+    # >>> _sin(0)
+    # 0    
+    return x
+
+
+# factorial
+def _factor(value):
+    # regulations 0! = 1
+    f = 1
+    if value:
+        for i in range(value):
+            f = f * (i + 1)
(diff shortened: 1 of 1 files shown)


cclauss:
https://travis-ci.com/github/TheAlgorithms/Python/builds/166676400#L369

maths/maths.py
```diff
@@ -0,0 +1,112 @@
...
+# trigonometric algorithm
+
+Pi = 3.14159265358979323846
+
+
+def _abs(value):
+    # >>> _abs(-10)
+    # 10
+    # >>> _abs(10)
+    # 10
```
cclauss, line 13:
```suggestion
    """
    >>> _abs(-10)
    10
    >>> _abs(-.5)
    0.5
    >>> _abs(0)
    0
    >>> _abs(.5)
    0.5
    >>> _abs(10)
    10
    """
```
Doctests must be in docstrings, not in comments.  Please do `python3 -m docstring -v maths.py` on your computer and make sure that all tests run and pass.

maths/maths.py
```diff
@@ -0,0 +1,112 @@
...
+        result = 1
+    else:
+        for i in range(n):
+            result = result * value
+
+    return result
+
+
+# newton iteration method
+def _sqrt(value):
```
cclauss, line 34:
```suggestion
def _sqrt(value):
    """
    >>> _sqrt(9)
    3
    >>> _sqrt(25)
    5
    """
```
The docstring must appear at the top of the function before any code.

maths/maths.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def _tan(value):
+    x = _sin(value) / _cos(value)
+    
+    # >>> _tan(90)
+    # 0
+    # >>> _tan(45)
+    # 1      
+    return x
```
cclauss, line 112:
```suggestion
def _tan(value):
    """
    >>> _tan(90)
    0
    >>> _tan(45)
    1
    """
    return _sin(value) / _cos(value)
```

maths/maths.py
```diff
@@ -0,0 +1,112 @@
...
+    # >>> _sqrt(9)
+    # 3
+    # >>> _sqrt(25)
+    # 5
+    return x
+
+
+# factorial
+def _factor(value):
+    # regulations 0! = 1
```
cclauss, line 54:
```suggestion
    """
    regulations 0! = 1
    >>> from math import factorial
    >>> for i in range 360:
    >>>     assert _factor(i) == factorial(i), (i, _factor(i), factorial(i))
    """
```

maths/maths.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _abs(value):
+    # >>> _abs(-10)
+    # 10
+    # >>> _abs(10)
+    # 10
+    return -value if value < 0 else value
+
+
+def _power(value, n):
```
cclauss, line 17:
```suggestion
def _power(value, n):
    """
    >>> from math import pow
    >>> for i in range 360:
    >>>     assert _power(i) == pow(i), (i, _power(i), pow(i))
    """
```

zhbi98: OK, I will take your suggestion

mquevill:
All of the trigonometric functions in the standard `math` library are in radians, which is why the tests are failing that try to compare `sin()` and `_sin()` (and other trig functions). Further more, even if we convert to radians, we will not get the exact same answer, due to float operations.

For example:
```python
maths._sin(720)
7.153372857113005e-13
>>> math.sin(math.radians(720))
-4.898587196589413e-16
```

In your test for `_power`, you end up with `0 ** -100`, which throws `ZeroDivisionError: 0.0 cannot be raised to a negative power`. However, there are larger issues with `_power`... You use `range(n)` to do repeat multiplication, but this will generate `[]` for a negative `n`, so any number raised to a negative `n` will return `1`.

In general, for checking floats, you need to check against a tolerance instead of just using `==`. For instance, `math.sqrt(95) = 9.746794344808963` but `maths._sqrt(95) = 9.746794344808965`. These two values are not equal with `==`, but are within 10^-14 of each other, which is often as good as you can do with floats.

cclauss:
https://docs.python.org/3/library/math.html#math.isclose

PR #2001, opened 2020-05-18
conversions/decimal_to_binary.py: Add type hints

sarcastic-verma:
### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


conversions/decimal_to_binary.py
```diff
@@ -1,7 +1,7 @@
 """Convert a Decimal Number to a Binary Number."""
 
 
-def decimal_to_binary(num):
+def decimal_to_binary(num: int) -> str:
```
cclauss, line 4: It might be even cooler to actually return a binary number. i.e. 0b123 instead of "0b123".

ShubhamKJha:
Python does have a `bin()` function which does the same work as the decimal_to_binary.py here, and since it is implemented within Python, it must be faster.

PR #2017, opened 2020-05-20
Update atbash cipher (doc, doctest, performance)

mateuszz0000:
### **Describe your change:**
Added update for atbash cipher:
- doctests,
- type hints
- little performance improvement

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/ciphers/atbash.py
+++ b/ciphers/atbash.py
@@ -2,6 +2,26 @@
 import string
 
 
+def atbash_slow(sequence: str) -> str:
+    """
+    >>> atbash_slow("ABCDEFG")
+    'ZYXWVUT'
+
+    >>> atbash_slow("aW;;123BX")
+    'zD;;123YC'
+    """
+    output = ""
+    for i in sequence:
+        extract = ord(i)
+        if 65 <= extract <= 90:
+            output += chr(155 - extract)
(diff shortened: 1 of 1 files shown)


mateuszz0000:
With small string speed is not visible - even worse. I've tried with 4 000 000 characters length and current implementation is little bit faster (from 3.2 sec -> 2.2 - 2.4 sec). That's why I wrote "little" in PR description :) 

I can put some performance tests I made If you want :)

mateuszz0000:
> ```
> Running performance benchmarks...
> > atbash_slow() 18.921338292999998 seconds
> >      atbash() 16.922152821999997 seconds
> ```
> 
> How could `atbash()` be modified to be much faster?

I thought about modulo operations and remove completely searching characters in `ascii_letters`. If statement in list comprehensions slowing down too. For now I don't have idea but I know it's possible :)

cclauss:
Would it help if we moved the definition of `letters` and `letters_reversed` outside the function?

mateuszz0000:
> Would it help if we moved the definition of `letters` and `letters_reversed` outside the function?

Perhaps but insignifficant, these strings are too small. The bottleneck is here:
`.. letters.index(c) .. if c in letters else c. ..`

PR #2020, opened 2020-05-20
Fixes in Bidirectional A*

guilyx:
In the current state Bidirectional A* was not working due to an issue in the available vertices check. Could've been highlighted sooner but I only realized it as I added more doctests. Sorry about that.

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/graphs/bidirectional_a_star.py
+++ b/graphs/bidirectional_a_star.py
@@ -234,6 +234,8 @@ class BidirectionalAStar:
 
 if __name__ == "__main__":
     # all coordinates are given in format [y,x]
+    import doctest
+    doctest.testmod()
     init = (0, 0)
     goal = (len(grid) - 1, len(grid[0]) - 1)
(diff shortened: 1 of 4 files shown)


cclauss:
Thanks for doing this.  I wrapped long lines w/ `# doctest: +NORMALIZE_WHITESPACE` and then needed to clean up some codespell issues in several other files.  I would encourage you to put doctests at the method level instead of the class level because it takes a lot of up and down scrolling to jump from the definition of a method to actually seeing how it is called and tested.

Note that this code confuses the flake8 linter because the return statements are of some functions are indented under if statements.  There is no explicit returns when the if statement fails.

As discussed in CONTRIBUTING.md, algorithm functions should not print().  They should raise exceptions or return None, etc. when bad things happen and leave it to the caller to print(), etc.

PR #2052, opened 2020-05-31
doubly_linked_list.py: Add doctests

cclauss:
### **Describe your change:**

* [ ] Add an algorithm?
* [x] Add tests to an existing algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/data_structures/linked_list/doubly_linked_list.py
+++ b/data_structures/linked_list/doubly_linked_list.py
@@ -12,12 +12,12 @@
 class LinkedList:  # making main class named linked list
     """
     >>> linked_list = LinkedList()
-    >>> linked_list.insert_head("a")
-    >>> linked_list.insert_tail("b")
-    >>> _ = linked_list.delete_tail()
+    >>> linked_list.insertHead("a")
+    >>> linked_list.insertTail("b")
+    >>> _ = linked_list.deleteTail()
     >>> linked_list.isEmpty()
     True
-    >>> linked_list.delete_head()
+    >>> linked_list.deleteHead()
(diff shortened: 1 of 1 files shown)


cclauss:
@WilliamHYZhang @AlexDvorak 
Could you please look at why these tests are failing?  Am I doing something wrong?

TheSuperNoob:
> Could you please look at why these tests are failing? Am I doing something wrong?

@cclauss 
I believe the problem here is with the code, not the tests. `line 44: self.head.previous = None` 
The problem here is that on `line 41: self.head = self.head.next` will be `None`, since there is only one element left in the list, head no longer has a next element.

Similarly the problem is also there in `deleteTail` on line 58 and 59, but this time tail will have no previous element.

cclauss:
https://travis-ci.com/github/TheAlgorithms/Python/builds/168928811#L754 is even worse than that.  It is saying that `self.head` is None.

TheSuperNoob:
> https://travis-ci.com/github/TheAlgorithms/Python/builds/168928811#L754 is even worse than that. It is saying that `self.head` is None.

Yes, because on line 41 `self.head` is set to `self.head.next` which would be `None` in this case, since there is no next node.

cclauss:
So,  Can you suggest a fix that allows the tests to pass?

TheSuperNoob:
You can make the tests pass by not trying to use `deleteHead` or `deleteTail` on the final element in the list, however I'd recommend fixing the algorithm so this is possible. I can open a PR later today that fixes this.

PR #2065, opened 2020-06-02
create sum_of_digits.py

vicky1999:
create sum_of_digits.py to find the sum of digits of a number
digit_sum(12345) ---> 15
digit_sum(12345) ---> 10

### DIFF
--- a/maths/sum_of_digits.py
+++ b/maths/sum_of_digits.py
@@ -1,0 +1,19 @@
+def digit_sum(n: int) -> int:
+    """
+    Find the sum of digits of a number.
+
+    >>> digit_sum(12345)
+    15
+    >>> digit_sum(123)
+    6
+    """
+    res=0
+    while n > 0:
+        res += n % 10
+        n = n // 10
+    
+    return res
+
+
+if __name__ == "__main__":
+    print(digit_sum(12345))   # ===> 15
--- a/maths/sum_of_digits.py
+++ b/maths/sum_of_digits.py
@@ -15,4 +15,4 @@ def sum_of_digits(n: int) -> int:
 
 
 if __name__ == "__main__":
-    print(digit_sum(12345))   # ===> 15
+    print(digit_sum(12345))  # ===> 15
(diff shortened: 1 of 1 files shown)


maths/sum_of_digits.py
```diff
@@ -0,0 +1,18 @@
+def digit_sum(n: int) -> int:
```
cclauss, line 1:
```suggestion
def sum_of_digits(n: int) -> int:
```
Seems like a more self documenting name.

maths/sum_of_digits.py
```diff
@@ -0,0 +1,18 @@
...
+    Find the sum of digits of a number.
+
+    >>> digit_sum(12345)
+    15
+    >>> digit_sum(123)
+    6
+    """
+    res = 0
+    while n > 0:
+        res += n % 10
```
cclauss, line 12: Could we optimize using Python builtin https://docs.python.org/3/library/functions.html#divmod ?

maths/sum_of_digits.py
```diff
@@ -0,0 +1,18 @@
...
+        n = n // 10
+    return res
+
+
+if __name__ == "__main__":
+    print(digit_sum(12345))   # ===> 15
```
cclauss, line 18:
```suggestion
    print(digit_sum(12345))  # ===> 15
```

maths/sum_of_digits.py
vicky1999: Yeah, divmod() method would be faster than mod operator.  But for small values, there is not much difference. Everytime, we mod a number with 10 until it becomes 0 will not have much effect on the running time I guess.

cclauss:
```python
def digits(n):
        while n:
            n, digit = divmod(n, 10)
            yield digit
         
print(sum(digits(12345)))
```

vicky1999: Yeah, that's the digit sum function using divmod()

PR #2076, opened 2020-06-06
created perfect_cube.py

vicky1999:
To find whether a number is a perfect cube or not.

perfect_cube(27)   ----> True
perfect_cube(4)     ----> False
perfect_cube(8)     ----> True

### DIFF
--- a/maths/perfect_cube.py
+++ b/maths/perfect_cube.py
@@ -1,15 +1,15 @@
 def perfect_cube(n: int) -> bool:
-	"""
-	Check if a number is a perfect cube or not.
-	>>>perfect_cube(27)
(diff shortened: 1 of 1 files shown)


maths/perfect_cube.py
```diff
@@ -0,0 +1,16 @@
...
+    >>> perfect_cube(27)
+    True
+    >>> perfect_cube(4)
+    False
+    """
+    val = n ** (1 / 3)
+    return (val * val * val) == n
+
+
+if(__name__ == '__main__'):
```
cclauss, line 14: Please format you code with psf/black to remove the clutter.

Issue #1809, opened 2020-03-22
implementation of computer vision algorithm in python  


hritik5102:
I want to work on this issue, please assign it to me.

cclauss:
You do not need to request permission... just submit pull requests.

jeffin07:
@cclauss Can you please give some examples of cv algorithms that you expecting ?

cclauss:
License plate recognition
Is there a tree in this picture?
How many trees are in this picture?
Is there a willow tree in this picture?
What is the species of the tree in the picture?
What are the species of all the trees in the picture?

cclauss:
All talk.  No action.  :-(

AyyESS1:
excuse me I'm have some other focuses in other things these days.So, I had
stopped my activity in git-hub .

Le dim. 31 mai 2020 à 10:44, Christian Clauss <notifications@github.com> a
écrit :

PR #2083, opened 2020-06-08
VGG Classification network

jeffin07:
### **Describe your change:**

Added a deep-learning method for image classification  #1809 using pytorch 

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/computer_vision/vgg.py
+++ b/computer_vision/vgg.py
@@ -1,0 +1,56 @@
+import torch
+import torch.nn as nn
+
+class VGG(nn.Module):
+
+	def __init__(self, cfg :list):
+		super(VGG, self).__init__()
+		self.net = self.build(cfg)
+
+	def build(self, cfg):
+		# Build the network based on the config
+		layers = []
+		n_channels = 1
+		for i in cfg:
+			if i == "2P":
+				layers.append(nn.MaxPool2d(kernel_size=2, stride=2))
+			else:
+				layers.append(nn.Conv2d(n_channels, i, kernel_size = 3, padding=1))
+				layers.append(nn.ReLU())
+				n_channels = i
+				# layers.append(nn.Conv2d(n_channels, i, kernel_size = 3, padding=1),nn.ReLU())
+		# layers.append(nn.Linear(4096,4096))
+		# layers.append(nn.ReLU())
+		# layers.append(nn.Linear(4096,4096))
+		# layers.append(nn.ReLU())
+		# layers.append(nn.Linear(4096,10))
+		# layers.append(nn.Softmax())
+		# print(layers)
+		return nn.Sequential(*layers)
+
+	def forward(self, x):
+		# Do the forward pass, return the output
+		x = x.view(-1,1,28,28)
+		return self.net(x)
+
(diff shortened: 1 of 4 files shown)


cclauss:
What is https://en.wikipedia.org/wiki/VGG and why does the reader need to guess?

jeffin07:
It's the implementation of https://arxiv.org/abs/1409.1556 a classification network

requirements.txt
```diff
@@ -17,3 +17,5 @@ sklearn
 sympy
 tensorflow
 xgboost
+torch
+torchvision
```
cclauss, line 21: Please keep the filenames in `requirements.txt`sorted to reduce the likelihood of duplicate entries.

cclauss:
> It's the implementation of https://arxiv.org/abs/1409.1556 a classification network

The problem is that the URL does not contain the acronym VGG and the code comments do not explain what VGG stands for.  It is going to confuse the reader to have class without knowing what the class name stands for.

computer_vision/vgg.py
```diff
@@ -0,0 +1,95 @@
...
+        # Can change according to your input size
+        # Here i sticked with the paper
+        x = x.view(-1, 1, 224, 224)
+        return self.net(x)
+
+    def __str__(self):
+        return f"VGG for feature extraction : \n {self.net}"
+
+
+cfg = {
```
deadshotsb, line 46: This repository is based on learning purpose. So, it will be better if you could provide an explanation of the usage of cfg parameter.

deadshotsb:
@jeffin07 You can change the VGG to VGG-16/19 term as they are mainly the standard ones and people refer to them mostly.

deadshotsb:
@cclauss VGG is a CNN classification, you can get it by using VGG-16/19. 

There is one problem though, the PR contains the model defination using pytorch and no insights to the layers(their functionality) but am not so sure that if you could call this as a work in algorithms.

cclauss:
* What does the V stand for?
* What does the first G stand for?
* What does the second G stand for?

deadshotsb:
VGG(Visual Geometry Group) or also OxfordNet. Similarly there are AlexNet or ResNET, etc

cclauss:
Thanks, please change the filename to `visual_geometry_group.py` or `oxford_net.py`.

deadshotsb:
@cclauss Actually VGG is often used and using a VGG-16/19 model and renaming the file as vgg_16 / vgg_19 would be beneficial, like using kmp instead of Knuth Morris Pratt.

cclauss:
In this repo we call it [Knuth Morris Pratt](https://github.com/TheAlgorithms/Python/blob/master/strings/knuth_morris_pratt.py) instead of kmp.  We do that _very_ deliberately because we are trying to teach the readers of our code.  We avoid unexplained acronyms because we want our readers to understand the origins and nomenclature of these algorithms.  The 16 and 19 are config choices that have proved to be effective but they are not the algorithm.  https://www.robots.ox.ac.uk/~vgg/research/very_deep

PR #2131, opened 2020-06-17
Remove unnecessary condition in collision_resolution method

marcoscannabrava:
### **Describe your change:**

Removed unnecessary condition in looping hash inside collision_resolution method.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/data_structures/hashing/hash_table.py
+++ b/data_structures/hashing/hash_table.py
@@ -47,7 +47,7 @@ class HashTable:
     def _collision_resolution(self, key, data=None):
         new_key = self.hash_function(key + 1)
 
-        while self.values[new_key] is not None and self.values[new_key] != key:
+        while self.values[new_key] is not None:
 
             if self.values.count(None) > 0:
                 new_key = self.hash_function(new_key + 1)


data_structures/hashing/hash_table.py
```diff
@@ -47,7 +47,7 @@ def _set_value(self, key, data):
     def _collision_resolution(self, key, data=None):
         new_key = self.hash_function(key + 1)
 
-        while self.values[new_key] is not None and self.values[new_key] != key:
+        while self.values[new_key] is not None:
```
mateuszz0000, line 50: Why this condition is unnecessary?

marcoscannabrava: I might be wrong so I'd appreciate if you could explain it to me or point me to any link. 
- How can a value be equal to a key? If it was, would it be a problem?

cclauss: d = {0: 0, 1: 1, 2: 2}  # is a dict where all keys equal their values.

marcoscannabrava: In this example d = {0: 0, 1: 1, 2: 2} if I want to insert a new value and the hashing function somehow returns key=2. 
- Wouldn't we want to go into the loop to produce a new key so it doesn't collide with the key-value pair 2-2? 
- It wouldn't anyway because "is not None" is sufficient condition, right?

PR #2135, opened 2020-06-19
Added minimum cost path algorithm

ruppysuppy:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


cclauss:
What is the use case for this algorithm?  What would be the real world value of solving this problem?

ruppysuppy:
I encountered it a long time back as a problem stating something like: "There is a gold digger travelling from the top left of the grid to the bottom right, help him maximize the gold he encounters on the path". Another variant of the problem can be considered quite useful, __minimum path sum for matrix__ (can be done by changing the max of the top and left to the min of the two). If we, consider the path of a traveler in a N x M grid, the value for in the matrix can be considered accommodation cost and we are trying to minimize it.

ruppysuppy:
Let me know if its not worth adding. I'll close it. Planning to LRU Cache, but I can't understand where (which folder) I should add it in. Could you please tell me @cclauss ?

itsvinayak:
> Let me know if its not worth adding. I'll close it. Planning to LRU Cache, but I can't understand where (which folder) I should add it in. Could you please tell me @cclauss ?

@ruppysuppy  please add a description for this algorithm 
if I am not wrong this problem is the same as[ this](https://www.geeksforgeeks.org/min-cost-path-dp-6/)

ruppysuppy:
@itsvinayak should I change it to minimum cost path from maximum cost path? I'll add the description for the problem

PR #2138, opened 2020-06-20
Added LRU Cache

ruppysuppy:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/other/lru_cache.py
+++ b/other/lru_cache.py
@@ -1,37 +1,42 @@
-class Double_Linked_List_Node():
+from typing import Optional, Callable
+
+
+class DoubleLinkedListNode:
     '''
     Double Linked List Node built specifically for LRU Cache
     '''
 
-    def __init__(self, key, val):
+    def __init__(self, key: int, val: int):
         self.key = key
         self.val = val
         self.next = None
         self.prev = None
 
 
-class Double_Linked_List():
+class DoubleLinkedList:
     '''
     Double Linked List built specifically for LRU Cache
     '''
 
     def __init__(self):
-        self.head = Double_Linked_List_Node(None, None)
-        self.rear = Double_Linked_List_Node(None, None)
+        self.head = DoubleLinkedListNode(None, None)
+        self.rear = DoubleLinkedListNode(None, None)
         self.head.next, self.rear.prev = self.rear, self.head
 
-    def add(self, node: Double_Linked_List_Node) -> None:
+    def add(self, node: DoubleLinkedListNode) -> None:
         '''
         Adds the given node to the end of the list (before rear)
         '''
+
         temp = self.rear.prev
         temp.next, node.prev = node, temp
         self.rear.prev, node.next = node, self.rear
 
-    def remove(self, node: Double_Linked_List_Node) -> Double_Linked_List_Node:
+    def remove(self, node: DoubleLinkedListNode) -> DoubleLinkedListNode:
         '''
         Removes and returns the given node from the list
         '''
+
         temp_last, temp_next = node.prev, node.next
         node.prev, node.next = None, None
         temp_last.next, temp_next.prev = temp_next, temp_last
@@ -39,11 +44,12 @@ class Double_Linked_List():
         return node
 
 
-class Lru_Cache:
+class LruCache:
     '''
-    LRU Cache to store a given capacity of data
+    LRU Cache to store a given capacity of data. Can be used as a stand-alone object
+    or as a function decorator.
 
-    >>> cache = Lru_Cache(2)
+    >>> cache = LruCache(2)
 
     >>> cache.set(1, 1)
 
@@ -54,17 +60,11 @@ class Lru_Cache:
 
     >>> cache.set(3, 3)
 
-    >>> cache.get(2)
-    Traceback (most recent call last):
-    ...
-    ValueError: Key '2' not found in cache
+    >>> cache.get(2)    # None returned
 
     >>> cache.set(4, 4)
 
-    >>> cache.get(1)
-    Traceback (most recent call last):
-    ...
-    ValueError: Key '1' not found in cache
+    >>> cache.get(1)    # None returned
 
     >>> cache.get(3)
     3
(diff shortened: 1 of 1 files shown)


cclauss:
Fascinating approach to us a DLL.  More Pythonic to use a dict.

It would be cool if this could support the decorator calling conventions of https://docs.python.org/3/library/functools.html#functools.lru_cache

ruppysuppy:
I'll add it as the decorator on monday. But I don't know how to do it using dictionary and O(1) time

cclauss:
Skip the dict for now.  Let’s land it as a doubly linked list implementation.

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
+class Double_Linked_List_Node():
```
onlinejudge95, line 1: Do not use `snake_cases` for class naming. Also, no `()` needed as you are not subclassing from any class.

You might wanna try
```
class DoubleLinkedListNode:
```

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
...
+    '''
+    Double Linked List Node built specifically for LRU Cache
+    '''
+
+    def __init__(self, key, val):
```
onlinejudge95, line 6: Type hits please

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        self.key = key
+        self.val = val
+        self.next = None
+        self.prev = None
+
+
+class Double_Linked_List():
```
onlinejudge95, line 13: Same as above

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
...
+    '''
+    Double Linked List built specifically for LRU Cache
+    '''
+
+    def __init__(self):
+        self.head = Double_Linked_List_Node(None, None)
```
onlinejudge95, line 19: How about a sensible default in the method declaration?

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        self.rear = Double_Linked_List_Node(None, None)
+        self.head.next, self.rear.prev = self.rear, self.head
+
+    def add(self, node: Double_Linked_List_Node) -> None:
```
onlinejudge95, line 23: Needs doctests

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        '''
+        Adds the given node to the end of the list (before rear)
+        '''
+        temp = self.rear.prev
+        temp.next, node.prev = node, temp
+        self.rear.prev, node.next = node, self.rear
+
+    def remove(self, node: Double_Linked_List_Node) -> Double_Linked_List_Node:
```
onlinejudge95, line 31: Needs doctests

other/lru_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        Removes and returns the given node from the list
+        '''
+        temp_last, temp_next = node.prev, node.next
+        node.prev, node.next = None, None
+        temp_last.next, temp_next.prev = temp_next, temp_last
+
+        return node
+
+
+class Lru_Cache:
```
onlinejudge95, line 42: Naming convention as suggested above

ruppysuppy:
Thanks a lot for the review @onlinejudge95, I completely overlooked the type hints. I didn't add the doctests for the linked lists as it doesn't have any utility of its own (made for the specific use case for the cache).

other/lru_cache.py
```diff
@@ -1,49 +1,55 @@
-class Double_Linked_List_Node():
+from typing import Optional, Callable
```
cclauss, line 1: Please sort imports so that Callable comes before Optional.  Doing so improves readability and makes it difficult to add duplications.

other/lru_cache.py
```diff
@@ -1,49 +1,55 @@
...
+
         temp_last, temp_next = node.prev, node.next
         node.prev, node.next = None, None
         temp_last.next, temp_next.prev = temp_next, temp_last
 
         return node
 
 
-class Lru_Cache:
+class LruCache:
```
cclauss, line 47:
```suggestion
class LRUCache:
```

other/lru_cache.py
```diff
@@ -54,58 +60,70 @@ class Lru_Cache:
...
+    ...         return 1
+    ...     return fib(num - 1) + fib(num - 2)
 
-    >>> cache.has_key(4)
-    True
+    >>> for i in range(1, 100):
+    ...     res = fib(i)
+
+    >>> fib.cache_info()
+    'CacheInfo(hits=194, misses=99, capacity=100, current size=99)'
```
cclauss, line 88: AWESOME!!!

other/lru_cache.py
```diff
@@ -114,11 +132,46 @@ def set(self, key: int, value: int) -> None:
             node.val = value
             self.list.add(node)
 
-    def has_key(self, key: int) -> bool:
+    def cache_info(self) -> str:
```
cclauss, line 135:
```suggestion
    def __str__(self) -> str:
```

other/lru_cache.py
```diff
@@ -114,11 +132,46 @@ def set(self, key: int, value: int) -> None:
...
         '''
-        Checks if the input key is present in cache
+        Returns the details for the cache instance
+        [hits, misses, capacity, current size]
```
cclauss, line 138:
```suggestion
        Return the details for the cache instance
        [hits, misses, capacity, current_size]
```

other/lru_cache.py
```diff
@@ -114,11 +132,46 @@ def set(self, key: int, value: int) -> None:
...
         '''
-        return key in self.cache
+
+        return f'CacheInfo(hits={self.hits}, misses={self.miss}, \
+capacity={self.capacity}, current size={self.num_keys})'
```
cclauss, line 142:
```suggestion
        return (f'CacheInfo(hits={self.hits}, misses={self.miss},'
                f'capacity={self.capacity}, current size={self.num_keys})')
```
PEP8 is clear that backslashes should not be used in Python code because one whitespace character to the right of the backslash will break the script on a change that is invisible to the reader.

other/lru_cache.py
```diff
@@ -114,11 +132,46 @@ def set(self, key: int, value: int) -> None:
...
+                if func not in LruCache.decorator_function_to_instance_map:
+                    LruCache.decorator_function_to_instance_map[func] = LruCache(size)
+
+                result = LruCache.decorator_function_to_instance_map[func].get(args[0])
+
+                if result is not None:
+                    return result
+
+                result = func(*args, **kwargs)
+                LruCache.decorator_function_to_instance_map[func].set(args[0], result)
```
cclauss, line 162:
```suggestion

                if result is not None:
                if not result:
                    result = func(*args, **kwargs)
                    LruCache.decorator_function_to_instance_map[func].set(args[0], result)
```

other/lru_cache.py
```diff
@@ -54,58 +60,70 @@ class Lru_Cache:
...
+    # class variable to map the decorator functions to their respective instance
+    decorator_function_to_instance_map = {}
+
+    def __init__(self, capacity: int):
+        self.list = DoubleLinkedList()
         self.capacity = capacity
         self.num_keys = 0
+        self.hits = 0
+        self.miss = 0
         self.cache = {}
```
cclauss, line 100: Why do we create both a cache and a decorator_function_to_instance_map?  Could we have one instead of two?

other/lru_cache.py
```diff
@@ -54,58 +60,70 @@ class Lru_Cache:
...
```
cclauss, line 101: def __contains__(key) -> bool:
        """
        >>> cache = LruCache(1)
        >>> 1 in cache
        False
        >>> set(1, 1)
        >>> 1 in cache
        True
        """
        return key in self.cache

cclauss: __contains__() is the modern version of has_key().  It enables the use of `in`.

other/lru_cache.py
```diff
@@ -54,58 +60,70 @@ class Lru_Cache:
 
     >>> cache.set(3, 3)
 
-    >>> cache.get(2)
-    Traceback (most recent call last):
-    ...
-    ValueError: Key '2' not found in cache
+    >>> cache.get(2)    # None returned
```
cclauss, line 63: Why get rid of the exception?
Now the caller is blind to whether 2 was in cache and set to None vs. 2 was not in cache.

other/lru_cache.py
```diff
@@ -114,11 +132,46 @@ def set(self, key: int, value: int) -> None:
...
+                if result is not None:
+                    return result
+
+                result = func(*args, **kwargs)
+                LruCache.decorator_function_to_instance_map[func].set(args[0], result)
+                return result
+
+            def cache_info():
+                if func not in LruCache.decorator_function_to_instance_map:
+                    return "Cache for function not initialized"
```
cclauss, line 167: Should this raise an Exception?

other/lru_cache.py
ruppysuppy: using __repr__ to use `>>> cache` in doctests

other/lru_cache.py
ruppysuppy: using `result is None` as it might also be 0

other/lru_cache.py
ruppysuppy: We need both as the cache maps the keys to the Double Linked List Node

other/lru_cache.py
ruppysuppy: Its given in the type hint, the function will return an integer or None if the key is absent

other/lru_cache.py
ruppysuppy: Yes, but later realized, its an impossible case, so removed it

other/lru_cache.py
ruppysuppy: > One of the coolest submissions ever!!

Thanks a lot :) 👍

ruppysuppy:
@cclauss won't it be merged? And should I add LFU Cache too (it runs in O(capacity) time not O(1))

cclauss:
I approved this PR as it is.  @onlinejudge95 should approve it as well before it is merged.

My sense is that LFU cache is not really needed.  If you want to add it, you can but in a separate Python file in a separate pull request.

PR #2151, opened 2020-06-24
Added LFU Cache

ruppysuppy:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/other/lfu_cache.py
+++ b/other/lfu_cache.py
@@ -1,0 +1,203 @@
+from typing import Callable, Optional
+
+
+class DoubleLinkedListNode:
+    '''
+    Double Linked List Node built specifically for LFU Cache
+    '''
+
+    def __init__(self, key: int, val: int):
+        self.key = key
+        self.val = val
+        self.freq = 0
+        self.next = None
+        self.prev = None
+
+
+class DoubleLinkedList:
+    '''
+    Double Linked List built specifically for LFU Cache
+    '''
+
+    def __init__(self):
+        self.head = DoubleLinkedListNode(None, None)
+        self.rear = DoubleLinkedListNode(None, None)
+        self.head.next, self.rear.prev = self.rear, self.head
+
+    def add(self, node: DoubleLinkedListNode) -> None:
+        '''
+        Adds the given node at the head of the list and shifting it to proper position
+        '''
+
+        temp = self.rear.prev
+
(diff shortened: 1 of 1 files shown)


cclauss:
The comment said that None was returned but the failing test demonstrates that is not the case.

ruppysuppy:
@cclauss, when None is returned in python interpreter or doctests, its not shown. For example try this code in the interpreter:
```
>>> def f():
	return None

>>> f()
>>> 
```
Even though f returns None, running f() won't display None in the interpreter or doctests

other/lfu_cache.py
```diff
@@ -0,0 +1,188 @@
+from typing import Callable, Optional
+
+
+class DoubleLinkedListNode:
+    '''
```
mateuszz0000, line 5: Please change `'''` to `"""` in docstrings

other/lfu_cache.py
```diff
@@ -0,0 +1,188 @@
...
+        '''
+
+        temp = self.rear.prev
+
+        self.rear.prev, node.next = node, self.rear
+        temp.next, node.prev = node, temp
+        node.freq += 1
+        self._position_node(node)
+
+    def _position_node(self, node):
```
mateuszz0000, line 39: Missing typehint

other/lfu_cache.py
```diff
@@ -0,0 +1,188 @@
...
+
+        return node
+
+
+class LFUCache:
+    '''
+    LFU Cache to store a given capacity of data. Can be used as a stand-alone object
+    or as a function decorator.
+
+    >>> cache = LFUCache(2)
```
mateuszz0000, line 63: Could `LFUCache` take optional cache parameter? Would be easier testable.

```
 def __init__(self, capacity: int, cache: dict = None):
        self.list = DoubleLinkedList()
        self.capacity = capacity
        self.num_keys = 0
        self.hits = 0
        self.miss = 0
        self.cache = cache or {}
```

other/lfu_cache.py
cclauss: autoblack will do this automatically so let's not slow the review process on this.

other/lfu_cache.py
cclauss: That seems like a big change.

PR #2161, opened 2020-07-02
double_linear_search algorithm

reinhold-b:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/searches/double_linear_search.py
+++ b/searches/double_linear_search.py
@@ -14,15 +14,15 @@ def double_linear_search(array, x):
     """:param array: the array to be searched
        :param x: the value to be searched (float, int or string)
        :returns index of x, if x is in array, else -1
-       
+
        Examples:
-       
+
        >>>double_linear_search([1, 5, 5, 10], 5)
        1
-       
+
        >>>double_linear_search([1, 5, 5, 10], 100)
        -1
-       
+
        >>>double_linear_search([1, 5, 5, 10], 10)
        3
        """
--- a/searches/double_linear_search.py
+++ b/searches/double_linear_search.py
@@ -36,7 +36,7 @@ def double_linear_search(array, x):
 
 if __name__ == "__main__":
     # should be at index 40
-    x = 100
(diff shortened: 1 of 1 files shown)


cclauss:
> There's no need of examples in the main code, all examples should be in the doctests

For me, I see no reason why we should limit the main code from having examples.

searches/double_linear_search.py
```diff
@@ -0,0 +1,59 @@
...
+Change ARRAY_LENGTH to a number you want to generate an example array,
+which goes up to the number you define:
+ARRAY_LENGTH = 50 --> [0, 1, 2, ..., 49, 50]
+"""
+
+# change for bigger/smaller arrays
+ARRAY_LENGTH = 100
+
+
+def double_linear_search(array, x):
```
cclauss, line 13:
```suggestion
def double_linear_search(array: list, search_term: int) -> int:
```
Let’s find a more self documenting variable name than x.  Using Python type hints as discussed in CONTRIBUTING.md

searches/double_linear_search.py
```diff
@@ -0,0 +1,59 @@
+"""Implementation of a double-linear-search,
+which iterates through the array from both sides: start and end.
+
+Change ARRAY_LENGTH to a number you want to generate an example array,
```
cclauss, line 4: Lines 4 - 10 should be moved into `__main__` because they are not used by your function.

searches/double_linear_search.py
```diff
@@ -0,0 +1,59 @@
...
+
+if __name__ == "__main__":
+    # should be at index 40
+    x = 40
+
+    # this creates an array, which goes up to the number
+    # ARRAY_LENGTH --> [0, 1, ... , 99, 100]
+    array = [x for x in range(ARRAY_LENGTH + 1)]
+
+    # print(array)
```
cclauss, line 57:
```suggestion
```

PR #2251, opened 2020-07-29
Renamed BinarySeach.py

avosa:
### **Change Description:**

Renamed BinarySeach.py to BinarySearch.py


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/usr/bin/python3"
```
cclauss, line 2: That is not the path that my python3 is on.  I use https://brew.sh and https://docs.python.org/3/library/venv.html which both mess with the path where Python can be found.

I would feel much more comfortable with `”usr/bin/env python3”` and would be surprised if that was not vscode’s default behavior.

cclauss:
This is a really out-of-date clone of this repo.

Issue #2128, opened 2020-06-17
Dev sprint ideas: More tests, type hints and less complexity


itsvinayak:
currently, some of the programs use static type checking like this [program](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/fast_fibonacci.py) but some of the programs did not use static typing.

it's a good practice to use static typing as it makes code more clear and readable, should we make it a standard for this repository.we can use [mypy](http://mypy-lang.org/) for testing code 

[more on static typing](https://medium.com/@ageitgey/learn-how-to-use-static-type-checking-in-python-3-6-in-10-minutes-12c86d72677b)
 
thank you

### Dev sprint ideas:
* [ ] [Add tests to Python files with <10% test coverage.](https://github.com/TheAlgorithms/Python/issues/2128#issuecomment-645231020)
* [ ] [Add static typing to functions and methods.](https://github.com/TheAlgorithms/Python/issues/2128)
* [ ] [Set `flake8 --max-complexity=15`](https://github.com/TheAlgorithms/Python/issues/2128#issuecomment-645190839) (Ensure files have strong tests ___before___ refactoring).  Test results from #2139...
    * [ ] ./boolean_algebra/quine_mc_cluskey.py:82:1: C901 'selection' is too complex (17)
    * [ ] ./digital_image_processing/edge_detection/canny.py:20:1: C901 'canny' is too complex (17) @lighttxu
    * [ ] ./graphs/minimum_spanning_tree_prims.py:5:1: C901 'PrimsAlgorithm' is too complex (21)
        * [ ] Add doctests aligned with https://en.wikipedia.org/wiki/Prim%27s_algorithm
        * [ ] In a ___separate___ PR reduce the McCabe complexity
    * [ ] ./linear_algebra/src/polynom-for-points.py:1:1: C901 'points_to_polynomial' is too complex (23) @nic-dern
    * [ ] ./machine_learning/linear_discriminant_analysis.py:251:1: C901 'main' is too complex (25)
    * [x] ./hashes/hamming_code.py:71:1: C901 'emitterConverter' is too complex (16) #2140
    * [x] ./hashes/hamming_code.py:153:1: C901 'receptorConverter' is too complex (20) #2140
    * [x] ./project_euler/problem_551/sol1.py:20:1: C901 'next_term' is too complex (16) #2141

cclauss:
We push all new contributions to use type hints as discussed in CONTRIBUTING.md.
[...]

PR #2259, opened 2020-08-01
added type hints and doctests to arithmetic_analysis/newton_method.py

spamegg1:
Continuing #2128
Also changed some variable names, made them more descriptive.

### **Describe your change:**

### **Checklist:**

### DIFF
--- a/arithmetic_analysis/newton_method.py
+++ b/arithmetic_analysis/newton_method.py
@@ -27,15 +27,12 @@ def newton(function: RealFunc, derivative: RealFunc, starting_int: int,) -> floa
     ...
     ZeroDivisionError: Could not find root
     """
-    prev_guess: float = starting_int
+    prev_guess float(starting_int)
     while True:
         try:
-            next_guess: float = prev_guess - function(prev_guess) / derivative(
-                prev_guess
-            )
+            next_guess = prev_guess - function(prev_guess) / derivative(prev_guess)
         except ZeroDivisionError:
             raise ZeroDivisionError("Could not find root")
-
         if abs(prev_guess - next_guess) < 10 ** -5:
             return next_guess
         prev_guess = next_guess
--- a/arithmetic_analysis/newton_method.py
+++ b/arithmetic_analysis/newton_method.py
@@ -33,11 +33,12 @@ def newton(function: RealFunc, derivative: RealFunc, starting_int: int,) -> floa
             next_guess: float = prev_guess - function(prev_guess) / derivative(
                 prev_guess
             )
-            if abs(prev_guess - next_guess) < 10 ** -5:
-                return next_guess
-            prev_guess = next_guess
         except ZeroDivisionError:
             raise ZeroDivisionError("Could not find root")
+
+        if abs(prev_guess - next_guess) < 10 ** -5:
+            return next_guess
+        prev_guess = next_guess
 
 
 def f(x: float) -> float:
--- a/arithmetic_analysis/newton_method.py
+++ b/arithmetic_analysis/newton_method.py
@@ -3,13 +3,11 @@
 # Newton's Method - https://en.wikipedia.org/wiki/Newton%27s_method
 from typing import Callable
 
+RealFunc = Callable[[float], float]  # type alias for a real -> real function
+
 
 # function is the f(x) and derivative is the f'(x)
-def newton(
-    function: Callable[[float], float],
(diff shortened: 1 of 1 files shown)


arithmetic_analysis/newton_method.py
```diff
@@ -1,23 +1,50 @@
...
-
-# function is the f(x) and function1 is the f'(x)
-def newton(function, function1, startingInt):
-    x_n = startingInt
+from typing import Callable
+
+
+# function is the f(x) and derivative is the f'(x)
+def newton(
+    function: Callable[[float], float],
```
cclauss, line 9: Is there a way to defined a type for `Callable[[float], float]` and then use that type to simplify the call parameters?

arithmetic_analysis/newton_method.py
```diff
@@ -1,23 +1,50 @@
...
+    prev_guess: float = starting_int
     while True:
-        x_n1 = x_n - function(x_n) / function1(x_n)
-        if abs(x_n - x_n1) < 10 ** -5:
-            return x_n1
-        x_n = x_n1
+        if derivative(prev_guess) == 0:
+            raise ZeroDivisionError("float division by zero, could not find root")
+
+        next_guess: float = prev_guess - function(prev_guess) / derivative(prev_guess)
```
cclauss, line 37:
```suggestion
        try:
            next_guess: float = prev_guess - function(prev_guess) / derivative(prev_guess)
        except ZeroDivisionError:
            raise ZeroDivisionError("Could not find root")
```
Avoid calculating the derivative twice.

arithmetic_analysis/newton_method.py
spamegg1: Hmm... I think there is. I could define a Type Variable as described here: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators
Let me try and see if it passes mypy.

arithmetic_analysis/newton_method.py
spamegg1: Makes sense, got it.

arithmetic_analysis/newton_method.py
spamegg1: What do you know? It does work!

arithmetic_analysis/newton_method.py
```diff
@@ -1,23 +1,50 @@
...
-
-
-def f(x):
+        try:
+            next_guess: float = prev_guess - function(prev_guess) / derivative(
+                prev_guess
+            )
+            if abs(prev_guess - next_guess) < 10 ** -5:
+                return next_guess
+            prev_guess = next_guess
```
cclauss, line 38: Lines 36-38 can not raise a ZeroDivisionError so they should not be in the try block as discussed in PEP8.

spamegg1: OK, didn't know that. Understood.

spamegg1:
Hi @cclauss I know this got merged, but something weird must have happened during the merging process. In your last commit, Line 30 changed from `prev_guess: float = starting_int` (which is what I committed) to `prev_guess float(starting_int)` which makes the file currently broken.

cclauss:
It was my mistake...  It is now fixed on master.  Thanks!

arithmetic_analysis/newton_method.py
dhruvmanila: Hey @spamegg1, from OSSU here. I was just going through this issue and looking at the changes made and this will create a problem where if there is a `ZeroDivisionError` then two exceptions will be raised. This PR is merged so it's better for @cclauss to make this small change.

```python
>>> try:
...     a = 1/0
... except ZeroDivisionError:
...     raise ZeroDivisionError("Not possible")
...
  File "<stdin>", line 2, in <module>
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
ZeroDivisionError: Not possible
```

Why? the `except` block is used to handle the exceptions, if you want to raise an exception inside it you need to use the `from` clause like so:

```python
>>> try:
...     a = 1/0
... except ZeroDivisionError:
...     raise ZeroDivisionError("Not possible") from None
...
  File "<stdin>", line 4, in <module>
ZeroDivisionError: Not possible
```

You can also use `from` in other ways which I won't go into detail much here but just wanted to point this out. 😁

PR #2261, opened 2020-08-01
Fixed bug with incorrect LU decomposition

### LINKED ISSUE #2257: arithmetic_analysis/lu_decomposition is wrong
The provided example matrix in the file is
```
[[2, -2, 1], 
[0, 1, 2], 
[5, 3, 1]]
```
The algorithm returns
```
L = [[1.  0.  0. ]
 [0.  1.  0. ]
 [2.5 0.  1. ]]
U = [[ 2.  -2.   1. ]
 [ 0.   1.   2. ]
 [ 0.   8.  -1.5]]
```
Multiplying `L` and `U` does give us back the original matrix. 

However as you can see the matrix `U` is NOT upper-triangular. 

[Wikipedia](https://en.wikipedia.org/wiki/LU_decomposition) says that a matrix admits LU-decomposition if and only if all its principal minors (in this case, the determinants of all the 2x2 matrices obtained by removing a row and a column) are nonzero. Upon inspection we can see that this holds true of the original matrix, so it should definitely admit LU-decomposition.

An [online calculator gives](https://matrixcalc.org/en/#LU-decomposition%28%7B%7B2,-2,1%7D,%7B0,1,2%7D,%7B5,3,1%7D%7D%29)
```
L = [[1.  0.  0. ]
       [0.  1.  0. ]
       [2.5 8.  1. ]]
U = [[ 2.  -2.   1. ]
       [ 0.   1.   2. ]
       [ 0.   0.  -17.5]]
```
Just to be sure I tried it on another matrix (it satisfies the principal minors condition):
```
[[1, 2, 3], 
[4, 5, 6], 
[7, 8, 9]]
```
The algorithm gives
```
L = [[1.  0.  0. ]
       [0.  1.  0. ]
       [7. 0.  1. ]]
U = [[ 1.  2.   3. ]
       [ 4.   5.   6. ]
       [ 0.   -6.  -12.]]
```
Once again `U` is not upper-triangular.

The [online calculator gives](https://matrixcalc.org/en/#LU-decomposition%28%7B%7B1,2,3%7D,%7B4,5,6%7D,%7B7,8,9%7D%7D%29)
```
L = [[1.  0.  0. ]
       [4.  1.  0. ]
       [7. 2.  1. ]]
U = [[ 1.  2.   3. ]
       [ 0.   -3.   -6. ]
       [ 0.   0.  0.]]
```

I'm not sure how to fix this algorithm, so help is definitely wanted.

Svn-Sp:
Fixes #2257  I fixed the Bug in the arithmetic_analysis/lu_decomposition.py file.

 **The Problem was:**
It seemed that the person who developed the algorithm had forgotten that range(i) will return [0, ... ,i-1]
Because of that the algorithm wasn't working properly, but  now it calculates correctly.

### DIFF
--- a/arithmetic_analysis/lu_decomposition.py
+++ b/arithmetic_analysis/lu_decomposition.py
@@ -13,15 +13,15 @@ def LUDecompose(table):
     if rows != columns:
         return []
     for i in range(columns):
-        for j in range(i - 1):
+        for j in range(i):
             sum = 0
-            for k in range(j - 1):
+            for k in range(j):
                 sum += L[i][k] * U[k][j]
             L[i][j] = (table[i][j] - sum) / U[j][j]
         L[i][i] = 1
-        for j in range(i - 1, columns):
+        for j in range(i, columns):
             sum1 = 0
-            for k in range(i - 1):
+            for k in range(i):
                 sum1 += L[i][k] * U[k][j]
             U[i][j] = table[i][j] - sum1
     return L, U
--- a/arithmetic_analysis/newton_method.py
+++ b/arithmetic_analysis/newton_method.py
@@ -27,7 +27,7 @@ def newton(function: RealFunc, derivative: RealFunc, starting_int: int,) -> floa
     ...
     ZeroDivisionError: Could not find root
     """
-    prev_guess float(starting_int)
+    prev_guess = float(starting_int)
     while True:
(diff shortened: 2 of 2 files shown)


spamegg1:
Ah, so it was something simple, that's good.

PR #2263, opened 2020-08-02
Refactored arithmetic_analysis/secant_method.py

spamegg1:
- added missing Wikipedia link
- added type hints and few more doctests
- moved the function `f` to inside `secant_method` as a parameter `func`, instead of calling a globally defined function inside the algorithm body
- renamed some variables and introduced some intermediate variables, to make it more readable and avoid recalculating `f(x1)`
- replaced unused for-loop variable `i` with `_`

### **Describe your change:**

### **Checklist:**

### DIFF
--- a/arithmetic_analysis/secant_method.py
+++ b/arithmetic_analysis/secant_method.py
@@ -24,17 +24,15 @@ def secant_method(
     ...
     ZeroDivisionError: Could not find root
     """
-    lower: float = lower_bound
-    upper: float = upper_bound
     for _ in range(repeats):
-        value: float = func(upper)
-        numerator: float = value * (upper - lower)
-        denominator: float = value - func(lower)
-        try:
-            lower, upper = upper, upper - numerator / denominator
-        except ZeroDivisionError:
+        value = func(upper_bound)
+        numerator = value * (upper_bound - lower_bound)
+        denominator = value - func(lower_bound)
+        if denominator == 0.0:
             raise ZeroDivisionError("Could not find root")
-    return upper
+        lower_bound, upper_bound = upper_bound, upper_bound - numerator / denominator
+
+    return upper_bound
 
 
 def f(x: float) -> float:
--- a/arithmetic_analysis/secant_method.py
+++ b/arithmetic_analysis/secant_method.py
@@ -1,11 +1,43 @@
 # Implementing Secant method in Python
+# https://en.wikipedia.org/wiki/Secant_method
 # Author: dimgrichr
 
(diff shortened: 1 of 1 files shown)


arithmetic_analysis/secant_method.py
```diff
@@ -1,28 +1,48 @@
...
+    2.094546116969569
+    >>> secant_method(lambda x: x ** 3 - 1, 0, 5, 5)
+    0.07992494940223759
+    >>> secant_method(lambda x: x ** 3 - 1, 0, 5, 500)
+    Traceback (most recent call last):
+    ...
+    ZeroDivisionError: Could not find root
+    """
+    lower: float = lower_bound
+    upper: float = upper_bound
```
cclauss, line 28: Why do this?  Why not just use `lower_bound` instead of `lower`?

arithmetic_analysis/secant_method.py
```diff
@@ -1,28 +1,48 @@
...
+    for _ in range(repeats):
+        value: float = func(upper)
+        numerator: float = value * (upper - lower)
+        denominator: float = value - func(lower)
```
cclauss, line 32:
```suggestion
        value = func(upper)
        numerator = value * (upper - lower)
        denominator = value - func(lower)
```
Type hints on local variables are not needed and they create a lot of visual and cognitive clutter.  Python is smart enough to figure out that func() returns a float so value is a float, etc.

arithmetic_analysis/secant_method.py
```diff
@@ -1,28 +1,48 @@
...
+        try:
+            lower, upper = upper, upper - numerator / denominator
+        except ZeroDivisionError:
+            raise ZeroDivisionError("Could not find root")
```
cclauss, line 36:
```suggestion
        if denominator == 0.0:
            raise ZeroDivisionError("Could not find root")
        lower, upper = upper, upper - numerator / denominator
```

cclauss (changes requested):
Looking good...

arithmetic_analysis/secant_method.py
spamegg1: Made all requested changes.

arithmetic_analysis/secant_method.py
spamegg1: Made the changes

arithmetic_analysis/secant_method.py
spamegg1: Removed them.

PR #2275, opened 2020-08-03
Adds new dynamic programming algorithm

kishan151999:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/dynamic_programming/longest_palindromic_subsequence.py
+++ b/dynamic_programming/longest_palindromic_subsequence.py
@@ -1,0 +1,54 @@
+"""
+Longest Palindromic Subsequence Problem:
+Find the longest sequence of characters from a given input string that is the same backwards and forwards
+"""
+
+
+def LongestPalindromicSubsequence(string, n):
+
+    """
+    Computes the longest Palindromic sequence of a string
+
+    :param string: str, the string we provide
+    :param n: int, the length of the string
+    :return L[0][n - 1]: int, the length of the longest Palindromic subsequence
+
+
+
+    >>> LongestPalindromicSubsequence("ABBCDABBC",9)
+    5
+    >>> LongestPalindromicSubsequence("ABACCG",6)
+    3
+    >>> LongestPalindromicSubsequence("55055901565109",14)
+    9
+    """
+
+    # creating an array to store the values generated
+    L = [[1 for i in range(n)] for i in range(n)]
+
+    # Filling in the array created
+    for x in range(2, n + 1):
+        for y in range(n + 1 - x):
+            z = y + x - 1
+            if string[y] == string[z] and x == 2:
+                L[y][z] = 2
+            elif string[y] == string[z]:
+                L[y][z] = L[y + 1][z - 1] + 2
+            else:
+                L[y][z] = max(L[y][z - 1], L[y + 1][z])
+
+    return L[0][n - 1]
+
+
+if __name__ == "__main__":
+    testCase = "ABAC"
+
+    n = len(testCase)
+    print(
+        "The longest palindromic subsequence is: "
+        + str(LongestPalindromicSubsequence(testCase, n))
+    )
+
+    import doctest
+
+    doctest.testmod()
--- a/dynamic_programming/longest_palindromic_subsequence.py
+++ b/dynamic_programming/longest_palindromic_subsequence.py
@@ -6,49 +6,49 @@ https://en.wikipedia.org/wiki/Longest_common_subsequence_problem
 """
 
 
-def longest_palindromic_subsequence(s: str, n: int) -> str:
+def longest_palindromic_subsequence(s: str, length: int) -> int:
 
     """
     Computes the longest Palindromic sequence of a string
 
-    :param string: str, the string we provide
-    :param n: int, the length of the string
+    :param s: str, the string we provide
+    :param length: int, the length of the string
     :return L[0][n - 1]: int, the length of the longest Palindromic subsequence
 
 
 
-    >>> LongestPalindromicSubsequence("ABBCDABBC",9)
+    >>> longest_palindromic_subsequence("ABBCDABBC",9)
     5
-    >>> LongestPalindromicSubsequence("ABACCG",6)
+    >>> longest_palindromic_subsequence("ABACCG",6)
     3
-    >>> LongestPalindromicSubsequence("55055901565109",14)
+    >>> longest_palindromic_subsequence("55055901565109",14)
     9
     """
 
     # creating an array to store the values generated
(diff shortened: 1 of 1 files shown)


dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
+"""
+Longest Palindromic Subsequence Problem:
+Find the longest sequence of characters from a given input string that
+is the same backwards and forwards
+https://en.wikipedia.org/wiki/Longest_common_subsequence_problem
+"""
+
+
+def LongestPalindromicSubsequence(string, n):
```
cclauss, line 9:
```suggestion
def longest_palindromic_subsequence(s: str, n: int) -> str:
```
Python function names are in snake_case as discussed in CONTRIBUTING.md.

Python type hints as discussed in CONTRIBUTING.md.

`string` is the name of a Python standard library module so should not be used as a variable name.

Can we come up with a more self-documenting variable name than `n`?

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
...
+            elif string[y] == string[z]:
+                L[y][z] = L[y + 1][z - 1] + 2
+            else:
+                L[y][z] = max(L[y][z - 1], L[y + 1][z])
+
+    return L[0][n - 1]
+
+
+if __name__ == "__main__":
+    testCase = "ABAC"
```
cclauss, line 46:
```suggestion
    test_case = "ABAC"
```

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
...
+    >>> LongestPalindromicSubsequence("ABBCDABBC",9)
+    5
+    >>> LongestPalindromicSubsequence("ABACCG",6)
+    3
+    >>> LongestPalindromicSubsequence("55055901565109",14)
+    9
+    """
+
+    # creating an array to store the values generated
+    L = [[1 for i in range(n)] for i in range(n)]
```
cclauss, line 29: Uppercase is reserved for constant literals in Python.

cclauss:
What is the use case for this algorithm?  Where would we want to use it and why?

dynamic_programming/longest_palindromic_subsequence.py
kishan151999: Sorry for the mistakes, I will fix the issues.

kishan151999:
> What is the use case for this algorithm? Where would we want to use it and why?
The uses are mostly to do with biology and more specifically DNA. They are used for DNA marking and permitting cutting. They can also be used to change a 1 dimensional DNA chain into a 2 or 3 dimensional structure. It can also be used to compress DNA sequences to save space. 

--https://pubmed.ncbi.nlm.nih.gov/11700586/

--https://stackoverflow.com/questions/23861436/real-life-situations-using-palindrome-algorithm#:~:text=Palindromes%20are%20used%20in%20DNA,2%20or%203%20dimensional%20structure.&text=Palindromes%20are%20strings%20that%20read,plan%2C%20a%20canal%2C%20Panama!

kishan151999:
I have fixed all the issues and committed the changes. Thank you for letting me know.

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
...
+https://en.wikipedia.org/wiki/Longest_common_subsequence_problem
+"""
+
+
+def longest_palindromic_subsequence(s: str, length: int) -> int:
+
+    """
+    Computes the longest Palindromic sequence of a string
+
+    :param s: str, the string we provide
```
cclauss, line 14: This comment can be removed because it tells us nothing that is not already in the call signature.

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
...
+    :param length: int, the length of the string
```
cclauss, line 15: We can calculate the length of the string with `len(s)` so why require the caller to provide it?

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
...
+    >>> longest_palindromic_subsequence("ABBCDABBC",9)
+    5
+    >>> longest_palindromic_subsequence("ABACCG",6)
+    3
+    >>> longest_palindromic_subsequence("55055901565109",14)
+    9
+    """
+
+    # creating an array to store the values generated
+    dp_table = [[1 for i in range(length)] for i in range(length)]
```
cclauss, line 29: Can we come up with a more self-documenting name than `dp_table`?

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
+"""
+Longest Palindromic Subsequence Problem:
+Find the longest sequence of characters from a given input string that
+is the same backwards and forwards
+https://en.wikipedia.org/wiki/Longest_common_subsequence_problem
+"""
+
+
+def longest_palindromic_subsequence(s: str, length: int) -> int:
```
cclauss, line 9:
```suggestion
def longest_palindromic_subsequence(s: str, length: int) -> str:
```
We should return the sequence, not the length of the sequence.  The function is not named `length_of_the_longest_palindromic_subsequence()` so I would not expect an int result.

The caller can always call `len(result)` if they want the length.

dynamic_programming/longest_palindromic_subsequence.py
```diff
@@ -0,0 +1,56 @@
...
+    :return L[0][n - 1]: int, the length of the longest Palindromic subsequence
+
+
+
+    >>> longest_palindromic_subsequence("ABBCDABBC",9)
+    5
+    >>> longest_palindromic_subsequence("ABACCG",6)
+    3
+    >>> longest_palindromic_subsequence("55055901565109",14)
+    9
```
cclauss, line 25: Please provide a negative test case that contains no palendrome.  Also, a test case for the empty string.

dynamic_programming/longest_palindromic_subsequence.py
kishan151999: Would array be fine? If not what would you suggest?

cclauss: What would we call the value that is in each cell of this table?  Are the _weights_ or _probabilities_ or _angstroms_ or what?

PR #2288, opened 2020-08-07
Create stalin_sort.py

GandelXIV:
I have added stalin_sort, a sorting algorithm with complexity of O(n)!!!
Its speed is based on eliminating items that are not in order!


cclauss:
Data loss is not to be encouraged.

PR #2291, opened 2020-08-08
Update scoring_functions.py

tanujdhiman:
We can find accuracy by manually if we are not going to use sklearn library.

### DIFF
--- a/machine_learning/scoring_functions.py
+++ b/machine_learning/scoring_functions.py
@@ -140,7 +140,5 @@ def mbd(predict, actual):
 def manual_accuracy(predict, actual):
     predict = np.array(predict)
     actual = np.array(actual)
-    
     score = np.mean(actual == predict)
-    
     return score
--- a/machine_learning/scoring_functions.py
+++ b/machine_learning/scoring_functions.py
@@ -135,3 +135,12 @@ def mbd(predict, actual):
     score = float(numerator) / denumerator * 100
 
     return score
+
+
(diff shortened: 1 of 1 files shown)


cclauss:
https://travis-ci.com/github/TheAlgorithms/Python/builds/178988952#L252

spamegg1:
@tanujdhiman All you need to do is remove whitespaces from lines 143 and 145

tanujdhiman:
@spamegg1 I think it is a good practice of not dependent on some library like Sklearn. It's my suggestion. What are you views ?

machine_learning/scoring_functions.py
```diff
@@ -135,3 +135,10 @@ def mbd(predict, actual):
     score = float(numerator) / denumerator * 100
 
     return score
+
+
+def manual_accuracy(predict, actual):
+    predict = np.array(predict)
+    actual = np.array(actual)
+    score = np.mean(actual == predict)
+    return score
```
cclauss, line 144:
```suggestion
    return np.mean(np.array(actual) == np.array(predict))
```

cclauss (changes requested):

machine_learning/scoring_functions.py
tanujdhiman: Yep we can do this by just writing a single line.
Thanks @cclauss

cclauss (approved):

tanujdhiman (commented):
Sure.

PR #2296, opened 2020-08-11
Stock market prediction using greadient boosting

abdoulayegk:
### **Describe your change:**
A program to predict the stock market of Microsoft using Gradient boosting regression algorithm
I took the data using the quandl api and then predict the Adj Close price.


cclauss:
See CONTRIBUTING.md
* No doctests
* No type hints
* Fails Travis CI tests
* No algorithmic functions defined...  this looks like a how-to-use instead of an algorithm.

abdoulayegk:
All right I will correct it soon according to the contribution guidelines

PR #2297, opened 2020-08-11
To reverse a string using stack

abdoulayegk:
### **Describe your change:**
A program that takes a string stack and then returns the reverse of that given string

### DIFF
--- a/data_structures/stacks/reverse_string_using_stack.py
+++ b/data_structures/stacks/reverse_string_using_stack.py
@@ -1,0 +1,39 @@
+#Write a program that take input as string stack and return the reverse of that string in python
+
+class Stack():
+    def __init__(self):
+        self.items = []
+# this function will take the item and push onto the top of the stack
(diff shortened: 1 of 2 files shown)


cclauss:
Thanks for the contribution but this is not really useful...  Reversing a string in Python is already fast and trivial: ` s[::-1]s

data_structures/stacks/reverse_string_using_stack.py
```diff
@@ -0,0 +1,39 @@
...
+# this empty string is where we are going to store the reversed string 
+        result = ''
+    while  not Stack.is_empty():
+        result = result + Stack.pop()
+    return result
+
+
+stack = Stack()
+#input_str_ = ' Abdoulaye'
+input_str_= input("Enter the string to reverse: ")
```
cclauss, line 36: As discussed in CONTRIBUTING.md, this line hangs our automated testing.

PR #2316, opened 2020-08-16
Add Leetcode questions in python

Yukti-09:
I have added the first question.


cclauss:
Sorry but I am not a fan of this PR.  There are tons of repos that focus on leetcode problems and this is not one of them.  Outside of the `project_euler` directory, this repo is focused on algorithms sorted by category and leetcode is not that.

Yukti-09:
This https://github.com/TheAlgorithms/C/tree/master/leetcode has leetcode questions in C and so it would be great if we could add the same in python as well.

cclauss:
Our C sister repo has different maintainers with different goals and with a different level of commitment to reviewing pull requests that are better reviewed in other forums.

PR #2389, opened 2020-09-02
capitalize

mohammadreza490:
Capitalization of the first character of a word or a sentence. It works like the capitalize() method of the strings in Python

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/strings/capitalize.py
+++ b/strings/capitalize.py
@@ -1,0 +1,20 @@
+def capitalize(sentence) -> str:
+    """
+    This function will capitalize the first character of a sentence
+
+    >>> capitalize("hello world") 
+        "Hello world"
+    >>> capitalize("123 hello world)
+        "123 hello world" 
+    >>> capitalize(" hello world") 
+        " hello world" 
+    """
+    
+    first_char = sentence[0]
+    new_sentence = str.upper(first_char) + sentence[1:]
+
+if "__name__" == __main__:
+  from doctest import testmod
+  
+  testmode()
+    


cclauss:
@mohammadreza490 Thanks for your submission but in its current form it is just a demonstration of how-to-use the builtin str.capitalize() method.  This repo focuses on building new algorithms, not on how-to-use existing ones.

So here is an alternative approach to consider...  Make a dict with the keys being [string.ascii_lowercase](https://docs.python.org/3/library/string.html#string.ascii_lowercase) and the values being string.ascii_uppercase.  You could then use that dict to capitalize the first letter (if that letter is a key in the dict) without needing to use str.capitalize().

strings/capitalize.py
```diff
@@ -1,3 +1,5 @@
+import string
```
cclauss, line 1:
```suggestion
from string import ascii_lowercase, ascii_uppercase
```

cclauss (changes requested):

mohammadreza490:
> @mohammadreza490 Thanks for your submission but in its current form it is just a demonstration of how-to-use the builtin str.capitalize() method. This repo focuses on building new algorithms, not on how-to-use existing ones.
> 
> So here is an alternative approach to consider... Make a dict with the keys being [string.ascii_lowercase](https://docs.python.org/3/library/string.html#string.ascii_lowercase) and the values being string.ascii_uppercase. You could then use that dict to capitalize the first letter (if that letter is a key in the dict) without needing to use str.capitalize().

Thank you for your comment. I changed the code as you suggested and sorry if I committed many changes because it's my first contribution.

cclauss:
No problems ...  Commit as many times as you want.  We are experimenting to get the right solution.

How could we streamline things using dict.get() ? https://docs.python.org/3/library/stdtypes.html#dict.get

mohammadreza490:
> No problems ... Commit as many times as you want. We are experimenting to get the right solution.
> 
> How could we streamline things using `dict.get()` ? https://docs.python.org/3/library/stdtypes.html#dict.get

I think we can do something like this:
``` 
lower_upper_dict = {
        lower: upper for lower, upper in zip(ascii_lowercase, ascii_uppercase)
    }
first_letter_upper  = lower_upper_dict.get(first_letter)
if(first_letter_upper != None):
    new_sentence = first_letter_upper + sentence[1:]
else:
    new_sentence =  sentence


```

cclauss:
Let's use the default parameter of dict.get().  Hint: define the dict first and define first_character second.

mohammadreza490:
> Let's use the default parameter of dict.get(). Hint: define the dict first and then define first_character second.

What about this?
```
def capitalize(sentence: str) -> str:
    lower_upper_dict = {
        lower: upper for lower, upper in zip(string.ascii_lowercase, string.ascii_uppercase)
    }
    first_letter = lower_upper_dict.get(sentence[0],sentence[0]) #default value is the first character itself
    new_sentence = first_letter + sentence[1:]
    return new_sentence
```

cclauss:
Nice! Now, can you shrink those last three lines into just one line?

mohammadreza490:
> Nice! Now, can you shrink those last three lines into just one line?

Well it seems it's possible:
```
def capitalize(sentence: str) -> str:
    lower_upper_dict = {
        lower: upper for lower, upper in zip(string.ascii_lowercase, string.ascii_uppercase)
    }
    return  lower_upper_dict.get(sentence[0],sentence[0]) + sentence[1:]
```

cclauss:
Looking great!!  We need two more doctests: The empty string ("") will raise an error and a single lowercase letter ("a").

mohammadreza490:
> Looking great!! We need two more doctests: The empty string ("") will raise an error and a single lowercase letter ("a").

I fixed them by checking the length of the string. It returns the capital of a single letter and returns an empty string for ""
```
from string import ascii_lowercase, ascii_uppercase


def capitalize(sentence: str) -> str:
    lower_upper_dict = {
        lower: upper for lower, upper in zip(ascii_lowercase, ascii_uppercase)
    }
    return (
        lower_upper_dict.get(sentence[0], sentence[0])
        + (sentence[1:] if (len(sentence) >= 1) else "")
        if (len(sentence) != 0)
        else ""
    )

```

strings/capitalize.py
```diff
@@ -10,11 +10,20 @@ def capitalize(sentence: str) -> str:
     '123 hello world'
     >>> capitalize(" hello world")
     ' hello world'
+    >>> capitalize("a")
+    'A'
+    >>> capitalize("")
+    ''
     """
     lower_upper_dict = {
```
cclauss, line 18: Let's fast-fail...
```suggestion
    if not sentence:
        return ""
    lower_upper_dict = {
```

cclauss (approved):
Awesome work!!  Thanks.

mohammadreza490:
> Awesome work!! Thanks.

Thanks for your help. I learned new things as well

PR #2422, opened 2020-09-13
Update linear_search.py

Ashley-J-George:
I've added a Python implementation of recursive search algorithm

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/searches/linear_search.py
+++ b/searches/linear_search.py
@@ -38,37 +38,39 @@ def linear_search(sequence, target):
     return None
 
 
-def rec_linear_search(sequence, low, high, target) -> int:
+def rec_linear_search(sequence: list, low: int, high: int, target: int) -> int:
     '''
     Pure implementation of recursive linear search algorithm in Python
 
-    :param sequence: a collection with comparable items (as sorted items not required
-        in Linear Search)
-    :param low: Lower bound of the sequence
-    :param high: Higher bound of the sequence
+    :param sequence: An array of items
+    :param low: Lower bound of the array
+    :param high: Higher bound of the array
     :param target: The element to be found
-    :return: Index of the target or None if target not found
+    :return: Index of the key or -1 if key not found or None in case of an exception
 
     Examples:
-    >>> linear_search([0, 5, 7, 10, 15], 0)
-    0
-
-    >>> linear_search([0, 5, 7, 10, 15], 15)
-    4
-
-    >>> linear_search([0, 5, 7, 10, 15], 5)
-    1
-
-    >>> linear_search([0, 5, 7, 10, 15], 6)
-
+    >>> rec_linear_search([0, 30, 500, 100, 700], 0, 4, 0)
+    The element 0 is present at index 0
+    
+    >>> rec_linear_search([0, 30, 500, 100, 700], 0, 4, 700)
+    The element 700 is present at index 4
+    
+    >>> rec_linear_search([0, 30, 500, 100, 700], 0, 4, 30)
+    The element 5 is present at index 1
+    
+    >>> rec_linear_search([0, 30, 500, 100, 700], 0, 4, -6)
+    The element -6 is not present in the array
     '''
     if high < low:
-        return None
-    if sequence[low] == target:
-        return low
-    if sequence[high] == target:
-        return high
-    return rec_linear_search(sequence, low + 1, high - 1, target)
+        return -1
+    try:
+        if sequence[low] == target:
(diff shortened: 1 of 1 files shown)


spamegg1:
@Ashley-J-George Hi there,
It looks like you did not provide any doctests for `rec_linear_search`.
`rec_linear_search` still has the same doctests as `linear_search`.
Can you provide some different doctests that use `rec_linear_search` instead?

There is an issue when, for example, `low` or `high` are greater than `len(sequence)`. It would give a list index out of bounds error. Maybe handle it with an exception?

Also, type hints for both `linear_search` and `rec_linear_search` would be appreciated. For example:
```python
def rec_linear_search(sequence: list, low: int, high: int, target: float) -> int:
```

searches/linear_search.py
```diff
@@ -38,6 +38,39 @@ def linear_search(sequence, target):
     return None
 
 
+def rec_linear_search(sequence, low, high, target) -> int:
```
cclauss, line 41: Type hints on function parameters?

searches/linear_search.py
```diff
@@ -38,6 +38,39 @@ def linear_search(sequence, target):
...
+    4
+
+    >>> linear_search([0, 5, 7, 10, 15], 5)
+    1
+
+    >>> linear_search([0, 5, 7, 10, 15], 6)
+
+    '''
+    if high < low:
+        return None
```
cclauss, line 66: Should this raise a ValueError for invalid input?

Ashley-J-George: This is the condition that tells us that high has crossed low and the element is not in the list

searches/linear_search.py
Ashley-J-George: I'll add it

searches/linear_search.py
cclauss: We promised on line 41 that this function would return an it but now we are returning an non-int.  We have two choices:
1. Return -1 as a signal that the element is not in the list --or--
2. Raise an error.

This is like [str.find()](https://docs.python.org/3/library/stdtypes.html#str.find) vs. [str.index](https://docs.python.org/3/library/stdtypes.html#str.index).

Ashley-J-George:
Hey @Ashley-J-George,
Something went wrong with the build.

TravisCI finished with status errored, which means the build failed because of something unrelated to the tests, such as a problem with a dependency or the build process itself.


@cclauss @spamegg1  How should I rectify this?

Ashley-J-George:
When the element is not found I have made the function return -1, and added
a try except to handle the index error

spamegg1:
@Ashley-J-George Build log says
> 3     W293 blank line contains whitespace

Very easy to fix! Check out [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md) about running `black` on your code.

searches/linear_search.py
```diff
@@ -38,6 +38,41 @@ def linear_search(sequence, target):
     return None
```
cclauss, line 38: Line 14 promises that we are going to return an int but line 38 returns None.  We should not surprise the caller.

searches/linear_search.py
```diff
@@ -38,6 +38,41 @@ def linear_search(sequence, target):
...
+
+    :param sequence: An array of items
+    :param low: Lower bound of the array
+    :param high: Higher bound of the array
+    :param target: The element to be found
+    :return: Index of the key or -1 if key not found or None in case of an exception
+
+    Examples:
+    >>> rec_linear_search([0, 30, 500, 100, 700], 0, 4, 0)
+    The element 0 is present at index 0
```
cclauss, line 53: This function returns an int.
```suggestion
    0
```

searches/linear_search.py
```diff
@@ -38,6 +38,41 @@ def linear_search(sequence, target):
...
+    if high < low:
+        return -1
+    try:
+        if sequence[low] == target:
+            return low
+        if sequence[high] == target:
+            return high
+        return rec_linear_search(sequence, low + 1, high - 1, target)
+    except IndexError:
+        print('Invalid upper or lower bound!')
```
cclauss, line 73: CONTRIBUTING.md says that `algorithmic functions should not print()` so raise Error.

cclauss (approved):
Nice work!!

Ashley-J-George:
Thank You!

PR #2433, opened 2020-09-15
Added binary_xor_operator.py and binary_and_operator.py

vivek9patel:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/bit_manipulation/binary_xor_operator.py
+++ b/bit_manipulation/binary_xor_operator.py
@@ -9,9 +9,9 @@ def binary_xor(a: int, b: int):
     >>> binary_xor(25, 32)
     '0b111001'
     >>> binary_xor(37, 50)
-    '0b10111'
+    '0b010111'
     >>> binary_xor(21, 30)
-    '0b1011'
+    '0b01011'
     >>> binary_xor(58, 73)
     '0b1110011'
     >>> binary_xor(0, 255)
--- a/bit_manipulation/binary_xor_operator.py
+++ b/bit_manipulation/binary_xor_operator.py
@@ -1,0 +1,50 @@
+
+def binary_xor(a: int, b: int):
+    """
+    Take in 2 integers, convert them to binary,
+    return a binary number that is the
+    result of a binary xor operation on the integers provided.
+
+    >>> binary_xor(25, 32)
+    '0b111001'
+    >>> binary_xor(37, 50)
+    '0b10111'
+    >>> binary_xor(21, 30)
+    '0b1011'
+    >>> binary_xor(58, 73)
+    '0b1110011'
+    >>> binary_xor(0, 255)
+    '0b11111111'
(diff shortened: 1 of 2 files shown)


vivek9patel:
@cclauss  can you please review this?

bit_manipulation/binary_xor_operator.py
```diff
@@ -0,0 +1,51 @@
...
+    TypeError: 'float' object cannot be interpreted as an integer
+    >>> binary_xor("0", "1")
+    Traceback (most recent call last):
+        ...
+    TypeError: '<' not supported between instances of 'str' and 'int'
+    """
+    if a < 0 or b < 0:
+        raise ValueError("the value of both input must be positive")
+
+    a_binary = str(bin(a))[2:]  # remove the leading "0b"
```
Abdujabbor, line 37: why not just use 

def binary_xor(a: int, b: int):
    xor = a ^ b

    return bin(xor)

would not it be faster or not? 

and I don't understand why xor operators binary representation in your result differ than core xor operator: 

for inputs: 
37 ^ 50 = 0b10111

but in your version:
37 ^ 50 = 0b010111

yes of course you would be right if you say that leading zeros would not change the value but I think it would be better if we remove leading zeros after 0b ? 

it's just my opinion, maybe @cclauss and others are agree with your version.

cclauss: Often in this repo we are trying to demonstrate ___how___ an algorithm works. Our primary goal is education not always optimized performance.  If we just use Python's builtin capabilities and its standard library we can often write code in a line or two but we lose all of the educational value.  Often the one or two-liner is great to use in our doctests to prove that the educational long-form gives the same results as the builtin short-form over a range of test values.

vivek9patel: The first thing using a^b gives us decimal representation on xor value
this function gives us a binary representation of xor function.
Totally agree with @cclauss ,this is just for learning and practicing things!

vivek9patel: > Often in this repo we are trying to demonstrate _**how**_ an algorithm works. Our primary goal is education not always optimized performance. If we just use Python's builtin capabilities and its standard library we can often write code in a line or two but we lose a lot of educational value. Often the one or two-liner is great to use in our doctests to prove that the educational long-form gives the same results as the builtin short-form over a range of test values.

But one more thing @cclauss does my PR is having any problems, because it is one week and it is not reviewed yet!?

cclauss (approved):
LGTM

bit_manipulation/binary_xor_operator.py
Abdujabbor: @cclauss ok, understand, thank you for your clarification

PR #2437, opened 2020-09-16
Added a file characters_unique

nishithshowri006:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


Kush1101 (commented):
Please run your code through `Black` code formatted as discussed in [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).

strings/charactes_unique.py
```diff
@@ -18,23 +27,31 @@ def isUniqueSet(s):
         hashSet.add(i)
     return True
 
-def isUniqueHash(s):
+
+def isUniqueHash(s: str) -> str:
```
Kush1101, line 31: Avoid using a single character as a variable names for variables whose scope extends just a few lines. Try making it more descriptive. `string`

strings/charactes_unique.py
```diff
@@ -18,23 +27,31 @@ def isUniqueSet(s):
...
     """
-    This uses Hashes
+    This function uses ascii values as index to determine
+    the string contains unique characters
     s -> string
```
Kush1101, line 35: After having Type hints, there is no need for this.

strings/charactes_unique.py
cclauss: Strings is the name of the directory,string is the name of a module in the Python standard library, and str is the name of a Python builtin data type so in this case, s is an acceptable.

strings/charactes_unique.py
cclauss: Sorry but Python 2 died on 1/1/2020.  One of the ways that Python 3 is different than Python 2 is that all str are Unicode.  Please include doctests that demonstrate that this code proper deals with sentences that include ãèíôü😀🎄

mateuszz0000 (changes requested):

strings/characters_unique.py
```diff
@@ -0,0 +1,53 @@
...
+    This function uses a set to determine the string contains unique characters
+    >>> isUniqueSet("train")
+    True
+    >>> isUniqueSet("test")
+    False
+    """
+
+    hashSet = set()
+    for i in string:
+        if i in hashSet:
```
mateuszz0000, line 23: `hashSet` -> `hash_set`

strings/characters_unique.py
```diff
@@ -0,0 +1,53 @@
...
+def isUniqueSet(string: str) -> bool:
+    """
+    This function uses a set to determine the string contains unique characters
+    >>> isUniqueSet("train")
+    True
+    >>> isUniqueSet("test")
+    False
+    """
+
+    hashSet = set()
```
mateuszz0000, line 21: We can use `set` in different way. We can just `return len(list) == len(set) <- pseudocode' to determine unique characters.

mateuszz0000 (changes requested):

PR #2442, opened 2020-09-17
Contains loops.py add

kanthuc:
### **Describe your change:**
Added an algorithm which checks a linked list for loops and returns true if one is found.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


cclauss (approved):

cclauss (changes requested):
We need at least two doctests.  One that returns True and one that returns False.  We need to defined what a loop is.  Is a loop just the same value twice in a row? Would 1, 2, 1, 2 be a loop?

data_structures/linked_list/contains_loop.py
```diff
@@ -13,8 +13,25 @@ def __init__(self, data: Any) -> None:
 
 def contains_loop(root: Node) -> bool:
     """
-    given a node, returns true if linked list contains a loop
+    A linked list contains a loop when traversing through a loop, no null is reached.
```
cclauss, line 16: This explanation is still unclear to me.  What is a null in this context?

cclauss (approved):

PR #2451, opened 2020-09-19
Create problem_54 in project Euler

dhruvmanila:
Add solution and test files (in pytest) for [project Euler problem 54](https://projecteuler.net/problem=54)

There are a couple of extra features of the PokerHand class:
- A public method `PokerHand.hand_name()` which returns the name of hand as stated in the [Wikipedia link](https://en.wikipedia.org/wiki/List_of_poker_hands)
- Added rich comparison operators such as `<, >, <=, >=, ==, !=` and this in turn gives us the ability to use the builtin `sort` method and `sorted` function on a list of PokerHand objects. 

The test file `test_poker_hand.py` should be picked up by Travis and it checks every function of `sol1.py` through `pytest`.

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


dhruvmanila:
Hi, I'll fix the codespell errors.

What does this codespell comment trying to say?

`'project_euler/problem_54/test_poker_hand.py': [(209, 'ans', 'and'), (217, 'ans', 'and'), (218, 'ans', 'and')]}`

Should I change the variable named `ans` to something more verbose like `answer`? This variable is in `test_poker_hand.py` file.

project_euler/problem_54/sol1.py
```diff
@@ -0,0 +1,366 @@
...
+
+Resources used:
+https://en.wikipedia.org/wiki/Texas_hold_%27em
+https://en.wikipedia.org/wiki/List_of_poker_hands
+
+Similar problem on codewars:
+https://www.codewars.com/kata/ranking-poker-hands
+https://www.codewars.com/kata/sortable-poker-hands
+"""
+from typing import Tuple, List
```
cclauss, line 43: Please run isort on the imports.

project_euler/problem_54/sol1.py
```diff
@@ -0,0 +1,366 @@
...
+        return self._hand
+
+    # Rich comparison operators (used in list.sort() and sorted() builtin functions)
+    # Note that this is not part of the problem but another extra feature where
+    # if you have a list of PokerHand objects, you can sort them just through
+    # the builtin functions.
+    def __eq__(self, other):
+        if isinstance(other, PokerHand):
+            return self.compare_with(other) == "Tie"
+        return NotImplemented
```
cclauss, line 343: Should this be NotImplementedError instead?

project_euler/problem_54/sol1.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def __lt__(self, other):
+        if isinstance(other, PokerHand):
+            return self.compare_with(other) == "Loss"
+        return NotImplemented
+
+    def __le__(self, other):
+        if isinstance(other, PokerHand):
```
cclauss, line 351: <=

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,221 @@
...
+    poker_hands = os.path.join(script_dir, "poker_hands.txt")
+    with open(poker_hands, "r") as file_hand:
+        for line in file_hand:
+            player_hand = line[:14].strip()
+            opponent_hand = line[15:].strip()
+            player, opponent = PokerHand(player_hand), PokerHand(opponent_hand)
+            output = player.compare_with(opponent)
+            if output == "Win":
+                ans += 1
+    assert ans == 376
```
cclauss, line 221: Does this match https://github.com/nayuki/Project-Euler-solutions/blob/master/Answers.txt#L54?

project_euler/problem_54/sol1.py
```diff
@@ -0,0 +1,366 @@
...
+        elif self._hand_type in [20, 16]:
+            join = "over" if self._hand_type == 20 else "and"
+            return name + f", {pair1}s {join} {pair2}s"
+        elif self._hand_type == 23:
+            return name
+        else:
+            return name + f", {high}"
+
+    def _compare_cards(self, other: "PokerHand") -> str:
+        # Comparing in reverse order as the list is sorted
```
cclauss, line 224: Why sort the list backwards?  If it is sorted in the way you want to use it then the logic is easier to follow and we can use enumerate.

project_euler/problem_54/sol1.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def _get_high_card(self) -> int:
+        return self._cards[-1][0]
+
+    def _card_value_list(self) -> list:
+        return list(map(itemgetter(0), self._cards))
+
+    def _is_flush(self) -> bool:
+        suit = self._cards[0][-1]
+        return all(map(lambda card: suit in card, self._cards))
```
cclauss, line 263:
```suggestion
        return all(suit in card for card in self._cards)
```
As discussed in CONTRIBUTING.md.

project_euler/problem_54/sol1.py
dhruvmanila: No, I know people get confused between `NotImplemented` and `NotImplementedError` but those are two different things and not to be confused with each other. The major difference is that `NotImplementedError` is an exception while `NotImplemented` is a way to tell python that this method on those two objects is not implemented.

One of the ways `raise NotImplementedError` is used is when you make an abstract base class and you want the child classes to implement a method you defined otherwise, it will raise this error. For example:

```python
class Random(object):
    def some_method(self):
        raise NotImplementedError

class ChildRandom(Random):
    def some_method(self):
        return some_value
```
In the above example, if I didn't implement `some_method` then python will complain by raising `NotImplementedError`. You can achieve the same effect using `@abstractmethod` decorator.

`NotImplemented` is used when you define those dunder methods for the operators such as `+, -, ==, <` and the user inputs an object which is not the same as `self`, it will raise this error where it tells you `operator is not supported between the instances of <self class> and <other class>`
For example, if you're trying `a == b`, python will call `a.__eq__(b)` and if this returns `NotImplemented` then python will try `b.__eq__(a)` and if that one returns `NotImplemented` as well then python will raise an appropriate error.

https://docs.python.org/3.6/library/constants.html#NotImplemented
https://docs.python.org/3.6/library/exceptions.html#NotImplementedError

project_euler/problem_54/test_poker_hand.py
dhruvmanila: Yes and all tests mentioned in `test_poker_hand.py` are picked up by Travis and being tested including this one.

project_euler/problem_54/sol1.py
dhruvmanila: I didn't sort the list backward. It is sorted in ascending order but in this case, I need the elements in reverse order. When two poker hands are tied you need to check the card values from highest to lowest. I don't think `enumerate` will work here as I am comparing elements from two different lists.

Let me see, I don't think it's feasible as I need it in `_is_five_high_straight` (I use `pop` to remove the last element) and also it's easy to look the hand in ascending order. Still, I'm open to suggestions.

project_euler/problem_54/sol1.py
dhruvmanila: Ah, missed that. Thanks for pointing it out and yes this looks more readable. I will make the following changes in my next commit:
- Remove `map` and use list comprehension
- Sort imports using `isort`
- Sort in reverse? (still in discussion)

project_euler/problem_54/test_poker_hand.py
dhruvmanila: I think you meant this line: [Problem 054: 376](https://github.com/nayuki/Project-Euler-solutions/blob/82d6bcb97aa2979ab8fa114dc47f6e4627f94aef/Answers.txt#L61)

cclauss:
> Should I change the variable named ans to something more verbose like answer? 

Yes.

project_euler/problem_54/sol1.py
cclauss: > I don't think enumerate will work here as I am comparing elements from two different lists.

enumerate() works fine.  It allows you to replace self.hand.cards[i] with card.

dhruvmanila: Oh right, but I still have to keep `other._cards[i][0]` and `enumerate` only takes a start argument. Also, I won't be able to just put `cards` in place of `self._cards[i][0]` as my internal state of the hand is a list of two tuple `[(value, suit), (value, suit) ...]`. And as I pointed out, I will have to change in a few other places where I'm assuming the list is in ascending order especially in `_is_five_high_straight`.

Oh damn, I got an idea which can simplify a lot of things: I am mostly using the card values except in `_is_flush` to determine whether all suit is the same. I have already separated the values and so what I will do is make a `set` of cards and if the length of `set` is one then it's a flush otherwise it is not. This will remove all the double indexing and I will incorporate `enumerate` into it as well. Thanks for all the suggestions!

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+    ("7C 7S KH 2H 7H", 17),
+    ("3C KH 5D 5S KH", 16),
+    ("QH 8H KD JH 8S", 15),
+    ("2D 6D 9D TH 7D", 14),
+]
+
+
+def generate_random_hands():
+    N_RANDOM_TESTS = 100
+    RANDOM_TESTS = []
```
cclauss, line 141:
```suggestion
    random_tests = []
```
This is not a constant.  It is a value that changes during runtime.

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+    for _ in range(N_RANDOM_TESTS):
+        play, oppo = randrange(len(SORTED_HANDS)), randrange(len(SORTED_HANDS))
+        expected = ["Loss", "Tie", "Win"][(play >= oppo) + (play > oppo)]
+        hand, other = SORTED_HANDS[play], SORTED_HANDS[oppo]
+        RANDOM_TESTS.append((hand, other, expected))
```
cclauss, line 146: Could lines 143 to 146 be reformulated as a function `generate_random_hand()` (singular)?

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
+import os
+from itertools import chain
+from random import randrange, shuffle
+
+import pytest
+
+from .sol1 import PokerHand
+
+SORTED_HANDS = [
```
cclauss, line 9: If the order will not be changed at runtime, then please make this a tuple instead of a list.  That will slightly more efficient and will signal the reader that the order will not be modified at runtime.

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.parametrize("hand, other, expected", generate_random_hands())
+def test_compare_random(hand, other, expected):
+    player, opponent = PokerHand(hand), PokerHand(other)
+    assert player.compare_with(opponent) == expected
+
+
+def test_hand_sorted():
+    POKER_HANDS = list(map(PokerHand, SORTED_HANDS))
```
cclauss, line 194:
```suggestion
    POKER_HANDS = [PokerHand(hand) for hand in SORTED_HANDS]
```

project_euler/problem_54/sol1.py
```diff
@@ -0,0 +1,366 @@
...
+        # 17: Three of a kind
+        # 16: Two pairs
+        # 15: One pair
+        # 14: High card
+        if self._is_flush():
+            if self._is_five_high_straight() or self._is_straight():
+                if sum(self._card_values) == 60:
+                    return 23
+                else:
+                    return 22
```
cclauss, line 249:
```suggestion
                return 23 if sum(self._card_values) == 60 else 22
```

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+        expected = ["Loss", "Tie", "Win"][(play >= oppo) + (play > oppo)]
+        hand, other = SORTED_HANDS[play], SORTED_HANDS[oppo]
+        RANDOM_TESTS.append((hand, other, expected))
+    return RANDOM_TESTS
+
+
+@pytest.mark.parametrize("hand, expected", TEST_FLUSH)
+def test_hand_is_flush(hand, expected):
+    player = PokerHand(hand)
+    assert player._is_flush() == expected
```
cclauss, line 153:
```suggestion
    assert PokerHand(hand)._is_flush() == expected
```
The player is not so important as the hand itself.

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.parametrize("hand, expected", TEST_STRAIGHT)
+def test_hand_is_straight(hand, expected):
+    player = PokerHand(hand)
+    assert player._is_straight() == expected
```
cclauss, line 159:
```suggestion
    assert PokerHand(hand)._is_straight() == expected
```

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+def test_hand_is_five_high_straight(hand, expected, cards):
+    player = PokerHand(hand)
+    assert player._is_five_high_straight() == expected
+    assert player._cards == cards
+
+
+@pytest.mark.parametrize("hand, expected", TEST_KIND)
+def test_hand_is_same_kind(hand, expected):
+    player = PokerHand(hand)
+    assert player._is_same_kind() == expected
```
cclauss, line 172:
```suggestion
    assert PokerHand(hand)._is_same_kind() == expected
```

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.parametrize("hand, expected", TEST_TYPES)
+def test_hand_values(hand, expected):
+    player = PokerHand(hand)
+    assert player._hand_type == expected
```
cclauss, line 178:
```suggestion
    assert PokerHand(hand)._hand_type == expected
```

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.parametrize("hand, other, expected", TEST_COMPARE)
+def test_compare_simple(hand, other, expected):
+    player, opponent = PokerHand(hand), PokerHand(other)
+    assert player.compare_with(opponent) == expected
```
cclauss, line 184:
```suggestion
    assert PokerHand(hand).compare_with(PokerHand(other)) == expected
```

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.parametrize("hand, other, expected", generate_random_hands())
+def test_compare_random(hand, other, expected):
+    player, opponent = PokerHand(hand), PokerHand(other)
+    assert player.compare_with(opponent) == expected
```
cclauss, line 190:
```suggestion
    assert PokerHand(hand).compare_with(PokerHand(other)) == expected
```

cclauss (approved):
This is so cool!

project_euler/problem_54/test_poker_hand.py
dhruvmanila: I will make the change but I don't see any difference. Can you tell me why are you suggesting this change?

project_euler/problem_54/test_poker_hand.py
dhruvmanila: Yes, it generates one pair of hand and it's expected output, one at a time, as many times as `N_RANDOM_TESTS` suggests. Do you want me to split this function in two: one generating a single pair plus the output and the other cumulating all into a list?

cclauss: Yes, Please.

project_euler/problem_54/test_poker_hand.py
```diff
@@ -0,0 +1,234 @@
...
+    ("8C 9C 5C 3C TC", 19),
+    ("JS QS 9H TS KH", 18),
+    ("7C 7S KH 2H 7H", 17),
+    ("3C KH 5D 5S KH", 16),
+    ("QH 8H KD JH 8S", 15),
+    ("2D 6D 9D TH 7D", 14),
+]
+
+
+def generate_random_hands():
```
cclauss, line 139:
```suggestion
def generate_random_hands(number_of_hands: int = 100):
    return (generate_random_hand() for _ in number_of_hands)
```

dhruvmanila: Ah, the generator object, awesome idea! 😁
I will incorporate this and other changes you suggested right now.

cclauss:
An awesome contribution!  Thanks for doing this.

PR #2439, opened 2020-09-16
Updated problem_06 in Project Euler

shellhub:
### **Describe your change:**

* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [x] Documentation change?

### **Checklist:**
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [ ] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [ ] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [ ] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #2463, opened 2020-09-23
Config Travis CI for two jobs

dhruvmanila:
As discussed in #2439 

Dividing Travis CI test into two jobs:
- First job will be testing everything except for the `project_euler` directory
- Second job will only run `pytest` on `project_euler` directory

**NOTE:** Before merging please tell me so that I can remove the `branches` section from `.travis.yml` file as that was done only for testing for which the results are here: https://travis-ci.org/github/dhruvmanila/Python/builds/729560689

**Extra:**
There's another error which I also wrote about in #2439 comment and I will paste it here:

The way `mypy` finds the python files are a bit tricky. In the Travis test, `mypy` isn't going inside the project_euler directory because there's no `__init__.py` file in it. We need to have the `__init__.py` file in every directory and subdirectory to tell `mypy` to look for python files in this directory as well which means `mypy` is skipping tons of files. It's checking only for **116** files while there are around **668** files with extension py with around **123** directories according to:

```console
$ fd -t f -e py . | wc -l        # find files with extension py and count the lines
       668

$ fd -t d . | wc -l                 # find directories and count the lines
       123
```

I have written a bash script which will add all the init files in the directories where it's missing.
Do you want me to make another PR for it or just add it in here?

### **Describe your change:**

### **Checklist:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ os: linux
 dist: focal
 language: python
 python: 3.8
-branches: # This section will be removed in the last commit
-  only:
-    - config-travis
 cache: pip
 before_install: pip install --upgrade pip setuptools six
 install: pip install black flake8 mypy pytest-cov
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ branches:
     - config-travis
 cache: pip
 before_install: pip install --upgrade pip setuptools six
-install: pip install black flake8
+install: pip install black flake8 mypy pytest-cov
 notifications:
     webhooks: https://www.travisbuddy.com/
     on_success: never
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: python
 python: 3.8
 cache: pip
 before_install: pip install --upgrade pip setuptools six
-install: pip install black flake8 mypy pytest-cov
+install: pip install black flake8
 notifications:
     webhooks: https://www.travisbuddy.com/
     on_success: never
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ jobs:
       script:
         - mypy --ignore-missing-imports .
         - pytest --doctest-modules --ignore=project_euler/ --durations=10 --cov-report=term-missing:skip-covered --cov=. .
-    - name: project-euler
+    - name: Project Euler
+      before_script: pip install pytest-cov
       script:
         - pytest --doctest-modules --durations=10 --cov-report=term-missing:skip-covered --cov=project_euler/ project_euler/
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,35 @@ os: linux
 dist: focal
 language: python
 python: 3.8
+branches:
+  only:
+    - config-travis
 cache: pip
 before_install: pip install --upgrade pip setuptools six
 install: pip install black flake8
 notifications:
     webhooks: https://www.travisbuddy.com/
     on_success: never
-before_script:
-  - black --check . || true
-  - flake8 --ignore=E203,W503 --max-complexity=25 --max-line-length=88 --statistics --count .
-  - scripts/validate_filenames.py  # no uppercase, no spaces, in a directory
-  - pip install -r requirements.txt  # fast fail on black, flake8, validate_filenames
-script:
-  - mypy --ignore-missing-imports .
-  - pytest --doctest-modules --durations=10 --cov-report=term-missing:skip-covered --cov=. .
+
+jobs:
+  include:
+    - name: build
+      before_script:
+        - black --check . || true 
+        - flake8 --ignore=E203,W503 --max-complexity=25 --max-line-length=88 --statistics --count .
+        - scripts/validate_filenames.py  # no uppercase, no spaces, in a directory
+        - pip install -r requirements.txt  # fast fail on black, flake8, validate_filenames
+      script:
+        - mypy --ignore-missing-imports .
+        - pytest --doctest-modules --ignore=project_euler/ --durations=10 --cov-report=term-missing:skip-covered --cov=. .
+    - name: project-euler
+      before_script:
+        - black --check project_euler/ || true
+        - flake8 --ignore=E203,W503 --max-complexity=25 --max-line-length=88 --statistics --count project_euler/
+        - scripts/validate_filenames.py
+      script:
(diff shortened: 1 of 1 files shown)


dhruvmanila:
Ignore the NOTE. I have removed the branches section.

One more thing about the `mypy` thing:
Out of **123** directories only **62** contains the `__init__.py` file and most of them are the `project_euler` directories.

.travis.yml
```diff
@@ -4,17 +4,25 @@ language: python
 python: 3.8
 cache: pip
 before_install: pip install --upgrade pip setuptools six
-install: pip install black flake8
+install: pip install black flake8 mypy pytest-cov
```
cclauss, line 7:
```suggestion
install: pip install black flake8
```
Let's preserve the fast fail.

.travis.yml
```diff
@@ -4,17 +4,25 @@ language: python
...
+    - name: build
+      before_script:
+        - black --check . || true 
+        - flake8 --ignore=E203,W503 --max-complexity=25 --max-line-length=88 --statistics --count .
+        - scripts/validate_filenames.py  # no uppercase, no spaces, in a directory
+        - pip install -r requirements.txt  # fast fail on black, flake8, validate_filenames
+      script:
+        - mypy --ignore-missing-imports .
+        - pytest --doctest-modules --ignore=project_euler/ --durations=10 --cov-report=term-missing:skip-covered --cov=. .
+    - name: project-euler
```
cclauss, line 23:
```suggestion
    - name: Project Euler
      before_script: pip install pytest-cov
```

.travis.yml
```diff
@@ -4,17 +4,25 @@ language: python
...
-  - flake8 --ignore=E203,W503 --max-complexity=25 --max-line-length=88 --statistics --count .
-  - scripts/validate_filenames.py  # no uppercase, no spaces, in a directory
-  - pip install -r requirements.txt  # fast fail on black, flake8, validate_filenames
-script:
-  - mypy --ignore-missing-imports .
-  - pytest --doctest-modules --durations=10 --cov-report=term-missing:skip-covered --cov=. .
+
+jobs:
+  include:
+    - name: build
```
cclauss, line 14:
```suggestion
    - name: Build
```

cclauss (changes requested):
Awesome!

.travis.yml
dhruvmanila: This will fail the `Project Euler` build always as there is no `pytest-cov` in it, we're installing it only on the `Build` job.
Another solution would be to include the `install` part in both jobs or install it in `before_script` part. Can we do the latter thing?

cclauss:
> Do you want me to make another PR for [adding `__init__.py` files] or just add it in here?

That is a completely separate topic so please make that a separate PR.

`Python` is a brand name so please capitalize except when referring to the command itself: `python3 -c print("Hello")`

cclauss:
Can you please open an issue on https://github.com/python/mypy to verify the requirement for `__init__.py` files to get mypy to work.  It seems to run against the current of not requiring those files to make packages and modules in modern version of Python.  It would be a bummer to create a ton of files that we do not need.

cclauss (approved):

dhruvmanila:
Sorry for not putting Python in capitalize. I will keep that in mind. Thanks for pointing it out.

There's already an open issue about it: https://github.com/python/mypy/issues/6385
Here's Guido van Rossum explaining what is going on: https://github.com/python/mypy/issues/6385#issuecomment-542033018

It seems not to be fixed and the only option right now is to add `__init__.py` files if we want `mypy` to check all the scripts.

cclauss:
I had to [put mypy into || true mode](https://github.com/TheAlgorithms/Python/blob/master/.travis.yml#L17) because of python/mypy#7907 so let's hold off on adding all the `__init__.py` files for a while.

dhruvmanila:
Ok. Sorry for the multiple deletions, I wanted to clean the history properly on my local copy.

dhruvmanila:
Do you want me to fix the `mypy` errors without adding the `__init__.py` file? 
I will add the `__init__.py` files and check for the errors and create PR only for the change through a different branch.
If so, should I create different PR for each file or group a few files together because the errors are tiny?

cclauss:
Let's hold off on mypy-related changes for now (unless you just want to fix the Project Euler ones).

Instead, let's focus on pytesting the return values of the Euler algorithms.

dhruvmanila:
Alright. So a few ideas right out of head are:

- Make a Python script `test_project_euler.py` in `project_euler` directory which will collect all the files and execute them comparing the answers to the actual ones (probably a bad idea).
- Make a test file in each folder testing for every solution files which seems more probable, but then what about new problems getting added? We could have a naming convention for the final function which will return the answer and compare that to the actual ones. I think the naming convention already exists for the files, but not mentioned in the `README.md` file.

cclauss:
Let's try to build #1 (single file) but cover just subdirectories problem_01 thru problem_09 with a single test file.  We can see how difficult it is and optimize the code.  If we create a test file per directory, it will be a ton of work to keep them in sync.

dhruvmanila:
Alright, I took a look at the files and found some inconsistencies:
(Should we do this conversation in a separate PR?)

- Some folder names contain trailing zeroes and some don't like from `problem_01` to `problem_09` and then there are none. So we need to have a naming convention which either says to have trailing zeroes in all the new directories or it should have none. I would prefer the latter (No trailing zeroes: `problem_1, problem_2, ...problem_99, problem_100...`
- Similar all solutions files should be like `sol1.py, sol2.py, ...` Everyone does follow this convention but I did see one file which is `problem_27_sol1.py`. No trailing zeroes in here as well: `sol1.py, sol2.py, ...sol9.py, sol10.py...`
- Next is that all solution files should contain a `solution` function which takes no input or defaults to the question's input (which is preferred as that allows `doctest` for different inputs) and returns the answer. Right now, some functions take input and some don't, the ones who take doesn't default to the question's input and so we would have to manually enter it.

What do you think about the above-mentioned points? If we implement them, testing will be easier from one file otherwise we will have to manually look at each file to see how it returns the answer and check accordingly.

cclauss:
Please open a new issue for this effort.

Let’s ___not___ change the directory names...
```python
for i in range(101):
    print(f"problem_{i:02}")
```

dhruvmanila:
Alright, I'll get on it.

PR #2470, opened 2020-09-24
Fixed remove duplicate

shellhub:
### **Describe your change:**

* [ ] Add an algorithm?
* [x] Documentation change?

### **Checklist:**

### DIFF
--- a/strings/remove_duplicate.py
+++ b/strings/remove_duplicate.py
@@ -1,14 +1,15 @@
-""" Created by sarathkaul on 14/11/19 """
-
-
 def remove_duplicates(sentence: str) -> str:
     """
-    Reomove duplicates from sentence
+    Remove duplicates from sentence
     >>> remove_duplicates("Python is great and Java is also great")
     'Java Python also and great is'
+    >>> remove_duplicates("Python   is      great and Java is also great")
+    'Java Python also and great is'
     """
-    return " ".join(sorted(set(sentence.split(" "))))
+    return " ".join(sorted(set(sentence.split())))
 
 
 if __name__ == "__main__":
-    print(remove_duplicates("INPUT_SENTENCE"))
+    import doctest
+
+    doctest.testmod()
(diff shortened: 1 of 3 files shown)


mateuszz0000 (approved):

strings/remove_duplicate.py
```diff
@@ -1,14 +1,15 @@
...
     'Java Python also and great is'
+    >>> remove_duplicates("Python   is      great and Java is also great")
+    'Java Python also and great is'
     """
-    return " ".join(sorted(set(sentence.split(" "))))
+    return " ".join(sorted(set(sentence.split())))
 
 
 if __name__ == "__main__":
-    print(remove_duplicates("INPUT_SENTENCE"))
```
cclauss, line 14: Why remove this line?  Without it, the user runs the program and nothing happens.

strings/remove_duplicate.py
```diff
@@ -1,14 +1,15 @@
-""" Created by sarathkaul on 14/11/19 """
-
-
 def remove_duplicates(sentence: str) -> str:
```
cclauss, line 1: We should rename the function to `remove_duplicate_whitespace()`

strings/remove_duplicate.py
shellhub: I think algorithm should not print anything. Because it have tested it using doctest.

strings/remove_duplicate.py
shellhub: Should I open another PR ?

strings/remove_duplicate.py
cclauss: The algorithmic function does not print() in alignment with CONTRIBUTING.md.  The algorithmic function instead returns a string.  However, it is more instructive (and intuitive) for the reader if main() does print the result of the algorithm.  We are writing for humans to learn, not fmerely or computers to test.

PR #2475, opened 2020-09-24
Heaps algorithm

grochedix:
### **Describe your change:**

Added the Heap's algorithm under the "divide and conquer" folder. It outputs all possible permutations from a list of elements.

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


divide_and_conquer/heaps_algorithm.py
```diff
@@ -0,0 +1,50 @@
...
+    Pure python implementation of the Heap's algorithm,
+    returning all permutations of a list.
+    >>> heaps([])
+    [[]]
+    >>> heaps([0])
+    [[0]]
+    >>> heaps([-1, 1])
+    [[-1, 1], [1, -1]]
+    >>> heaps([1, 2, 3])
+    [[1, 2, 3], [2, 1, 3], [3, 1, 2], [1, 3, 2], [2, 3, 1], [3, 2, 1]]
```
cclauss, line 21: Instead of manual doctest results, please test this against 
https://docs.python.org/3/library/itertools.html#itertools.permutations to ensure they both deliver identical results.

BriseBalloches: Done. Added one comparison between my algo and itertools permutations. Also outputed my permutations as list of tuples instead of list of lists to match itertools.

cclauss: Why the use of `set()`? Perhaps you should be comparing against
https://docs.python.org/3/library/itertools.html#itertools.combinations or
https://docs.python.org/3/library/itertools.html#itertools.combinations_with_replacement

BriseBalloches: I use `set()` since i have to compare the result of a generator (itertools) vs mine (list of tuple), as well as the order of tuples might also be different from one version to the other but can still be correct. Maybe you see another "cleaner" way of doing this ?
I don't think comparing with combinations is a good idea, the Heap's algorithm outputs permutations (order of elements matters).

cclauss: `tuple(permutations(data))` and if that does not work `sorted(permutations(data))`

BriseBalloches: Used `sorted()` but had to used it on both `permutations()` and `heaps()`.

grochedix: Updated with your comment relative to the iterative version.

cclauss (approved):
LGTM

PR #2486, opened 2020-09-26
Initial commit of conjugate gradient method

zakademic:
### **Describe your change:**

This is an implementation of the conjugate gradient method for solving a linear system given an SPD matrix A. 

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/linear_algebra/src/conjugate_gradient.py
+++ b/linear_algebra/src/conjugate_gradient.py
@@ -1,11 +1,12 @@
+"""
+Resources:
+- https://en.wikipedia.org/wiki/Conjugate_gradient_method
+- https://en.wikipedia.org/wiki/Definite_symmetric_matrix
+"""
 import numpy as np
-
-# https://en.wikipedia.org/wiki/Conjugate_gradient_method
-# https://en.wikipedia.org/wiki/Definite_symmetric_matrix
 
 
 def _is_matrix_spd(matrix: np.array) -> bool:
-
     """
     Returns True if input matrix is symmetric positive definite.
     Returns False otherwise.
@@ -56,24 +57,24 @@ def _create_spd_matrix(dimension: np.int64) -> np.array:
     >>> _is_matrix_spd(spd_matrix)
     True
     """
-
     random_matrix = np.random.randn(dimension, dimension)
     spd_matrix = np.dot(random_matrix, random_matrix.T)
-
     assert _is_matrix_spd(spd_matrix)
-
     return spd_matrix
 
 
 def conjugate_gradient(
-    spd_matrix: np.array, b: np.array, max_iterations=1000, tol=1e-8
+    spd_matrix: np.array,
+    load_vector: np.array,
+    max_iterations: int = 1000,
+    tol: float = 1e-8,
 ) -> np.array:
     """
     Returns solution to the linear system np.dot(spd_matrix, x) = b.
 
     Input:
     spd_matrix is an NxN Symmetric Positive Definite (SPD) matrix.
-    b is an Nx1 vector that is the load vector.
+    load_vector is an Nx1 vector.
 
     Output:
     x is an Nx1 vector that is the solution vector.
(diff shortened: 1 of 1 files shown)


linear_algebra/src/conjugate_gradient.py
```diff
@@ -0,0 +1,172 @@
...
+    assert np.shape(A)[0] == np.shape(A)[1]
+
+    # Get eigenvalues and eignevectors for a symmetric matrix.
+    eigen_values, _ = np.linalg.eigh(A)
+
+    # Check sign of all eigenvalues.
+    if np.all(eigen_values > 0):
+        return True
+    else:
+        return False
```
cclauss, line 36:
```suggestion
    return np.all(eigen_values > 0)
```

linear_algebra/src/conjugate_gradient.py
```diff
@@ -0,0 +1,172 @@
...
+    >>> N = 3
+    >>> A = _create_spd_matrix(N)
+    >>> _is_matrix_spd(A)
+    True
+    """
+
+    A = np.random.randn(N, N)
+    A = np.dot(A, A.T)
+
+    assert _is_matrix_spd(A) is True
```
cclauss, line 59:
```suggestion
    assert _is_matrix_spd(A)
```

linear_algebra/src/conjugate_gradient.py
```diff
@@ -0,0 +1,172 @@
...
+    else:
+        return False
+
+
+def _create_spd_matrix(N: np.int64) -> np.array:
+    """
+    Returns a symmetric positive definite matrix given a dimension.
+
+    Input:
+    N is an integer.
```
cclauss, line 44: This line is redundant now that we have type hints so let's instead give `N` a more self-documenting variable name so that this comment is no longer required or make this comment document the what & why, not the datatype.

linear_algebra/src/conjugate_gradient.py
```diff
@@ -0,0 +1,172 @@
+import numpy as np
```
cclauss, line 1: Please add a URL that can help the reader learn about the problems that your algorithm is trying to solve...
https://en.wikipedia.org/wiki/Conjugate_gradient_method
https://en.wikipedia.org/wiki/Definite_symmetric_matrix

cclauss (changes requested):
This looks really great.  I am not a numpy guy so I can not comment on the usage of numpy.  In Python, UPPERCASE variable names are reserved for constants that never change during the runtime of the script.  Function parameters should never be uppercase.  Also, please use more self-documenting variable names.  Single-letter variable names look so old school.  Instead of `A`, use `matrix` or `spd_matrix` or some other name that is even more self-documenting of the data that the variable contains.  It is OK to have a comment that links the Python variable name back to the formula name: _spd_matrix is often called A in discussions of the conjugate gradient algorithm_.

linear_algebra/src/conjugate_gradient.py
zakademic: Added :)

linear_algebra/src/conjugate_gradient.py
zakademic: Yes good point. I changed this.

zakademic:
> This looks really great. I am not a numpy guy so I can not comment on the usage of numpy. In Python, UPPERCASE variable names are reserved for constants that never change during the runtime of the script. Function parameters should never be uppercase. Also, please use more self-documenting variable names. Single-letter variable names look so old school. Instead of `A`, use `matrix` or `spd_matrix` or some other name that is even more self-documenting of the data that the variable contains. It is OK to have a comment that links the Python variable name back to the formula name: _spd_matrix is often called A in discussions of the conjugate gradient algorithm_.

Totally agree. I changed many namings to be more meaningful. Thanks!

53jk1 (approved):

linear_algebra/src/conjugate_gradient.py
```diff
@@ -0,0 +1,176 @@
...
+        return False
+
+    # Get eigenvalues and eignevectors for a symmetric matrix.
+    eigen_values, _ = np.linalg.eigh(matrix)
+
+    # Check sign of all eigenvalues.
+    return np.all(eigen_values > 0)
+
+
+def _create_spd_matrix(N: np.int64) -> np.array:
```
cclauss, line 43: N?  A more self-documenting name please with no uppercase in variable names.

linear_algebra/src/conjugate_gradient.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    # Ensure proper dimensionality.
+    assert np.shape(spd_matrix)[0] == np.shape(spd_matrix)[1]
+    assert np.shape(b)[0] == np.shape(spd_matrix)[0]
+    assert _is_matrix_spd(spd_matrix)
+
+    N = np.shape(b)[0]
+
+    # Initialize solution guess, residual, search direction.
+    x0 = np.zeros((N, 1))
```
cclauss, line 103: When you create a variable and then use that variable only once on the very next line then try to remove that variable unless 1. The second line is _very_ long or 2. the variable name documents something that is unclear.

`N` in this context does neither.

```suggestion
    # Initialize solution guess, residual, search direction.
    x0 = np.zeros((np.shape(b)[0], 1))
```

linear_algebra/src/conjugate_gradient.py
zakademic: Changed this :)

xcodz-dot (approved):
@dhruvmanila, check this

cclauss (approved):

dhruvmanila:
@algorithms-keeper review

dhruvmanila (approved):

Issue #2502, opened 2020-09-28
Is this repo using mypy correctly?


cclauss:
@dhruvmanila Is [mypy](http://mypy-lang.org) running properly [in our Travis CI runs](https://travis-ci.com/github/TheAlgorithms/Python/jobs/391696041#L398)?  I believe that I made a mistake in #2464 which mypy is complaining about but it only seems to be raising issues in the `maths` subdirectory.  Are we doing something wrong here?

dhruvmanila:
Let me take a look at it and I'll get back to you.

dhruvmanila:
I see the problem. 
As mentioned in https://github.com/TheAlgorithms/Python/pull/2463#issue-491639954, `mypy` ignores all the directories which don't contain the `__init__.py` file, and as the `maths` subdirectory contains the `__init__.py` file, `mypy` is only complaining about the errors in that specific directory. I will put the highlights here:

>There's already an open issue about it (Why do we need `__init__.py` files?): python/mypy#6385
>Here's Guido van Rossum explaining what is going on: python/mypy#6385 (comment)

>Out of **123** directories only **62** contains the `__init__.py` file and most of them are the `project_euler/..` directories.

> It's checking only for **116** files while there are around **668** files with extension `py` with around **123** directories according to:

```console
$ fd -t f -e py . | wc -l        # find files with extension py and count the lines
       668

$ fd -t d . | wc -l                 # find directories and count the lines
       123
```

By running a simple `bash` script, I found that `__init__.py` exists only in these directories, but `mypy` will stop recursing into the subdirectories if it doesn't find `__init__.py` file in the parent directory. What I mean is it looks for Python files in `./Python/maths/` but not in `./Python/data_structures/stacks/` as there is no `init` file in `data_structures` directory.
```
__init__.py exists in ./Python/data_structures/hashing/number_theory/
__init__.py exists in ./Python/data_structures/linked_list/
__init__.py exists in ./Python/data_structures/stacks/
__init__.py exists in ./Python/digital_image_processing/
__init__.py exists in ./Python/digital_image_processing/dithering/
__init__.py exists in ./Python/digital_image_processing/edge_detection/
__init__.py exists in ./Python/digital_image_processing/filters/
__init__.py exists in ./Python/digital_image_processing/resize/
__init__.py exists in ./Python/digital_image_processing/rotation/
__init__.py exists in ./Python/maths/
__init__.py exists in ./Python/project_euler/problem_01/

...

[...]

PR #2503, opened 2020-09-28
Add __init__.py files in all the directories

dhruvmanila:
Fixes: #2502 

**Please check if there's any directory where it's not needed.**

### **Describe your change:**

### **Checklist:**


dhruvmanila:
That was a disaster! Let me see what's going on in there.

dhruvmanila:
@cclauss Ok, I found the reason for all the errors. 
The errors are happening because of how namespace packages work. As we're adding all the `__init__.py` files, we're basically creating sub-packages in our parent directory (`Python/`). What this means is that if we want to use `import` in a file inside a sub-directory, then we need to use absolute import from our parent directory. 

For example:
In `ciphers/affine_cipher.py`, instead of `import cryptomath_module as cryptomath` we will have to do `import ciphers.cryptomath_module as cryptomath`

This doesn't seem like a viable solution so I will take a further look into this.

dhruvmanila:
I don't think there's any other choice if we're adding the `__init__.py` files.
It also seems that in some of the files we're using relative imports and in others we're using absolute. It would be better, in the long run, to stick with one. For more information on relative imports: https://docs.python.org/3/reference/import.html#package-relative-imports

If we're importing a module from the current directory we can use either of the following:
- `from parent.module import function` 
- `import .module`. 

Remember that a `module` is just another Python file.

dhruvmanila:
I fixed all the imports on my local copy and it fixes everything up.
Shall I make a PR to fix the imports? @cclauss

cclauss (approved):

cclauss:
Yes. please.  Just the changes to get Travis CI to be green again.

AayushDadhich:
Typo

in data structures directory folder is named as trie instead of tree

cclauss:
@AayushDadhich https://en.wikipedia.org/wiki/Trie

annsleyshakeah66:
https://amzn.to/3wG6XcY
https://amzn.to/3eLecL2
https://amzn.to/38c7OZ2
https://amzn.to/3dG5sVl
https://amzn.to/3BcOacw
https://amzn.to/3jJ0Lhp
https://amzn.to/2RgUPQY
https://amzn.to/2TzhM31
https://amzn.to/3jQ2guj
https://amzn.to/3yr1nfB
https://amzn.to/3jkJn1V
https://amzn.to/3ks6V4b

kmkurumi:
annsleyshakeah66 ***@***.***>於 2021年9月7日 週二，下午6:58寫道：

PR #2507, opened 2020-09-29
Update sorts/quick_sort_3_partition.py

taseikyo:
Another quick sort algorithm, returns a new sorted list

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/sorts/quick_sort_3_partition.py
+++ b/sorts/quick_sort_3_partition.py
@@ -1,4 +1,4 @@
-def quick_sort_3partition(sorting: list, left: int, right: int):
+def quick_sort_3partition(sorting: list, left: int, right: int) -> None:
     if right <= left:
         return
     a = i = left
--- a/sorts/quick_sort_3_partition.py
+++ b/sorts/quick_sort_3_partition.py
@@ -1,4 +1,4 @@
-def quick_sort_3partition(sorting: list, left: int, right: int) -> list:
+def quick_sort_3partition(sorting: list, left: int, right: int):
     if right <= left:
         return
     a = i = left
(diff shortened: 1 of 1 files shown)


taseikyo:
> I am not a fan of this change. What about `three_way_quicksort()` ?
> 
> Are we trying to do https://en.wikipedia.org/wiki/Quicksort#Three-way_radix_quicksort or https://en.wikipedia.org/wiki/Multi-key_quicksort ?
> Or is there some other webpage that describes the algorithm that we are implementing?

Yes, we are implementing https://en.wikipedia.org/wiki/Quicksort#Three-way_radix_quicksort. First divide the list into three parts. Then recursively sort the "less than" and "greater than" partitions.

sorts/quick_sort_3_partition.py
```diff
@@ -1,4 +1,4 @@
-def quick_sort_3partition(sorting: list, left: int, right: int) -> list:
+def quick_sort_3partition(sorting: list, left: int, right: int):
```
cclauss, line 1: Why this change?

taseikyo: `quick_sort_3partition` does not return a new list, it just updates in place. So I remove `-> list`.

sorts/quick_sort_3_partition.py
```diff
@@ -1,4 +1,4 @@
-def quick_sort_3partition(sorting, left, right):
+def quick_sort_3partition(sorting: list, left: int, right: int):
```
cclauss, line 1:
```suggestion
def quick_sort_3partition(sorting: list, left: int, right: int) -> None:
```

cclauss (approved):

PR #2509, opened 2020-09-29
Create instagram_crawler.py

yogeshwaran01:
### **Describe your change:**
Crawl the Instagram users information

* [X] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**
* [X] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [X] This pull request is all my own work -- I have not plagiarized.
* [X] I know that pull requests will not be merged if they fail the automated tests.
* [X] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [X] All new Python files are placed inside an existing directory.
* [X] All filenames are in all lowercase characters with no spaces or dashes.
* [X] All functions and variable names follow Python naming conventions.
* [X] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [X] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [X] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [X] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #2511, opened 2020-09-30
Add pre-commit hook for TheAlgorithms/Python

dhruvmanila:
As mentioned in https://github.com/TheAlgorithms/Python/pull/2509#issuecomment-700778237

### **Describe your change:**

* [x] Add an algorithm?

### **Checklist:**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: trailing-whitespace
       - id: requirements-txt-fixer
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: stable
     hooks:
       - id: black
   - repo: https://github.com/PyCQA/isort
@@ -42,8 +42,8 @@ repos:
           - --quiet-level=2
         exclude: |
           (?x)^(
-              other/dictionary.txt|
-              other/words|
+              other/dictionary.txt |
+              other/words |
               project_euler/problem_22/p022_names.txt
           )$
   - repo: local
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
     hooks:
       - id: isort
         args:
-          - --profile black
+          - --profile=black
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.3
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,12 @@ repos:
       - id: check-executables-have-shebangs
       - id: check-yaml
       - id: end-of-file-fixer
+        types: [python]
       - id: trailing-whitespace
+        exclude: |
+          (?x)^(
+            data_structures/heap/binomial_heap.py
+          )$
       - id: requirements-txt-fixer
   - repo: https://github.com/psf/black
     rev: stable
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,55 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: check-executables-have-shebangs
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: requirements-txt-fixer
+  - repo: https://github.com/psf/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.5.3
+    hooks:
+      - id: isort
+        args:
+          - --profile black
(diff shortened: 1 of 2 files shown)


dhruvmanila:
There are a bunch of files that will be fixed if I run this.
Should I open another PR for those fixes?

dhruvmanila:
We also need to inform people to enable it in their local copy by:
```console
$ pip install pre-commit
$ pre-commit install
```

Also, there are badges for [pre-commit](https://pre-commit.com/#badging-your-repository) and [black](https://img.shields.io/badge/code%20style-black-000000.svg).
Shall I add them?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,55 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: check-executables-have-shebangs
```
cclauss, line 5: Do we have something in CONTRIBUTING.md that describes the proper shebang
(`#!/usr/bin/env python3`)?

cclauss (changes requested):
This is awesome!  How long does it take to run on your machine?  (If it is not fast, contributors will turn it off.)  Please add the badges to README.md in a separate follow-up PR.  Please fix the issues in a separate PR that we should land before this one.

.pre-commit-config.yaml
dhruvmanila: No. I thought as executable files are made by us for mostly testing purposes, to include this in case we forget to add it.
Should I remove it?

dhruvmanila:
By default, `pre-commit` runs only on the new files or modified files. We can change it to run on all the files all the time but that would be inefficient.

.pre-commit-config.yaml
cclauss: How many files will this flag?

cclauss:
OK.  Yes now I understand.  After this lands, we can add something like this to do all files.  https://github.com/cclauss/itinerant-tester/blob/master/.github/workflows/pre-commit.yml

.pre-commit-config.yaml
dhruvmanila: 3 files:
- machine_learning/scoring_functions.py
- maths/sum_of_arithmetic_series.py
- scheduling/round_robin.py

cclauss: Let’s leave it in.

dhruvmanila:
I don't think so we need a workflow file for this as all this is done locally. The workflow file is run after the commit has been pushed and mainly used to check if every test passed. It's mostly just for the visuals as far as I can understand.

dhruvmanila:
TODO:
[ ] Separate PR for all the fixes BEFORE this is merged
[ ] Follow-up PR for adding badges to README.md

Is that correct?

cclauss:
A fraction of all contributors will install and run pre-commit.  The workflow would check the work of the contributors wo do not install the tool and it will allow us to remove all that complexity from .travis.yml which should speed up our test runs.

dhruvmanila:
Got it!

cclauss (approved):

cclauss:
Do you have rights to click the `Squash and merge` button above?

dhruvmanila:
Missed it! But I don't think so as that will be available only to members and owners.

cclauss:
OK.  We need the fix PR before we merge.

Also I added to the GitHub Action to your task list above.

.pre-commit-config.yaml
dhruvmanila: So, shall I include the shebang in those files or remove the executable bit from them?

dhruvmanila:
@cclauss DON'T MERGE THIS YET.
We need to ignore some files which require whitespace and newlines.

dhruvmanila:
@cclauss Please take a look at the changes:
EOF newline will be checked only for Python files and ignoring whitespace for that specific file as it requires it in `pytest`

dhruvmanila:
I can see the merge option on the app but not on the browser.

cclauss:
THANKS!!

DmytroLitvinov:
Hi @dhruvmanila ,

I guess we should warn peope about pre-commit at [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md) file.

cclauss:
Let's not force the complexities of a local install of pre-commit on all contributors.  Some folks are pretty junior and adding pre-commit to their machines might turn them off from contributing.  The README should however show them how to find and solve issues when the GitHub Action does not love their contribution.

dhruvmanila:
Maybe we should put it in `Travis_CI_tests_are_failing` and change the name of the file to `tests_are_failing` and put the instructions as to how to find the error for all the automatic tests we're performing in that file.

The other idea would be to separate out the files for each test but having a lot of files doesn't seem attractive.

PR #2515, opened 2020-09-30
Add GitHub action for pre-commit

dhruvmanila:
Follow-up on #2511 
As mentioned in https://github.com/TheAlgorithms/Python/pull/2511#issuecomment-701163537

### **Describe your change:**

* [x] Add an algorithm?

### **Checklist:**


dhruvmanila:
It's working perfectly: https://github.com/TheAlgorithms/Python/pull/2515/checks?check_run_id=1186742542

A few small changes required:
- Fix `instagram_crawler.py` using `black`
- Either add `shebang` or remove the executable bit from those files

dhruvmanila:
I think the name of the file should be ***lint***, what do you think? @cclauss

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+
```
cclauss, line 11:
```suggestion
```

cclauss: Not needed.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
```
cclauss, line 19:
```suggestion
      - uses: actions/setup-python@v2
```

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Install dependencies
```
cclauss, line 20:
```suggestion
      - name: Install pre-commit
```

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,26 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade pre-commit
+
+      - name: Run pre-commit
+        run: pre-commit run --verbose --all-files --show-diff-on-failure
```
cclauss, line 26:
```suggestion
      - run: pre-commit run --verbose --all-files --show-diff-on-failure
```

cclauss (changes requested):
Simplifying by always using the latest and greatest Python and by dropping names that are less self-documenting than the commands that they run.

cclauss:
Why not just add the shebang instead of removing the execution bit?

cclauss:
> I think the name of the file should be _**lint**_, what do you think? @cclauss

I think the current name is good enough.  The script runs pre-commit so it should be called pre-commit.

dhruvmanila:
Except for the `scheduling/round_robin.py` file, other files don't output anything when I run `python <filename>` so I thought there's no point in keeping the executable bit on. 

If you think otherwise, I will turn them on?

cclauss (approved):

PR #2523, opened 2020-09-30
Travis CI: Remove redundant tests

cclauss:
### **Describe your change:**

Many tests (flake8, black, valid filenames, etc.) have migrated to a pre-commit hook #2511 and GitHub Action to run it #2515.

We can now remove those tests from Travis CI so that our automated tests run even faster.  @dhruvmanila 

* [ ] Add an algorithm?
* [X] Testing change?

### **Checklist:**


poyea (approved):
Thanks

.travis.yml
```diff
@@ -3,27 +3,20 @@ dist: focal
...
       script:
-        - mypy --ignore-missing-imports . || true  # https://github.com/python/mypy/issues/7907
         - pytest --doctest-modules --ignore=project_euler/ --durations=10 --cov-report=term-missing:skip-covered --cov=. .
     - name: Project Euler
-      before_script:
-        - pip install pytest-cov pytest-subtests
-        - pytest --tb=no --no-summary --capture=no project_euler/validate_solutions.py || true  # fail fast on wrong solution
       script:
+        - pytest --tb=no --no-summary --capture=no project_euler/validate_solutions.py || true  # fail fast on wrong solution
         - pytest --doctest-modules --durations=10 --cov-report=term-missing:skip-covered --cov=project_euler/ project_euler/
```
dhruvmanila, line 17: By keeping the testing of Project Euler solutions in `script` section, Travis won't exit if it fails. If the test is in `before_script` section, Travis will exit.

.travis.yml
```diff
@@ -3,27 +3,20 @@ dist: focal
 language: python
 python: 3.8
 cache: pip
-before_install: pip install --upgrade pip setuptools six
-install: pip install black flake8
+install:
+  - pip install --upgrade pip setuptools six
+  - pip install -r requirements_pytest.txt
```
dhruvmanila, line 8: By convention, those test requirements file are called `requirements_dev.txt`

cclauss: That file was not needed after all.

.travis.yml
cclauss: Why do we run that `project_euler/validate_solutions.py` thru pytest?  It seems like a lot of complexity and overhead.

.travis.yml
dhruvmanila: It's better to keep a separate file for test requirements for people who actually use it, they can install it directly. We can include all our test requirements in it, but it's up to you. 😁

.travis.yml
dhruvmanila: I did do it without `pytest` but you mentioned (https://github.com/TheAlgorithms/Python/pull/2471#issuecomment-698533067) to include it in as that is already included.

Currently, the algorithm is set in such a way to utilize the `pytest.mark.parameterize` decorator which calls the function with the arguments as `problem_number` and `expected` and at the end uses `pytest.fixture` to print all the error messages. This can be done without using `pytest`.

dhruvmanila: It would be similar to this without `pytest`: https://github.com/TheAlgorithms/Python/blob/f1d60aca9e8c6cba013e5e5571d180f062bb3452/project_euler/validate_solutions.py

The script will be self-contained and will need no dependencies.

cclauss: It was run automatically by pytest when its name was `project_euler/solution_test.py` but that stopped being the case when the filename was changed to something that does not obey pytest discovery rules.  Let's not mess with it for now because we ___need___ to run in `|| true` mode.

.travis.yml
cclauss: I sense we have three sets of requirements:
1. requirements.txt for our code's requirements
2. requirements like black, flake8, isort that are embedded in the pre-commit yaml file
3. the Travis CI requirements pytest and its plugins

I am reluctant to install requirements that we do not need both on contributors machines and on CI machines.

dhruvmanila: - `pre-commit` doesn't require any installation, it is self-contained
- `test-requirements.txt` is only for the user who wants to install all the test tools used in this repository
- Travis CI won't install packages with `test-requirements.txt`, we will specify which ones to install

How does this sound?
This means we will have only two requirements file:
1. Code requirements
2. Test requirements

.travis.yml
cclauss: Are you OK with the current state of this PR or is there something that I should change?

dhruvmanila: I think `validate_solutions.py` should be in `before_script` section so that it fails fast when we remove `|| true`

.travis.yml
cclauss: I believe that we are both good with our current `requirements.txt`.

That leaves open the question, what should go in the other Test requirements file.  If I was running the tests on my machine, I would just type `pytest` because I do not need coverage and I do not need subtests (whatever that is).  So, do we create a separate Test requirements file that only contains `-r requirements.txt \n pytest`?

I would prefer a called requirements_dev.txt or requirements_test.txt because:
1. we have no files in the repo that have "`-`" in the filename
2. all the requirements files should appear next to each other in a code listing.

cclauss:
Both Travis runs execute in parallel in approx 3 min 30 sec.  Awesome work @dhruvmanila

.travis.yml
dhruvmanila: `flake8`, `black`, `isort`, `codespell` fall in the `requirements_dev.txt` file as those are the tools used to develop the code as well in terms of linting and formating.

We will tell Travis to install `requirements.txt` and `pytest` and its plugins.

`pytest-subtests` are used when you're having tests for multiple iterations in a given function. By default, one function equals one test for `pytest` but if the function contains loops that iterate over many values, those will be considered as different tests. `pytest` will stop if any of those loops fail and exit immediately saying that the function failed. We want to know from all the iterations which ones are going to fail. I don't know if I can explain properly on text so you would have to do some research 😅

dhruvmanila: This is also one of the benefits of using `pipenv` for managing your environments. It separates out your common packages and dev packages and you don't have to worry about anything but that's a discussion for another day.

dhruvmanila:
Thank you! 😁 🎉 👍

PR #2619, opened 2020-10-01
Create fibonacci_sequence_recursion_with_memoization.py

SamueldaCostaAraujoNunes:
### **My algorithm relates the application of the Fibonacci sequence (using Recursion) to the Memoization technique. Thus, it is possible to reach the complexity O (1), to calculate the Fibonacci value of very high numbers and in an incredibly shorter time interval than using only the common recursion.**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
+++ b/dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
@@ -1,5 +1,3 @@
-import doctest
-
 # Fibonacci Sequence Using Recursion and Memoization
 
 # Memoization Concept: https://en.wikipedia.org/wiki/Memoization
@@ -45,5 +43,7 @@ def main() -> None:
 
 
 if __name__ == "__main__":
+    import doctest
     doctest.testmod()
+    
     main()
--- a/dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
+++ b/dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
@@ -6,13 +6,10 @@
 
 class Fibonacci():
 
-    def __init__(self):
+    def __init__(self) -> None:
         self.cache: Dict[int, int] = dict()
 
     def __getitem__(self, key: int) -> int:
-        return self.__calculate_fibonacci(key)
-
-    def __calculate_fibonacci(self, n: int) -> int:
         """
         Return number of index n in sequence fibonacci
         >>> fibonacci = Fibonacci()
@@ -21,34 +18,37 @@ class Fibonacci():
         >>> fibonacci[12]
         144
         """
-
-        if type(n) != int:
+        if type(key) != int:
             raise "An integer value is expected"
 
-        elif n < 0:
+        elif key < 0:
             raise "The value must be greater than or equal to zero"
 
-        if n in self.cache:
-            return self.cache[n]
+        if key in self.cache:
+            return self.cache[key]
 
-        if n in {0, 1}:
-            ans = n
+        if key in {0, 1}:
+            ans = key
         else:
-            ans = self.__calculate_fibonacci(n - 1) + self.__calculate_fibonacci(n - 2)
-        self.cache[n] = ans
+            ans = self.__getitem__(key - 1) + self.__getitem__(key - 2)
+        self.cache[key] = ans
         return ans
 
 
-def main() -> None:
+def main(number: int) -> None:
+    """
+    >>> main(3)
+    The first 3 Fibonacci sequence terms are: [0, 1, 1]
+    The 3 index of the fibonacci sequence: 2
+    """
     fibonacci = Fibonacci()
-    number = int(input("Enter an integer greater than or equal to 0: "))
-    print(f"The first {number} Fibonacci sequence terms are:")
+    print(f"The first {number} Fibonacci sequence terms are:", end=" ")
     print([fibonacci[n] for n in range(number)])
-    print(f"The {number} index of the fibonacci sequence: ", end="")
+    print(f"The {number} index of the fibonacci sequence:", end=" ")
     print(fibonacci[number])
 
 if __name__ == "__main__":
     import doctest
 
     doctest.testmod()
-    main()
+    main(int(input("Enter an integer greater than or equal to 0: ")))
--- a/dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
+++ b/dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
@@ -3,47 +3,50 @@
 # Memoization Concept: https://en.wikipedia.org/wiki/Memoization
 # Fibonacci Concept : https://pt.wikipedia.org/wiki/Sequ%C3%AAncia_de_Fibonacci
 
+class Fibonacci():
+    def __init__(self):
+        self.cache: Dict[int, int] = dict()
 
-def fibonacci(n: int, cache: dict = dict()) -> int:
-    """
-    Return number of index n in sequence fibonacci
-    >>> [fibonacci(i) for i in range(12)]
-    [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89]
-    >>> fibonacci(12)
-    144
-    """
-    aux = n
+    def __getitem__(self, key: int) -> int:
+        return self.__calculate_fibonacci(key)
 
-    try:
-        if aux in cache:
-            return cache[aux]
+    
+    def __calculate_fibonacci(self, n: int) -> int:
+        """
+        Return number of index n in sequence fibonacci
+        >>> fibonacci = Fibonacci()
+        >>> [fibonacci[i] for i in range(12)]
+        [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89]
+        >>> fibonacci[12]
+        144
+        """
+
+        if type(n) != int:
+            raise "An integer value is expected"
+
+        elif(n < 0):
+            raise "The value must be greater than or equal to zero"
+
+        if n in self.cache:
+            return self.cache[n]
 
         if n in {0, 1}:
             ans = n
         else:
-            ans = fibonacci(n - 1, cache) + fibonacci(n - 2, cache)
-        cache[aux] = ans
-
+            ans = self.__calculate_fibonacci(n - 1) + self.__calculate_fibonacci(n - 2)
+        self.cache[n] = ans
         return ans
-    except TypeError:
-        print("Error: An integer value is expected")
-    except ValueError:
-        print("Error: The value must be greater than or equal to zero")
 
 
 def main() -> None:
-    try:
-        number = int(input("Enter an integer greater than or equal to 0: "))
-        print(f"The first {number} Fibonacci sequence terms are:")
-        print([fibonacci(n) for n in range(number)])
-        print(f"The {number} index of the fibonacci sequence: ", end="")
-        print(fibonacci(number))
-    except ValueError:
-        print("Error: Input value should be an integer greater than or equal to 0")
-
+    fibonacci = Fibonacci()
+    number = int(input("Enter an integer greater than or equal to 0: "))
+    print(f"The first {number} Fibonacci sequence terms are:")
+    print([fibonacci[n] for n in range(number)])
+    print(f"The {number} index of the fibonacci sequence: ", end="")
+    print(fibonacci[number])
 
 if __name__ == "__main__":
     import doctest
-
     doctest.testmod()
     main()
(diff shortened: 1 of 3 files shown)


maths/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,36 @@
+# Fibonacci Sequence Using Recursion and Memoization
+
+# Memoization Concept: https://en.wikipedia.org/wiki/Memoization
+# Fibonacci Concept : https://pt.wikipedia.org/wiki/Sequ%C3%AAncia_de_Fibonacci
+
+cache = dict()
+
+
+def fibonacci(n):
```
dhruvmanila, line 9:
```suggestion
def fibonacci(n: int) -> int:
```

maths/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Return number of index n in sequence fibonacci
+    >>> [fibonacci(i) for i in range(12)]
+    [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89]
+    """
```
dhruvmanila, line 14: Please have multiple doctest.

maths/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,36 @@
...
+    aux = n
+
+    if aux in cache:
+        return cache[aux]
```
dhruvmanila, line 18: Please check whether the input is of type `int`, if not then raise `TypeError` and for negative numbers raise `ValueError`.

maths/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,36 @@
...
+
+    if (n == 0) or (n == 1):
```
dhruvmanila, line 20:
```suggestion
    if n in {0, 1}:
```

`set` are hashed, so fast lookup.
```console
$ python -m timeit -s "n = 1" "(n == 0) or (n == 1)"
10000000 loops, best of 5: 32.5 nsec per loop

$ python -m timeit -s "n = 1" "n in {0, 1}"
10000000 loops, best of 5: 21 nsec per loop

$ python -m timeit -s "n = 1" "n in (0, 1)"
10000000 loops, best of 5: 27.4 nsec per loop
```

maths/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,36 @@
...
+        ans = n
+    else:
+        ans = fibonacci(n - 1) + fibonacci(n - 2)
+    cache[aux] = ans
+
+    return ans
+
+
+def main() -> None:
+    number = int(input("Enter an integer greater than or equal to 0: "))
```
dhruvmanila, line 30:
```suggestion
    try:
        number = int(input("Enter an integer greater than or equal to 0: "))
    except ValueError:
        print("Error: Input value should be an integer greater than or equal to 0")
```

NumberPiOso: This try does not assert that the value is greater than or equal to 0, so the message of error is misleading.

NumberPiOso:
A simpler approach is to use the decorator
`from functools import lru_cache`

This uses the **memoization technique** withouth the overhead of having to implement the dictionary and looking for values.

There is a _example_ of this for Fibonacci numbers in the Python documentation.
https://docs.python.org/3/library/functools.html

SamueldaCostaAraujoNunes:
> A simpler approach is to use the decorator
> `from functools import lru_cache`
> 
> This uses the **memoization technique** withouth the overhead of having to implement the dictionary and looking for values.
> 
> There is a _example_ of this for Fibonacci numbers in the Python documentation.
> https://docs.python.org/3/library/functools.html
> 
> ```
> @lru_cache(maxsize=None)
> def fib(n):
>     if n < 2:
>         return n
>     return fib(n-1) + fib(n-2)
> ```

I really agree that it would be much simpler to implement, but my intention with this code was to be more didactic, applying memoization directly within the Fibonacci sequence! I liked to know that there is a native solution within Python itself, but I chose to keep the logic of my algorithm in this way, thank you!

poyea:
Thanks. Did you take a look at the existing implementation in `dynamic_programming`? What's the major difference between that two versions and yours?

SamueldaCostaAraujoNunes:
> Thanks. Did you take a look at the existing implementation in `dynamic programming`? What's the major difference between that two versions and yours?

My program is in the application of memoization in the fibonacci sequence using a single recursive function. The fast_fibonacci example applies recursion, but does not apply any algorithm focused on saving data, as is done in the memoizationtechnique. In the fibonacci example, an array is used that will be generated at the beginning of the program execution, with the size defined by the user and only after that, that the user can request the desired fibonacci value! My program uses a dictionary to store the corresponding values ​​of the fibonacci sequence, only when necessary! So, my program allows you to find the desired value in O (1), right on the first run, and it is possible to save this same dictionary for future runs! The limitation of this program is due to the amount of memory available in the system!

poyea:
> > Thanks. Did you take a look at the existing implementation in `dynamic programming`? What's the major difference between that two versions and yours?
> 
> My program is in the application of memoization in the fibonacci sequence using a single recursive function. The fast_fibonacci example applies recursion, but does not apply any algorithm focused on saving data, as is done in the memoizationtechnique. In the fibonacci example, an array is used that will be generated at the beginning of the program execution, with the size defined by the user and only after that, that the user can request the desired fibonacci value! My program uses a dictionary to store the corresponding values ​​of the fibonacci sequence, only when necessary! So, my program allows you to find the desired value in O (1), right on the first run, and it is possible to save this same dictionary for future runs! The limitation of this program is due to the amount of memory available in the system!

Yes, but is it still, classified as `dynamic programming`?

SamueldaCostaAraujoNunes:
> dynamic programming

I didn't see anything similar in the dynamic programming folder, do you think it would be better to put this program in the dynamic programming folder instead of the Math folder?

SamueldaCostaAraujoNunes:
Hello, would you have any further comments for this program?

poyea:
> > dynamic programming
> 
> I didn't see anything similar in the dynamic programming folder, do you think it would be better to put this program in the dynamic programming folder instead of the Math folder?

Yes. I'd say since the function is returning one number, maybe it should go to `dynamic programming`.

SamueldaCostaAraujoNunes:
I'm changing the folder to Dynamic Programming

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,48 @@
+# Fibonacci Sequence Using Recursion and Memoization
+
+# Memoization Concept: https://en.wikipedia.org/wiki/Memoization
+# Fibonacci Concept : https://pt.wikipedia.org/wiki/Sequ%C3%AAncia_de_Fibonacci
+
+cache: dict = dict()
+
+
+def fibonacci(n: int) -> int:
```
poyea, line 9:
```suggestion



def fibonacci(n: int, cache: dict = dict()) -> int:
```

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,48 @@
...
+    aux = n
+
+    try:
+        if aux in cache:
+            return cache[aux]
+
+        if n in {0, 1}:
+            ans = n
+        else:
+            ans = fibonacci(n - 1) + fibonacci(n - 2)
```
poyea, line 26:
```suggestion
            ans = fibonacci(n - 1, cache) + fibonacci(n - 2, cache)
```

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,48 @@
...
+
+def main() -> None:
+    try:
+        number = int(input("Enter an integer greater than or equal to 0: "))
+        print(f"The first {number} Fibonacci sequence terms are:")
+        print([fibonacci(n) for n in range(number)])
+        print(f"The {number} index of the fibonacci sequence:")
+        print(fibonacci(number))
+    except ValueError:
+        print("Error: Input value should be an integer greater than or equal to 0")
```
poyea, line 44: We can move these into `if __name__ == "__main__":`. (optional)

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == "__main__":
+    main()
```
poyea, line 48:
```suggestion
if __name__ == "__main__":
    import doctest
    doctest.testmod()
    main()
```

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
SamueldaCostaAraujoNunes: I really loved this suggestion, much better than declaring a dictionary in a global scope!

poyea (approved):

53jk1 (approved):

SamueldaCostaAraujoNunes:
Hello, @ Kush1101, would it be possible to perform a review of my PR?

SamueldaCostaAraujoNunes:
Hello, @Kush1101 , would it be possible to perform a review of my PR?

sarveshggn (approved):

SamueldaCostaAraujoNunes:
Hello, @Kush1101 , would it be possible to perform a review of my PR? Or if possible, designate a Dev to evaluate my PR

todlove:
G_i

todlove:
SamueldaCostaAraujoNunes:master

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,49 @@
+# Fibonacci Sequence Using Recursion and Memoization
+
+# Memoization Concept: https://en.wikipedia.org/wiki/Memoization
+# Fibonacci Concept : https://pt.wikipedia.org/wiki/Sequ%C3%AAncia_de_Fibonacci
+
+
+def fibonacci(n: int, cache: dict = dict()) -> int:
```
dhruvmanila, line 7: Cache is an implementation detail and not something which needs to be exposed to the client. So, we should not use that as a function parameter and instead make a variable `cache` either globally or if using a class, a private instance variable.

Another thing is you need to provide type hints for the dictionary keys and values like so: `dict[keytype, valuetype]`

Also, never keep a mutable object as the default argument value. It can have unforeseen side effects.

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Return number of index n in sequence fibonacci
+    >>> [fibonacci(i) for i in range(12)]
+    [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89]
+    >>> fibonacci(12)
+    144
+    """
+    aux = n
```
dhruvmanila, line 15: I don't think there's any need to assign another variable, we can directly use `n`.

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
```diff
@@ -0,0 +1,49 @@
...
+            ans = n
+        else:
+            ans = fibonacci(n - 1, cache) + fibonacci(n - 2, cache)
+        cache[aux] = ans
+
+        return ans
+    except TypeError:
+        print("Error: An integer value is expected")
+    except ValueError:
+        print("Error: The value must be greater than or equal to zero")
```
dhruvmanila, line 31: In the above scenario, instead of using the `try ... except ...` pattern, please use the pattern like so,
```python
# pseudocode
if "some condition did not satisfy":
    raise "an error with the appropriate message"
```

dhruvmanila (changes requested):

SamueldaCostaAraujoNunes (commented):
Corrigido!!

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
SamueldaCostaAraujoNunes: Corrected, thanks for the remarks

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
SamueldaCostaAraujoNunes: Corrected, thanks for the remarks

dynamic_programming/fibonacci_sequence_recursion_with_memoization.py
SamueldaCostaAraujoNunes: Corrected, thanks for the remarks

SamueldaCostaAraujoNunes (commented):
Aprovado!

SamueldaCostaAraujoNunes:
F

PR #2632, opened 2020-10-02
Fix: Corrected test. List in test must be ordered.

NumberPiOso:
What the title says.

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [ X All functions and variable names follow Python naming conventions.

### DIFF
--- a/searches/simple_binary_search.py
+++ b/searches/simple_binary_search.py
@@ -33,6 +33,12 @@ def binary_search(a_list: list[int], item: int) -> bool:
     False
     >>> print(binary_search([-.1, .1 , .8], .1))
     True
+    >>> binary_search(range(-5000, 5000, 10), 80)
+    True
+    >>> binary_search(range(-5000, 5000, 10), 1255)
+    False
+    >>> binary_search(range(0, 10000, 5), 2)
+    False
     """
     if len(a_list) == 0:
         return False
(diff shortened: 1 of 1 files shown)


dhruvmanila:
Awesome! Thanks for pointing this out. Can you add a few more random tests by using `range` and different `step` values? Maybe we can use `sorted(random.choices(range(-n, +n), k=100))` to select 100 pseudo-random numbers from a list of 1000 numbers ranging from negative to positive.

NumberPiOso:
@poyea are you working in this one? or should I propose the next commit ?

poyea:
You may consider adding the test cases suggested. I'll review it once you've done.

NumberPiOso:
@dhruvmanila could you please elaborate a little more that idea?
In particular,  how do I know in advance if the number is in the list.

```
>>> n = 500
>>> print(binary_search(sorted(random.choices(range(-n, +n), k=100)), 100))
>>> ?
```

dhruvmanila:
Ah, did not see that. Sorry about that. A better option would be to just use `random(range(-100, 100, <step_value>))` or something like that. Put whichever step value you think would be good enough.

dhruvmanila (approved):
Looks good! Thank you for your contribution!

PR #2641, opened 2020-10-02
math/fermat_little_theorem: update proof statement

pikulet:
### **Describe your change:**

From wikipedia, the definition of fermat little theorem is that a**p = a (mod p)

In the old code, the author checked if:
(a/b) == a * (b**(p-2)) (mod p)

This is a rather convoluted way of expressing the fermat little theorem, which only involves two params a and p in the first place. The equivalent transformation was to do b**2/a to the original LHS and RHS.

I have also added the demonstration code for the special case when a is not divisible by p, then:
a**(p-1) = 1 (mod p)

* [x] Add an algorithm?
* [x] Documentation change?

### **Checklist:**

### DIFF
--- a/maths/fermat_little_theorem.py
+++ b/maths/fermat_little_theorem.py
@@ -2,7 +2,7 @@
 Python program to show the usage of Fermat's little theorem in a division
 
 If p is a prime number, then given any integer a,
-(a raised to the power of p) is congruent to (a) modulo the prime p.
+(a**p) is congruent to (a) modulo the prime p.
 
 Wikipedia reference: https://en.wikipedia.org/wiki/Fermat%27s_little_theorem
 """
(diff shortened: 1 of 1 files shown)


maths/fermat_little_theorem.py
```diff
@@ -18,14 +30,60 @@ def binary_exponentiation(a, n, mod):
...
+    assert a % p != 0
+
+    a_power_p_minus_one = binary_exponentiation(a, p - 1, p)
+    return a_power_p_minus_one == 1
 
-a = 1000000000
-b = 10
 
-# using binary exponentiation function, O(log(p)):
-print((a / b) % p == (a * binary_exponentiation(b, p - 2, p)) % p)
```
pikulet, line 28: This was the previous check done!

maths/fermat_little_theorem.py
```diff
@@ -1,12 +1,24 @@
-# Python program to show the usage of Fermat's little theorem in a division
-# According to Fermat's little theorem, (a / b) mod p always equals
-# a * (b ^ (p - 2)) mod p
```
pikulet, line 3: This is the original definition the author was working with. It's very convoluted and a is unnecessary in this equation (appears once on both LHS and RHS)

53jk1 (approved):

PR #2659, opened 2020-10-02
Fix: Multiple errors in fibonacci search.

NumberPiOso:
- Test lists were not ordered, this is required for Fibonacci search
- Place documentation of function inside function
- Create multiple different tests including, float, char and negatives
- Add type hints in line with #2128

### **Describe your change:**

### **Checklist:**

### DIFF
--- a/searches/fibonacci_search.py
+++ b/searches/fibonacci_search.py
@@ -10,7 +10,6 @@ python3 -m doctest -v fibonacci_search.py
 For manual testing run:
 python3 fibonacci_search.py
 """
-import sys
 from functools import lru_cache
 
 
@@ -104,9 +103,12 @@ def fibonacci_search(arr: list, val: int) -> int:
     """
     len_list = len(arr)
     # Find m such that F_m >= n where F_i is the i_th fibonacci number.
-    greater_fibb_index = next(
-        x for x in range(sys.maxsize ** 10) if fibonacci(x) >= len_list
-    )
+    i = 0
+    while True:
+        if fibonacci(i) >= len_list:
+            greater_fibb_index = i
+            break
+        i += 1
     fibb_k = greater_fibb_index
     offset = 0
     while fibb_k > 0:
--- a/searches/fibonacci_search.py
+++ b/searches/fibonacci_search.py
@@ -106,10 +106,9 @@ def fibonacci_search(arr: list, val: int) -> int:
     i = 0
     while True:
         if fibonacci(i) >= len_list:
-            greater_fibb_index = i
+            fibb_k = i
             break
         i += 1
-    fibb_k = greater_fibb_index
     offset = 0
     while fibb_k > 0:
         index_k = min(
--- a/searches/fibonacci_search.py
+++ b/searches/fibonacci_search.py
@@ -10,12 +10,11 @@ python3 -m doctest -v fibonacci_search.py
 For manual testing run:
 python3 fibonacci_search.py
 """
-from __future__ import annotations
-from functools import lru_cache
 import sys
+from functools import lru_cache
 
 
-@lru_cache()
+@lru_cache
 def fibonacci(k: int) -> int:
     """Finds fibonacci number in index k.
 
@@ -46,12 +45,12 @@ def fibonacci(k: int) -> int:
         return fibonacci(k - 1) + fibonacci(k - 2)
 
 
-def fibonacci_search(arr: List[int], val: int) -> int:
+def fibonacci_search(arr: list, val: int) -> int:
     """A pure Python implementation of a fibonacci search algorithm.
 
     Parameters
     ----------
-    arr : List[int]
+    arr : list
         List of sorted elements.
     val : int
         Element to search in list.
--- a/searches/fibonacci_search.py
+++ b/searches/fibonacci_search.py
@@ -1,51 +1,101 @@
-# run using python fibonacci_search.py -v
-
 """
-@params
-arr: input array
-val: the value to be searched
-output: the index of element in the array or -1 if not found
-return 0 if input array is empty
+This is pure Python implementation of fibonacci search.
+
+Resources used:
+https://en.wikipedia.org/wiki/Fibonacci_search_technique
+
+For doctests run following command:
+python3 -m doctest -v fibonacci_search.py
+
+For manual testing run:
+python3 fibonacci_search.py
 """
+from __future__ import annotations
+from functools import lru_cache
+import sys
 
 
-def fibonacci_search(arr, val):
+@lru_cache()
+def fibonacci(k: int) -> int:
+    """Finds fibonacci number in index k.
 
-    """
-    >>> fibonacci_search([1,6,7,0,0,0], 6)
-    1
-    >>> fibonacci_search([1,-1, 5, 2, 9], 10)
-    -1
-    >>> fibonacci_search([], 9)
+    Parameters
+    ----------
+    k : int
+        Index of fibonacci.
+
+    Returns
+    -------
+    int
+        Fibonacci number in position k.
+
+    >>> print(fibonacci(0))
     0
+    >>> print(fibonacci(2))
+    1
+    >>> print(fibonacci(5))
+    5
+    >>> print(fibonacci(15))
+    610
     """
-    fib_N_2 = 0
-    fib_N_1 = 1
-    fibNext = fib_N_1 + fib_N_2
-    length = len(arr)
-    if length == 0:
+    if k == 0:
         return 0
-    while fibNext < len(arr):
-        fib_N_2 = fib_N_1
-        fib_N_1 = fibNext
-        fibNext = fib_N_1 + fib_N_2
-    index = -1
-    while fibNext > 1:
-        i = min(index + fib_N_2, (length - 1))
-        if arr[i] < val:
-            fibNext = fib_N_1
-            fib_N_1 = fib_N_2
-            fib_N_2 = fibNext - fib_N_1
-            index = i
-        elif arr[i] > val:
-            fibNext = fib_N_2
-            fib_N_1 = fib_N_1 - fib_N_2
-            fib_N_2 = fibNext - fib_N_1
-        else:
-            return i
-    if (fib_N_1 and index < length - 1) and (arr[index + 1] == val):
-        return index + 1
-    return -1
+    elif k == 1:
+        return 1
+    else:
+        return fibonacci(k - 1) + fibonacci(k - 2)
+
+
+def fibonacci_search(arr: List[int], val: int) -> int:
+    """A pure Python implementation of a fibonacci search algorithm.
+
+    Parameters
+    ----------
+    arr : List[int]
+        List of sorted elements.
+    val : int
+        Element to search in list.
+
+    Returns
+    -------
+    int
+        The index of the element in the array.
+        -1 if the element is not found.
+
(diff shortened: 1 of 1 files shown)


searches/fibonacci_search.py
```diff
@@ -1,51 +1,100 @@
...
+
+
+@lru_cache
+def fibonacci(k: int) -> int:
+    """Finds fibonacci number in index k.
 
+    Parameters
+    ----------
+    k : int
+        Index of fibonacci.
```
dhruvmanila, line 24:
```suggestion
    Parameters
    ----------
    k : Index of fibonacci.
```
Already mentioned in type hints.

searches/fibonacci_search.py
```diff
@@ -1,51 +1,100 @@
...
 
-def fibonacci_search(arr, val):
+    Returns
+    -------
+    int
+        Fibonacci number in position k.
```
dhruvmanila, line 29:
```suggestion
    Returns
    -------
    Fibonacci number in position k.
```
Same as above

searches/fibonacci_search.py
```diff
@@ -1,51 +1,100 @@
...
+    >>> print(fibonacci(2))
+    1
+    >>> print(fibonacci(5))
+    5
+    >>> print(fibonacci(15))
+    610
     """
-    >>> fibonacci_search([1,6,7,0,0,0], 6)
+    if k == 0:
+        return 0
```
dhruvmanila, line 41: Please check for whether the input is `int` and greater than 0 and write an appropriate `doctest` for it.

searches/fibonacci_search.py
```diff
@@ -1,51 +1,100 @@
...
+    arr : list
+        List of sorted elements.
+    val : int
+        Element to search in list.
+
+    Returns
+    -------
+    int
+        The index of the element in the array.
+        -1 if the element is not found.
```
dhruvmanila, line 62:
```suggestion
    Parameters
    ----------
    arr : List of sorted elements.
    val : Element to search in the list.

    Returns
    -------
    The index of the element in the array.
    -1 if the element is not found.
```
As mentioned in the above comment.

searches/fibonacci_search.py
```diff
@@ -1,51 +1,100 @@
...
+    n = len(arr)
+    # Find m such that F_m >= n where F_i is the i_th fibonacci number.
+    m = next(x for x in range(sys.maxsize ** 10) if fibonacci(x) >= n)
+    k = m
+    offset = 0
+    while k != 0:
+        if arr[offset + fibonacci(k - 1)] == val:
+            return fibonacci(k - 1)
+        elif val < arr[offset + fibonacci(k - 1)]:
+            k = k - 1
```
dhruvmanila, line 92:
```suggestion
            k -= 1
```

searches/fibonacci_search.py
```diff
@@ -1,51 +1,100 @@
...
+        elif val > arr[offset + fibonacci(k - 1)]:
+            offset += fibonacci(k - 2)
+            k = k - 2
```
dhruvmanila, line 95:
```suggestion
            k -= 2
```

dhruvmanila (changes requested):

searches/fibonacci_search.py
NumberPiOso: I am using NumPy Style Python Docstrings, and it is said that the return type must be duplicated in the docstring to comply.

URL
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

@dhruvmanila why do you thinks it is better this other way?

searches/fibonacci_search.py
```diff
@@ -35,7 +35,17 @@ def fibonacci(k: int) -> int:
...
+    Traceback (most recent call last):
+    ValueError: k must be an integer.
+    >>> print(fibonacci(-5))
+    Traceback (most recent call last):
+    ValueError: k integer must be greater or equal to zero.
     """
+    if not isinstance(k, int):
+        raise ValueError("k must be an integer.")
+    if k < 0:
+        raise ValueError("k integer must be greater or equal to zero.")
```
NumberPiOso, line 49: I have a question in this one.

Do you think is it better to use Assertions or this conditionals?

searches/fibonacci_search.py
dhruvmanila:
```python
>>> import this
...
There should be one-- and preferably only one --obvious way to do it.
...
```

No duplications.

I have no idea about the NumPy Style Python Docstrings but if it is clear to read then it's fine.

searches/fibonacci_search.py
dhruvmanila: Assertions will raise `AssertionError` which is ambiguous, the user has no idea what the function is asserting. By raising specific errors, users can understand and the code is easier to read.

You can use `try ... except ...` block but that would become cumbersome and unnecessary for such small cases.

NumberPiOso:
I think we are done here. Are not we?

searches/fibonacci_search.py
```diff
@@ -1,51 +1,110 @@
...
+    >>> print(fibonacci('a'))
+    Traceback (most recent call last):
+    ValueError: k must be an integer.
+    >>> print(fibonacci(-5))
+    Traceback (most recent call last):
+    ValueError: k integer must be greater or equal to zero.
     """
-    >>> fibonacci_search([1,6,7,0,0,0], 6)
+    if not isinstance(k, int):
+        raise ValueError("k must be an integer.")
```
dhruvmanila, line 47:
```suggestion
        raise TypeError("k must be an integer.")
```
As the user entered the wrong type.

searches/fibonacci_search.py
```diff
@@ -1,51 +1,110 @@
...
-            fib_N_1 = fib_N_1 - fib_N_2
-            fib_N_2 = fibNext - fib_N_1
-        else:
-            return i
-    if (fib_N_1 and index < length - 1) and (arr[index + 1] == val):
-        return index + 1
-    return -1
+    n = len(arr)
+    # Find m such that F_m >= n where F_i is the i_th fibonacci number.
+    m = next(x for x in range(sys.maxsize ** 10) if fibonacci(x) >= n)
```
dhruvmanila, line 95: `sys.maxsize` is already a huge number. Can you tell me why is this necessary?
And please use descriptive variable names instead of the single letter names as mentioned in CONTRIBUTING.md. Apply the changes wherever the variables were used.

dhruvmanila (changes requested):

searches/fibonacci_search.py
NumberPiOso: `sys.maxsize` is a simple way to represent that the generator produced can go up to infinity. Probably is not the best way to find it.

It is equal to do something like this.

```
i = 0
while True:
    if fibbonacci(i) >= n:
        m = i
        break
    i += 1
```

Both are loops that can go up to infinity. But I do not know which one of them is clearer.

NumberPiOso: I commented it inside the conflict, but I will paste the message here, because I think there it is not easy to find.

`sys.maxsize` is a simple way to represent that the generator produced can go up to infinity. Probably is not the best way to find it.

It is equal to do something like this.

```
i = 0
while True:
    if fibbonacci(i) >= n:
        m = i
        break
    i += 1
```

Both are loops that can go up to infinity. But I do not know which one of them is clearer.

dhruvmanila: As far as I understand, you are finding the index of a fibonacci number which is greater than the length of the input array, right? If that's the case then why are you using a generator just to find one value?

It seems like I probably found a bug:
```python
print(fibonacci_search(list(range(100)), 63))

...
  File "fibonacci_search.py", line 117, in <module>
    print(fibonacci_search(list(range(100)), 63))
  File "fibonacci_search.py", line 106, in fibonacci_search
    offset += fibonacci(fibb_k - 2)
  File "fibonacci_search.py", line 49, in fibonacci
    raise ValueError("k integer must be greater or equal to zero.")
ValueError: k integer must be greater or equal to zero.
```

NumberPiOso: > As far as I understand, you are finding the index of a fibonacci number which is greater than the length of the input array, right? If that's the case then why are you using a generator just to find one value?

Because when I use, next of a generator it will only find the first value of the iterator. And the first value of that generator is the value which I am looking for.

I think is OK but the code in this repo should be clean and understandable. Do you think that it is better for me to change it to something easier? like the snippet above.

dhruvmanila: I think the above mentioned snippet is more readable (with descriptive variable names). Also, take a look at the error produced in the above comment.

NumberPiOso: > I think the above mentioned snippet is more readable (with descriptive variable names).
Agreed, I change it for the sake of clarity.

>  Also, take a look at the error produced in the above comment.
Thank you for pointing that out, it is a special case of the fibonacci search. I fixed it and added your example and many more with big list as doctests.

anilbaday:
What is this ?

iOS için Outlook<https://aka.ms/o0ukef> uygulamasını edinin
________________________________
Gönderen: Dhruv <notifications@github.com>
Gönderildi: Wednesday, October 7, 2020 5:28:49 PM
Kime: TheAlgorithms/Python <Python@noreply.github.com>
Bilgi: Subscribed <subscribed@noreply.github.com>
Konu: Re: [TheAlgorithms/Python] Fix: Multiple errors in fibonacci search. (#2659)


@dhruvmanila commented on this pull request.

________________________________

In searches/fibonacci_search.py<https://github.com/TheAlgorithms/Python/pull/2659#discussion_r501056383>:

> -            fibNext = fib_N_1
-            fib_N_1 = fib_N_2
-            fib_N_2 = fibNext - fib_N_1
-            index = i
-        elif arr[i] > val:
-            fibNext = fib_N_2
-            fib_N_1 = fib_N_1 - fib_N_2
-            fib_N_2 = fibNext - fib_N_1
-        else:
-            return i
-    if (fib_N_1 and index < length - 1) and (arr[index + 1] == val):
-        return index + 1
-    return -1
+    n = len(arr)
+    # Find m such that F_m >= n where F_i is the i_th fibonacci number.
+    m = next(x for x in range(sys.maxsize ** 10) if fibonacci(x) >= n)


As far as I understand, you are finding the index of a fibonacci number which is greater than the length of the input array, right? If that's the case then why are you using a generator just to find one value?

It seems like I probably found a bug:

print(fibonacci_search(list(range(100)), 63))

...
  File "fibonacci_search.py", line 117, in <module>
    print(fibonacci_search(list(range(100)), 63))
  File "fibonacci_search.py", line 106, in fibonacci_search
    offset += fibonacci(fibb_k - 2)
  File "fibonacci_search.py", line 49, in fibonacci
    raise ValueError("k integer must be greater or equal to zero.")
ValueError: k integer must be greater or equal to zero.

searches/fibonacci_search.py
```diff
@@ -1,51 +1,130 @@
...
-    return -1
+    len_list = len(arr)
+    # Find m such that F_m >= n where F_i is the i_th fibonacci number.
+    i = 0
+    while True:
+        if fibonacci(i) >= len_list:
+            greater_fibb_index = i
+            break
+        i += 1
+    fibb_k = greater_fibb_index
```
dhruvmanila, line 112:
```suggestion
            fibb_k = i
            break
        i += 1
```
You can directly assign it in the loop.

dhruvmanila (approved):
Looks good now! Thank you for your contribution!

PR #2665, opened 2020-10-02
Issue fixed #2661

### LINKED ISSUE #2661: Issue related to addition.
In the Maths folder there is add. py file.  So that is used for addition of two numbers. You must have to implement it and use it in a way so that user can enter N number and get the addition. Also you have to use try/except just because if someone uses string.

MANYAJAIN195:
Modified add.py .
Now user can enter N number and get the addition and also used try/except just because if someone uses string.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/maths/add.py
+++ b/maths/add.py
@@ -1,10 +1,16 @@
 ans="yes"
 l1=[]
 Sum=i=0
+'''
+Enter n numbers
+'''
 while ans=="yes":
+  '''
+  check for invalid input
+  '''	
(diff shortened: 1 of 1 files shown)


Kush1101 (commented):
It's a bit too trivial. We can add n numbers then we can simply make recursive calls to the `add()` function. If you want to contribute then please bring a genuine idea/implementation.

PR #2706, opened 2020-10-03
Two Sum with hash method

adityaSINGHrox:
Added a new method for Two Sum problem


adityaSINGHrox:
i didnt quite understand what is wrong 
i am kinda new to all this 
i'll learn what's wrong and the issue another pr

PR #2794, opened 2020-10-05
Sherlock and the Valid String

AARTISHELAR:
Sherlock considers a string to be valid if all characters of the string appear the same number of times. It is also valid if he can remove just 1 character at 1  index in the string, and the remaining characters will occur the same number of times. Given a string s , determine if it is valid. If so, return YES, otherwise return NO.
For example, if s=abc, it is a valid string because frequencies are {a:1,b:1,c:1} . So is s=abc because we can remove one c and have  1 of each character in the remaining string. If s=abccc  however, the string is not valid as we can only remove 1 occurrence of c . That would leave character frequencies of  {a:1,b:1,c:2} .

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/Sherlock and the Valid String
+++ b/Sherlock and the Valid String
@@ -1,0 +1,32 @@
+from collections import Counter
+
+s=input()
+d=Counter(s)
+s1=set(d.values())
+s1=list(s1)
+l=list(d.values())
+
+try:
+    one=l.count(s1[0])
+    two=l.count(s1[1])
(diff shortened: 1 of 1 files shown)


Kush1101 (commented):
Please go through the contribution guidelines and then if you want to contribute then submit a new PR.

PR #2821, opened 2020-10-05
Create a program to obtain length and breadth of rectangle and calcul…

pankaj1278:
…ate its area

a program to obtain length and breadth of rectangle and calculate its area


Kush1101 (commented):
This is not necessary. Please go through the contribution guidelines before making any PR.

PR #2875, opened 2020-10-05
Add solution for Project Euler problem 191

darkstar:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/project_euler/problem_191/sol1.py
+++ b/project_euler/problem_191/sol1.py
@@ -26,8 +26,74 @@ References:
 """
 
 
+# a cache to speed up calculation
+cache = {}
+
+
+def _calculate(days: int, absent: int, late: int) -> int:
+    """
+    A small helper function for the recursion, mainly to have
+    a clean interface for the solution() function below.
+
+    It should get called with the number of days (corresponding
+    to the desired length of the 'prize strings'), and the
+    initial values for the number of consecutive absent days and
+    number of total late days.
+
+    >>> _calculate(days=4, absent=0, late=0)
+    43
+    >>> _calculate(days=30, absent=2, late=0)
+    0
+    >>> _calculate(days=30, absent=1, late=0)
+    98950096
+    """
+
+    # if we are absent twice, or late 3 consecutive days,
+    # no further prize strings are possible
+    if late == 3 or absent == 2:
+        return 0
+
+    # if we have no days left, and have not failed any other rules,
+    # we have a prize string
+    if days == 0:
+        return 1
+
+    # No easy solution, so now we need to do the recursive calculation
+
+    # First, check if the combination is already in the cache, and
+    # if yes, return the stored value from there since we already
+    # know the number of possible prize strings from this point on
+    key = (days, absent, late)
+    if key in cache:
+        return cache[key]
+
+    # now we calculate the three possible ways that can unfold from
+    # this point on, depending on our attendance today
+
+    # 1) if we are late (but not absent), the "absent" counter stays as
+    # it is, but the "late" counter increases by one
+    state_late = _calculate(days - 1, absent, late + 1)
+
+    # 2) if we are absent, the "absent" counter increases by 1, and the
+    # "late" counter resets to 0
+    state_absent = _calculate(days - 1, absent + 1, 0)
+
+    # 3) if we are on time, this resets the "late" counter and keeps the
+    # absent counter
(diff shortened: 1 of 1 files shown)


project_euler/problem_191/sol1.py
```diff
@@ -0,0 +1,87 @@
...
+How many "prize" strings exist over a 30-day period?
+
+References:
+    - The original Project Euler project page:
+      https://projecteuler.net/problem=191
+"""
+
+
+def solution(days: int = 30) -> int:
+    """Returns the number of possible prize strings for a particular number
```
dhruvmanila, line 30:
```suggestion
    """
    Returns the number of possible prize strings for a particular number
```

project_euler/problem_191/sol1.py
```diff
@@ -0,0 +1,87 @@
...
+    1918080160
+    >>> solution(4)
+    43
+    """
+
+    # a cache to speed up calculation
+    cache = {}
+
+    # we will be using a simple recursive function
+    def calculate(days: int, absent: int, late: int) -> int:
```
dhruvmanila, line 43: Please separate out this function and include `doctest` and `docstring` for it.

project_euler/problem_191/sol1.py
```diff
@@ -0,0 +1,87 @@
...
+        # "late" counter resets to 0
+        state_absent = calculate(days - 1, absent + 1, 0)
+
+        # if we are on time, this resets the "late" counter and keeps the
+        # absent counter
+        state_ontime = calculate(days - 1, absent, 0)
+
+        # the total number of prize strings is the sum of these three
+        # possible states
+        prizestrings = state_late + state_absent + state_ontime
```
dhruvmanila, line 77: General comment: Please remove or improve the comments which are self-explanatory as mentioned in CONTRIBUTING.md. The above comment is literally a pseudocode.

For your reference:
The following are considered to be bad and may be requested to be improved:
```python
x = x + 2	# increased by 2
```
This is too trivial. Comments are expected to be explanatory.

project_euler/problem_191/sol1.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # now store the value we calculated in the cache
```
dhruvmanila, line 79: Same as above

dhruvmanila (changes requested):

project_euler/problem_191/sol1.py
darkstar: I wanted to keep the helper function "inside" the solution function, since it is an implementation detail and not part of the (public) interface. Is this not usually done in the python world?

dhruvmanila: Oh no, you don't actually need to hide the implementation details here but in the real world if you want to achieve that then you use underscores. You can google it up as it's quite a deep dive. Also, function inside a function is mostly used just for a wrapper function (you can google that up as well).

darkstar: but this basically *is* just a wrapper function, as the "solution()" function is now pretty much  just a 1-liner

dhruvmanila:
Please remove the comments which are self-explanatory. There's a good saying regarding it:
> Good code is self-documenting

darkstar:
I'm not sure which comments you deem self-explanatory since you didn't point them out, but maybe it's better now?

project_euler/problem_191/sol1.py
```diff
@@ -0,0 +1,105 @@
...
+    >>> _calculate(days=30, absent=2, late=0)
+    0
+    >>> _calculate(days=30, absent=1, late=0)
+    98950096
+    """
+
+    # if we are absent twice, or late 3 consecutive days,
+    # no further prize strings are possible
+    if late == 3 or absent == 2:
+        return 0
```
dhruvmanila, line 53: > I cannot hold your hand throughout your life :)

I will help you out here with this code as an example. The code here says if the `late` variable is 3 or the `absent` variable is 2 then we will `return` 0 and the comment says the same thing. What I want you to see is that comments are meant to explain the code **ONLY** if it's necessary but here the code is pretty self-explanatory. The comment here says the same thing in English.

What I basically mean here is that your code is perfect, you shouldn't need a ton of comments to explain what you have written. You don't want to keep referring to a dictionary when you're reading a book otherwise it breaks your flow.

dhruvmanila (approved):

PR #2880, opened 2020-10-05
Add solution for Project Euler problem 188

darkstar:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/project_euler/problem_188/sol1.py
+++ b/project_euler/problem_188/sol1.py
@@ -1,6 +1,7 @@
 """
+Project Euler Problem 188: https://projecteuler.net/problem=188
+
 The hyperexponentiation of a number
-Problem 188
 
 The hyperexponentiation or tetration of a number a by a positive integer b,
 denoted by a↑↑b or b^a, is recursively defined by:
@@ -14,10 +15,7 @@ Thus we have e.g. 3↑↑2 = 3^3 = 27, hence 3↑↑3 = 3^27 = 7625597484987 and
 Find the last 8 digits of 1777↑↑1855.
 
 References:
-    - The original Project Euler project page:
-      https://projecteuler.net/problem=188
-    - Wikipedia article about Hyperexponentiation, aka. Tetration
-      https://en.wikipedia.org/wiki/Tetration
+    - https://en.wikipedia.org/wiki/Tetration
 """
 
 
@@ -67,4 +65,4 @@ def solution(base: int = 1777, height: int = 1855, digits: int = 8) -> int:
 
 
 if __name__ == "__main__":
-    print(solution())
+    print(f"{solution() = }")
--- a/project_euler/problem_188/sol1.py
+++ b/project_euler/problem_188/sol1.py
@@ -22,29 +22,31 @@ References:
 
 
 # small helper function for modular exponentiation
-def modexpt(base: int, exponent: int, modulo_value: int) -> int:
-    """Returns the modular exponentiation, that is the value
+def _modexpt(base: int, exponent: int, modulo_value: int) -> int:
+    """
+    Returns the modular exponentiation, that is the value
     of `base ** exponent % modulo_value`, without calculating
     the actual number.
-    >>> modexpt(2, 4, 10)
+    >>> _modexpt(2, 4, 10)
     6
-    >>> modexpt(2, 1024, 100)
+    >>> _modexpt(2, 1024, 100)
     16
-    >>> modexpt(13, 65535, 7)
+    >>> _modexpt(13, 65535, 7)
     6
     """
 
     if exponent == 1:
         return base
     if exponent % 2 == 0:
-        x = modexpt(base, exponent / 2, modulo_value) % modulo_value
+        x = _modexpt(base, exponent / 2, modulo_value) % modulo_value
         return (x * x) % modulo_value
     else:
-        return (base * modexpt(base, exponent - 1, modulo_value)) % modulo_value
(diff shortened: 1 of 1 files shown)


project_euler/problem_188/sol1.py
```diff
@@ -0,0 +1,53 @@
+"""
+The hyperexponentiation of a number
+Problem 188
```
dhruvmanila, line 3: Include the project link and any resources used, such as a wikipedia article or something which points the user to understand about this topic

project_euler/problem_188/sol1.py
```diff
@@ -0,0 +1,53 @@
...
+a↑↑(k+1) = a(a↑↑k).
+
+Thus we have e.g. 3↑↑2 = 3^3 = 27, hence 3↑↑3 = 3^27 = 7625597484987 and
+3↑↑4 is roughly 103.6383346400240996*10^12.
+
+Find the last 8 digits of 1777↑↑1855.
+"""
+
+
+def solution(a=1777, k=1855, digits=8):
```
dhruvmanila, line 18: Type hints: https://docs.python.org/3/library/typing.html

Please use descriptive names instead of one letter names like `a` and `k`

project_euler/problem_188/sol1.py
```diff
@@ -0,0 +1,53 @@
...
+    97484987
+    """
+
+    # we calculate everything modulo 10^8, since we only care about the
+    # last 8 digits
+    modulo_value = 10 ** digits
+
+    # small helper function for modular exponentiation, to keep the result
+    # values small enough
+    def modexpt(base, exponent):
```
dhruvmanila, line 35: You can keep this function separately.
Include type hints, `doctest` for this function

dhruvmanila (changes requested):

project_euler/problem_188/sol1.py
darkstar: Is there any project euler problem which I can use as template? I looked at several of them and used these as a template to model my solution after

dhruvmanila: Some algorithms don't require any external resource to solve or refer to but as this is introducing some topic, people reading this code will want to know more about it.

This is all done to make life easier for future readers of the code including yourself.

project_euler/problem_188/sol1.py
darkstar: But it's literally "a" and "k" in the problem itself, should I add that as a comment then? like 

> "a" in the problem is called "base" here

or similar?

project_euler/problem_188/sol1.py
darkstar: Yes, I understand that. I just want to know where to put it. After the problem statement? In a separate section? With some heading or just as plain link? That's why I was asking about an example

darkstar:
I updated everything with the requested changes from #2875 which probably apply here as well, even though not explicitly stated

project_euler/problem_188/sol1.py
```diff
@@ -0,0 +1,74 @@
...
+        return (base * _modexpt(base, exponent - 1, modulo_value)) % modulo_value
+
+
+def solution(base: int = 1777, height: int = 1855, digits: int = 8) -> int:
+    """
+    Returns the last 8 digits of the hyperexponentiation of base by
+    height, i.e. the number base↑↑height:
+
+    >>> solution()
+    95962097
```
dhruvmanila, line 53: You don't need to test the answer here as it is being tested by this [script](https://github.com/TheAlgorithms/Python/blob/master/project_euler/validate_solutions.py). Provide `doctest` to test the function and not the answer.

darkstar: I don't think I understand. Do you want me to remove that one test and leave the other two in place? Isn't the answer to the problem also a test of the function?

dhruvmanila: I am going to write guidelines regarding code style for Project Euler to maintain consistency throughout the directory. What I'm proposing here is that the above-linked script is already checking whether the answer is correct or not. Now, if we include it in `doctest` again then Travis CI will run that test and so we will have duplicate tests. To remove that I'm proposing to not include the test for the answer but include tests for the function. Check whether the function is working appropriately by passing arguments other than the default ones (which gives us the answer).

Example: If `solution(n = 100)` gives the answer to the problem, test the function for `solution(n = 5)`, `solution(n = 0)`, `solution(n = 20)` and so on. Check whether the answer to those calls matches what you expect from the function.

darkstar:
This should address all requested changes now

darkstar:
Why is this failing in a file that I didn't even change?

dhruvmanila (commented):
### Announcement:

@darkstar

*This message is sent to all pull requests which are **OPEN, NOT MERGED, not containing 'wontfix' label and contains 'Project Euler' or 'Project' or 'Euler' (case doesn't matter) in their title**. If this message doesn't apply to your pull request, please ignore this.*

### Message:

This is to notify all the users submitting a pull request for the Project Euler solution that the [README.md](https://github.com/TheAlgorithms/Python/blob/master/project_euler/README.md) file has been updated with all the style requirements necessary for the directory and solution files. ***Please read through it and make all the necessary changes required to your submission.***

dhruvmanila:
Remove the label.

dhruvmanila (approved):
Thank you for your contribution and for your persistence!

Issue #2786, opened 2020-10-05
[Project Euler] Algorithms do not follow coding styles


kalpanajangra:
Many algorithms in Project Euler do not follow coding styles as mentioned by @dhruvmanila [here](https://github.com/TheAlgorithms/Python/issues/2695#issuecomment-703122388), or even the standard coding guidelines.

Notes:
- A lot of inconsistencies with the `solution` functions (a lot of them aren't named as "solution")
- docstring inconsistencies:
```
>```docstring```
vs
>```
docstring```
```
- [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md) says:
```
The following are considered to be bad and may be requested to be improved:
>>> x = x + 2	# increased by 2
This is too trivial. Comments are expected to be explanatory. For comments, you can write them above, on or below a line of code, as long as you are consistent within the same piece of code.
```
  But there are places where this isn't practised. https://github.com/TheAlgorithms/Python/blob/b934da4516a18814e2645413a20eedb3ac03a170/project_euler/problem_30/soln.py#L36
- Python typehints don't seem to be present in the initial problem solutions.
https://github.com/TheAlgorithms/Python/blob/b934da4516a18814e2645413a20eedb3ac03a170/project_euler/problem_01/sol1.py#L9 https://github.com/TheAlgorithms/Python/blob/b934da4516a18814e2645413a20eedb3ac03a170/project_euler/problem_01/sol2.py#L9
- Filename inconsistencies. [example](https://github.com/TheAlgorithms/Python/blob/master/project_euler/problem_30/soln.py)

The list isn't complete.
As a beginner who's looking to either understand/contribute, this makes it a bit difficult.

dhruvmanila:
This is a completely valid issue. The thing is:
- Those initial solutions were submitted a long time ago
- I don't think consistency was maintained back then

Now as the number of solutions is increasing we need to enforce some of the consistency. One of the main ones was as I mentioned in https://github.com/TheAlgorithms/Python/issues/2695#issuecomment-703122388 but as you said there are many more. We can start with this and fix one issue at a time.

Due to some of the solutions not containing `solution()` function or requires some number of positional arguments we have decided to skip this [script](https://github.com/TheAlgorithms/Python/blob/master/project_euler/validate_solutions.py) for now which checks for the solutions for every problem present. We can start fixing this first and then move on to other issues mentioned. 

[...]

PR #2958, opened 2020-10-07
Hacktoberfest 2020 - Style Improvements for project_euler/problem_26

nstoik:
### **Describe your change:**

Style and testing improvements for Project Euler problem 26

* add completed problem description
* add solution() function to pass tests in travis-ci
* add type hints and default argument to solution() function
* corrected invocation function of sol1.py as per the Contribution guide

Related to: #2786 

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


project_euler/problem_26/sol1.py
```diff
@@ -35,8 +53,22 @@ def find_digit(numerator: int, digit: int) -> int:
     return the_digit
 
 
-# Tests
-if __name__ == "__main__":
-    import doctest
+def solution(n: int = 1000) -> int:
```
dhruvmanila, line 56: Instead of making another function, change the name of the original function and add default arguments to it.
Also, please don't remove the doctest part in `if __name__ == "__main__":` block

dhruvmanila (changes requested):

project_euler/problem_26/sol1.py
nstoik: Thanks for the review! I thought that the `solution()` function needed to only have one argument. But after looking at the test script more I see that it just needs to have default arguments. 

I have made the changes you requested. 

I was a little unsure how to to modify the `print(solution(int(str(input()).strip())))` section in the `if __name__ == "__main__":` block to deal with multiple arguments. So I tried to make it as similar to the problem description as I could. Let me know if that's okay.

Thanks.

project_euler/problem_26/sol1.py
```diff
@@ -35,8 +53,8 @@ def find_digit(numerator: int, digit: int) -> int:
     return the_digit
 
 
-# Tests
 if __name__ == "__main__":
     import doctest
 
     doctest.testmod()
+    print(solution(digit=int(str(input()).strip())))
```
dhruvmanila, line 60: I don't see this in the original file, so please remove this.

nstoik: got it, done

dhruvmanila (approved):
Thank you for your contribution!

PR #2992, opened 2020-10-07
[Project Euler] Fix code style in Problem 41

JoanRosell:
### **Describe your change:**
Addresses #2786 
 * Improved the problem description with a link
 * Added solution function that calls the old function.
 * Fixed some doctests, their use was inconsistent with the functions they tested

I have some questions @dhruvmanila :
 1. Is it correct to use an assignment inside an f-string? PyCharm is complaining and I can't execute the tests with the current solution call, but I'm leaving it as I found it in the original file.
 2. There is one test case (with n = 7) that is cumbersome to put in a doctest, as the call generates a huge list that wraps around many lines, I have left it out and tested directly the solution output, which passes the test case. Should I leave it as it was and undo the changes?
----
* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/project_euler/problem_41/sol1.py
+++ b/project_euler/problem_41/sol1.py
@@ -36,20 +36,21 @@ def is_prime(n: int) -> bool:
     return True
 
 
-def solution(n: int = 7) -> list[int]:
+def solution(n: int = 7) -> int:
     """
     Returns a list of all pandigital prime numbers of length n.
     >>> solution(2)
-    []
-    >>> max(solution(4))
+    0
+    >>> solution(4)
     4231
-    >>> max(solution(7))
+    >>> solution(7)
     7652413
     """
     pandigital_str = "".join(str(i) for i in range(1, n + 1))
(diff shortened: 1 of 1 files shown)


dhruvmanila:
`print(f"{max(compute_pandigital_primes(7)) = }")`

If you're referring to this then this is not an assignment statement. The `=` sign in an f-string means that it will produce an output where the expression will be mentioned first then ` = ` and then the output of the expression.

Example:
```python
>>> print(f"{1 + 2 = }")
1 + 2 = 3
```

For 2: leave it as it was

moveax3 (approved):
Good increase of pythonicity and code documentation.

JoanRosell:
Alright, I removed the extra function I added and left the old code as it was.

dhruvmanila:
The answer is actually the maximum value in the list but the function returns a list, we just need the value. Please make the necessary changes as required.

dhruvmanila (approved):
Thank you for your contribution!

Issue #2695, opened 2020-10-03
Add more solutions to Project Euler problems


aceking007:
Project Euler has about 700 problems, but the current repository hosts solutions for only about 60-70 problems.   
I'm willing to work on the issue and also if someone wants to help they can join in.  
We can create guidelines on how many minimum solutions a PR should have to prevent spammy PRs with just one or two easy solutions. (open to discussion)

dhruvmanila:
As of now, we are only accepting PRs with one algorithm in it as it is easier to review one file at a time.
Feel free to open PR for whichever problem you have solved provided the answer is correct.

Anyone reading this can join and open PRs as well. We accept multiple solutions to a problem as everyone has their own approach to problem-solving but it should be their own and unique compared to other solutions. 

**NOTE:** Please give us time to review as there are so many to go through. If possible submit one PR at a time and when that is merged provided it passes all the automatic tests, you can submit another PR.

***One step at a time***

RGarg2002:
Where can I get the problems from?

berry-thawson:
https://projecteuler.net/
Hope this helps

RGarg2002:
Thanks it helped🙂

hbarovertwo:
I would love to join and contribute. Project Euler was always fun to me back in college :)

berry-thawson:
Hi,
When there is a program for which input is presented as a file, how should the input be coded?
Should it be hardcoded in the file or should it be reading from a file relative to the directory?

dhruvmanila:
@berry-thawson Please refer to [problem_54#L367-L369](https://github.com/TheAlgorithms/Python/blob/master/project_euler/problem_54/sol1.py#L367-L369)

```python
import os  # with other imports
...

def solution():
    script_dir = os.path.abspath(os.path.dirname(__file__))
    <file_name_without_extension> = os.path.join(script_dir, "<file_name>")
    with open(<file_name_without_extension> , "r") as file_hand:
        # ... your code ...
```

- Instead of `<file_name>` add the name of the file you want to import.
- Instead of `<file_name_without_extension>` add the variable name which is descriptive. If possible choose the name as the file name but without the extension.
- Then using `with` do whatever you want to do with the contents of the file.

I hope this explains your doubt. If you still have any problem, don't hesitate to ask.

berry-thawson:
Thanks @dhruvmanila 
This is helpful

appsx13:
Hi, can we add solution to any problem?

[...]

PR #3018, opened 2020-10-08
Hacktoberfest: Add project euler problem 51

Hyftar:
### **Describe your change:**

This pull request adds the Project Euler problem 51

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/project_euler/problem_51/sol1.py
+++ b/project_euler/problem_51/sol1.py
@@ -1,0 +1,106 @@
+from collections import Counter
+
+"""
+https://projecteuler.net/problem=51
+Prime digit replacements
+Problem 51
+
+By replacing the 1st digit of the 2-digit number *3, it turns out that six of
+the nine possible values: 13, 23, 43, 53, 73, and 83, are all prime.
+
+By replacing the 3rd and 4th digits of 56**3 with the same digit, this 5-digit
+number is the first example having seven primes among the ten generated numbers,
+yielding the family: 56003, 56113, 56333, 56443, 56663, 56773, and 56993.
+Consequently 56003, being the first member of this family, is the smallest prime
+with this property.
+
+Find the smallest prime which, by replacing part of the number (not necessarily
+adjacent digits) with the same digit, is part of an eight prime value family.
+"""
+
+
+def sieve(n: int) -> list:
+    """
+    Sieve of Erotosthenes
+    Function to return all the prime numbers up to a certain number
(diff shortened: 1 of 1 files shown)


project_euler/problem_51/sol1.py
```diff
@@ -0,0 +1,106 @@
+from collections import Counter
```
NSachida, line 1: Problem=151

NSachida (approved):
First prime in the eight prime family is 121313

Hyftar:
> By replacing the first degit of 51 which is (5)
> *3 =15 
> Joined with the last degit 1
> That's 151

I don't understand what you're trying to explain

NSachida (approved):
You can find the compete source code [Here](www.mathblog.dk/files/euler/Problem51.cs)

Hyftar:
> 
> 
> You can find the compete source code [Here](www.mathblog.dk/files/euler/Problem51.cs)

I don't understand what you're trying to explain, my code achieved the same result.

Hyftar:
This PR helping the issue #2695

project_euler/problem_51/sol1.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return replacements
+
+
+def solution() -> int:
+    """
+    Returns the solution of the problem
+
+    >>> solution()
+    121313
```
dhruvmanila, line 81: Please don't provide a test for answers as it will be checked by this [script](https://github.com/TheAlgorithms/Python/blob/master/project_euler/validate_solutions.py).

Make a positional argument and keep the default value of that as the question input (in this case it might be `8`?). Provide `doctest` for different inputs to the `solution()` function as we want to test the function.

project_euler/problem_51/sol1.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    numbers_checked = set()
+    primes = set(sieve(1_000_000))
+
+    # Filter primes with less than 3 replaceable digits
+    primes = {x for x in primes if len(str(x)) - len(set(str(x))) >= 3}
```
dhruvmanila, line 87:
```suggestion

    # Filter primes with less than 3 replaceable digits
    primes = {x for x in set(sieve(1_000_000)) if len(str(x)) - len(set(str(x))) >= 3}
```

dhruvmanila (changes requested):

dhruvmanila (approved):
Awesome work! Thank you for your contribution!

PR #3055, opened 2020-10-08
Update ternary_search.py

zyrch:
### **Describe your change:**

Reduces the overall complexity of the algorithm for discrete searches (like in a list)

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/searches/ternary_search.py
+++ b/searches/ternary_search.py
@@ -3,7 +3,7 @@ This is a type of divide and conquer algorithm which divides the search space in
 3 parts and finds the target value based on the property of the array or list
 (usually monotonic property).
 
-Time Complexity  : O(log3 N)
+Time Complexity  : O(log2 N)
 Space Complexity : O(1)
 """
 import sys
@@ -31,8 +31,8 @@ def ite_ternary_search(A: List[int], target: int):
             if right - left < precision:
                 return lin_search(left, right, A, target)
 
-            oneThird = (left + right) / 2 - 1
-            twoThird =  (left + right) / 2 + 1
+            oneThird = (left + right) // 2
+            twoThird =  (left + right) // 2 + 1
 
             if A[oneThird] == target:
                 return oneThird
@@ -58,8 +58,8 @@ def rec_ternary_search(left: int, right: int, A: List[int], target: int):
         if right - left < precision:
             return lin_search(left, right, A, target)
 
-        oneThird = (left + right) / 2 - 1
-        twoThird = (left + right) / 2 + 1
+        oneThird = (left + right) // 2
+        twoThird = (left + right) // 2 + 1
 
         if A[oneThird] == target:
             return oneThird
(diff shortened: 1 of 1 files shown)


searches/ternary_search.py
```diff
@@ -31,8 +31,8 @@ def ite_ternary_search(A: List[int], target: int):
             if right - left < precision:
                 return lin_search(left, right, A, target)
 
-            oneThird = (left + right) / 3 + 1
-            twoThird = 2 * (left + right) / 3 + 1
+            oneThird = (left + right) / 2 - 1
+            twoThird =  (left + right) / 2 + 1
```
peteryao7, line 35: Ternary search is a divide and conquer algorithm that's supposed to split an array into three equal portions. Here, you're splitting it into two halves and a subarray of either 0 or 1 in the middle.

Using floating point division `/` over integer division `//` is causing an entirely different bug, but the index calculations themselves are fine.

Another way to calculate `oneThird` and `twoThird` that prevents overflow problems in other languages is by doing it this way:
`oneThird = left + (right - left) // 3`
`twoThird = right - (right - left) // 3`

searches/ternary_search.py
```diff
@@ -58,8 +58,8 @@ def rec_ternary_search(left: int, right: int, A: List[int], target: int):
         if right - left < precision:
             return lin_search(left, right, A, target)
 
-        oneThird = (left + right) / 3 + 1
-        twoThird = 2 * (left + right) / 3 + 1
+        oneThird = (left + right) / 2 - 1
+        twoThird = (left + right) / 2 + 1
```
peteryao7, line 62: Same as the iterative change above.

peteryao7 (changes requested):
The changes don't make the functions ternary search anymore.

searches/ternary_search.py
zyrch: By dividing the search space into three in ternary search we are able to find the minimum or maximum of convex functions, (if it is not convex then binary search is directly applicable). Now comparing three values in the search space allows us to eliminate one portion at a time. Now to increase the efficiency of the program we can decrease the distance two intervals and still get the same result, this method is also known as dichotomous search.

peteryao7: Binary and ternary search are examples of dichotomous search. What ternary search lets you do over binary search is omit two sections, or 2/3, of the search space on each iteration (as opposed to 1/2 in binary search), which is correctly done at the end of the loop:

```
elif target < A[oneThird]:
    right = oneThird - 1
elif A[twoThird] < target:
    left = twoThird + 1
else:
    left = oneThird + 1
    right = twoThird - 1
```

The goal of the function is simply to find a specified value, so binary search will suffice, but by changing how the boundaries are calculated and changing the time complexity to O(log2(n)), you implemented binary search, not ternary search.

zyrch: You are correct I just realized we are searching in a sorted array, and my method is a better only in the case when ternary search is used to min or max of convex functions or data sets.

PR #3065, opened 2020-10-08
Created max_sum_sliding_window in Python/other

Kush1101:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/other/max_sum_sliding_window.py
+++ b/other/max_sum_sliding_window.py
@@ -19,12 +19,12 @@ def max_sum(array: List[int], k: int) -> int:
     24
     >>> k = 10
     >>> max_sum(arr,k)
-    Invalid input
-    -1
+    Traceback (most recent call last):
+        ...
+    ValueError: Invalid Input
     """
     if len(array) < k or k < 0:
-        print("Invalid input")
-        return -1
+        raise ValueError("Invalid Input")
     max_sum = -sys.maxsize
     current_sum = sum(array[:k])
     for i in range(len(array) - k):
--- a/other/max_sum_sliding_window.py
+++ b/other/max_sum_sliding_window.py
@@ -22,6 +22,10 @@ def max_sum(array: List[int], k: int) -> int:
     Traceback (most recent call last):
         ...
     ValueError: Invalid Input
+    >>> arr = [1, 4, 2, 10, 2, 13, 1, 0, 2]
+    >>> k = 4
+    >>> max_sum(arr, k)
+    27
     """
     if len(array) < k or k < 0:
         raise ValueError("Invalid Input")
--- a/other/max_sum_sliding_window.py
+++ b/other/max_sum_sliding_window.py
@@ -1,0 +1,43 @@
+"""
+Given an array of integer elements and an integer 'k', we are required to find the
+maximum sum of 'k' consecutive elements in the array.
+
+Instead of using a nested for loop, in a Brute force approach we will use a technique
+called 'Window sliding technique' where the nested loops can be converted to a single
+loop to reduce time complexity.
+"""
+import sys
+from typing import List
+
+
+def max_sum(array: List[int], k: int) -> int:
+    """
+    Returns the maximum sum of k consecutive elements
+    >>> arr = [1, 4, 2, 10, 2, 3, 1, 0, 20]
+    >>> k = 4
+    >>> max_sum(arr, k)
+    24
+    >>> k = 10
+    >>> max_sum(arr,k)
+    Invalid input
+    -1
+    """
+    if len(array) < k or k < 0:
+        print("Invalid input")
+        return -1
+    max_sum = -sys.maxsize
+    current_sum = sum(array[:k])
+    for i in range(len(array) - k):
+        current_sum = current_sum - array[i] + array[i + k]
+        current_sum = max(max_sum, current_sum)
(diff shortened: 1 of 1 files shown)


Kush1101:
@dhruvmanila @shellhub, Please review it.

other/max_sum_sliding_window.py
```diff
@@ -0,0 +1,43 @@
...
+    24
+    >>> k = 10
+    >>> max_sum(arr,k)
+    Invalid input
+    -1
+    """
+    if len(array) < k or k < 0:
+        print("Invalid input")
+        return -1
+    max_sum = -sys.maxsize
```
dhruvmanila, line 28: Can you tell me why is this useful?
How is an extremely low number be `max_size`?

other/max_sum_sliding_window.py
```diff
@@ -0,0 +1,43 @@
+"""
+Given an array of integer elements and an integer 'k', we are required to find the
+maximum sum of 'k' consecutive elements in the array.
+
+Instead of using a nested for loop, in a Brute force approach we will use a technique
+called 'Window sliding technique' where the nested loops can be converted to a single
+loop to reduce time complexity.
```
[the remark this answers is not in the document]
Kush1101: No, it isn't a problem form some platform. Its an application of the window sliding technique as mentioned

other/max_sum_sliding_window.py
Kush1101: You can see its `-max_size()`

dhruvmanila: If you want a negative infinity number then just use `float('-inf')`

Kush1101: This way is more commonly used, and so easy to understand

dhruvmanila: > You can see its -max_size()

What do you mean by this?

dhruvmanila:
I might be wrong but it seems that this is an incorrect algorithm.

As far as I understand you are finding the maximum sum of k consecutive integers in a list of integers which means:
```python
>>> arr = [1, 4, 2, 10, 2, 13, 1, 0, 2]
>>> k = 4
>>> max_sum(arr, k)  # This should give me 27
16
```
Your algorithm provides 16 as the answer but the answer clearly is 27. Your algorithm is just giving the sum of the last k integers.

Kush1101:
There was a bug in the program. I corrected it and added more tests. Now the algorithm works just the way it should.

other/max_sum_sliding_window.py
```diff
@@ -0,0 +1,47 @@
...
+    ValueError: Invalid Input
+    >>> arr = [1, 4, 2, 10, 2, 13, 1, 0, 2]
+    >>> k = 4
+    >>> max_sum(arr, k)
+    27
+    """
+    if len(array) < k or k < 0:
+        raise ValueError("Invalid Input")
+    max_sum = -sys.maxsize
+    current_sum = sum(array[:k])
```
dhruvmanila, line 33:
```suggestion
    current_sum = max_sum = sum(array[:k])
```
You can remove `import sys` as well.
Also, change either the variable name or the function name as they are the same.

dhruvmanila (changes requested):

dhruvmanila:
Also, can you take a look at #3094 ?
Thanks.

other/max_sum_sliding_window.py
dhruvmanila: You just added my test :)
Try to do some random testing. Maybe use `unittest` or `pytest` as `doctest` can get quite cumbersome when you want to do large scale testing.

Kush1101: Actually, its a more conventional way of doing the same thing. Beginners are told to do just like that, so I think they may find it a bit easier to understand if I leave it unchanged.

Kush1101: I am using random testing

dhruvmanila: Try to think yourself: what do you think is more readable and easier to understand?
1. Putting the first value as the reference
2. Putting an arbitrary negative value depending on the system which the program is running on as the reference

I would choose 1 all the time and I think beginners will understand 1 more easily.

And if you're referring to the block `if __name__ == "__main__":` then that's just a demo for your program, it's not actually testing anything.

Kush1101: About that first suggestion, most beginner-friendly solutions are written that way. I personally will find that as easier to read and understand. (Maybe its personal preferences). It just makes the code more beginner-friendly (IMO).
Yeah, I am not familiar with `unittest` and `pytest`. Will surely look it up and add it from now on if needed. 👍

dhruvmanila: Also, as `sys.maxsize` is a fix value, what would happen when the maximum sum value in the array is less than that value? That would give us the wrong answer.

dhruvmanila: If you want to use that approach then I recommend using `float('-inf')` otherwise a problem can arise as mentioned in the previous comment.

Kush1101: Yep. I made the change

dhruvmanila (approved):

PR #3100, opened 2020-10-09
Create distributed quick sort algorithm

zhe-thoughts:
This algorithm extends from quick_sort.py in this same directory. It uses Ray to make quick sort execute in distributed manner.

On my local machine, it has 3x speedup:

Sequential execution: 16.010289192199707
Distributed execution: 5.844448089599609

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ sklearn
 sympy
 tensorflow
 xgboost
+ray
--- a/sorts/quick_sort_distributed.py
+++ b/sorts/quick_sort_distributed.py
@@ -1,0 +1,82 @@
+"""
+This is a pure Python implementation of the quick sort algorithm,
+in distributed manner with Ray.
+
+NOTE: it extends from quick_sort.py in this same directory
+
+For doctests run following command:
+python -m doctest -v quick_sort_distributed.py
+or
+python3 -m doctest -v quick_sort_distributed.py
+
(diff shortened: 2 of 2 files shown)


zhe-thoughts:
NOTE: this algorithm extends from the existing quick_sort.py in the same directory. Please let me know how we usually handles this (should I import it? seems an overkill since the code is very simple)

seelam:
I gave this a try... all looks good but the WARNINGS are annoying and it will be good to limit the concurrency via an API or config call: 

ubuntu@seelam-ansible-host:~$ python quicksort.py 
....
Sequential execution: 20.511595964431763
2020-10-18 22:02:01,225	WARNING worker.py:1047 -- WARNING: 24 PYTHON workers have been started. This could be a result of using a large number of actors, or it could be a consequence of using nested tasks (see https://github.com/ray-project/ray/issues/3644) for some a discussion of workarounds.
....
Distributed execution: 10.588677167892456

In your case, since you are using the regular quick sort for <200K elements and total elements are 1M, only ~10 actors. When I changed the 200k to 20k, the distributed version took 4x longer than sequential version:
`Distributed execution: 84.4353358745575`

I guess it is important to figure out at what granularity of computation, you need to use the distributed version.

PR #3162, opened 2020-10-10
Alternative implementation of Lempel-Ziv-Welch compression algorithm

OmbraDiFenice:
### **Describe your change:**

* [x ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/compression/lzw/common.py
+++ b/compression/lzw/common.py
@@ -1,0 +1,49 @@
+import compression.lzw.symbol_generator as symbol_generator
+
+
+def generate_initial_dictionary():
+    """
+    Generate an initial dictionary to be used for encoder and decoder.
+
+    The dictionary holds all the encoding of single "letters" in the vocabulary that
+    the encoder and decoder can understand, and it must be the same for both the
+    encoder and decoder.
+
+    This implementation creates a dictionary with only the capital english letters.
+    It's useful for testing, but not much more.
+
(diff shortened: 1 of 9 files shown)


OmbraDiFenice:
This implements Lempel-Ziv-Welch compression algorithm. I took the algorithm description and example test case from Wikipedia (link in the test itself).

When I started working on this I didn't realize that there were already an implementation of this algorithm in the repo... anyway I had a quick look at the code and I think my implementation has some interesting aspcts:

- it only passes through the input once
- it uses a library implemented in C to store and manage the bit representation of the data
- it completely isolates the algorithm from the internal data structure used to represent the data

The first two points are about performances but I believe that the last is the most interesting one.

About the fields I didn't check in the checklist above:

- Type hints would be maybe useful (and I did use them in some places), but for some parameters they would be rather involved (for example the initial symbol mapping doesn't strictly requires the keys to be strings) and probabably not add much on the clarity of the code. Furthermore, one of the strengths of Python I believe is the ability to use ducktyping, and indeed I put some effort in explicitly defining and documenting the interface the Symbol class should have.

- I never used doctest before, but I am used to write unit tests for my code (and I did it here as well of course). From the Travis report you can see that my code coverage is above 95% for all the "production" code, so I think it's fine like this already :)

Issue #3163, opened 2020-10-10
Fix needed for Project Euler problem 234 sol1.py


dhruvmanila:
### Problem:

According to the [problem description](https://projecteuler.net/problem=234), the input value for the function `solution()` should be `999966663333`. Now, if we follow the control flow of the problem we see that there's going to be a loop that will have `999966663333` many iterations and for each iteration [this line](https://github.com/TheAlgorithms/Python/blob/master/project_euler/problem_234/sol1.py#L43) asks the user for input. As mentioned in CONTRIBUTING.md, we cannot ask for input and this solution will be asking for the input for `999966663333` times.

```python
for x in range(n):  # n = 999966663333 according to the problem description
    l = [i for i in input().split()]  # noqa: E741
```

### Proposed solution:

1. Remove the current solution and implement a new one.
2. Take a look at the current solution and fix it.

The second solution is highly unlikely as I am not able to deduce what is actually going on in the function and why do we even need to ask for the input. If anyone can figure it out, they can submit a PR for this.

***I propose the first solution.***

Once the issue is fixed, we can start running this [script](https://github.com/TheAlgorithms/Python/blob/master/project_euler/validate_solutions.py) which checks for all answers to Project Euler problems.

Harivind:
Hi @dhruvmanila, can I take up the issue?

dhruvmanila:
@fpringle Can you take a look at this? Thanks if you do.

Hyftar:
Hi, I will attempt to implement a new solution for this problem.

PR #3179, opened 2020-10-11
Fixes: #3163 Adds solution to problem 102

pastre:
### **Describe your change:**

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


pastre:
Is there anything missing for this PR to be merged? :)

53jk1 (approved):

poyea:
For `project_euler` you should have a solution class.

pastre:
@poyea what class should I create? As far as I can see, my submission perfectly matches all requirements in the [README](https://github.com/TheAlgorithms/Python/blob/master/project_euler/README.md) inside `project_euler`

poyea:
It requires default parameter for solution function, no? Also can the txt file be read locally?

pastre:
@poyea  `def solution() -> int` does not require any parameter. It can read from the local text file, but since the CI was failing at it, I chose to read from a remote URL

poyea:
ah ok. Although there're other ways, I think this is good enough.

poyea (approved):

poyea:
I'll review this in a moment

dhruvmanila:
I will include a guide in the README about how to use local files in the solution soon. As we are going to get the input from a file, there won't be any requirement of arguments to the `solution` function as the main goal to provide a parameter is to add the ability to test the function with different input.

As of now please take a look at this comment on how to use a local file in your algorithm: https://github.com/TheAlgorithms/Python/issues/2695#issuecomment-704237225

pastre:
@dhruvmanila checkout `solution.py` on 7e93dfe9026e16f07cc721d7a0d40fcf5d3d284a, I was trying to load locally before fetching from the web. If nothing changed on Travis, it will not work. If there's anything I can do, just let me know

dhruvmanila:
Please don't take things literally as specified. There should be a file extension to the actual path. What I meant by `<file name without extension>` is not to literally name it that but to make the variable name the same as that of the file so that it is easier to understand. You can use any name you like, that was just a framework.
```python
    script_dir = os.path.abspath(os.path.dirname(__file__))
    filepath = os.path.join(script_dir, "triangles_euler102.txt")
    with open(filepath, "r") as file_hand:
```

dhruvmanila (changes requested):
Please make the necessary changes regarding the file input.

Also, there are many problems in the type hints specified and there are functions that do not contain type hints. Please provide the return type hints as well. Use `List[type]` where you return or input `list`. Take a look at the official documentation regarding type hints.

PR #3215, opened 2020-10-12
Start running validate_solutions script for Travis CI

dhruvmanila:
### Removed print error_msgs at the end of the test:

This was done only to reduce the message clutter produced by 60
failing tests. As that is fixed, we can produce the traceback in
short form and allow pytest to print the captured error message
output at the end of the test.

### Start validate_solutions script for Travis CI

I am separating out the solution testing and doctest as validating
the solutions for the current number of solutions present is
taking 2 minutes to run which is purely due to the solutions provided.

### **Describe your change:**

* [x] Documentation change?

### **Checklist:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,14 @@ jobs:
         - pytest --doctest-modules --ignore=project_euler/ --durations=10 --cov-report=term-missing:skip-covered --cov=. .
     - name: Project Euler
       install:
-        - pip install pytest-cov pytest-subtests
-      before_script:
-        - pytest --tb=no --no-summary --capture=no project_euler/validate_solutions.py || true  # fail fast on wrong solution
+        - pip install pytest-cov
       script:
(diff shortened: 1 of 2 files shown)


dhruvmanila:
I'm also planning to write down guidelines regarding Project Euler solutions. There are more than 700 of them and people are starting to submit a lot of them now so we need some consistency to maintain throughout the directory. Some of the points I have in my mind are as follows:

- Consistency in directory names: `problem_001, problem_002, ..., problem_120, ..., problem_450, ...`. This way the directories will be sorted and not be jumbled as it is now.
- `solution()` function should be present and it should contain one or many positional arguments whose default values are the question input. This is proposed because the `doctest` should test the function and not check for the answer as that will be done by us using `validate_solutions.py`.
- `doctest` provided by the user should not check for the answer as that will increase the load on Travis. We are already validating the answer and so the user should test only for the function by providing values for the positional argument different from the question input, preferably small values.
- The problem link should be present in the module-level docstring and reference link to the topic used in the solution if any.

cclauss (approved):

PR #3230, opened 2020-10-12
Created reverse_string.py

ProgrammerSnigdho:
String reversing is complete

I created a reverse_string.py.
This file contains how to reverse a string.


Kush1101:
It's a bit too trivial in Python. We can accomplish the same just by doing `reversed_string = my_string[ : :-1]`. So I think its unnecessary to have a separate file for this. Thanks

PR #3238, opened 2020-10-12
Add skew heap data structure.

Phil9l:
### Add skew heap data structure.

A data structure that allows inserting a new value and to pop the smallest
values. Both operations take O(logN) time where N is the size of the structure.
  - Wiki: https://en.wikipedia.org/wiki/Skew_heap
  - Visualisation: https://www.cs.usfca.edu/~galles/visualization/SkewHeap.html

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


data_structures/heap/skew_heap.py
```diff
@@ -0,0 +1,120 @@
...
+    @staticmethod
+    def from_list(data: Iterable[T]) -> SkewHeap[T]:
+        """Get the sorted list from the heap. Heap will be cleared afterwards."""
+        result = SkewHeap()
+        for item in data:
+            result.insert(item)
+
+        return result
+
+    def to_sorted_list(self) -> List[T]:
```
cclauss, line 104: It might be cooler to define an [`.__init__()`](https://github.com/TheAlgorithms/Python/search?q=__iter__) method

data_structures/heap/skew_heap.py
```diff
@@ -0,0 +1,120 @@
...
+    A data structure that allows inserting a new value and to pop the smallest
+    values. Both operations take O(logN) time where N is the size of the structure.
+    - Wiki: https://en.wikipedia.org/wiki/Skew_heap
+    - Visualisation: https://www.cs.usfca.edu/~galles/visualization/SkewHeap.html
+
+    >>> SkewHeap.from_list([2, 3, 1, 5, 1, 7]).to_sorted_list()
+    [1, 1, 2, 3, 5, 7]
+
+    >>> sh = SkewHeap()
+    >>> sh.insert(1)
```
cclauss, line 58: Please test `.pop()` __before__ `insert()`ing any data to ensure that it behaves like `[].pop()`.

Insert data out of order and then test to ensure that `.to_sorted_list()` returns data in order.

Please add tests for `.clear()`, `.from_list()`, and `bool()`.

data_structures/heap/skew_heap.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+from typing import Generic, Iterable, List, Optional, TypeVar
+
+__all__ = ["SkewHeap"]
```
cclauss, line 7: Not really needed.

data_structures/heap/skew_heap.py
cclauss: Please do this...  It will be more Pythonic.

Phil9l: Ah, I see. I guess you mean `__iter__`, not `__init__`, right?

cclauss: Yes.

Phil9l: What do you think about this approach? 
https://github.com/TheAlgorithms/Python/pull/3241#discussion_r504559805

cclauss (approved):
Out of curiosity, how would you create a `SkewHeap.__len__()` method?

Phil9l:
> Out of curiosity, how would you create a `SkewHeap.__len__()` method?

- One way that requires `O(1)` memory would be to just add a field that keeps track of it. Insert increases it by one. Pop decreases. Clear sets to 0. Merging results as the sum of two heap sizes.  But in this case, we can't merge for example the left child of the heap with another heap, because we don't know the size of the children.
- Another one is to save the size of the subtree. Leaves are `1`s, any node is `left.size + right.size + 1`. And then it is easy to get the heap size just as the root size. But it requires `O(n)` additional memory to store `n` ints.

cclauss:
>  just add a field that keeps track of it

Not a fan...  Too many places for bugs to pop up.

Phil9l:
> Not a fan... Too many places for bugs to pop up.

Yep, agreed,  I always implement the second approach when I need it.

PR #3240, opened 2020-10-12
Added pytorch.py for irish dataset

Satyam-kumar-yadav:


Satyam-kumar-yadav:
Added pytorch.py for machine learning model

PR #3241, opened 2020-10-12
Add randomized heap.

Phil9l:
### Add randomized heap data structure

A data structure that allows inserting a new value and to pop the smallest values. Both operations take O(logN) time where N is the size of the structure.
Wiki: https://en.wikipedia.org/wiki/Randomized_meldable_heap

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


data_structures/heap/randomized_heap.py
```diff
@@ -0,0 +1,188 @@
...
+        >>> rh = RandomizedHeap([3, 1, 3, 7])
+        >>> rh.clear()
+        >>> rh.pop()
+        Traceback (most recent call last):
+            ...
+        IndexError: Can't get top element for the empty heap.
+        """
+        self._root = None
+
+    def to_sorted_list(self) -> List[T]:
```
cclauss, line 154: https://github.com/TheAlgorithms/Python/pull/3238#discussion_r503505422

data_structures/heap/randomized_heap.py
```diff
@@ -0,0 +1,188 @@
...
+        >>> rh.to_sorted_list()
+        [1, 3, 3, 7]
+        """
+        result = []
+        while self:
+            result.append(self.pop())
+
+        return result
+
+    def __bool__(self) -> bool:
```
cclauss, line 168: Please move the `__methods__()` to just under `__init__()` for readability.  It is easier to read all dunder methods before reading the other methods.

cclauss (changes requested):

data_structures/heap/randomized_heap.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        Returns sorted list containing all the values in the heap.
+
+        >>> rh = RandomizedHeap([3, 1, 3, 7])
+        >>> rh.to_sorted_list()
+        [1, 3, 3, 7]
+        """
+        result = []
+        while self:
+            result.append(self.pop())
```
cclauss, line 164: It would surprise me as a user of this class that requesting a list would remove all data from the heap.

Phil9l: Do you suggest to add all the items back to the heap at the end of the method?

cclauss: I would suggest that you replace `.to_sorted_list()` with `.__iter__()` and that should _not_ use `.pop()` but should just yield values as it walks thru the heap.  Please search this repo for instances of `__intr__()` to see how this is done.

data_structures/heap/randomized_heap.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class RandomizedHeap(Generic[T]):
+    """
+    A data structure that allows inserting a new value and to pop the smallest
+    values. Both operations take O(logN) time where N is the size of the
+    structure.
+    Wiki: https://en.wikipedia.org/wiki/Randomized_meldable_heap
+
+    >>> RandomizedHeap([2, 3, 1, 5, 1, 7]).to_sorted_list()
```
cclauss, line 56:
```suggestion
    >>> list(RandomizedHeap([2, 3, 1, 5, 1, 7]))
```
Should pass if `.__iter__()` is properly implemented.

data_structures/heap/randomized_heap.py
Phil9l: I mean I know how `__iter__` works in python. But we can't iterate over heap to get sorted values. In BST we know where is the next element but in the heap, we can only pop the smallest. We can't get the second element.

I can do it like this since `__iter__` can work in time larger than O(1). And we don't need the structure to be thread-safe, we can modify it. So poping and pushing back works.

WDYT?
```
    def __iter__(self) -> Iterator[T]:
        """
        Returns sorted list containing all the values in the heap.

        >>> sh = RandomizedHeap([3, 1, 3, 7])
        >>> list(sh)
        [1, 3, 3, 7]
        """
        result = []
        while self:
            result.append(self.pop())

        # Pushing items back to the heap not to clear it.
        for item in result:
            self.insert(item)

        return iter(result)
```

cclauss: Good enough for now.  Thanks for doing this!

cclauss (approved):
Well done!!

PR #2399, opened 2020-09-06
octal_to_decimal converter

mohammadreza490:
octal to decimal converter

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [x] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #3243, opened 2020-10-12
Modified octal_to_decimal

cybov:
### **Describe your change:**

 * [ ] Add an algorithm?

* [x] Fix a bug or typo in an existing algorithm? #2399
* [ ] Documentation change?

### **Checklist:**


conversions/octal_to_decimal.py
```diff
@@ -9,26 +9,40 @@ def oct_to_decimal(oct_string: str) -> int:
     >>> oct_to_decimal("-45")
     -37
     >>> oct_to_decimal("2-0Fm")
+    Traceback (most recent call last):
```
cclauss, line 12: These are optional and do not add much value

conversions/octal_to_decimal.py
```diff
@@ -9,26 +9,40 @@ def oct_to_decimal(oct_string: str) -> int:
...
     >>> oct_to_decimal("")
+    Traceback (most recent call last):
+    ...
     ValueError: Empty string value was passed to the function
     >>> oct_to_decimal("19")
+    Traceback (most recent call last):
+    ...
     ValueError: Non-octal value was passed to the function
     """
+    # Strip oct_string of whitespaces
```
cclauss, line 24: This comment repeats what the code says.  Comments should not be about _what_ the code is doing, but should be about _why_ we are doing something.  Why are we removing whitespace?  Why are we removing the minus sign.

conversions/octal_to_decimal.py
```diff
@@ -9,26 +9,40 @@ def oct_to_decimal(oct_string: str) -> int:
...
     oct_string = str(oct_string).strip()
     if not oct_string:
-        raise ValueError("Empty string was passed to the function")
+        raise ValueError("Empty string value was passed to the function")
+
+    # Check if oct_string is a negative value
```
cclauss, line 29: This comment repeats what the code says.

conversions/octal_to_decimal.py
```diff
@@ -9,26 +9,40 @@ def oct_to_decimal(oct_string: str) -> int:
...
     is_negative = oct_string[0] == "-"
     if is_negative:
+        # Remove (-) from oct_string
```
cclauss, line 32: This comment repeats what the code says.

conversions/octal_to_decimal.py
```diff
@@ -9,26 +9,40 @@ def oct_to_decimal(oct_string: str) -> int:
...
         oct_string = oct_string[1:]
-    if not all(0 <= int(char) <= 7 for char in oct_string):
+
+    # check if oct_string is an octal value and convert
+    if oct_string.isdecimal() and all(0 <= int(char) <= 7 for char in oct_string):
```
cclauss, line 36: These are redundant tests.  The first one cannon pass if the second one does not pass.  The original formulation is more consise.

conversions/octal_to_decimal.py
```diff
@@ -9,26 +9,40 @@ def oct_to_decimal(oct_string: str) -> int:
...
+        decimal_number = 0
+        for char in oct_string:
+            decimal_number = 8 * decimal_number + int(char)
+        if is_negative:
+            decimal_number = -decimal_number
+        return decimal_number
+    else:
+        # else raise exception
```
cclauss, line 44: This comment repeats what the code says.

cclauss (commented):
I am not seeing an improvement in these changes.

conversions/octal_to_decimal.py
cybov: Thanks for reviewing. In the original, int(char) in the iterable passed to the all() function returns error that is not handled by the ValueError raised when char is not an integer value. It's why I modified it so.

cclauss: A ValueError is still raised.  This is why I encourage doctests.  It is not necessary that we raise all exceptions ourselves.  It is only that an appropriate exception is raised when garbage data is received.
```python
>>> int('a')
  File "<string>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'a'
```

conversions/octal_to_decimal.py
cybov: I added them because the doctest failed without them.

conversions/octal_to_decimal.py
cybov: Ok

conversions/octal_to_decimal.py
cybov: Ok

PR #3253, opened 2020-10-13
Improve validate solutions script & fix pre-commit error

dhruvmanila:
### **Reason:**

I wanted to add the ability to log the time it took for each solution. It started by adding this functionality to the existing script but that got quite cumbersome as you can see in my previous commit on this branch. Then I got an aha moment to change the way I am testing the solution which leads me to this PR.

### **Describe your change:**

I have completely changed the way I was performing the tests. Instead of parameterizing the problem numbers and expected output, I will parametrize the solution file path. This way we won't need the `subtests` dependency. This also means that we are only testing the files present in the directory and no need to skip for solutions not submitted. Now, we can also look at the time it takes for each solution with the help of the `--durations` flag from `pytest`.

Steps:
- Collect all the solution file paths
- Convert the paths into a Python module
- Call solution on the module
- Assert the answer with the expected results

For getting the expected value, it was needed to convert the JSON list object to a Python dictionary object 
which required changing the JSON file itself.

* [x] Improve an algorithm

### **Checklist:**


cclauss (approved):

dhruvmanila:
Wow, thanks for such a quick response!

cclauss:
This is great work!

PR #3287, opened 2020-10-14
String Check.py

aniketprasadgit:
1. Insert the given string in a set
2. Check if the set characters consist of 1 and/or 0 only.

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/String Check.py
+++ b/String Check.py
@@ -1,0 +1,34 @@
+# Python program to check 
+# if a string is binary or not 
+
+# function for checking the 
+# string is accepted or not 
+def check(string) : 
+
+	# set function convert string 
+	# into set of characters . 
+	p = set(string) 
+
+	# declare set of '0', '1' . 
+	s = {'0', '1'} 
+
+	# check set p is same as set s 
+	# or set p contains only '0' 
+	# or set p contains only '1' 
+	# or not, if any one condition 
+	# is true then string is accepted 
+	# otherwise not . 
+	if s == p or p == {'0'} or p == {'1'}: 
+		print("Yes") 
+	else : 
+		print("No") 
+
+
+		
+# driver code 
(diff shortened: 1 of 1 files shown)


cozek:
you can also check with ( perhaps within a `try` and `except`)
```python
int(some_string,2)
```
Also, tests are failing. Please read the contributions guidelines and rectify.

String Check.py
```diff
@@ -0,0 +1,34 @@
+# Python program to check 
+# if a string is binary or not 
+
+# function for checking the 
+# string is accepted or not 
+def check(string) :
```
mateuszz0000, line 6: Better to return `bool` instead print True/False

String Check.py
```diff
@@ -0,0 +1,34 @@
+# Python program to check 
+# if a string is binary or not 
+
+# function for checking the 
+# string is accepted or not 
+def check(string) : 
+
+	# set function convert string
```
mateuszz0000, line 8: PLease add doctests and typehints

mateuszz0000 (changes requested):

String Check.py
```diff
@@ -0,0 +1,34 @@
...
+	# declare set of '0', '1' . 
+	s = {'0', '1'} 
+
+	# check set p is same as set s 
+	# or set p contains only '0' 
+	# or set p contains only '1' 
+	# or not, if any one condition 
+	# is true then string is accepted 
+	# otherwise not . 
+	if s == p or p == {'0'} or p == {'1'}:
```
mateuszz0000, line 21:
```
try:
    int(string, 2)
    return True
except:
   return False
```

mateuszz0000 (changes requested):

PR #345, opened 2018-09-15
different take on bubble sort

ayush246:
this sorting algorithm sorts an array using the principle of bubble sort , but does it both from left to right and right to left , hence i decided to call it "double sort"

PR #3336, opened 2020-10-15
Added a Backtracking Solution for Project Euler Problem #345

nan011:
### **Describe your change:**
I added a backtracking solution for Project Euler problem 345 

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


nan011:
@Kush1101 @dhruvmanila Please review this pull request :), thank you!

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,177 @@
...
+813 883 451 509 615  77 281 613 459 205 380 274 302  35 805
+"""
+from functools import cmp_to_key
+
+
+def solve(
+    matrix, enhanced_matrix=None, i=0, has_booked=None, value=0, max_value=float("-inf")
+):
+    """
+    The approach is using brute force through recursive but cut some tree before reach
```
dhruvmanila, line 41: Please move the solution explanation to module-level docstring.

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,177 @@
...
+821 461 843 513  17 901 711 993 293 157 274  94 192 156 574
+ 34 124   4 878 450 476 712 914 838 669 875 299 823 329 699
+815 559 813 459 522 788 168 586 966 232 308 833 251 631 107
+813 883 451 509 615  77 281 613 459 205 380 274 302  35 805
+"""
+from functools import cmp_to_key
+
+
+def solve(
+    matrix, enhanced_matrix=None, i=0, has_booked=None, value=0, max_value=float("-inf")
```
dhruvmanila, line 38: Type hints?

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,177 @@
...
+    [414, 456, 310, 312, 798, 104, 566, 520, 302, 248, 694, 976, 430, 392, 198],
+    [184, 829, 373, 181, 631, 101, 969, 613, 840, 740, 778, 458, 284, 760, 390],
+    [821, 461, 843, 513, 17, 901, 711, 993, 293, 157, 274, 94, 192, 156, 574],
+    [34, 124, 4, 878, 450, 476, 712, 914, 838, 669, 875, 299, 823, 329, 699],
+    [815, 559, 813, 459, 522, 788, 168, 586, 966, 232, 308, 833, 251, 631, 107],
+    [813, 883, 451, 509, 615, 77, 281, 613, 459, 205, 380, 274, 302, 35, 805],
+]
+
+
+def solution(matrix: list = matrix2) -> int:
```
dhruvmanila, line 163: This function doesn't do anything except for calling the `solve` function. So just remove this function and rename `solve` -> `solution`. Read through the solution guidelines (project_euler/README.md)

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    Returns the maximum possible sum of matrix values
+    with no value share a same row or column.
+
+    >>> solution(matrix1)
+    3315
+    >>> solution(matrix2)
+    13938
```
dhruvmanila, line 171:
```suggestion
```
As mentioned, do not test for the answer.

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    return solve(matrix)
+
+
+if __name__ == "__main__":
+    print(solution())
```
dhruvmanila, line 177:
```suggestion
if __name__ == "__main__":
    print(f"{solution() = }")
```

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,177 @@
...
+                )
+
+            # Ensure current node is not booked anymore and others can use it
+            has_booked[enhanced_matrix[i][j][1]] = False
+
+    return max_value
+
+
+# Problem inputs
+matrix1 = [
```
dhruvmanila, line 136: These are constants so they will be in uppercase and comes after all the imports.

dhruvmanila (changes requested):

project_euler/problem_345/sol1.py
nan011: So I only test for the first matrix? it seems other problems test for the answer

project_euler/problem_345/sol1.py
nan011: I thought it's only for the solution method also I don't notice it on README.md, okay I will do

nan011:
@Kush1101 @dhruvmanila Please review this PR again :), thanks so much for the time you spent on the last review.

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,176 @@
...
+    [184, 829, 373, 181, 631, 101, 969, 613, 840, 740, 778, 458, 284, 760, 390],
+    [821, 461, 843, 513, 17, 901, 711, 993, 293, 157, 274, 94, 192, 156, 574],
+    [34, 124, 4, 878, 450, 476, 712, 914, 838, 669, 875, 299, 823, 329, 699],
+    [815, 559, 813, 459, 522, 788, 168, 586, 966, 232, 308, 833, 251, 631, 107],
+    [813, 883, 451, 509, 615, 77, 281, 613, 459, 205, 380, 274, 302, 35, 805],
+]
+
+
+def solution(
+    matrix: list = MATRIX_2,
```
dhruvmanila, line 85: Please specify the type of elements inside the list like so: `List[type]`

You will have to import this: `from typing import List`

Make the same changes to other parameters of list type.

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,176 @@
...
+                            in tuple form and each value is paired
+                            with its column index
+    i                   --  i-th row
+    booked_list         --  List of j's flag
+    value               --  Accumulative value from parent to current node
+    max_value           --  Maximum value is stored right now by the tree
+    """
+
+    # Get length of a row and length of a column, respectively denoted by m and n
+    m, n = len(matrix), len(matrix[0])
```
dhruvmanila, line 111: Descriptive names would be awesome: `row_length`, `column_length`

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Create list of j's flag at the first time
+    if booked_list is None:
+        booked_list = [False for _ in range(m)]
```
dhruvmanila, line 115:
```suggestion
        booked_list = [False] * m
```

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,176 @@
...
+                )
+
+            # Ensure current node is not booked anymore and others can use it
+            booked_list[enhanced_matrix[i][j][1]] = False
+
+    return max_value
+
+
+if __name__ == "__main__":
+    print(f"{solution()}")
```
dhruvmanila, line 176:
```suggestion
    print(f"{solution() = }")
```

project_euler/problem_345/sol1.py
```diff
@@ -0,0 +1,176 @@
...
+    # Create enhanced matrix at the first time
+    if enhanced_matrix is None:
+        # Pair value with its column index
+        enhanced_matrix = [[(row[j], j) for j in range(len(row))] for row in matrix]
+
+        # Sort each row by considering the value as comparison
+        for k in range(len(enhanced_matrix)):
+            enhanced_matrix[k] = sorted(
+                enhanced_matrix[k],
+                key=cmp_to_key(lambda a, b: 1 if a[0] < b[0] else -1),
```
dhruvmanila, line 126: Can you explain the use case of `cmp_to_key` in here? According to the Contributing guidelines we avoid the use of `lambda` so I am just trying to understand what's going on in here.

dhruvmanila (changes requested):
Small changes required and a question asked.

project_euler/problem_345/sol1.py
nan011: Just to make it clear, so each row in enchance_matrix is a list of a tuple and each tuple contains two elements. Those codes tell to sort each row in ascending order by considering the first element as a comparison. I'm not sure what cmp_to_key is doing but in order to sort by looking at the first element only, I need to use lambda and ```cmp_to_key``` to make it work in the ```sorted``` function. I don't know about another way to replace this behavior, but if you have one let me know so I can change it and will follow the guideline. Thank you!

nan011: I just realized that NumPy can do the work, yea I think I will use NumPy for the array

project_euler/problem_345/sol1.py
nan011: My bad, I forgot

project_euler/problem_345/sol1.py
nan011: Alright

project_euler/problem_345/sol1.py
nan011: Noted

project_euler/problem_345/sol1.py
nan011: I'm afraid I'm doing it wrong but seems the suggestion has a syntax error. Can you explain what you are going to achieve here?

project_euler/problem_345/sol1.py
nan011: Fixed: I will use NumPy for each row only, not the whole matrix

project_euler/problem_345/sol1.py
nan011: @dhruvmanila

dhruvmanila: You must be running a Python version < 3.8.
Try running this command on the terminal and see what output you get: `python -V` or `python3 -V`

dhruvmanila:
I think you just included a dependency which is not present in any of the Project Euler solutions: `numpy`

I will have to change the test first for you to be able to use the dependency as currently, we only use the `requirements.txt` file in non-project-euler algorithms.

project_euler/problem_345/sol1.py
nan011: I see, will try later

nan011:
@dhruvmanila Yes that would be helpful, thanks. For this problem, I only used this for sorting but actually, there are some Euler Problems that require ```numpy``` to get better performance rather than using ```list```. For example, generate prime numbers by sieve approach. It's a slightly different time for 1e6 million generations compare to ```list```, it's like ~100ms, but some Euler Problems need to compute until 1e12 or more. Check [this](https://stackoverflow.com/questions/2068372/fastest-way-to-list-all-primes-below-n/3035188) out for the performance comparison of the generation of prime numbers between ```list``` and ```numpy```

PR #3347, opened 2020-10-15
Project euler problem 1 solution

SaiSudhaV:
It returns the sum of all the multiples of 3 or 5 below the given integer.


dhruvmanila:
As mentioned, we are only accepting solutions that do not exist in this repository. Thank you for your contribution.

PR #3350, opened 2020-10-15
Create euclidean_distance.py

ac-tam:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


maths/distance_formula.py
```diff
@@ -0,0 +1,15 @@
+def distance_formula(x1, y1, x2, y2):
```
cozek, line 1: add type hints.
Change function name and file name to "euclidean_distance" or whatever distance formula this is. Update the comments to reflect this change.

maths/distance_formula.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Calculate the distance between the two coordinates (x1,y1) and (x2,y2).
+    >>> distance_formula(0, 0, 2, 2)
+    2.8284271247461903
+    >>> distance_formula(1, 0, 5, 2)
+    4.47213595499958
+    >>> distance_formula(0, 0, 5, 0)
+    5.0
+    """
+    return ((y2 - y1) ** 2 + (x2 - x1) ** 2) ** (1 / 2)
```
cozek, line 11: Can you generalize this calculate the same for vectors of any dimensions instead of just 2-dimensional coordinates?
You can use numpy to implement it. 
Or you can call NumPy's inbuilt distance function. (which I don't recommend)

cozek (changes requested):

maths/distance_formula.py
ac-tam: I generalized it for any 2 vectors.

ac-tam (commented):
Updated name and generalized

maths/euclidean_distance.py
```diff
@@ -0,0 +1,22 @@
...
+    >>> euclidean_distance(np.array([0, 0, 0]), np.array([2, 2, 2]))
+    3.4641016151377544
+    >>> euclidean_distance(np.array([1, 2, 3, 4]), np.array([5, 6, 7, 8]))
+    8.0
+    """
+
+    sumSquared = 0 
+    for i in range(len(v1)):
+        sumSquared += (v2[i] - v1[i]) ** 2
+    return sumSquared ** (1 / 2)
```
cozek, line 17: You don't need to loop. You can use a one-liner. 
`numpy.sqrt(numpy.sum((x-y)**2))`

You can also use an equivalent linear algebra trick
```
diff = point_a - point_b
numpy.dot(diff.T, diff) ** (1/2)
```

I will leave it to you verify the correctness of the above. xD Don't take my math blindly.

maths/euclidean_distance.py
```diff
@@ -0,0 +1,22 @@
...
+
+def euclidean_distance(v1, v2) -> double:
+    """
+    Calculate the distance between the two = endpoints of two vectors, v1 and v2.
+    >>> euclidean_distance(np.array([0, 0]), np.array([2, 2]))
+    2.8284271247461903
+    >>> euclidean_distance(np.array([0, 0, 0]), np.array([2, 2, 2]))
+    3.4641016151377544
+    >>> euclidean_distance(np.array([1, 2, 3, 4]), np.array([5, 6, 7, 8]))
+    8.0
```
cozek, line 11: You can test if your implementation is correct by using an actual library function.
```euclidean_distance(v1,v2) == numpy.linalg.norm(v1-v2)```
I use this trick to check if my implementation of some function is equivalent to an actual tried and tested library function. If you do so, you will need to use a flooring function since it is impossible to match every point after the decimal.

File: maths/euclidean_distance.py [hunk 5]
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+
+def euclidean_distance(v1, v2) -> double:
```
cozek, line 3: Your typing is not accurate. You cannot directly use double like this.
Read more about typing here https://docs.python.org/3/library/typing.html

How about 

```python
def euclidean_distance(vector_a: np.ndarray, vector_b: np.ndarray ) -> np.float64:
```

If you would also like to support lists and tuples, the typing would get a little complicated.
And you would need to add corresponding doctests as well.
```python
from typing import Union, List, Tuple


vector_tuple = Tuple[Union[float,int]]
vector_list = List[Union[float,int]]
Vector = Union[vector_tuple, vector_list ,np.ndarray]
VectorOut = Union[np.float64, int, float]

def euclidean_distance(vector_a: Vector, vector_b: Vector ) -> VectorOut: 
# convert vector_a and vector_b to np.array and carry out calculations
```

@cclauss Could you please check if this typing is ok or perhaps we can do better?

cozek (changes requested):

cozek:
Hi @ac-tam . You have made good progress. Just read the Travis error logs and make your code comply with the contribution guidelines.

ac-tam (commented):
Added suggestions

maths/euclidean_distance.py
```diff
@@ -0,0 +1,27 @@
...
+
+Vector = Union[List[float], List[int], Tuple[float], Tuple[int], np.ndarray]
+VectorOut = Union[np.float64, int, float]
+
+
+def euclidean_distance(vector_1: Vector, vector_2: Vector) -> VectorOut:
+    """
+    Calculate the distance between the two endpoints of two vectors.
+    A vector is defined as a list, tuple, or numpy 1D array.
+    2.8284271247461903
```
cozek, line 13: Perhaps you meant to add a test case here with the ```2.8284271247461903```

cclauss:
This is looking cool!  Please add or delete the doctest for 2.8284271247461903 as suggested by @cozek.  Let’s try to benchmark a numpy version of this algorithm against a non-numpy version.  Please rename the current function to `euclidean_distance_np()` and create a new `euclidean_distance()` function that has the same function signature but does not use numpy.  Then we can add a `timeit` benchmark to see the performance difference.

maths/euclidean_distance.py
```diff
@@ -0,0 +1,27 @@
+from typing import List, Tuple, Union
+
+import numpy as np
+
+Vector = Union[List[float], List[int], Tuple[float], Tuple[int], np.ndarray]
```
cclauss, line 5:
```suggestion
from typing import Iterable Union

import numpy as np

Vector = Union[Iterable[float], Iterable[int], np.ndarray]
```
Would mypy consider a np.ndarray considered an Iterable

ac-tam (commented):
done

ac-tam (commented):
added no numpy function

cozek (approved):
Looks good from my side. 
You might want to add some benchmarks on the np and non_np versions as @cclauss suggested.
This might be helpful 
https://realpython.com/python-timer/#creating-a-python-timer-decorator

cclauss:
https://docs.python.org/3.9/library/timeit.html is what is suggested because there are several examples in this repo https://github.com/TheAlgorithms/Python/search?q=timeit

53jk1 (approved):

poyea (approved):

cclauss (changes requested):
Needs a `benchmark()` function that uses timeit.

ac-tam (commented):
Hi, could you please review the PR? IT has been marked as stale and I added the benchmark function. Thanks!

maths/euclidean_distance.py
```diff
@@ -0,0 +1,65 @@
...
+    A vector is defined as a list, tuple, or numpy 1D array.
+    >>> euclidean_distance((0, 0), (2, 2))
+    2.8284271247461903
+    >>> euclidean_distance([1, 2, 3, 4], [5, 6, 7, 8])
+    8.0
+    """
+    the_sum = 0
+    for i in range(len(vector_1)):
+        the_sum += (vector_2[i] - vector_1[i]) ** 2
+    return the_sum ** (1 / 2)
```
cclauss, line 37: This function should use two Python builtin functions... sum() and zip().

maths/euclidean_distance.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+    )
+    print(
+        timeit(
+            "euclidean_distance([1, 2, 3], [4, 5, 6])",
+            setup="from __main__ import euclidean_distance",
+            number=10000,
+        )
+    )
+
```
dhruvmanila, line 57: Please put this directly in the `if __name__ == "__main__":` block.

dhruvmanila: Also, use the `globals()` dictionary in the `timeit` function like so: `timeit(..., globals=globals())` instead of the `setup`  parameter.

dhruvmanila: One more thing: Maybe print out what those numbers actually mean?
```
$ python euclidean_distance.py
0.019355665999999994
0.12133019300000003
```
This would be pretty confusing when someone runs this script.

maths/euclidean_distance.py
```diff
@@ -0,0 +1,61 @@
...
+    >>> euclidean_distance_no_np((0, 0), (2, 2))
+    2.8284271247461903
+    >>> euclidean_distance_no_np([1, 2, 3, 4], [5, 6, 7, 8])
+    8.0
+    """
+    return sum([(y - x) ** 2 for (y, x) in list(zip(vector_1, vector_2))]) ** (1 / 2)
+
+
+if __name__ == "__main__":
+  def benchmark() -> None:
```
dhruvmanila, line 38: I am sorry if I was not clear but I meant to remove the function and just keep the code.

maths/euclidean_distance.py
```diff
@@ -0,0 +1,62 @@
...
+def euclidean_distance_no_np(vector_1: Vector, vector_2: Vector) -> VectorOut:
+    """
+    Calculate the distance between the two endpoints of two vectors without numpy.
+    A vector is defined as a list, tuple, or numpy 1D array.
+    >>> euclidean_distance_no_np((0, 0), (2, 2))
+    2.8284271247461903
+    >>> euclidean_distance_no_np([1, 2, 3, 4], [5, 6, 7, 8])
+    8.0
+    """
+    return sum([(y - x) ** 2 for (y, x) in list(zip(vector_1, vector_2))]) ** (1 / 2)
```
cclauss, line 34:
```suggestion
    return sum((y - x) ** 2 for y, x in zip(vector_1, vector_2)) ** (1 / 2)
```

cclauss (approved):
Nice work!  Cool to see the benchmark.

PR #3378, opened 2020-10-16
Create GitHub action only for Project Euler

dhruvmanila:
### **Describe your change:**

Move Project Euler tests to GitHub actions.

### **Reason:**

Testing the Project Euler directory even if there's no change in any of its files is a pretty waste of time and energy. Travis CI doesn't provide the ability to test only if some files are changed but GitHub actions do. Both jobs will only run if any of the files within the `project_euler` directory is changed and this action file.

### **Status:**

- [x] GitHub action performing as expected
  - [x] Run only if the specified files change
  - [x] Don't run if any other files change
- [x] Remove `#` from `.travis.yml`


dhruvmanila:
### **Proposal:**

We can do the same thing where the job without Project Euler runs only if any of the files in the directories except for `project_euler` is changed.

cclauss (approved):
Awesome work as always.

PR #3408, opened 2020-10-17
Add solution to problem 81 - project euler

ayoubc:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


project_euler/problem_081/sol1.py
```diff
@@ -0,0 +1,48 @@
...
+    >>> solution()
+    427337
+    """
+    with open(os.path.join(os.path.dirname(__file__), filename), "r") as file:
+        data = file.read().strip().split("\n")
+
+    grid = []
+    for line in data:
+        row = line.split(",")
+        grid.append(list(map(int, row)))
```
cclauss, line 31: OPTIONAL:
```suggestion
def solution(filename: str = "matrix.txt") -> int:
    """
    Returns the minimal path sum from the top left to the bottom right of the matrix.
    >>> solution()
    427337
    """
    with open(os.path.join(os.path.dirname(__file__), filename), "r") as in_file:
        data = in_file.read()

    grid = [[int(cell) for cell in row.split(",")] for row in data.strip().splitlines()]
    dp = [[0 for cell in row] for row in grid]
    n = len(grid[0])
```
1. Remove `n` because we can calculate it by looking at the length of row 0 of the grid.
    * This the use of our algorithm and allows us to be flexible to handle other datasets.
    * This reduces the chances that the caller will give us bad data. (n = 79, n = 81, n = -5)
2. Used `in_file` instead of `file` to make it easier to find bugs.
    * `in_file.write()` would be a mistake because we can not write to a file opened in read mode.
    * `out_file.read()` would be a mistake because we can not read a file opened in write mode.
3. Do not strip() and split() the data right away because it can be useful to keep the raw data available.
4. Used a list comprehension inside a list comprehension to build the matrix.
    * As discussed in CONTRIBUTING.md, this repo (and Py3 in general) prefers comprehensions over `list(map())`

cclauss (approved):

project_euler/problem_081/sol1.py
```diff
@@ -0,0 +1,48 @@
...
+    for i in range(1, n):
+        dp[0][i] = grid[0][i] + dp[0][i - 1]
+    for i in range(1, n):
+        dp[i][0] = grid[i][0] + dp[i - 1][0]
+
+    for i in range(1, n):
+        for j in range(1, n):
+            dp[i][j] = grid[i][j] + min(dp[i - 1][j], dp[i][j - 1])
+
+    return dp[n - 1][n - 1]
```
cclauss, line 44:
```suggestion
    return dp[-1][-1]
```
The last cell in the last row...

PR #3410, opened 2020-10-17
Create .gitattributes for Cross OS compatibility

acoder77:
With this set, Windows users will have text files converted from Windows style line endings (\r\n) to Unix style line endings (\n) when they’re added to the repository.
https://www.edwardthomson.com/blog/git_for_windows_line_endings.html

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto


cclauss (approved):
Nice!  Thanks.

PR #3415, opened 2020-10-17
Update README.md

KUNU26262:
It does not looks good .Please change it


cozek:
I fail to see any merit to this PR.

PR #3495, opened 2020-10-18
Binary To Decimal Conversion

MeenakshyBS:
b_num = list(input("Input a binary number: "))
value = 0

for i in range(len(b_num)):
	digit = b_num.pop()
	if digit == '1':
		value = value + pow(2, i)
print("The decimal value of the number is", value)[](url)


shellhub:
See this file: [conversions/binary_to_decimal.py](https://github.com/TheAlgorithms/Python/blob/master/conversions/binary_to_decimal.py)

PR #3632, opened 2020-10-21
Add breadth first search

Rolv-Apneseth:
Added a .py file in the searches folder for breadth-first search algorithm with a pathfind visualisation to show the result. 

Included doctests and have already run black, mypy and doctests locally.

In relation to #3260

Issue #3260, opened 2020-10-13
Pathfinding algorithms


Rolv-Apneseth:
Hi, just wondering if adding some pathfinding algorithms implemented in python would be something I could contribute to this project? I have experience with these as I have built a pathfinding algorithm visualiser using pygame (repo here: https://github.com/Rolv-Apneseth/pathfind-visualiser)

NumberPiOso:
Nice repo!

This kind of algorithms, like **A\*** and **Greedy** algorithms are included in searches section.

I suppose they would belong there.

Rolv-Apneseth:
Ok, so should I do it as a grid of nodes which is printed to the console or it returns a path or what is best

Rolv-Apneseth:
Or just an int of the shortest path length?

PR #3648, opened 2020-10-22
Added breadth_first_search.py

Rolv-Apneseth:
Added a .py file in the searches folder for breadth-first search algorithm with a pathfind visualisation to show the result.

Included doctests and have already run black, mypy and doctests locally.

In relation to #3260, and retrying from #3632 since I accidentally included commits from another pr in that one.

### DIFF
--- a/searches/breadth_first_search.py
+++ b/searches/breadth_first_search.py
@@ -1,0 +1,238 @@
+"""
+Python implementation of the breadth-first search algorithm for pathfinding.
+
+Find an explanation of this algorithm here: 
+https://en.wikipedia.org/wiki/Breadth-first_search
+
+Also included is a simple node class, maze generation and maze visualisation
+to show the results of the algorithm.
+
+For doctest testing run: python3 -m doctest -v breadth_first_search.py
+
+For manual testing and visualisation run: python3 breadth_first_search.py
+"""
+import random
+import queue
+
+
+class Node:
+    """
+    Objects of this class make up the grid (2d array) which will be
+    traversed by the algorithm.
+    """
+
+    def __init__(self, value, row, col):
+        self.value = value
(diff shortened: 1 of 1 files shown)


Rolv-Apneseth:
Thank you to the comments on #3632

spamegg1:
I see there are already 4 BFS implementations in the `graphs` folder. Is this significantly different from those? Not sure, a maintainer would have to look over.

Rolv-Apneseth:
Oh my god I did not even see that folder, I just went off the comment under the issue I posted (#3260) that said to put it in searches and nobody else replied to that issue so I just went with it.

Rolv-Apneseth:
For a maintainer to decide then, looking at the other breadth first search algorithms though, it looks like they have to be fed a specific linked list, whereas mine generates a simple maze(optional) where the user only has to specify size, starting and ending location and it will print it the maze before and after it is solved. I could do the same for a different algorithm though it looks like most are covered in that graphs folder. Again, my bad, thought it would be in the searches folder

spamegg1:
@Rolv-Apneseth Ah don't feel bad, the repo is not perfectly organized and contains a lot of redundant stuff. (We actually need people to improve existing files much more than we need new algorithms being submitted, at least in my opinion. Take a look at #2128 )

But as you can see it definitely pays to browse through the repo FIRST.

I guess [mazes can be converted to graphs](https://www.youtube.com/watch?v=rop0W4QDOUI) and vice versa, but your maze version might be worth having. But I guess it will take a long time to be reviewed.

There are other much much less popular TheAlgorithms repos where you could submit this: F#, Haskell, OCaml, Scala, Elm and so on. In the spirit of Hacktoberfest, learn a new language!

PR #3663, opened 2020-10-23
Added new solution for Project Euler problem 027

fpringle:
### **Describe your change:**

 Project Euler Problem 27: Quadratic primes

Problem Statement:

Euler discovered the remarkable quadratic formula:

n² + n + 41

It turns out that the formula will produce 40 primes for the consecutive values n = 0 to 39. However, when n = 40, 40² + 40 + 41 = 40(40 + 1) + 41 is divisible by 41, and certainly when n = 41, 41² + 41 + 41 is clearly divisible by 41.

The incredible formula  n2 − 79n + 1601 was discovered, which produces 80 primes for the consecutive values n = 0 to 79. The product of the coefficients, −79 and 1601, is −126479.

Considering quadratics of the form:
n² + an + b, where |a| < 1000 and |b| <= 1000 where |n| is the modulus/absolute value of ne.g. |11| = 11 and |−4| = 4

Find the product of the coefficients, a and b, for the quadratic expression that produces the maximum number of primes for consecutive values of n, starting with n = 0.

Solution: This is almost the same algorithm as sol1.py authored by @cclauss. The main difference is that we generate the primes at the beginning rather than checking for primeness on the fly. since for n=b, n^2 + a*n + b will always be divisible by b, the theoretical maximum number we have to check for primeness is (b_limit - 1)^2 + a_limit * b_limit + b_limit. For the real problem, a_limit = b_limit = 1000 so we need to generate primes up to 1998000, just under 2m. In practice, the largest prime we actually end up checking for primeness is 12989, but I can't see any way we could have determined that before running the algorithm. If there was, we'd only need to generate primes up to 13000 instead of 2000000, which would result in a code speedup of about 50%

Reference:  [https://projecteuler.net/problem=27](url)

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/project_euler/problem_027/sol2.py
+++ b/project_euler/problem_027/sol2.py
@@ -37,6 +37,13 @@ from typing import Set
 
 
 def sieve(limit: int) -> Set[int]:
+    """
+    Return a set containing all the primes less than limit.
+    >>> sieve(10)
+    {2, 3, 5, 7}
+    >>> sieve(50)
+    {2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47}
+    """
     primes = set(range(3, limit, 2))
     primes.add(2)
     p = 3
@@ -80,4 +87,4 @@ def solution(a_limit: int = 1000, b_limit: int = 1000) -> int:
 
 
 if __name__ == "__main__":
-    print(f"solution() = {solution()}")
+    print(f"{solution() = }")
--- a/project_euler/problem_027/sol2.py
+++ b/project_euler/problem_027/sol2.py
@@ -1,0 +1,83 @@
+"""
+Project Euler Problem 27
+https://projecteuler.net/problem=27
+
+Problem Statement:
+
+Euler discovered the remarkable quadratic formula:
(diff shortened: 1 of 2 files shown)


project_euler/problem_027/sol2.py
```diff
@@ -0,0 +1,83 @@
...
+If there was, we'd only need to generate primes up to 13000 instead of 2000000,
+which would result in a code speedup of about 50%
+
+"""
+
+from typing import Set
+
+
+def sieve(limit: int) -> Set[int]:
+    primes = set(range(3, limit, 2))
```
xcodz-dot, line 40: If possible could you please add a doctest here

fpringle: Thanks, done

xcodz-dot (approved):

xcodz-dot:
Now you will have to wait for a member approval or changes

dhruvmanila:
Thank you for your contribution but currently, we are not accepting any solution to existing problems. Take a look at the `project_euler` directory and submit a solution to non-existing ones.

PR #3677, opened 2020-10-23
Added solution for Project Euler problem 112

fpringle:
### **Describe your change:**

Project Euler problem 112: Bouncy numbers

Working from left-to-right if no digit is exceeded by the digit to its left it is called an increasing number; for example, 134468.

Similarly if no digit is exceeded by the digit to its right it is called a decreasing number; for example, 66420.

We shall call a positive integer that is neither increasing nor decreasing a "bouncy" number; for example, 155349.

Clearly there cannot be any bouncy numbers below one-hundred, but just over half of the numbers below one-thousand (525) are bouncy. In fact, the least number for which the proportion of bouncy numbers first reaches 50% is 538.

Surprisingly, bouncy numbers become more and more common and by the time we reach 21780 the proportion of bouncy numbers is equal to 90%.

Find the least number for which the proportion of bouncy numbers is exactly 99%.

Reference: [https://projecteuler.net/problem=112](url)

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -619,6 +619,7 @@
     * [Sol1](https://github.com/TheAlgorithms/Python/blob/master/project_euler/problem_031/sol1.py)
     * [Sol2](https://github.com/TheAlgorithms/Python/blob/master/project_euler/problem_031/sol2.py)
   * Problem 032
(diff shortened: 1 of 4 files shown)


dhruvmanila:
This pull request contains a lot of extra files, maybe you created the branch from a wrong commit. As such there's already a solution for problem 112. Thank you for your contribution!

PR #3679, opened 2020-10-23
added text_to_speech

asperduti:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/other/tests/test_text_to_speech.py
+++ b/other/tests/test_text_to_speech.py
@@ -1,0 +1,43 @@
+import unittest
+
+from text_to_speech import get_mp3_filename, request_fromtexttospeech
+
+
+class TestTextToSpeech(unittest.TestCase):
+    def test_request_fromtexttospeech_status_code(self):
+        r = request_fromtexttospeech()
+        self.assertEqual(r.status_code, 200)
+
+    def test_get_mp3_file_valid_html(self):
+        html = "<p>Listen to the MP3 file  :<BR></p> <!> \
+                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000' width='470' height='24' id='single1' name='single1'> \
+                <param name='movie' value='player.swf'> \
+                <param name='allowfullscreen' value='true'> \
+                <param name='allowscriptaccess' value='always'> \
+                <param name='wmode' value='transparent'> \
+                <param name='flashvars' value='file=/output/0360061001606401513/59405670.mp3'> \
+                <embed \
+                    id='single2' \
+                    name='single2' \
+                    src='player.swf' \
+                    width='470' \
+                    height='24' \
+                    bgcolor='#000000' \
+                    allowscriptaccess='always' \
+                    allowfullscreen='true' \
+                    flashvars='file=/output/0360061001606401513/59405670.mp3' \
+                /> \
+                </object>"
+        mp3_file_name = get_mp3_filename(html)
+        self.assertEqual("/output/0360061001606401513/59405670.mp3", mp3_file_name)
+
+    def test_get_mp3_file_invalid_html(self):
+        html = "<p>Listen to the MP3 file  :<BR></p> <!> \
+                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000' width='470' height='24' id='single1' name='single1'> \
+                <param name='movie' value='player.swf'> \
+                <param name='allowfullscreen' value='true'> \
+                <param name='allowscriptaccess' value='always'> \
+                <param name='wmode' value='transparent'> \
+                </object>"
+        mp3_file_name = get_mp3_filename(html)
+        self.assertIsNone(mp3_file_name)
--- a/other/tests/test_text_to_speech.py
+++ b/other/tests/test_text_to_speech.py
@@ -1,6 +1,6 @@
 import unittest
 
-from text_to_speech import get_mp3_filename, request_fromtexttospeech
+from other.text_to_speech import get_mp3_filename, request_fromtexttospeech
 
 
 class TestTextToSpeech(unittest.TestCase):
@@ -10,12 +10,14 @@ class TestTextToSpeech(unittest.TestCase):
 
     def test_get_mp3_file_valid_html(self):
         html = "<p>Listen to the MP3 file  :<BR></p> <!> \
-                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000' width='470' height='24' id='single1' name='single1'> \
+                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000' \
+                        width='470' height='24' id='single1' name='single1'> \
                 <param name='movie' value='player.swf'> \
                 <param name='allowfullscreen' value='true'> \
                 <param name='allowscriptaccess' value='always'> \
                 <param name='wmode' value='transparent'> \
-                <param name='flashvars' value='file=/output/0360061001606401513/59405670.mp3'> \
+                <param name = 'flashvars' \
+                       value='file=/output/0360061001606401513/59405670.mp3'> \
                 <embed \
                     id='single2' \
                     name='single2' \
@@ -33,7 +35,8 @@ class TestTextToSpeech(unittest.TestCase):
 
     def test_get_mp3_file_invalid_html(self):
         html = "<p>Listen to the MP3 file  :<BR></p> <!> \
-                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000' width='470' height='24' id='single1' name='single1'> \
+                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000' \
+                        width='470' height='24' id='single1' name='single1'> \
                 <param name='movie' value='player.swf'> \
                 <param name='allowfullscreen' value='true'> \
                 <param name='allowscriptaccess' value='always'> \
--- a/other/text_to_speech.py
+++ b/other/text_to_speech.py
@@ -9,9 +9,6 @@ Example:
 """
 
 import re
-import subprocess
-import urllib.parse
-
 import requests
 
 LANGUAGES_VOICES = {
@@ -50,7 +47,7 @@ def get_mp3_filename(html: str) -> str:
     """Returns the  mp3 file name of an HTML response from fromtexttospeech.com"""
     try:
         filename = re.search(r"(\/output\/\d+\/\d+\.mp3)", html).group(0)
-    except:
+    except AttributeError:
         filename = None
     return filename
 
@@ -73,7 +70,8 @@ def request_fromtexttospeech(
     headers = {
         "Accept-Language": "en-US,en;q=0.5",
         "Upgrade-Insecure-Requests": "1",
-        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:66.0) Gecko/20100101 Firefox/66.0",
+        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:66.0) Gecko/20100101 " +
+                      "Firefox/66.0",
         "Content-Type": "application/x-www-form-urlencoded",
         "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
         "Referer": "http://www.fromtexttospeech.com/",
--- a/other/text_to_speech.py
+++ b/other/text_to_speech.py
@@ -9,6 +9,7 @@ Example:
 """
 
 import re
+
 import requests
 
 LANGUAGES_VOICES = {
--- a/other/text_to_speech.py
+++ b/other/text_to_speech.py
@@ -46,6 +46,44 @@ VOICES = {
 }
 
 
+def get_mp3_filename(html: str) -> str:
+    """Returns the  mp3 file name of an HTML response from fromtexttospeech.com"""
+    try:
+        filename = re.search(r"(\/output\/\d+\/\d+\.mp3)", html).group(0)
+    except:
+        filename = None
+    return filename
+
+
+def request_fromtexttospeech(
+    text: str = "Hello world",
+    language: str = "British",
+    voice: str = "Emma",
+) -> requests.Response:
+    """Makes POST request to fromtexttospeech and returns a requests.Response."""
+    # prepare data
+    data = {
+        "input_text": text,
+        "language": language,
+        "voice": VOICES[voice],
+        "speed": 0,
+        "action": "process_text",
+    }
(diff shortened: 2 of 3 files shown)


other/text_to_speech.py
```diff
@@ -0,0 +1,119 @@
...
+        raise ValueError(f"Voice '{voice}' not available")
+
+    # prepare data
+    text_encoded = urllib.parse.quote(text.encode("utf8"))
+    data = f"input_text={text_encoded}&language={language}\
+            &voice={VOICES[voice]}&speed=0&action=process_text"
+
+    # make request to www.fromtexttospeech.com
+    # it gives an html with the mp3's url
+    args = [
```
dhruvmanila, line 69: This is a Python algorithm repository so it makes more sense to use the `requests` library instead of `curl`. Please make the necessary changes.

Here's the reference to `requests` docs: https://requests.readthedocs.io/en/master/

other/text_to_speech.py
```diff
@@ -0,0 +1,119 @@
...
+        "5",
+        "--max-time",
+        "120",
+    ]
+    process = subprocess.Popen(args, stdout=subprocess.PIPE)
+
+    output = process.communicate()[0]
+
+    # extract the mp3's url
+    mp3_file = re.search(r"(\/output\/\d+\/\d+\.mp3)", output.decode("utf-8")).group(0)
```
dhruvmanila, line 104: Please include additional tests for the regex pattern.

dhruvmanila (changes requested):
Please provide tests from the following, whichever you prefer:
`doctest`, `unittest`, `pytest`

other/text_to_speech.py
xcodz-dot: You are right, also curl is not available by default on Windows (Among all the PCs in world, Windows account for 88%, this code will be compatible for 12% people only)

asperduti: > This is a Python algorithm repository so it makes more sense to use the `requests` library instead of `curl`. Please make the necessary changes.
> 
> Here's the reference to `requests` docs: https://requests.readthedocs.io/en/master/
HI @dhruvmanila , yes you are right. I know about requests, I use it in the script to download the mp3. But I was not able to implement the download of the page with requests, even using the same headers as in curl. I will give a new view.

other/text_to_speech.py
asperduti: > Please include additional tests for the regex pattern.

Can you give me a hand? 
How I can only add tests for only that part of the code? And not the entire function?

other/text_to_speech.py
dhruvmanila: Please take a look at all the parameters accepted by the `request` method: https://requests.readthedocs.io/en/latest/api/#requests.request

All your headers (`-H`) should be provided in a dictionary to the `headers` parameter. All your data (`-d`) should also be in a dictionary to the `data` parameter. 

As you can see, it is pretty intuitive to understand. In a similar way, you can construct your other request parameters.

other/text_to_speech.py
dhruvmanila: Take a look at the Python `re` module: https://docs.python.org/3/library/re.html#module-re

You compile your regex pattern first using [this method](https://docs.python.org/3/library/re.html#re.compile) and then you can use that as usual but instead of doing `re.search()` you would do `pattern.search()` where `pattern` is the variable name assigned to the [compiled pattern object](https://docs.python.org/3/library/re.html#regular-expression-objects).

This will separate out the pattern object which you can use to test ***only*** the pattern.

dhruvmanila (changes requested):
You would also require to write tests for the function. This would be a bit difficult so if you face any problem, ping me, but before that please try first.

asperduti:
> You would also require to write tests for the function. This would be a bit difficult so if you face any problem, ping me, but before that please try first.

Well, after some troubles, I did it. Please tell me if you need more tests or something else.

I structured the code differently, I split it up to make easier the testing.

other/text_to_speech.py
```diff
@@ -0,0 +1,116 @@
...
+    "Mateo": "IVONA Enrique22 (Spanish [Modern])",
+    "Rodrigo": "IVONA Cristiano22 (Portuguese)",
+}
+
+
+def get_mp3_filename(html: str) -> str:
+    """Returns the  mp3 file name of an HTML response from fromtexttospeech.com"""
+    try:
+        filename = re.search(r"(\/output\/\d+\/\d+\.mp3)", html).group(0)
+    except AttributeError:
```
dhruvmanila, line 51: Why will there be an `AttributeError` in here?

Also, it seems that the group is unnecessary in the regex as you're getting the entire match as output:
```suggestion
    try:
        filename = re.search(r"\/output\/\d+\/\d+\.mp3", html)
    except AttributeError:
```

Make sure the regex is correct, I might be wrong.

other/text_to_speech.py
```diff
@@ -0,0 +1,116 @@
...
+    "Daisy": "IVONA Salli22",
+    "Alessandra": "IVONA Carla22 (Italian)",
+    "Giovanni": "IVONA Giorgio22 (Italian)",
+    "Isabella": "IVONA Conchita22 (Spanish [Modern])",
+    "Mateo": "IVONA Enrique22 (Spanish [Modern])",
+    "Rodrigo": "IVONA Cristiano22 (Portuguese)",
+}
+
+
+def get_mp3_filename(html: str) -> str:
```
dhruvmanila, line 47:
```suggestion
def get_mp3_filename(html: str) -> Optional[str]:
```
As you're returning `None` or `str`.

other/text_to_speech.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if language not in LANGUAGES_VOICES:
+        raise ValueError(f"Language '{language}' not available")
+
+    if voice not in LANGUAGES_VOICES[language]:
+        raise ValueError(f"Voice '{voice}' not available")
+
+    r = request_fromtexttospeech(text, language, voice)
+    # extract the mp3's url
+    mp3_file = get_mp3_filename(r.text)
```
dhruvmanila, line 101: You seem to be returning `Optional[str]` and there doesn't seem to be any checking done whether the value is `None` or we got the filename.

If the value is `None` which means either the pattern did not work or there is no file present, the user should know about it and we should exit the function as well.

other/text_to_speech.py
```diff
@@ -0,0 +1,116 @@
...
+        "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+        "Referer": "http://www.fromtexttospeech.com/",
+        "Connection": "keep-alive",
+    }
+    # make request to www.fromtexttospeech.com
+    # it gives an html with the mp3's url
+    return requests.post("http://www.fromtexttospeech.com/", headers=headers, data=data)
+
+
+def text_to_speech(
```
dhruvmanila, line 86: Please write the docstring mentioning the contract between you and the user of this function. As you're returning a bool, the user should know about that and also what does `True` or `False` even mean.

other/text_to_speech.py
```diff
@@ -0,0 +1,116 @@
...
+    text: str = "Hello world",
+    language: str = "British",
+    voice: str = "Emma",
+    audio_path: str = "output.mp3",
+) -> bool:
+
+    if language not in LANGUAGES_VOICES:
+        raise ValueError(f"Language '{language}' not available")
```
dhruvmanila, line 94:
```suggestion
        raise ValueError(f"Language {language!r} not available")
```

other/text_to_speech.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if voice not in LANGUAGES_VOICES[language]:
+        raise ValueError(f"Voice '{voice}' not available")
```
dhruvmanila, line 97:
```suggestion
        raise ValueError(f"Voice {voice!r} not available")
```

dhruvmanila (changes requested):

other/text_to_speech.py
asperduti: The `group` is needed to it, because it returns me the string(without it, I get a Match object), but I can take out the index(in the `group` method) and the `()`(in the regex) because I will take the first occurrence.

asperduti: > Why will there be an `AttributeError` in here?

I put the `AttributeError` because if the `re.search`doesn't found anything I got `None` and `None` doesn't have `group()`
Do you suggest another way?

dhruvmanila: In that case:
```python
if match := re.search(r"\/output\/\d+\/\d+\.mp3", html):
    filename = match.group()
```
You don't actually need to put the parenthesis to get the entire match. Also, if you don't pass any argument to the `group` method, it will assume to be 0 thus returning the entire match.

asperduti: > In that case:
> 
> ```python
> if match := re.search(r"\/output\/\d+\/\d+\.mp3", html):
>     filename = match.group()
> ```

Thanks, I didn't know about the `:=`.
In general, I use the try/except for this: https://docs.python.org/3.9/glossary.html#term-eafp

asperduti:
Hi @dhruvmanila I worked on all the requested changes.

asperduti:
Hi,  @ xcodz-dot , @dhruvmanila .

Did you check it?
I worked in the things that you told me.

xcodz-dot (approved):
looks good

xcodz-dot:
cc: @cclauss

dhruvmanila:
Hi, I will take a look at this in a day or two.

other/tests/test_text_to_speech.py
```diff
@@ -0,0 +1,46 @@
...
+                    name='single2' \
+                    src='player.swf' \
+                    width='470' \
+                    height='24' \
+                    bgcolor='#000000' \
+                    allowscriptaccess='always' \
+                    allowfullscreen='true' \
+                    flashvars='file=/output/0360061001606401513/59405670.mp3' \
+                /> \
+                </object>"
```
cclauss, line 32: Triple-quoted multiline strings work in Python and as PEP8 says, backslashes should be avoided in Python code.
```suggestion
        html = """<p>Listen to the MP3 file  :<BR></p> <!>
                <object classid='clsid:D27CDB6E-AE6D-11cf-96B8-444553540000'
                        width='470' height='24' id='single1' name='single1'>
                <param name='movie' value='player.swf'>
                <param name='allowfullscreen' value='true'>
                <param name='allowscriptaccess' value='always'>
                <param name='wmode' value='transparent'>
                <param name = 'flashvars'
                       value='file=/output/0360061001606401513/59405670.mp3'>
                <embed
                    id='single2'
                    name='single2'
                    src='player.swf'
                    width='470'
                    height='24'
                    bgcolor='#000000'
                    allowscriptaccess='always'
                    allowfullscreen='true'
                    flashvars='file=/output/0360061001606401513/59405670.mp3'
                />
                </object>"""
```

other/text_to_speech.py
```diff
@@ -0,0 +1,137 @@
...
+    text_to_speech(text="¡Hola Mundo!", language="Spanish", voice="Mateo")
+"""
+
+import re
+from typing import Optional
+
+import requests
+
+LANGUAGES_VOICES = {
+    "US English": ["Alice", "Daisy", "George", "Jenna", "John"],
```
cclauss, line 17:
```suggestion
    "US English": ("Alice", "Daisy", "George", "Jenna", "John"),
```
Tuples take less RAM than lists and they indicate to the reader that the values will not change at runtime.

other/text_to_speech.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def get_mp3_filename(html: str) -> Optional[str]:
+    """Returns the  mp3 file name of an HTML response from fromtexttospeech.com"""
+    if match := re.search(r"\/output\/\d+\/\d+\.mp3", html):
+        filename = match.group()
+    else:
+        filename = None
+
+    return filename
```
cclauss, line 55:
```suggestion
    match = re.search(r"\/output\/\d+\/\d+\.mp3", html):
    return match.group() if match else None
```

other/text_to_speech.py
```diff
@@ -0,0 +1,137 @@
...
+    "Daisy": "IVONA Salli22",
+    "Alessandra": "IVONA Carla22 (Italian)",
+    "Giovanni": "IVONA Giorgio22 (Italian)",
+    "Isabella": "IVONA Conchita22 (Spanish [Modern])",
+    "Mateo": "IVONA Enrique22 (Spanish [Modern])",
+    "Rodrigo": "IVONA Cristiano22 (Portuguese)",
+}
+
+
+def get_mp3_filename(html: str) -> Optional[str]:
```
cclauss, line 48: Why two different return types?  Why not just return an empty str instead of `None`?

other/text_to_speech.py
```diff
@@ -0,0 +1,137 @@
...
+        raise ValueError(f"Voice {voice!r} not available")
+
+    r = request_fromtexttospeech(text, language, voice)
+    # extract the mp3's url
+    if (mp3_file := get_mp3_filename(r.text)) is None:
+        return False
+
+    audio_url = f"http://www.fromtexttospeech.com{mp3_file}"
+
+    mp3data = requests.get(audio_url)
```
cclauss, line 125:
```suggestion
    mp3data = requests.get(f"http://www.fromtexttospeech.com{mp3_file}")
```

cclauss: This is not really `mp3data`.  Instead, it is a `requests.Response` so perhaps `response` would be a better variable name than `mp3data`.

other/text_to_speech.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # save the data as an mp3 file
+    if mp3data.status_code != 404:
+        with open(audio_path, "w+b") as f:
+            f.write(mp3data.content)
+        return True
+
+    return False
```
cclauss, line 133:
```suggestion
    try:
        response.raise_for_status()
        # save the data as an mp3 file
        with open(audio_path, "w+b") as f:
            f.write(mp3data.content)
        return True
    except requests.HTTPError:
        return False
```

other/tests/test_text_to_speech.py
```diff
@@ -0,0 +1,46 @@
...
+        mp3_file_name = get_mp3_filename(html)
+        self.assertEqual("/output/0360061001606401513/59405670.mp3", mp3_file_name)
+
+    def test_get_mp3_file_invalid_html(self):
+        html = "<p>Listen to the MP3 file  :<BR></p> <!> \
```
cclauss, line 37: Triple-quoted str with no backslashes please.

PR #3680, opened 2020-10-23
Added solution to Project Euler problem 101

fpringle:
### **Describe your change:**

Project Euler problem 101: Optimum polynomial

If we are presented with the first k terms of a sequence it is impossible to say with certainty the value of the next term, as there are infinitely many polynomial functions that can model the sequence.

As an example, let us consider the sequence of cube numbers. This is defined by the generating function,
u(n) = n^3: 1, 8, 27, 64, 125, 216, ...

Suppose we were only given the first two terms of this sequence. Working on the principle that "simple is best" we should assume a linear relationship and predict the next term to be 15 (common difference 7). Even if we were presented with the first three terms, by the same principle of simplicity, a quadratic relationship should be assumed.

We shall define OP(k, n) to be the nth term of the optimum polynomial generating function for the first k terms of a sequence. It should be clear that OP(k, n) will accurately generate the terms of the sequence for n ≤ k, and potentially the first incorrect term (FIT) will be OP(k, k+1); in which case we shall call it a bad OP (BOP).

As a basis, if we were only given the first term of sequence, it would be most sensible to assume constancy; that is, for n ≥ 2, OP(1, n) = u(1).

Hence we obtain the following OPs for the cubic sequence:

OP(1, n) = 1                            1, 1, 1, 1, ...
OP(2, n) = 7n - 6                     1, 8, 15, ...
OP(3, n) = 6n^2 - 11n + 6       1, 8, 27, 58, ...
OP(4, n) = n^3                        1, 8, 27, 64, 125, ...
Clearly no BOPs exist for k ≥ 4.

By considering the sum of FITs generated by the BOPs (indicated in red above), we obtain 1 + 15 + 58 = 74.

Consider the following tenth degree polynomial generating function:

u(n) = 1 - n + n^2 - n^3 + n^4 - n^5 + n^6 - n^7 + n^8 - n^9 + n^10

Find the sum of FITs for the BOPs.

Reference: [https://projecteuler.net/problem=101](url)

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


fpringle:
@dhruvmanila @Kush1101 
Checks are failing because I'm using the 3rd party library numpy in my code, how do I get around this? I see that numpy is listed in [requirements.txt](https://github.com/TheAlgorithms/Python/tree/master/requirements.txt) so I'm not sure how to include it.

dhruvmanila:
As requested before, do not open another PR before your other PRs are merged. You need to close all your recent PRs. We're already being flooded with PRs and on top of that you're opening dozens of them even after requesting not to do so. If you cannot follow such a simple instruction, I'm sorry I cannot help you. Thank you for your contribution.

fpringle:
Fine, I won't submit any more PRs until all of my open ones have been dealt
with.

Btw, if you want help with the flood of recent PRs I'd be more than happy
to contribute as a reviewer :) idk if you need to enable me somehow, I'm
not entirely sure how that works!

PR #3762, opened 2020-10-26
New file prime_numbers_in_a_given_interval.py has been added

Shashank5665:
Added new algorithm

### DIFF
--- a/maths/prime_numbers_in_a_given_interval.py
+++ b/maths/prime_numbers_in_a_given_interval.py
@@ -1,0 +1,11 @@
+#Take the input from the user:   
+lower = int(input("Enter lower range: "))  
+upper = int(input("Enter upper range: "))  
+  
+for num in range(lower,upper + 1):  
+   if num > 1:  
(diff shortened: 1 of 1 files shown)


Shashank5665:
I am new and i am not able to figure out why is it showing "Merging is blocked", please help me

Ankuraxz:
Trim Trailing White space error
-#Take the input from the user:   
-lower = int(input("Enter lower range: "))  
-upper = int(input("Enter upper range: "))  
-  
-for num in range(lower,upper + 1):  
-   if num > 1:  
-       for i in range(2,num):  
-           if (num % i) == 0:  
-               break  
-       else:  
-           print(num) 
\ No newline at end of file
+# Take the input from the user:
+lower = int(input("Enter lower range: "))
+upper = int(input("Enter upper range: "))
+
+for num in range(lower, upper + 1):
+    if num > 1:
+        for i in range(2, num):
+            if (num % i) == 0:
+                break
+        else:
+            print(num)

try this --> Flake8

PR #3811, opened 2020-10-28
Changed how the Visited nodes are tracked

akashgkrishnan:
Updated the code to track visited Nodes with Set data structure instead of Lists to bring down the lookup time in visited  from O(N) to O(1)
as doing O(N) lookup each time in the visited List will become significantly slow when the graph grows

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/graphs/bfs_shortest_path.py
+++ b/graphs/bfs_shortest_path.py
@@ -1,8 +1,6 @@
 """Breadth-first search shortest path implementations.
-
     doctest:
     python -m doctest -v bfs_shortest_path.py
-
     Manual test:
     python bfs_shortest_path.py
 """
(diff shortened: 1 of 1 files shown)


poyea (approved):

dhruvmanila:
Can you show us some data regarding the change? Data might include the difference in time in running the same algorithm on the same input with same configuration, if any.

dhruvmanila:
I don't see the difference you mentioned `O(n) -> O(1)`:
```console
$ time python graphs/bfs_shortest_path.py  # with set
['G', 'C', 'A', 'B', 'D']
4

real    0m0.094s
user    0m0.044s
sys     0m0.036s


$ time python graphs/bfs_shortest_path.py  # with list
['G', 'C', 'A', 'B', 'D']
4

real    0m0.097s
user    0m0.046s
sys     0m0.036s
```

akashgkrishnan:
@dhruvmanila running on smaller inputs will yield similar results. the speed will be significantly faster if the graph has a 10^5 nodes

As you know the lookup time in an array is O(N) if we are using the **in** keyword.
```
4 in [1,2,3,4] 
```
will be O(N). getting items based on an index is O(1) other than that it is O(N) if we are using the **in** keyword

sets
```
4 in {1,2,3,4} 
```
will be O(1) as sets use hash function and store the items for fast lookups the only caveat being that all items need to be unique but here in this implementation that would be the same case even if we use the list ds

https://wiki.python.org/moin/TimeComplexity
the same is specified here.
you can check the time complexity of the in for both array and sets here

dhruvmanila:
If possible can you show us with some practical examples? 
Also, please restore all the newlines.

theroyakash (approved):
looks okay

PR #3042, opened 2020-10-08
Project Euler 206 Solution 

peteryao7:
### **Describe your change:**

Adds a solution and explanation for Project Euler 206. Part of issue #2695.

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [x] Documentation change?

### **Checklist:**
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [x] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #3829, opened 2020-10-30
Project Euler 206 Solution

peteryao7:
### **Describe your change:**

Adds a solution and explanation for Project Euler 206. Part of issue #2695 and duplicate of #3042.

* [x] Add an algorithm?
* [x] Documentation change?

### **Checklist:**


peteryao7:
`solution()` has no tests because it doesn't make sense for this function to take any input. There's only a certain range of numbers that could possibly be the answer, which is [101010103, 138902663], and I start from the end and count down to the start point. Adding tests would introduce a new layer of logic and complexity that I think would make the solution harder to understand.

dhruvmanila (approved):
Awesome job! 🙌
Thank you for your contribution 🥳

PR #3860, opened 2020-11-02
Update caesar_cipher.py

MrGeek1337:
improved for conciseness and readability

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?
* [x] Improved existing algorithm?

### **Checklist:**

### DIFF
--- a/ciphers/caesar_cipher.py
+++ b/ciphers/caesar_cipher.py
@@ -193,26 +193,13 @@ def brute_force(input_string: str, alphabet=None) -> dict:
     # Set default alphabet to lower and upper case english chars
     alpha = alphabet or ascii_letters
 
-    # The key during testing (will increase)
-    key = 1
-
-    # The encoded result
-    result = ""
-
     # To store data on all the combinations
     brute_force_data = {}
 
     # Cycle through each combination
-    while key <= len(alpha):
-        # Decrypt the message
-        result = decrypt(input_string, key, alpha)
-
-        # Update the data
-        brute_force_data[key] = result
-
-        # Reset result and increase the key
-        result = ""
-        key += 1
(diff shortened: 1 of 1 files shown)


mrmaxguns:
This seems like the easiest and most pythonic way to do brute force.

ciphers/caesar_cipher.py
```diff
@@ -193,26 +193,13 @@ def brute_force(input_string: str, alphabet=None) -> dict:
...
-        # Decrypt the message
-        result = decrypt(input_string, key, alpha)
-
-        # Update the data
-        brute_force_data[key] = result
-
-        # Reset result and increase the key
-        result = ""
-        key += 1
+    for key in range(1, len(alpha) + 1):
```
mrmaxguns, line 200: I think the better way to implement this loop is to use enumerate:
```python
for key, _ in enumerate(alpha, 1):
    ...
```

MrGeek1337: I don't personally think `enumerate` is useful here, since this is clearly a looping through a range of numbers, hence the logical approach to this would be a `range`, and besides, `enumerate`'s unused generated values (`_`) are also a concern, why generate them if you're not going to use them?

mrmaxguns: I think you are right. The generated value is wasteful, and the `range` is still pretty readable.

mrmaxguns:
@cclauss Could you review this PR?

xcodz-dot:
@dhruvmanila, this seems good to go

dhruvmanila:
@algorithms-keeper review-all

dhruvmanila:
@algorithms-keeper review-all

dhruvmanila (approved):

PR #3889, opened 2020-11-14
Move CI tests from Travis to GitHub

dhruvmanila:
### ~**Proposal for using `pipenv`:**~

~***Goal: Add `Pipfile` and `Pipfile.lock` (keeping `requirements.txt`) and speed up our builds***~

~`pipenv` is a tool which combines the power of `pip` and `venv`. My main goal for using this is only to speed up our builds. Why? Take a look at: https://github.com/dhruvmanila/algorithms-keeper/runs/1394249008?check_suite_focus=true. The entire virtual environment directory is cached and so we just skip the installing part which saves approximately [60 seconds](https://travis-ci.com/github/TheAlgorithms/Python/jobs/439797775#L222) in every run.~

### ~What happens if we update our dependencies?~

~The cache works based on a key and the key will depend on one of the dependency file which in this case can be `Pipfile` or `Pipfile.lock` (lock file is preferable). When we update any dependency, we update the lock file which means the key will change and thus the cache won't be found. In this case, the virtual environment will be created and cached again, thus updating the cache.~

~@cclauss Please take a look at the proposal and if there are any doubts, do ask. ***DO NOT MERGE*** this PR yet.~

# Move CI tests to GitHub Actions

### **Describe your change:**

* [x] Test changes

### **Checklist:**


cclauss:
I am not a fan of pipenv...  Pining dependencies is for projects that need reliable builds on a rigidly fixed set of dependencies.  This is for production software than is heavily supported and must not break.

I much prefer for our dependencies to always be the latest-and-greatest.  Locking in pined dependencies runs contrary to that goal.  If a new release of Python or one of our dependencies breaks our algorithms, I would like to know that as soon as our continuous integration is run.

So, I am OK using pipenv if we ___never put version numbers in `Pipfile`___ but if we need to manually add/change version numbers int that file, then I would be interested to avoid it.

dhruvmanila:
To avoid using version numbers, we just put an asterisk symbol like so: `numpy = "*"` but if that is the case then I should say that we will have to manually update the lock file using the command `pipenv lock` or we could just avoid the lock file and get every information from the `Pipfile` itself. 

We will just add the `Pipfile` file and use that to install and update all our dependencies.

cclauss:
What if we continue to use requirements.txt and trust the cache if `$(pip freeze)` does not change.

dhruvmanila:
There's nothing to worry about if we are using requirements.txt as we will just be using `pip` to install the dependencies.

If that is the case then please take a look at the workflow file in this PR which is basically a mirror of Travis configuration except for the Travis Buddy notification part which I think I can incorporate into our bot. ***Only if it ever gets accepted 😕***

xcodz-dot (approved):
Awesome 😄

dhruvmanila:
@cclauss this seems good to go, can you take a quick look at the changes?

web_programming/instagram_crawler.py
```diff
@@ -23,7 +23,7 @@ class InstagramUser:
     """
     Class Instagram crawl instagram user information
 
-    Usage: (doctest failing on Travis CI)
+    Usage: (doctest failing on GitHub actions)
```
cclauss, line 26:
```suggestion
    Usage: (doctest failing on GitHub Actions)
```
It is a brand name.

web_programming/instagram_crawler.py
```diff
@@ -102,10 +102,10 @@ def test_instagram_user(username: str = "github") -> None:
     A self running doctest
     >>> test_instagram_user()
     """
-    from os import getenv
+    import os
 
-    if getenv("CONTINUOUS_INTEGRATION"):
-        return  # test failing on Travis CI
+    if os.environ.get("GITHUB_ACTIONS"):
+        return None  # test failing on GitHub actions
```
cclauss, line 108:
```suggestion
        return None  # test failing on GitHub Actions
```

web_programming/instagram_crawler.py
```diff
@@ -102,10 +102,10 @@ def test_instagram_user(username: str = "github") -> None:
     A self running doctest
     >>> test_instagram_user()
     """
-    from os import getenv
+    import os
 
-    if getenv("CONTINUOUS_INTEGRATION"):
-        return  # test failing on Travis CI
+    if os.environ.get("GITHUB_ACTIONS"):
```
cclauss, line 107:
```suggestion
    if os.environ.get("CI"):
```
They all seem to do `CI` so let's remain flexible.  https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

.travis.yml
```diff
@@ -1,17 +0,0 @@
...
-jobs:
-  include:
-    - name: Build
-      install: pip install pytest-cov -r requirements.txt
-      script:
-        - pytest --doctest-modules --ignore=project_euler/ --cov-report=term-missing:skip-covered --cov=. .
-after_success:
-  - scripts/build_directory_md.py 2>&1 | tee DIRECTORY.md
-notifications:
-  webhooks: https://www.travisbuddy.com/
```
cclauss, line 16: Thank God that this silly thing is leaving us!

web_programming/instagram_crawler.py
dhruvmanila: GitHub suggests otherwise:

> GITHUB_ACTIONS | Always set to true when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.

cclauss: I do not think that it is suggesting differently.  I like the description next to CI even more. ;-)  It is always set to True.  I trust that more.  CI is supported by Travis CI, CircleCI, and GHA so it is less fragile than hard coding GHA.

Also, why the change https://stackoverflow.com/questions/10952507/when-would-os-environfoo-not-match-os-getenvfoo and https://stackoverflow.com/questions/16924471/difference-between-os-getenv-and-os-environ-get

dhruvmanila: Alright, also take a look at https://stackoverflow.com/a/55940480

One more reason is habit. I was working on a feature in https://github.com/brettcannon/gidgethub where `os.environ` was used everywhere. 

`getenv` is just a wrapper around `environ`

cclauss (approved):
Nice!!

PR #3925, opened 2020-11-22
Replace base64_cipher.py with an easy to understand version

hfz1337:
### **Describe your change:**
This PR changes the existing `base64_cipher.py` with an easy to understand script that explains how the Base64 algorithm works under the hood. The solution is not necessarily faster, but I believe it will help people understand how the algorithm works and why it is possible to use Base64 in Steganography to conceal the existence of a piece of data.
Further more, I believe the `base64_cipher.py` name can be misleading as Base64 is considered an encoding, and not a cipher.

* [ ] Add an algorithm?
* [x] Improve an existing solution.
* [ ] Documentation change?

### **Checklist:**


xcodz-dot (approved):

xcodz-dot:
@cclauss, check this

ciphers/base64_encoding.py
```diff
@@ -0,0 +1,113 @@
...
+    >>> from base64 import b64encode
+    >>> a = b"This pull request is part of Hacktoberfest20!"
+    >>> b = b"https://tools.ietf.org/html/rfc4648"
+    >>> c = b"A"
+    >>> base64_encode(a) == b64encode(a)
+    True
+    >>> base64_encode(b) == b64encode(b)
+    True
+    >>> base64_encode(c) == b64encode(c)
+    True
```
cclauss, line 29: Please add a test that proves that this algorithm behaves like `b64encode("abc")`.

ciphers/base64_encoding.py
```diff
@@ -0,0 +1,113 @@
...
+    >>> from base64 import b64decode
+    >>> a = "VGhpcyBwdWxsIHJlcXVlc3QgaXMgcGFydCBvZiBIYWNrdG9iZXJmZXN0MjAh"
+    >>> b = "aHR0cHM6Ly90b29scy5pZXRmLm9yZy9odG1sL3JmYzQ2NDg="
+    >>> c = "QQ=="
+    >>> base64_decode(a) == b64decode(a)
+    True
+    >>> base64_decode(b) == b64decode(b)
+    True
+    >>> base64_decode(c) == b64decode(c)
+    True
```
cclauss, line 73: Please add a test that proves that this algorithm behaves like `b64decode("abc")`.

cclauss (changes requested):
Please add these two tests to show how the algorithms deal with bad input data.

ciphers/base64_encoding.py
```diff
@@ -27,8 +27,18 @@ def base64_encode(data: bytes) -> bytes:
...
+    >>> base64_encode("abc")
+    Traceback (most recent call last):
+      ...
+    TypeError: a bytes-like object is required, not 'str'
     """
-    binary_stream = "".join(bin(char)[2:].zfill(8) for char in data)
+    # Make sure the supplied data is a bytes-like object
+    if not isinstance(data, bytes):
+        raise TypeError(
+            "a bytes-like object is required, not '{}'".format(data.__class__.__name__)
```
cclauss, line 38:
```suggestion
            f"a bytes-like object is required, not '{data.__class__.__name__}'"
```

ciphers/base64_encoding.py
```diff
@@ -71,7 +81,27 @@ def base64_decode(encoded_data: str) -> bytes:
...
+    >>> base64_decode("abc")
+    Traceback (most recent call last):
+      ...
+    AssertionError: Incorrect padding
     """
+    # Make sure encoded_data is either a string or a bytes-like object
+    if not isinstance(encoded_data, bytes) and not isinstance(encoded_data, str):
+        raise TypeError(
+            "argument should be a bytes-like object or ASCII string, not '{}'".format(
+                encoded_data.__class__.__name__
```
cclauss, line 93:
```suggestion
            f"argument should be a bytes-like object or ASCII string, not '{encoded_data.__class__.__name__}'"
```

ciphers/base64_encoding.py
hfz1337: Thanks, I made the necessary changes, hopefully.
But there's something that I find weird in the official implementation:
```python
>>> from base64 import b64decode
>>> # this throws an exception
>>> b64decode("\xff")
  ...
ValueError: string argument should contain only ASCII characters
>>> # while this returns an empty bytes-like object
>>> b64decode(b"\xff")
b''
>>> b64decode("#")
b''
```
I contrast to my implementation where I made sure to throw an exception when needed.

shouldn't `b64decode("#")` for example throw an exception?

ciphers/base64_encoding.py
hfz1337: I was using f-strings at the beginning, but doing so exceeds the 88 chars per line in this case, and flake8 reports an error.
What should I do?

ciphers/base64_encoding.py
cclauss: Perhaps you should open an issue at https://bugs.python.org ;-)

ciphers/base64_encoding.py
```diff
@@ -0,0 +1,143 @@
...
+    >>> base64_decode("abc")
+    Traceback (most recent call last):
+      ...
+    AssertionError: Incorrect padding
+    """
+    # Make sure encoded_data is either a string or a bytes-like object
+    if not isinstance(encoded_data, bytes) and not isinstance(encoded_data, str):
+        raise TypeError(
+            "argument should be a bytes-like object or ASCII string, not '{}'".format(
+                encoded_data.__class__.__name__
```
cclauss, line 93:
```suggestion
            "argument should be a bytes-like object or ASCII string, not "
            f"'{encoded_data.__class__.__name__}'"
```

cclauss (approved):
An awesome contribution!  Thanks for doing this.

PR #3934, opened 2020-11-22
Ohm's Law algorithm added

erdum:
### **Describe your change:**
Finnaly i did it now i hope all checks should be clear

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


erdum:
Thank god i made my first ever github contribution, i was very difficult for a beginner, but i did it, jus please review it and merged it i will more happy if you will merged it thanks.

electronics/ohms_law.py
```diff
@@ -0,0 +1,32 @@
+# Importing doctest for testing our function
+from doctest import testmod
+
+# Defining function
+
+
+def ohms_law(voltage: float, current: float, resistance: float) -> float:
```
xcodz-dot, line 7:
```suggestion
def ohms_law(voltage: float = 0, current: float = 0, resistance: float = 0) -> float:
```
Default Values be 0 (It will not affect the solution but is recommended by me)

erdum: Yes i did it in previous commit but the max line length was exceeded by this because the parameters name also very long then the styling spaces , i just try again hope it should work, thank you so much for your interest.

electronics/ohms_law.py
```diff
@@ -0,0 +1,32 @@
+# Importing doctest for testing our function
+from doctest import testmod
```
cclauss, line 2: Please move this import inside `__main__` so I want to call your function from a bigger program, I do not need to import doctest.

Also, please replace the comment with something that helps the reader understand your algorithm like https://en.wikipedia.org/wiki/Ohm%27s_law

electronics/ohms_law.py
```diff
@@ -0,0 +1,32 @@
...
+
+# Defining function
+
+
+def ohms_law(voltage: float = 0, current: float = 0, resistance: float = 0) -> float:
+
+    """
+    This function calculates the any one
+    of the three ohms_law values voltage,
+    current, resistance, of electronics.
```
cclauss, line 12:
```suggestion
    For an electrical system, calculate the missing ohms_law values: voltage, current,
    or resistance.
```

electronics/ohms_law.py
```diff
@@ -0,0 +1,32 @@
...
+    >>> ohms_law(voltage=10, resistance=5,current=0 )
+    2.0
+    >>> ohms_law(current=1, resistance=10, voltage=0)
+    10.0
+    """
```
cclauss, line 17: The very first thing that this function should do is raise ValueError if 0.0 does not appear once and only once in the three input parameters.  Returning 0 on bad data is not cool.

electronics/ohms_law.py
```diff
@@ -0,0 +1,32 @@
...
+    if voltage == 0:
+        result = float(current * resistance)
+        return result
```
cclauss, line 20: I think your function should return a dict or even better a namedtuple which provides both the name of the value and the value itself.

Creating `result` on one line and getting rid of it on the next is not helpful.

```suggestion
        return {"voltage": float(current * resistance)}
```

cclauss (changes requested):

electronics/ohms_law.py
```diff
@@ -0,0 +1,32 @@
...
+def ohms_law(voltage: float = 0, current: float = 0, resistance: float = 0) -> float:
+
+    """
+    This function calculates the any one
+    of the three ohms_law values voltage,
+    current, resistance, of electronics.
+    >>> ohms_law(voltage=10, resistance=5,current=0 )
+    2.0
+    >>> ohms_law(current=1, resistance=10, voltage=0)
+    10.0
```
cclauss, line 16: Test for numbers with fractional parts (1.5) and negative numbers, etc.

erdum: No problem i will be right back

electronics/ohms_law.py
erdum: Yes i am thinking about it before even you tell me, so now i will do something for it

erdum:
But cclauss thank you so much showing interest in my code before you nobody even tell me the problems with my code thank you again and i will be add these changes as you tell me .

erdum:
So i got it what you suggest me is 
1. return dictionary with value name instead value
2. doctest import is made inside main
3. Comments should be helpful for reader 
4. And for wrong value input value error should be raised instead returning 0

erdum:
cclauss i did it finally

dhruvmanila:
Hey, @cclauss this is creating a new directory in the root. 

If he gives an ok then @erdum please add an empty `__init__.py` file in the directory as well.

cclauss:
> this is creating a new directory in the root.

Thanks for the heads up on that.  I missed that detail and we should seriously consider if making a new root directory is the right thing to do.  Would we consider this a `conversion` -- Not really.  What other existing root directories would make sense for this contribution or should we really create a new root dir?

dhruvmanila:
If this really needs a new directory then I think it should be `physics`.

Thinking about it, we can keep the number of directories same by putting the `quantum` directory in the `physics`.

cclauss:
There are tons of quantum algorithms that have nothing to do with physics.

cclauss:
I would be OK with adding `electronics` (there are tons of possible algorithms in that category) if we could get rid of something else...  What about `knapsack`?  That seems to be the weakest topic for a root directory.  Where could we stuff the knapsack algorithms to make room for electronics?

electronics/ohms_law.py
```diff
@@ -0,0 +1,46 @@
+def ohms_law(voltage: float, current: float, resistance: float) -> float:
```
cclauss, line 1: Need a URL to more info as discussed in CONTRIBUTING.md
https://en.wikipedia.org/wiki/Ohm%27s_law or other.

electronics/ohms_law.py
```diff
@@ -0,0 +1,46 @@
...
+    >>> ohms_law(voltage=0, current=0, resistance=10)
+    Traceback (most recent call last):
+      File "<stdin>", line 14, in <module>
+    ValueError: Only one argument can be 0 at the time
+    >>> ohms_law(resistance=0, voltage=-10, current=1)
+    {'resistance': -10.0}
+    >>> ohms_law(voltage=0, current=-1.5, resistance=2)
+    {'voltage': -3.0}
+    """
+    if voltage == 0:
```
cclauss, line 17: Let's fast-fail.  This leaves the remaining logic simple and easy to reason about.
```suggestion
    if (voltage, current, resistance).count(0) != 1:
        raise ValueError("One and only one argument must be 0")
    if resistance < 0:
        raise ValueError("Resistance cannot be negative")
    if voltage == 0:
```

electronics/ohms_law.py
```diff
@@ -0,0 +1,46 @@
...
+        if current == 0 or resistance == 0:
+            raise ValueError("Only one argument can be 0 at the time")
+        elif resistance <= 0:
+            raise ValueError("Resistance can't be 0 or in negative")
+        else:
+            result = {"voltage": float(current * resistance)}
+            return result
```
cclauss, line 24: Simply remove variables that are created on one line and deleted on the next line.
```suggestion
            return {"voltage": float(current * resistance)}
```

electronics/ohms_law.py
```diff
@@ -0,0 +1,46 @@
...
+    elif current == 0:
+        if voltage == 0 or resistance == 0:
+            raise ValueError("Only one argument can be 0 at the time")
+        elif resistance <= 0:
+            raise ValueError("Resistance can't be 0 or in negative")
+        else:
+            result = {"current": voltage / resistance}
+            return result
```
cclauss, line 32:
```suggestion
            return {"current": voltage / resistance}
```

electronics/ohms_law.py
```diff
@@ -0,0 +1,46 @@
...
+    elif resistance == 0:
+        if voltage == 0 or current == 0:
+            raise ValueError("Only one argument can be 0 at the time")
+        else:
+            result = {"resistance": voltage / current}
+            return result
```
cclauss, line 38:
```suggestion
            return {"resistance": voltage / current}
```

cclauss (changes requested):
Let's fast-fail.

electronics/ohms_law.py
xcodz-dot: Hello, You can commit my changes and then run the following commands in root directory of repository. I assume you are making changes to a local copy of repository.

* `pip install pre-commit`
* `pre-commit install`
* `pre-commit run -a`

Then upload the changes

electronics/ohms_law.py
cclauss: Why do we cast as `float()` on this branch but not on the other two branches?

dhruvmanila:
It seems that the `images` directory should be removed as we are not using Travis anymore.

dhruvmanila:
There's also `greedy_knapsack` in the `greedy_method` directory.

This is my proposal:
- Remove `images` and `greedy_method` directory.
- Move `greedy_knapsack` to `knapsack` directory as there are multiple variations for the knapsack problem according to [Wikipedia](https://en.wikipedia.org/wiki/Knapsack_problem#Variations)

electronics/ohms_law.py
erdum: Because if you divide two values its gives you float but not in case for the multiplication , and remember i give return type float in type hint

erdum:
And now i think you may don't want to add my contribution because i saw other PR just simple and you see it and just merged without any exception but with my case you nearly figured out every small detail, i don't think this is a great place for beginner to start with that's not the way to teach a beginner its just like you are boss and i am your employ, finally i just say what i realize is you have specific problem with me.

erdum:
Very disappointment not expecting from a person who has lot's of experience, i expecting kindness and good teaching attitude but not always people are good

cclauss:
@erdum I think this PR is extremely close to being landed.  I know it is a lot of work but we have a great repo here because we sweat the details.  We were seriously overwhelmed by Hacktoberfest so we might have let thru things that were not at the highest level of quality but that is the exception, not the rule.

Thanks massively for your hard work and patience.

cclauss (approved):

erdum:
@cclauss if you all say to me before then i am not going to comment that but sorry if my words hutt you but i jus lost my patience .

cclauss:
Patience is a fragile thing.  Your algorithm is now part of the repo.  Thanks for that!

erdum:
I am beginner to github its my first time ever contribution i don't even about the CI for testing my code i figured out by my own how to solve issue because my checks are failing but in first time i jus understand that the more your code is clean then more your code is to be good

cclauss:
Agreed.  Our high standards make this a difficult repo for beginners to contribute to but we are trying to help people to quickly learn things like testing, type hints, good variable names, f-strings because they can make code more readable and more reliable.  Other repos can be easier to contribute to but this repo tries to teach best practices.

erdum:
yes I learned a lot from you and your repo thanks

cclauss:
The thing that I love is that this is not my repo...  It is ___our___ repo.  Now that your algorithm is in the repo, you are part of the club.  The [`electronics` directory](../tree/master/electronics) currently only has one algorithm so please consider what else could be added.

erdum:
Yes my next goal will be on kcl law, it is complicated so i will try to run it locally then i will make PR so you can test or debug my code

erdum:
Now i realize you are a very nice person actually, so sorry for the previous comments

Issue #3956, opened 2020-11-25
Let's lose the images and greedy_method directories


cclauss:
As discussed at https://github.com/TheAlgorithms/Python/pull/3934#issuecomment-733606682

There's also `greedy_knapsack` in the `greedy_method` directory.

Proposal:
- Remove `images` and `greedy_method` directory.
- Move `greedy_knapsack` to `knapsack` directory as there are multiple variations for the knapsack problem according to [Wikipedia](https://en.wikipedia.org/wiki/Knapsack_problem#Variations) and [multiple ways](https://en.wikipedia.org/wiki/Knapsack_problem#Solving) of solving as well.

PR #3972, opened 2020-11-27
fixes: #3956 Cleaned up knapsack and images directory

dhruvmanila:
Fixes: #3956 

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [ ] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanation.


dhruvmanila:
There's this [README](https://github.com/TheAlgorithms/Python/blob/master/knapsack/README.md) file as well which probably either needs updating or should we just remove it?

cclauss:
Perhaps we could get the original poster @tbsschroeder to update the README to reflect that all implementations of the knapsack problem are going to be put in one directory.

Let's do this.  Move the README along with the code and we will see if someone is willing to update it appropriately.  If it is not updated by yearend, we can whack the README.

dhruvmanila:
Well, then this is good to go. I haven't touched the README so it is present in the `knapsack` directory.

knapsack/tests/test_greedy_knapsack.py
```diff
@@ -1,6 +1,6 @@
 import unittest
 
-from . import greedy_knapsack as kp
+from knapsack import greedy_knapsack as kp
```
cclauss, line 3: Is this a vital change?  There are two ways to go.  Hardcode the name of the parent directory or use relative imports.  Each has its pros and cons.  Relative imports are better for easy refactoring and absolute imports are better for libraries and projects that need stability.

cclauss (approved):

knapsack/tests/test_greedy_knapsack.py
dhruvmanila: Relative import would be like so: `from .. import greedy_knapsack as kp` as I created a `tests` directory for both test files. Here's the tree:
```
knapsack/
├── README.md
├── __init__.py
├── greedy_knapsack.py
├── knapsack.py
└── tests
    ├── __init__.py
    ├── test_greedy_knapsack.py
    └── test_knapsack.py
```

dhruvmanila: I suppose for our use case relative imports should be de facto as we don't know from where the script is going to be called by the user.

cclauss:
Nice work.  Thanks for doing this.

PR #3977, opened 2020-11-28
Validate only submitted Project Euler solution

dhruvmanila:
Messing around with GitHub Actions, trying to see if this is possible:
We will run checks only on the submitted Project Euler solution and not all.

### **Describe your change:**

* [x] Improve Project Euler validate solutions script
* [ ] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**
* [ ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [ ] This pull request is all my own work -- I have not plagiarized.
* [ ] I know that pull requests will not be merged if they fail the automated tests.
* [ ] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [ ] All new Python files are placed inside an existing directory.
* [ ] All filenames are in all lowercase characters with no spaces or dashes.
* [ ] All functions and variable names follow Python naming conventions.
* [ ] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [ ] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [ ] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [ ] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #3981, opened 2020-11-28
Instagram Video and IGTV downloader

Epic-R-R:
Download Video and IGTV from Instagram

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/web_programming/instagram_video.py
+++ b/web_programming/instagram_video.py
@@ -14,4 +14,4 @@ if __name__ == "__main__":
     file_name = f"{datetime.now():%Y-%m-%d_%H:%M:%S}.mp4"
     with open(file_name, "wb") as fp:
         fp.write(download_video(url))
-    print(f"Done. Video saved to disk as {file_name}.")
+    print("Done. Video saved to disk as {file_name}.")
--- a/web_programming/instagram_video.py
+++ b/web_programming/instagram_video.py
@@ -3,8 +3,8 @@ from datetime import datetime
 
 import requests
 
-
-def get_video(link) -> None:
+if __name__ == "__main__":
+    link = input("Enter Video/IGTV url: ").strip()
     url = f"https://downloadgram.net/wp-json/wppress/video-downloader/video?url={link}"
     req = requests.get(url)
     res = req.content
@@ -31,8 +31,3 @@ def get_video(link) -> None:
     with open(file_name, "wb") as fp:
         fp.write(video_data)
     print(f"Done. Video saved to disk as {file_name}.")
-
-
-if __name__ == "__main__":
-    url = input("Enter Video/IGTV url: ").strip()
-    get_video(link=url)
--- a/web_programming/instagram_video.py
+++ b/web_programming/instagram_video.py
@@ -9,24 +9,8 @@ if __name__ == "__main__":
     req = requests.get(url)
     res = req.content
     json_data = json.loads(res)
-    information = [
-        json_data[0]["urls"][0]["quality"],
-        json_data[0]["urls"][0]["ext"],
-        json_data[0]["urls"][0]["size"],
-        json_data[0]["urls"][0]["src"],
-    ]
-    print(
-        f"""
-                This Is Information About Video
-        
-        Quality : {information[0]}
-        Format  : {information[1]}
-        Size    : {information[2]} 
-        SRC     : {information[3]}
-    """
-    )
     print(f"Downloading video from {link} ...")
-    video_data = requests.get(information[3]).content
+    video_data = requests.get(json_data[0]["urls"][0]["src"]).content
     file_name = f"{datetime.now():%Y-%m-%d_%H:%M:%S}.mp4"
     with open(file_name, "wb") as fp:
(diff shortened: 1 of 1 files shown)


web_programming/instagram_video.py
```diff
@@ -0,0 +1,17 @@
+import json
```
cclauss, line 1:
```suggestion
```

web_programming/instagram_video.py
```diff
@@ -0,0 +1,17 @@
...
+from datetime import datetime
+
+import requests
+
+if __name__ == "__main__":
+    link = input("Enter Video/IGTV url: ").strip()
+    url = f"https://downloadgram.net/wp-json/wppress/video-downloader/video?url={link}"
+    req = requests.get(url)
+    res = req.content
+    json_data = json.loads(res)
```
cclauss, line 11:
```suggestion
    json_data = requests.get(url).json()
```

cclauss (changes requested):
Requests has json builtin.

The first thing that I would do after running this would be to rename the file to something relevant.  Would there be a way to parse a more useful file name out of the user input?

dhruvmanila:
FYI, I am not a big fan of the way this code is written just to fool the bot to remove the `require ...` labels. Also, if possible please write tests as the code could break in a matter of months depending on the API used. Take a look at how I wrote it for `fetch_github_info`: https://github.com/TheAlgorithms/Python/blob/master/web_programming/test_fetch_github_info.py

xcodz-dot:
this is more of a command line tool, please implement the following things:
* function is used to get the data (with doctests, and typehints)
* no print or input function is used in function
* file is returned through function and not saved
* You can still have the input output (command line interface) in the `if __name__ == "__main__"` block

web_programming/instagram_video.py
```diff
@@ -0,0 +1,17 @@
+import json
+from datetime import datetime
+
+import requests
+
+if __name__ == "__main__":
+    link = input("Enter Video/IGTV url: ").strip()
+    url = f"https://downloadgram.net/wp-json/wppress/video-downloader/video?url={link}"
```
xcodz-dot, line 8:
```suggestion
def download_video(url: str) -> bytes:
    url = f"https://downloadgram.net/wp-json/wppress/video-downloader/video?url={url}"
```

web_programming/instagram_video.py
```diff
@@ -0,0 +1,17 @@
...
+    req = requests.get(url)
+    res = req.content
+    json_data = json.loads(res)
+    print(f"Downloading video from {link} ...")
```
xcodz-dot, line 12:
```suggestion
```

web_programming/instagram_video.py
```diff
@@ -0,0 +1,17 @@
...
+    video_data = requests.get(json_data[0]["urls"][0]["src"]).content
+    file_name = f"{datetime.now():%Y-%m-%d_%H:%M:%S}.mp4"
+    with open(file_name, "wb") as fp:
+        fp.write(video_data)
+    print(f"Done. Video saved to disk as {file_name}.")
```
xcodz-dot, line 17:
```suggestion
    video_data = requests.get(json_data[0]["urls"][0]["src"]).content
    return video_data
```

xcodz-dot (changes requested):
Please make the following changes first then you can add the doctests (or you can leave them because it is a downloading proggram and doctest might not be possible here)

web_programming/instagram_video.py
cclauss: My sense is that it is too confusing for the readers if we reuse `url` as both a function parameter and a local variable.

Epic-R-R:
I Update the code but in pre-commit, I have some error in the download_video function.
I use isort and black on code and it's ok

cclauss (approved):
Nice!

dhruvmanila:
So, no tests?

Epic-R-R:
Great, thank you for Update the code

xcodz-dot:
Wow, the code lines shrank too much 😄

cclauss:
;-). That’s what I do... https://github.com/cclauss/Ten-lines-or-less

dhruvmanila:
***Explicit is better than implicit.*** 😉

cclauss:
What is implicit in this code?

dhruvmanila:
Nothing, just putting this line out there :)

cclauss:
[_Namespaces are one honking great idea -- let's do more of those!_](https://www.python.org/dev/peps/pep-0020/)

cclauss:
> So, no tests?

As I wrote you before, web programming tests are difficult to write without mocking so that basically makes doctests a nonstarter.  If someone would be willing to add tests to this or another web programming algorithm, then we could use that as a guide for future web programming contributors.

dhruvmanila:
I will try to do that. In the meantime, if possible can you give a small glance to #3977 just to determine whether it's worth doing?

Also, I am in the middle of redesigning the parser for the bot and it will do more than just a comment. :)

PR #4031, opened 2020-12-12
Added solution for Project Euler problem 59

fpringle:
### **Describe your change:**

Project Euler problem 59: [XOR decryption](https://projecteuler.net/problem=59)

Each character on a computer is assigned a unique code and the preferred standard is ASCII (American Standard Code for Information Interchange). For example, uppercase A = 65, asterisk (*) = 42, and lowercase k = 107.

A modern encryption method is to take a text file, convert the bytes to ASCII, then XOR each byte with a given value, taken from a secret key. The advantage with the XOR function is that using the same encryption key on the cipher text, restores the plain text; for example, 65 XOR 42 = 107, then 107 XOR 42 = 65.

For unbreakable encryption, the key is the same length as the plain text message, and the key is made up of random bytes. The user would keep the encrypted message and the encryption key in different locations, and without both "halves", it is impossible to decrypt the message.

Unfortunately, this method is impractical for most users, so the modified method is to use a password as a key. If the password is shorter than the message, which is likely, the key is repeated cyclically throughout the message. The balance for this method is using a sufficiently long password key for security, but short enough to be memorable.

Your task has been made easy, as the encryption key consists of three lower case characters. Using p059_cipher.txt (right click and 'Save Link/Target As...'), a file containing the encrypted ASCII codes, and the knowledge that the plain text must contain common English words, decrypt the message and find the sum of the ASCII values in the original text.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


project_euler/problem_059/sol1.py
```diff
@@ -0,0 +1,121 @@
...
+    cipher_file: str = os.path.join(script_dir, "p059_cipher.txt")
+    ciphertext: List[int]
+    possibles: List[str]
+    common_word: str
+    decoded_text: str
+
+    with open(cipher_file, "r") as file:
+        data = file.read()
+
+    ciphertext = list(map(int, data.strip().split(",")))
```
dhruvmanila, line 108:
```suggestion
    ciphertext = [int(d) for d in data.strip().split(",")]
```

project_euler/problem_059/sol1.py
```diff
@@ -0,0 +1,121 @@
...
+
+    possibles = filter_valid_chars(ciphertext)
+    for common_word in COMMON_WORDS:
+        possibles = filter_common_word(possibles, common_word)
+        if len(possibles) == 1:
+            break
+
+    decoded_text = possibles[0]
+    return sum(map(ord, decoded_text))
```
dhruvmanila, line 117:
```suggestion
    return sum(ord(char) for char in decoded_text)
```

project_euler/problem_059/sol1.py
```diff
@@ -0,0 +1,121 @@
...
+
+def solution() -> int:
+    """
+    Test the ciphertext against all possible 3-character keys, then narrow down the
+    possibilities by filtering using common words until there's only one possible
+    decoded message.
+    """
+    data: str
+    script_dir: str = os.path.abspath(os.path.dirname(__file__))
+    cipher_file: str = os.path.join(script_dir, "p059_cipher.txt")
```
dhruvmanila, line 99:
```suggestion
    data: str = Path(__file__).parent.joinpath("p059_cipher.txt").read_text(encoding="utf-8")
```
Please import: `from pathlib import Path`

project_euler/problem_059/sol1.py
```diff
@@ -0,0 +1,121 @@
...
+    ciphertext: List[int]
+    possibles: List[str]
+    common_word: str
+    decoded_text: str
+
+    with open(cipher_file, "r") as file:
+        data = file.read()
+
```
dhruvmanila, line 107:
```suggestion
```
We already read it in one go :)

dhruvmanila (changes requested):

project_euler/problem_059/sol1.py
fpringle: Thanks, didn't know that!

project_euler/problem_059/sol1.py
```diff
@@ -0,0 +1,128 @@
...
+    containing common_word will be returned.
+    >>> filter_common_word(['asfla adf', 'I am here', '   !?! #a'], 'am')
+    ['I am here']
+    >>> filter_common_word(['athla amf', 'I am here', '   !?! #a'], 'am')
+    ['athla amf', 'I am here']
+    """
+    return [possible for possible in possibles if common_word in possible.lower()]
+
+
+def solution() -> int:
```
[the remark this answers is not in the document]
fpringle: @dhruvmanila I can't see any way to provide a doctest for this function since the input is read from a file. Advice?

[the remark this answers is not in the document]
dhruvmanila: If possible, add another text file containing test input and write a doctest for that file. A parameter would be required for the function whose default value will be the original text file.

This is not compulsory so if you plan not to do it inform me so that I can merge this.

dhruvmanila (approved):
LGTM! 🎉

PR #4038, opened 2020-12-16
Optimization for shell sort

shellhub:
### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [x] All function parameters and return values are annotated with Python [type hints](hxtps://docs.python.org/3/library/typing.html).

### DIFF
--- a/sorts/shell_sort.py
+++ b/sorts/shell_sort.py
@@ -1,13 +1,5 @@
 """
-This is a pure Python implementation of the shell sort algorithm
-
-For doctests run following command:
-python -m doctest -v shell_sort.py
-or
-python3 -m doctest -v shell_sort.py
-
-For manual testing run:
-python shell_sort.py
+https://en.wikipedia.org/wiki/Shellsort#Pseudocode
 """
 
 
@@ -25,15 +17,16 @@ def shell_sort(collection):
     [-45, -5, -2]
     """
     # Marcin Ciura's gap sequence
+
     gaps = [701, 301, 132, 57, 23, 10, 4, 1]
     for gap in gaps:
         for i in range(gap, len(collection)):
             insert_value = collection[i]
-            j = i - gap
-            while j >= 0 and insert_value < collection[j]:
-                collection[j + gap] = collection[j]
+            j = i
+            while j >= gap and insert_value < collection[j - gap]:
+                collection[j] = collection[j - gap]
                 j -= gap
-            collection[j + gap] = insert_value
+            collection[j] = insert_value
     return collection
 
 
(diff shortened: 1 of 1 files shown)


cclauss:
I do not quite understand the title of this PR.  In what way is the current implementation broken?

shellhub:
> I do not quite understand the title of this PR. In what way is the current implementation broken?

We should't insert value instead of swapping elements of array.

mateuszz0000:
Is your implementation improve performance or what? If so please write proper tests for proof that your solution is better than already exists.

mateuszz0000 (changes requested):
Need additional tests for performance

poyea:
Maybe we can strictly follow this?: https://en.wikipedia.org/wiki/Shellsort#Pseudocode 
Sometimes it's just a matter of styles.

poyea (changes requested):
We may add a reference to the wiki page as well.

shellhub:
@poyea Done. Please review again.

sorts/shell_sort.py
```diff
@@ -1,13 +1,5 @@
...
-This is a pure Python implementation of the shell sort algorithm
-
-For doctests run following command:
-python -m doctest -v shell_sort.py
-or
-python3 -m doctest -v shell_sort.py
-
-For manual testing run:
-python shell_sort.py
+https://en.wikipedia.org/wiki/Shellsort#Pseudocode
```
poyea, line 2: I would leave the first line like this
```suggestion
"""
This is a Python implementation of the shell sort algorithm.

Reference: https://en.wikipedia.org/wiki/Shellsort#Pseudocode
```

poyea (approved):
I would approve this as the original version has really no reference to the algorithm (or at least, the implementation didn't strictly follow the algorithm). Thank you for your pull request!🤩

sorts/shell_sort.py
```diff
@@ -25,15 +17,16 @@ def shell_sort(collection):
...
+
     gaps = [701, 301, 132, 57, 23, 10, 4, 1]
     for gap in gaps:
         for i in range(gap, len(collection)):
             insert_value = collection[i]
-            j = i - gap
-            while j >= 0 and insert_value < collection[j]:
-                collection[j + gap] = collection[j]
+            j = i
+            while j >= gap and insert_value < collection[j - gap]:
```
poyea, line 26:
```suggestion
            while j >= gap and collection[j - gap] > insert_value:
```
for readability

shellhub:
@poyea Done

shellhub:
@mateuszz0000 Can you review again. Thanks :)

mateuszz0000 (approved):

PR #4051, opened 2020-12-22
Added solution for Project Euler problem 102

fpringle:
### **Describe your change:**

Project Euler problem 102: [Triangle containment](https://projecteuler.net/problem=102) 

Three distinct points are plotted at random on a Cartesian plane, for which -1000 ≤ x, y ≤ 1000, such that a triangle is formed.

Consider the following two triangles:

A(-340,495), B(-153,-910), C(835,-947)
X(-175,41), Y(-421,-714), Z(574,-645)

It can be verified that triangle ABC contains the origin, whereas triangle XYZ does not.

Using triangles.txt (right click and 'Save Link/Target As...'), a 27K text file containing the co-ordinates of one thousand "random" triangles, find the number of triangles for which the interior contains the origin.

NOTE: The first two examples in the file represent the triangles in the example given above.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


project_euler/problem_102/sol1.py
```diff
@@ -0,0 +1,83 @@
...
+the number of triangles for which the interior contains the origin.
+
+NOTE: The first two examples in the file represent the triangles in the
+example given above.
+"""
+
+from pathlib import Path
+from typing import List, Tuple
+
+vector = Tuple[int, int]
```
dhruvmanila, line 25: Hey, I fixed this for the bot. Now if the assignment is for a type alias, the bot won't complain.
Please change the name to CamelCase and also end it with a suffix `T` or `Type` to denote it as a type alias: `VectorT` or `VectorType`

dhruvmanila (changes requested):

project_euler/problem_102/sol1.py
dhruvmanila: Also, I don't think a point or coordinate represents a `Vector`. 

Another tip: Use type alias only if it is complex nested types and multiple return variables from a function (as a tuple). I will leave it up to you to decide whether we need a type alias and what the name should be.

dhruvmanila (approved):
LGTM! 🎉

PR #4066, opened 2020-12-26
Added solution for Project Euler problem 107

fpringle:
#### Project Euler problem 107: [Minimal network](https://projecteuler.net/problem=107)

The following undirected network consists of seven vertices and twelve edges with a total weight of 243.

The same network can be represented by the matrix below.

```
        A	B	C	D	E	F	G
A	-	16	12	21	-	-	-
B	16	-	-	17	20	-	-
C	12	-	-	28	-	31	-
D	21	17	28	-	18	19	23
E	-	20	-	18	-	-	11
F	-	-	31	19	-	-	27
G	-	-	-	23	11	27	-
```
However, it is possible to optimise the network by removing some edges and still ensure that all points on the network remain connected. The network which achieves the maximum saving is shown below. It has a weight of 93, representing a saving of 243 - 93 = 150 from the original network.

Using network.txt (right click and 'Save Link/Target As...'), a 6K text file containing a network with forty vertices, and given in matrix form, find the maximum saving which can be achieved by removing redundant edges whilst ensuring that the network remains connected.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class Graph:
+    """
+    A class representing an undirected weighted graph.
+    """
+
+    def __init__(self, vertices: Set[int], edges: Dict[Edge, int]) -> None:
+        """
+        >>> graph = Graph({3,4,1,2},{(2,1):5, (1,3):10, (4,1):20, (2,4):30, (4,3):1})
```
dhruvmanila, line 40: There's no need for tests in the `__init__` function unless you do some extra steps like validation and all. In that case, you should test that the attributes are set correctly.

project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+Using network.txt (right click and 'Save Link/Target As...'), a 6K text file
+containing a network with forty vertices, and given in matrix form, find the maximum
+saving which can be achieved by removing redundant edges whilst ensuring that the
+network remains connected.
+"""
+
+import os
+from typing import Dict, List, Set, Tuple
+
+Edge = Tuple[int, int]
```
dhruvmanila, line 30: `EdgeT` or `EdgeType`: To make it easier to differentiate between a type alias and a variable.

project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class Graph:
+    """
+    A class representing an undirected weighted graph.
+    """
+
+    def __init__(self, vertices: Set[int], edges: Dict[Edge, int]) -> None:
```
dhruvmanila, line 38: Use covariant or less specific types for parameters as much as possible. In the case of `edges`, we want to either use `Mapping` or `MutableMapping` depending on what actions you want to perform. This will mean a user can pass in any Dict like object and we are not enforcing to pass exactly the `dict` object.

project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+        [1, 2, 3, 4]
+        >>> sorted([(v,k) for k,v in graph.edges.items()])
+        [(1, (3, 4)), (5, (1, 2)), (10, (1, 3)), (20, (1, 4)), (30, (2, 4))]
+        """
+        self.vertices: Set[int] = vertices
+        self.edges: Dict[Edge, int] = {
+            (min(edge), max(edge)): weight for edge, weight in edges.items()
+        }
+
+    def add_edge(self, edge: Edge, weight: int) -> None:
```
dhruvmanila, line 51: Please provide a short description of what this method does.

project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+        [1, 2, 3]
+        >>> sorted([(v,k) for k,v in graph.edges.items()])
+        [(4, (1, 2)), (5, (1, 3))]
+        """
+        self.vertices.add(edge[0])
+        self.vertices.add(edge[1])
+        self.edges[(min(edge), max(edge))] = weight
+
+
+def prims_algorithm(graph: Graph) -> Graph:
```
dhruvmanila, line 65: Why don't we make this a method of the `Graph` class as you actually are working on the `Graph` object?

project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    Run Prim's algorithm on a Graph to find the minimum spanning tree.
+    Reference: https://en.wikipedia.org/wiki/Prim%27s_algorithm
```
dhruvmanila, line 68: Please provide the reference link in the module-level docstring as well.

project_euler/problem_107/sol1.py
```diff
@@ -0,0 +1,131 @@
...
+    >>> sorted(mst.vertices)
+    [1, 2, 3, 4]
+    >>> sorted(mst.edges)
+    [(1, 2), (1, 3), (3, 4)]
+    """
+    subgraph: Graph = Graph({min(graph.vertices)}, {})
+    min_edge: Edge
+    min_weight: int
+    edge: Edge
+    weight: int
```
dhruvmanila, line 80: This is just a suggestion and you don't have to make any changes:

There's no need to provide a type hint for every variable as whichever static type checker you are using will probably be able to infer the variable type according to the assignment. In simple cases like these, it is not required.

dhruvmanila (changes requested):
Good start! 👍

project_euler/problem_107/sol1.py
fpringle: When I do this python throws an error on the return type hint "Graph", as it hasn't been fully declared yet

dhruvmanila: That is why we have [forward references](https://www.python.org/dev/peps/pep-0563/) ;)
You need to quote it like so: `def prims_algorithm(self) -> "Graph":`. Also, from Python 3.10, this won't be required as this will become the default behavior.

fpringle: Cool, thanks!

dhruvmanila (approved):

PR #4069, opened 2020-12-27
Format Code

shellhub:
### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -499,6 +499,7 @@
   * [Minimum Cut](https://github.com/TheAlgorithms/Python/blob/master/networking_flow/minimum_cut.py)
 
 ## Neural Network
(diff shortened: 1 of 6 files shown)


dhruvmanila:
This does not seem like an enhancement. You're removing the declaration of type for the `left` and `right` attributes. Another thing is assigning the same value to multiple variables can lead to surprising behavior in Python when dealing with mutable objects. As `left` and `right` attributes are also `Node` instances, they are mutable objects. Also, declaring types for multiple assignments is not possible.

shellhub:
I learn this skill from https://www.w3schools.com/python/gloss_python_assign_value_to_multiple_variables.asp

dhruvmanila:
Yes, it is possible but you have to see the context as well. `left` and `right` variables can be two values as suggested by the type hint: `Optional[Node]` and we're moving towards a fully annotated repository which makes this an invalid syntax:
```python
self.left: Optional[int] = self.right: Optional[int] = None
```

Issue #4052, opened 2020-12-23
[mypy] Need help to fix all `mypy` errors in the codebase


dhruvmanila:
# Just one left to fix...
https://github.com/TheAlgorithms/Python/blob/master/mypy.ini#L5
* [ ] other/least_recently_used.py
* [x] other/lfu_cache.py #5755
* [x] other/lru_cache.py #5755

---

__UPDATE:__ Our GitHub Actions now run `mypy --ignore-missing-imports` excluding those directories that fail that test.
* https://github.com/TheAlgorithms/Python/blob/master/mypy.ini#L5

Currently, we are not running `mypy` in our regular CI tests as there are a lot of errors in the entire codebase, which needs to be fixed. This won't be a one-person job, so we are asking for help from you. I cannot paste the entire message in here as there are around 600 of them, so here's just a gist of it:

```console
$ mypy --ignore-missing-imports .
strings/word_occurrence.py:17: error: Need type annotation for 'occurrence'
strings/min_cost_string_conversion.py:36: error: No overload variant of "__setitem__" of "list" matches argument types "int", "str"
strings/min_cost_string_conversion.py:36: note: Possible overload variants:
strings/min_cost_string_conversion.py:36: note:     def __setitem__(self, int, int) -> None
strings/min_cost_string_conversion.py:36: note:     def __setitem__(self, slice, Iterable[int]) -> None
strings/min_cost_string_conversion.py:40: error: No overload variant of "__setitem__" of "list" matches argument types "int", "str"
strings/min_cost_string_conversion.py:40: note: Possible overload variants:
strings/min_cost_string_conversion.py:40: note:     def __setitem__(self, int, int) -> None
strings/min_cost_string_conversion.py:40: note:     def __setitem__(self, slice, Iterable[int]) -> None
...
backtracking/n_queens_math.py:109: error: List comprehension has incompatible type List[str]; expected List[int]
backtracking/n_queens_math.py:110: error: Argument 1 to "append" of "list" has incompatible type "List[int]"; expected "List[str]"
backtracking/n_queens_math.py:149: error: Need type annotation for 'boards' (hint: "boards: List[<type>] = ...")
backtracking/minimax.py:15: error: "list" is not subscriptable, use "typing.List" instead
backtracking/knight_tour.py:6: error: "tuple" is not subscriptable, use "typing.Tuple" instead
backtracking/knight_tour.py:6: error: "list" is not subscriptable, use "typing.List" instead
...
```

[...]

PR #4143, opened 2021-01-24
[mypy] Add/fix type annotations for hashes. Fix for binary_tree_traversals

barklan:
### **Describe your change:**

Related Issue:  #4052 
Fix incorrect confusing type annotations in hashes

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/hashes/chaos_machine.py
+++ b/hashes/chaos_machine.py
@@ -6,7 +6,8 @@ t = 3
 m = 5
 
 # Buffer Space (with Parameters Space)
-buffer_space, params_space = [], []
+buffer_space: list[int] = []
+params_space: list[int] = []
 
 # Machine Time
 machine_time = 0
--- a/traversals/binary_tree_traversals.py
+++ b/traversals/binary_tree_traversals.py
@@ -188,7 +188,7 @@ def pre_order_iter(node: TreeNode) -> None:
     """
     if not isinstance(node, TreeNode) or not node:
         return
-    stack: List[TreeNode] = []
+    stack: list[TreeNode] = []
     n = node
     while n or stack:
         while n:  # start from root node, find its left child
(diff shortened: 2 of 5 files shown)


hashes/chaos_machine.py
```diff
@@ -6,7 +6,8 @@
 m = 5
 
 # Buffer Space (with Parameters Space)
-buffer_space, params_space = [], []
+buffer_space: list[int] = []
```
poyea, line 9:
```suggestion
buffer_space: List[int] = []
```
perhaps?

traversals/binary_tree_traversals.py
```diff
@@ -188,7 +188,7 @@ def pre_order_iter(node: TreeNode) -> None:
     """
     if not isinstance(node, TreeNode) or not node:
         return
-    stack: List[TreeNode] = []
+    stack: list[TreeNode] = []
```
poyea, line 191:
```suggestion
    stack: List[TreeNode] = []
```

hashes/chaos_machine.py
mrmaxguns: This works in python 3.9, since [type hinting generics were added in standard collections](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections), although this probably still needs to be refactored to `List[]` since this feature isn't implemented in python 3.7.

Rada8732 (approved):

hashes/chaos_machine.py
dhruvmanila: We are switching to Python 3.9 generic types, so this is fine. Also, `mypy` supports it as well.

dhruvmanila:
Hi @barklan, did you run `mypy` locally because running it on my machine yields a lot of errors in some of the files. Can you take a look into them? You will have to install `mypy` and the required imports, if there are any in all the pull request files.

dhruvmanila (changes requested):
Hi @barklan, did you run mypy locally because running it on my machine yields a lot of errors in some of the files. Can you take a look into them? You will have to install mypy and the required imports, if there are any in all the pull request files.

cclauss:
This work was done in other pull requests.

PR #4181, opened 2021-02-06
Improved emails from url

AyushSehrawat:
### **Describe your change:**

This version finds more email from the url and also find the urls from the given url and also find emails from them. It all depends on the range given by the user.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/web_programming/emails_from_url.py
+++ b/web_programming/emails_from_url.py
@@ -2,16 +2,20 @@
 Author - AyushSehrawat
 """
 
-# Imports the needed modules
-from bs4 import BeautifulSoup
-import requests
-import requests.exceptions
+import re
 import urllib.parse
 from collections import deque
-import re
+
+import requests
+import requests.exceptions
+
+# Imports the needed modules
+from bs4 import BeautifulSoup
 
 """ Function to get emails from the url """
-def getMails(num : int) -> None:
+
+
+def getMails(num: int) -> None:
     count = 0
     """
     Keep a check on range
@@ -29,44 +33,50 @@ def getMails(num : int) -> None:
             base_url = "{0.scheme}://{0.netloc}".format(parts)
             # Forms a base url
 
-            path = url[:url.rfind('/') + 1] if '/' in parts.path else url
+            path = url[: url.rfind("/") + 1] if "/" in parts.path else url
 
             try:
                 response = requests.get(url)
-            except (requests.exceptions.MissingSchema,
-                    requests.exceptions.ConnectionError):
+            except (
+                requests.exceptions.MissingSchema,
+                requests.exceptions.ConnectionError,
+            ):
                 continue
 
             new_emails = set(
-                re.findall(r"[a-zA-Z0-9+._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+",
-                        response.text, re.I))
+                re.findall(
+                    r"[a-zA-Z0-9+._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+",
+                    response.text,
+                    re.I,
+                )
+            )
             # A regex search to find emails
             emails.update(new_emails)
 
-            soup = BeautifulSoup(response.text, features='lxml')
+            soup = BeautifulSoup(response.text, features="lxml")
 
             """
             Find links
             """
             for anchor in soup.find_all("a"):
-                link = anchor.attrs['href'] if 'href' in anchor.attrs else ''
-                if link.startswith('/'):
+                link = anchor.attrs["href"] if "href" in anchor.attrs else ""
+                if link.startswith("/"):
                     link = base_url + link
-                elif not link.startswith('http'):
+                elif not link.startswith("http"):
                     link = path + link
-                if not link in urls and not link in scraped_urls:
+                if not (link in urls or link in scraped_urls):
                     urls.append(link)
 
     except KeyboardInterrupt:
-        print('[-] Closing')
+        print("[-] Closing")
 
 
-if __name__ == '__main__':
-    user_url = "https://google.com" # Example Url
+if __name__ == "__main__":
+    user_url = "https://google.com"  # Example Url
     urls = deque([user_url])
 
     """
-    These set store emails and prevent repeating emails 
+    These set store emails and prevent repeating emails
     """
     scraped_urls = set()
     emails = set()
--- a/web_programming/emails_from_url.py
+++ b/web_programming/emails_from_url.py
@@ -1,103 +1,86 @@
-"""Get the site emails from URL."""
-__author__ = "Muhammad Umer Farooq"
-__license__ = "MIT"
-__version__ = "1.0.0"
-__maintainer__ = "Muhammad Umer Farooq"
-__email__ = "contact@muhammadumerfarooq.me"
-__status__ = "Alpha"
+"""
+Author - AyushSehrawat
+"""
 
-import re
-from html.parser import HTMLParser
-from urllib import parse
-
+# Imports the needed modules
+from bs4 import BeautifulSoup
 import requests
+import requests.exceptions
+import urllib.parse
+from collections import deque
+import re
 
-
-class Parser(HTMLParser):
-    def __init__(self, domain: str):
-        HTMLParser.__init__(self)
-        self.data = []
-        self.domain = domain
-
-    def handle_starttag(self, tag: str, attrs: str) -> None:
-        """
-        This function parse html to take takes url from tags
-        """
-        # Only parse the 'anchor' tag.
-        if tag == "a":
-            # Check the list of defined attributes.
-            for name, value in attrs:
-                # If href is defined, and not empty nor # print it.
-                if name == "href" and value != "#" and value != "":
-                    # If not already in data.
-                    if value not in self.data:
-                        url = parse.urljoin(self.domain, value)
-                        self.data.append(url)
-
-
-# Get main domain name (example.com)
-def get_domain_name(url: str) -> str:
+""" Function to get emails from the url """
+def getMails(num : int) -> None:
+    count = 0
     """
-    This function get the main domain name
-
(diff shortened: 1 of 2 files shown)


AyushSehrawat:
Finally solved all the issues 👍

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
...
 
-# Get sub domain name (sub.example.com)
-def get_sub_domain_name(url: str) -> str:
+def getMails(num: int) -> None:
+    count = 0
     """
-    >>> get_sub_domain_name("https://a.b.c.d/e/f?g=h,i=j#k")
-    'a.b.c.d'
-    >>> get_sub_domain_name("Not a URL!")
-    ''
```
cclauss, line 57: Why are we REMOVING tests?!?

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
...
+    Keep a check on range
     """
-    return parse.urlparse(url).netloc
+    try:
+        while len(urls):
+            count += 1
+            if count == num:
+                break
```
cclauss, line 27:
```suggestion
        while urls:
```

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
...
+            """
+            Find links
+            """
+            for anchor in soup.find_all("a"):
+                link = anchor.attrs["href"] if "href" in anchor.attrs else ""
+                if link.startswith("/"):
+                    link = base_url + link
+                elif not link.startswith("http"):
+                    link = path + link
+                if not (link in urls or link in scraped_urls):
```
cclauss, line 67:
```suggestion
                if link not in urls and link not in scraped_urls:
```
See PEP8.

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
...
+                    urls.append(link)
+
+    except KeyboardInterrupt:
+        print("[-] Closing")
```
cclauss, line 71: Closing what?!?

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
...
+    """
+    These set store emails and prevent repeating emails
+    """
+    scraped_urls = set()
+    emails = set()
 
-    # Finally return a sorted list of email addresses with no duplicates.
-    return sorted(valid_emails)
+    """
+    Num = How much loops you want )
```
cclauss, line 85: Let's make this a comment instead of a str.

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
...
 
+    """
+    num = 10
+    getMails(num)
 
-if __name__ == "__main__":
-    emails = emails_from_url("https://github.com")
-    print(f"{len(emails)} emails found:")
-    print("\n".join(sorted(emails)))
+    """
```
cclauss, line 91: Let's make this a comment instead of a str.

web_programming/emails_from_url.py
```diff
@@ -1,103 +1,96 @@
-"""Get the site emails from URL."""
-__author__ = "Muhammad Umer Farooq"
-__license__ = "MIT"
-__version__ = "1.0.0"
-__maintainer__ = "Muhammad Umer Farooq"
-__email__ = "contact@muhammadumerfarooq.me"
-__status__ = "Alpha"
+"""
+Author - AyushSehrawat
```
cclauss, line 2: Perhaps instead of deleting the work of someone else, you should create a `emails_from_url_alt.py` implementation.  I like the current implementation and it would be good to be able to see the two side-by-side.

cclauss (changes requested):

web_programming/emails_from_url.py
AyushSehrawat: Can you elaborate more? You mean a test case?

web_programming/emails_from_url.py
AyushSehrawat: Oops, i was adding sys.exit(1) there. I will do

web_programming/emails_from_url.py
AyushSehrawat: So you mean, close this pr and create a new one?

AyushSehrawat:
I will consider your changes 😊.

web_programming/emails_from_url.py
AyushSehrawat: @cclauss So close this pr ? Or make a new file and commit using this pr , and let the original be same ?

AyushSehrawat:
`getmails` function returns nothing. Instead it adds things to the sets. So, ig there will be no doctest for it.

AyushSehrawat:
Waiting for review :)

web_programming/emails_from_url_alt.py
```diff
@@ -0,0 +1,104 @@
...
+            try:
+                response = requests.get(url)
+            except (
+                requests.exceptions.MissingSchema,
+                requests.exceptions.ConnectionError,
+            ):
+                continue
+
+            new_emails = set(
+                re.findall(
```
cclauss, line 56:
```suggestion
                email_pattern = r"[a-zA-Z0-9+._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+"
                new_emails = set(re.findall(email_pattern, response.text, re.I))
```

cclauss (changes requested):

web_programming/emails_from_url_alt.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+Info -
+Function to get emails from the url,
+It first find the email and links on first page , then go to the
+links find and find some more links
+and emails. This process goes on , depending on the loops
+"""
+
+
+def getmails(num: int) -> None:
```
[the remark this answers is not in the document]
cclauss: Should the list be ordered by first found or alphabetically?

web_programming/emails_from_url_alt.py
```diff
@@ -0,0 +1,104 @@
...
+Author - AyushSehrawat
+"""
+
+import re
+import sys
+import urllib.parse
+from collections import deque
+
+import requests
+import requests.exceptions
```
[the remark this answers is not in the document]
cclauss: Will make code cleaner and faster.

web_programming/emails_from_url.py
```diff
@@ -39,7 +39,6 @@ def handle_starttag(self, tag: str, attrs: str) -> None:
 def get_domain_name(url: str) -> str:
     """
     This function get the main domain name
-
```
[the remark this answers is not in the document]
AyushSehrawat: I didn't change the original file. It is same as before @cclauss

web_programming/emails_from_url_alt.py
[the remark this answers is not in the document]
AyushSehrawat: max_emails can be a wrong variable , since the int value is a loop , which runs to find the links and emails.
We can't tell the max emails.

web_programming/emails_from_url_alt.py
```diff
@@ -0,0 +1,104 @@
...
+                    urls.append(link)
+
+    except KeyboardInterrupt:
+        print("[-] Closing")
+        sys.exit(1)
+
+
+if __name__ == "__main__":
+    user_url = "https://google.com"  # Example Url
+    urls = deque([user_url])
```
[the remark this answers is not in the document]
AyushSehrawat: this get_emails() function doesn't return anything. It just add them to set ( to prevent same emails )

web_programming/emails_from_url_alt.py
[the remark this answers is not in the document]
cclauss: `while urls and len(emails) < max_emails:`

AyushSehrawat:
I tried to make code as much simple as possible. I can also return the set ( to add doctest )

web_programming/emails_from_url_alt.py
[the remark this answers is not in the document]
cclauss: Now that I seen how this works, I think we need to change the function signature:
`def get_mails(urls: list, max_emails: int = 5000) -> list:`

AyushSehrawat:
@cclauss I tried to create a doctest for it, but my code is dependant on sets which are declared in 
```bash
if __name__ == "__main__":`
```
Also, now the getmails function now returns set of emails

web_programming/emails_from_url_alt.py
```diff
@@ -88,6 +87,8 @@ def getmails(num: int) -> None:
             print(mail)
         sys.exit(1)
 
+    return emails
+
 
 if __name__ == "__main__":
     user_url = "https://google.com"  # Example Url
```
cclauss: Let's just replace ___all___ of `__main__` with:
```suggestion
if __name__ == "__main__":
    print("\n".join(getmails("https://google.com")))
```

web_programming/emails_from_url_alt.py
```diff
@@ -88,6 +87,8 @@ def getmails(num: int) -> None:
             print(mail)
         sys.exit(1)
 
+    return emails
```
cclauss, line 90:
```suggestion
    return list(sorted(emails))
```

web_programming/emails_from_url_alt.py
AyushSehrawat: It takes `num` as argument. I tried to add url too , but didn't work the way i want, and also `popleft` is faster then `pop(0)`  (if using list). So set would be faster.

AyushSehrawat:
@cclauss Been 8 days since i asked for review. Adding doctest is like changing the whole code. Also this is an alternative to the original code. Please look into it :)

AyushSehrawat:
@cclauss Shall i close the pull request ? Since no action is being taken. I fulfilled most of the needs except `doctest` :(

cclauss:
No.  Please leave it open.

PR #4207, opened 2021-02-15
Implementation of the algorithm for the Koch snowflake

algobytewise:
Implementation of the algorithm for the Koch snowflake

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


algobytewise:
Hello, it would be great if I could get some feedback on how to respond to the failed tests.
(1) algorithms-keeper complains that VECTOR_1, VECTOR_2 and VECTOR_1 are not snake_case. But they are intended as constants, not as variables.
(2) I'm not sure what doctest to implement for the plot-function. It is not part of the algorithm, it just draws a graph and it has no return value.

mrmaxguns:
The bot just gives guidance. Tests failed because you didn't run isort. Here's the isort diff:
```
-import numpy
 
+import numpy
 
 # initial triangle of Koch snowflake
 VECTOR_1 = numpy.array([0, 0])
```
As you can see, a newline was added between `from __future__ import annotations` and `import numpy`. The constants are fine. For the tests, you should probably add a small note in the relevant function why you chose not to include tests, so that future contributors know.

algobytewise:
Dear mrmaxguns, thank you for the feedback, I have implemented the changes you suggested.

cclauss:
This looks good.  Nice work.  I am not a fan of uncategorized algorithms in the `other` directory because it makes it harder for visitors to find algorithms of interest.  Would it make sense to move this to the `cellular_automata` or `graphics` directory instead?

other/koch_snowflake.py
```diff
@@ -0,0 +1,123 @@
...
+    axes.set_aspect("equal")
+
+    # matplotlib.pyplot.plot takes a list of all x-coordinates and a list of all
+    # y-coordinates as inputs, which need to be constructed from our vector-list
+    x_coordinates = []
+    for vector in vectors:
+        x_coordinates.append(vector[0])
+    y_coordinates = []
+    for vector in vectors:
+        y_coordinates.append(vector[1])
```
cclauss, line 111: Please make these `list comprehensions`.

other/koch_snowflake.py
```diff
@@ -0,0 +1,123 @@
...
+    Loops through each pair of adjacent vectors. Each line between two adjacent
+    vectors is divided into 4 segments by adding 3 additional vectors in-between
+    the original two vectors. The vector in the middle is constructed through a
+    60 degree rotation so it is bent outwards.
+    >>> iteration_step([numpy.array([0, 0]), numpy.array([1, 0])])
+    [array([0, 0]), array([0.33333333, 0.        ]), array([0.5       , \
+0.28867513]), array([0.66666667, 0.        ]), array([1, 0])]
+    """
+    new_vectors = []
+    for i in range(len(vectors) - 1):
```
cclauss, line 64: `range(len())` is almost always a sign that `enumerate()` could be used instead.

other/koch_snowflake.py
cclauss: Or maybe even slicker, use `zip()` instead.

algobytewise:
I have implemented the changes, thanks for the advice on better looping. I think `graphics` is a better fit so I've put it there. My preferred solution would be to have a new directory for `fractals` and move it there together with `other/sierpinski_triangle.py`.

cclauss (approved):
LGTM

PR #4244, opened 2021-02-28
Add geometric_mean.py

ayushbisht2001:
### **Describe your change:**
Added an algorithm to compute the geometric mean of series

* [x] Add an algorithm

### **Checklist:**


maths/series/geometric_mean.py
```diff
@@ -0,0 +1,55 @@
...
+    True
+    >>> is_geometric_series([3, 6, 12, 24])
+    True
+    >>> is_geometric_series([1, 2, 3])
+    False
+
+    """
+    if len(series) == 1:
+        return True
+    common_ratio = series[1] / series[0]
```
mateuszz0000, line 15: Will fail with empty series.

mateuszz0000: Also if series[0] is 0 will fail as well

maths/series/geometric_mean.py
```diff
@@ -0,0 +1,55 @@
...
+    ValueError: Input list must be a non empty list
+
+    """
+    if len(series) == 0:
+        raise ValueError("Input list must be a non empty list")
+    if not is_geometric_series(series):
+        raise ValueError("Input list is not a geometric series")
+    answer = 1
+    for _ in series:
+        answer *= _
```
mateuszz0000, line 48: I don't like it named `_`. `_` means that it's not important but actually it's affecting `answer` variable.

maths/series/geometric_mean.py
```diff
@@ -0,0 +1,55 @@
+def is_geometric_series(series: list) -> bool:
```
mateuszz0000, line 1: Please add link with some theory about algorithm.

mateuszz0000 (changes requested):

maths/series/geometric_mean.py
ayushbisht2001: > Will fail with empty series.

for that I  have check the size of list. If it is zero then It will raise an error. 

```python
if len(series) == 0:
     raise ValueError("Input list must be a non empty list")
```
Would it be fine or  ... how can I do it in a better way ?

maths/series/geometric_mean.py
ayushbisht2001: > I don't like it named `_`. `_` means that it's not important but actually it's affecting `answer` variable.

sorry for this silly mistake.

maths/series/geometric_mean.py
mateuszz0000: Right, didn't notice.

mateuszz0000 (approved):

PR #4251, opened 2021-03-02
Add decimal_to_binary.py

ayushbisht2001:
### **Describe your change:**
I have added an algorithm to find the binary string representation of decimal numbers.

* [x] Add an algorithm?

### **Checklist:**

### DIFF
--- a/bit_manipulation/decimal_to_binary.py
+++ b/bit_manipulation/decimal_to_binary.py
@@ -32,7 +32,6 @@ def decimal_to_binary(number: int) -> str:
     while number:
         answer = str(number % 2) + answer
         number = number >> 1
-    # answer = answer[-1::-1]
     if flag:
         answer = "-" + answer
     return answer
--- a/bit_manipulation/decimal_to_binary.py
+++ b/bit_manipulation/decimal_to_binary.py
@@ -1,0 +1,44 @@
+def decimal_to_binary(number: int) -> str:
+    """
+    Return a binary string
+    >>> decimal_to_binary(0)
+    '0'
+    >>> decimal_to_binary(1)
+    '1'
+    >>> decimal_to_binary(8)
+    '1000'
+    >>> decimal_to_binary(110)
+    '1101110'
+    >>> decimal_to_binary(-20)
+    '-10100'
+    >>> decimal_to_binary("abc")
+    Traceback (most recent call last):
+        ...
+    ValueError: Input is not valid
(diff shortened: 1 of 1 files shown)


algobytewise:
It seems to me that this performs the same function as the built-in function "bin".

ayushbisht2001:
> It seems to me that this performs the same function as the built-in function "bin".

Exactly,

bit_manipulation/decimal_to_binary.py
```diff
@@ -0,0 +1,43 @@
+def decimal_to_binary(number: int) -> str:
```
mateuszz0000, line 1: I don't see any decimal number in the tests. Please provide tests for numbers like `13.321312312`

mateuszz0000 (changes requested):

bit_manipulation/decimal_to_binary.py
Joe-Sin7h, line 1: number need not be `int`, it can be `float` too since you have named your function as `decimal_to_binary`.  
1 and 1.5 are both decimal numbers because 1 can also be written as 1.0 or 0.01 * 10**2. 
if you are confused about decimal numbers check this [Wikipedia link](https://en.wikipedia.org/wiki/Decimal)

Joe-Sin7h (changes requested):

bit_manipulation/decimal_to_binary.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    if not isinstance(number, int):
+        raise ValueError("Input is not valid")
+    if number == 0:
+        return "0"
+    flag = 0
+    if number < 0:
+        number *= -1
+        flag = 1
+    answer = ""
```
Joe-Sin7h, line 31: Please write code that can actually convert decimal to binary, your code just converts int to binary. Even the build-in function `bin` can do that.

I have written a sample code to do so since I only have view permission, I want you to make some amendments.

```Python
a = ""
r = ""
count = 0
d = number - int(number)
number = int(number)
while(d and count<10):
     d = d*2
     r+=str(int(d))
     if int(d)==d:
            break
     else:
            d = d -int(d)
            count+=1
while number:
     a = str(number%2) + a
     number = number >> 1
a = a+"."+r
print(a)
```

Joe-Sin7h (changes requested):

PR #4292, opened 2021-03-23
[mypy] fix small folders

algobytewise:
### **Describe your change:**
Related Issue: #4052

comment for strassen_matrix_multiplication.py: I couldn't use Tuple as return type since one branch of the function returns a tuple an another a list/matrix. So I turned the tuple into a list instead.

comment for electric_power.py & ohms_law.py: I added a final else-clause for raising an error to make sure that each branch either has a return-value or throws an error

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


.github/workflows/build.yml
```diff
@@ -29,12 +29,16 @@ jobs:
...
+           electronics
+           file_transfer
            fractals
            fuzzy_logic
            genetic_algorithm
            geodesy
            knapsack
            networking_flow
+           quantum
            scheduling sorts
```
cclauss, line 42:
```suggestion
           scheduling
           sorts
```

file_transfer/receive_file.py
```diff
@@ -13,7 +13,7 @@
         print("Receiving data...")
         while True:
             data = sock.recv(1024)
-            print(f"data={data}")
+            print(f"data={str(data)}")
```
cclauss, line 16: Why this change?

file_transfer/send_file.py
```diff
@@ -13,7 +13,7 @@ def send_file(filename: str = "mytext.txt", testing: bool = False) -> None:
         conn, addr = sock.accept()  # Establish connection with client.
         print(f"Got connection from {addr}")
         data = conn.recv(1024)
-        print(f"Server received {data}")
+        print(f"Server received {str(data)}")
```
cclauss, line 16: Why this change?

divide_and_conquer/max_difference_pair.py
```diff
@@ -1,7 +1,7 @@
-from typing import List
+from typing import List, Tuple
```
cclauss, line 1: @dhruvmanila Do we need these imports for basic types in Python 3.9 and the current mypy?

file_transfer/receive_file.py
dhruvmanila: This and similar changes are redundant as string interpolation converts the object to string using the `__str__` method of the object.

divide_and_conquer/max_difference_pair.py
dhruvmanila: No, I updated the issue with the message: https://github.com/TheAlgorithms/Python/issues/4052#issuecomment-783919171

file_transfer/receive_file.py
algobytewise: I'm not sure why but without it we get the following mypy-errors:
file_transfer/send_file.py:16: error: On Python 3 '{}'.format(b'abc') produces "b'abc'", not 'abc'; use '{!r}'.format(b'abc') if this is desired behavior
file_transfer/receive_file.py:16: error: On Python 3 '{}'.format(b'abc') produces "b'abc'", not 'abc'; use '{!r}'.format(b'abc') if this is desired behavior

divide_and_conquer/max_difference_pair.py
algobytewise: I see, I made the corresponding changes. One difference is that the basic types are spelled lower case with this import

file_transfer/receive_file.py
algobytewise: I guess it has something to do with the method from the socket-package that returns the data-object.

cclauss: `bytes` vs. `str` is the most time-consuming part of porting code from Python 2 to Python 3.
% __python3__
```
>>> type(b"abc")
<class 'bytes'>
>>> type("abc")
<class 'str'>
```

file_transfer/send_file.py
cclauss, line 16:
```suggestion
        print(f"Server received: {data = }")
```

cclauss (approved):
LGTM...

PR #4294, opened 2021-03-27
Merge directories

algobytewise:
### **Describe your change:**
The directories "boolean_algebra" & "traversals" each contained only one file. I renamed the bit_manipulation-directory to "binary" and moved these 2 files there to decrease the number of root-directories.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


cclauss:
### Too many files changed in a single PR.

https://en.wikipedia.org/wiki/Boolean_algebra and https://en.wikipedia.org/wiki/Bit_manipulation but there is no entry for binary algorithms and a search on `binary algorithms wiki` leads mostly to binary search which is not in this directory.

My sense is that:
1. binary/binary_tree_traversals.py --> searches/binary_tree_traversal.py (remove the trailing `s`)
    * This is not idle tourism -- we are traversing the tree to find something(s).
2. We should leave the other two directories as they are today.
3. We should open a new issue encouraging the contributions of new `boolean algebra` algorithms
    * Ensure that https://en.wikipedia.org/wiki/Boolean_algebra is added to the issue to spark ideas
    * @mention in that issue whoever has contributed to the existing `boolean algebra` algorithm.

algobytewise:
I'm not sure about the boolean_algebra-directory. The corresponding [Wikipedia category](https://en.wikipedia.org/wiki/Category:Boolean_algebra) contains a few more algorithms, though. But I agree that the searches-directory is a better fit for `binary_tree_traversals.py`. I'll get to work on that.

cclauss:
The category cited above does not include bit manipulation.  My sense is that bit manipulation is more of a computer science thing while boolean algebra is more of a mathematics thing.  The former might be useful for solve the latter on computers but they are two separate things.  The category cited above shows that there are tons of possibilities for contributions to make the boolean algebra directory richer.  Thanks for all your hard work!!

PR #4315, opened 2021-04-05
feat: Add mypy configuration file

dhruvmanila:
### **Describe your change:**

* [x] New feature
* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,9 +21,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip setuptools six wheel
           python -m pip install mypy pytest-cov -r requirements.txt
-      # FIXME: #4052 fix mypy errors in the exclude directories and remove them below
-      - run: mypy --ignore-missing-imports
-          --exclude '(data_structures|digital_image_processing|dynamic_programming|graphs|maths|matrix|other|project_euler|scripts|searches|strings*)/$' .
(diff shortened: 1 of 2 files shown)


cclauss (commented):
If we only run mypy in our GitHub Action then I would be opposed to storing configuration in two separate files.  However if we add mypy to https://github.com/TheAlgorithms/Python/blob/master/.pre-commit-config.yaml then I would be for this change.

dhruvmanila:
For now, we are only using `ignore-missing-imports` flag which is available for the CLI `mypy` but some flags are not available and as we get more stricter in the future we would need to have additional flags and I would prefer it to be added in its own configuration file.

Also, when we test `mypy` locally we would not need to pass any flags and directly run `mypy` as: `$ mypy .` which is better and shorter to type than `$ mypy --ignore-missing-imports --warn-unused-imports --disallow-untyped-defs ...`

dhruvmanila:
Hmm, weird. It seems that all of the builds are stuck from my last commit to master: https://github.com/TheAlgorithms/Python/actions

cclauss:
Something strange is happening...

dhruvmanila:
I have reported to GitHub. Lets see what's going on.

cclauss (approved):
LGTM

PR #4321, opened 2021-04-06
Pre-commit: add mypy

cclauss:
### **Describe your change:**

* [ ] Add mypy tests to pre-commit
* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,8 @@ repos:
     rev: v0.812
     hooks:
       - id: mypy
+        args:
+          - --config-file mypy.ini
   - repo: https://github.com/codespell-project/codespell
     rev: v2.0.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,13 +30,10 @@ repos:
           - --ignore=E203,W503
           - --max-complexity=25
           - --max-line-length=88
-#     FIXME: fix mypy errors and then uncomment this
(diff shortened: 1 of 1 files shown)


dhruvmanila:
As mentioned in this comment https://github.com/TheAlgorithms/Python/pull/4315#issuecomment-813341184, there's going to be problem when running `mypy` in `pre-commit`.

> The problem with running `mypy` in pre-commit is that pre-commit runs every hook in its own virtual environment and as `mypy` needs the information from project dependencies, they need to be installed there as well. `pre-commit` does provide an option: `- additional_dependencies: ['numpy', ...]` but that would mean we need to maintain the dependencies in `requirements.txt` and in `pre-commit-config.yaml`.

```yaml
- repo: https://github.com/pre-commit/mirrors-mypy
    rev: v0.812
    hooks:
      - id: mypy
        additional_dependencies: [numpy, beautifulsoup4, requests, ...]
```
Ref:
- https://pre-commit.com/#config-additional_dependencies
- https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit (3rd point)

dhruvmanila:
`mypy` or any other type checkers should be run in an environment which contains the project dependencies for it to correctly report errors. If we are using any function/class/variable from other modules then `mypy` will know whether we are passing the correct type of arguments or assigning the correct type to the variable, etc.

cclauss:
Makes sense.

PR #4336, opened 2021-04-17
binary_tree_traversals.py: Simplify with dataclasses

cclauss:
### **Describe your change:**

* [ ] Add an algorithm?
* [x] Optimize a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/data_structures/binary_tree/binary_tree_traversals.py
+++ b/data_structures/binary_tree/binary_tree_traversals.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 
 
 @dataclass
-class Node(object):
+class Node:
     data: int
     left: "Node" = None
     right: "Node" = None
--- a/data_structures/binary_tree/binary_tree_traversals.py
+++ b/data_structures/binary_tree/binary_tree_traversals.py
@@ -1,24 +1,16 @@
 # https://en.wikipedia.org/wiki/Tree_traversal
+from dataclasses import dataclass
 
 
-class Node:
-    """
-    A Node has data variable and pointers to its left and right nodes.
-    """
-
-    def __init__(self, data):
-        self.left = None
-        self.right = None
-        self.data = data
+@dataclass
+class Node(object):
+    data: int
+    left: "Node" = None
+    right: "Node" = None
 
 
 def make_tree() -> Node:
-    root = Node(1)
-    root.left = Node(2)
-    root.right = Node(3)
-    root.left.left = Node(4)
-    root.left.right = Node(5)
-    return root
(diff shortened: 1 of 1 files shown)


data_structures/binary_tree/binary_tree_traversals.py
```diff
@@ -1,24 +1,16 @@
...
-    """
-    A Node has data variable and pointers to its left and right nodes.
-    """
-
-    def __init__(self, data):
-        self.left = None
-        self.right = None
-        self.data = data
+@dataclass
+class Node(object):
```
dhruvmanila, line 6:
```suggestion
class Node:
```
By default, we are subclassing `object` so there's no point in including it.

dhruvmanila (changes requested):
Awesome work!

This is out of scope for this PR but just wanted to point out that I see a lot of Node classes containing a left and right node. I don't think that makes sense.

In case of tree structures a node can only contain key and data, key to identify the node and data is the containing information. The tree has left and right subtrees and not the node.

data_structures/binary_tree/binary_tree_traversals.py
```diff
@@ -1,24 +1,16 @@
...
-    A Node has data variable and pointers to its left and right nodes.
-    """
-
-    def __init__(self, data):
-        self.left = None
-        self.right = None
-        self.data = data
+    data: int
+    left: "Node" = None
+    right: "Node" = None
```
dhruvmanila, line 9: I did not notice this the first time but shouldn't this be `Optional["Node"]`?

cclauss: My sense was that if neither Python nor mypy was complaining about it then I would not either.  This would be the only location where we would set these values to `None` unless we started deleting Nodes which I assumed we were not gonna do.

dhruvmanila: Python is never going to complain as it simply does not care about type hints. `mypy` will complain when you run it in a bit stricter modes. Plain `mypy` will not complain about a lot of stuff as that's how gradual typing goes.

And no, as per this definition, the leaves of the trees (the last nodes) will have its left and right node `None`

cclauss: > This would be the only location where we would set these values

dhruvmanila: Yes, but that still won't allow the file to be type checked later on. The variable's correct type is `Optional["Node"]`. Just because it is being set `None` once, does not change the type of that variable. That's what 'static' means in static typing :)

dhruvmanila (approved):

PR #4337, opened 2021-04-18
Fix mypy matrix/matrix_operation.py

cclauss:
### **Describe your change:**

Subtask of #4052

Fix `mypy matrix/matrix_operation.py`

@StephenGemin Your review, please.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/matrix/matrix_operation.py
+++ b/matrix/matrix_operation.py
@@ -4,9 +4,18 @@ Functions for 2D matrix operations
 
 from __future__ import annotations
 
+from collections.abc import Iterable
+from typing import Optional
+
 
 def add(*matrix_s: list[list]) -> list[list]:
     """
+    >>> add([])
+    [[]]
+    >>> add([[]])
+    [[]]
+    >>> add([[1, 2]])
+    [[1, 2]]
     >>> add([[1,2],[3,4]],[[2,3],[4,5]])
     [[3, 5], [7, 9]]
     >>> add([[1.2,2.4],[3,4]],[[2,3],[4,5]])
@@ -14,10 +23,12 @@ def add(*matrix_s: list[list]) -> list[list]:
     >>> add([[1, 2], [4, 5]], [[3, 7], [3, 4]], [[3, 5], [5, 7]])
     [[7, 14], [12, 16]]
     """
-    if all(_check_not_integer(m) for m in matrix_s):
+    if matrix_s and matrix_s[0] and all(_check_not_integer(m) for m in matrix_s):
         for i in matrix_s[1:]:
             _verify_matrix_sizes(matrix_s[0], i)
         return [[sum(t) for t in zip(*m)] for m in zip(*matrix_s)]
+    else:
+        return [[]]
 
 
 def subtract(matrix_a: list[list], matrix_b: list[list]) -> list[list]:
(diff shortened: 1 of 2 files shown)


matrix/matrix_operation.py
```diff
@@ -4,20 +4,31 @@
 
 from __future__ import annotations
 
+from collections.abc import Iterable
+from typing import Optional
+
 
 def add(*matrix_s: list[list]) -> list[list]:
```
dhruvmanila, line 11: This is an incorrect type hint for a `*` (stared) argument. It should be `tuple[<arg1>, <arg2>, ...]` In this case, it should probably be `tuple[list[list[int]], ...]`

Also, please provide the element types for return type.

matrix/matrix_operation.py
```diff
@@ -4,20 +4,31 @@
...
-    if all(_check_not_integer(m) for m in matrix_s):
+    if matrix_s and matrix_s[0] and all(_check_not_integer(m) for m in matrix_s):
         for i in matrix_s[1:]:
             _verify_matrix_sizes(matrix_s[0], i)
         return [[sum(t) for t in zip(*m)] for m in zip(*matrix_s)]
+    else:
+        return [[]]
 
 
 def subtract(matrix_a: list[list], matrix_b: list[list]) -> list[list]:
```
dhruvmanila, line 34: If possible, please add the element type for the provided types.

matrix/matrix_operation.py
```diff
@@ -33,9 +44,11 @@ def subtract(matrix_a: list[list], matrix_b: list[list]) -> list[list]:
         and _verify_matrix_sizes(matrix_a, matrix_b)
     ):
         return [[i - j for i, j in zip(*m)] for m in zip(matrix_a, matrix_b)]
+    else:
+        return [[]]
 
 
-def scalar_multiply(matrix: list[list], n: int) -> list[list]:
+def scalar_multiply(matrix: Iterable[list], n: float) -> list[list]:
```
dhruvmanila, line 51: I don't think a matrix should _just_ be an `Iterable` type. `list` seems to be an appropriate type for it. And if we do plan to just keep it `Iterable` then it should be the same where we require and return a matrix.

dhruvmanila (changes requested):
Good job!
Can you provide the element types for all the container types like `list[<element type>]`?

I would also argue to not use `Iterable` for a Matrix type as its not just an Iterable. If you're looking for a covariant type, I would suggest `Sequence` which allows more methods then `Iterable`. In our case of matrices, `list` seems to be fine.

matrix/matrix_operation.py
```diff
@@ -148,11 +163,13 @@ def _check_not_integer(matrix: list[list]) -> bool:
     raise TypeError("Expected a matrix, got int/list instead")
 
 
-def _shape(matrix: list[list]) -> list:
+def _shape(matrix: list[list]) -> tuple[int, int]:
     return len(matrix), len(matrix[0])
```
pedrolopesc, line 167: In case of matrix = [[1, 2], [1, 2, 3]] it would return [2, 2]

Wouldn't this help?
`def _shape(matrix: list[list]) -> tuple[int, int]:`
`if (len(set([len(row) for row in matrix]) == 1):`
`return len(matrix), len(matrix[0])`
`raise TypeError("Matrix is not rectangular")`

PR #4382, opened 2021-04-30
Drawing Julia sets

alexn11:
### **Describe your change:**

Drawing Julia sets

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


algobytewise:
Hi alexn11, I couldn't find the images produced by your program neither on the Wikipedia article nor through a short google search. Do you know of such a website to have a visual confirmation that the program does what it is supposed to do?

alexn11:
Hi algobytewise,
The first two images are of the Cauliflower Julia set: it is visible on this wikipedia article https://en.wikipedia.org/wiki/List_of_fractals_by_Hausdorff_dimension (9th row of 1st table), specifically this image:https://en.wikipedia.org/wiki/File:Julia_z2%2B0,25.png)
For the two other ones, there are similar images: http://www.math.univ-toulouse.fr/~cheritat/GalII/galery.html, or p.4 of this article. All these sets are proved to be ambiently homeomorphic. If you think it's better I could change the parameter c to match one of these examples exactly.
Thanks for reviewing my PR.

algobytewise:
Thanks for the prompt response. Visually it looks fine. Correct me if I'm wrong, but I think for the examples you picked, it's not important to have two plots per example (one for the absolute value & one whether it escapes) because there is not much of a color gradient for the absolute value. Since it's in the directory `fractals`, you could add one example that looks a little more "fractal-like", for example, something from https://en.wikipedia.org/wiki/Julia_set#Quadratic_polynomials. For them, the two plots should make sense, I guess.

You still need to add the type-hints for mypy. https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md has some basic information on that.

By the way, I get the following warnings:
 `julia_sets.py:39: RuntimeWarning: overflow encountered in multiply
  return z * z + c`
`julia_sets.py:39: RuntimeWarning: invalid value encountered in multiply
  return z * z + c`
`julia_sets.py:27: RuntimeWarning: overflow encountered in exp
  return numpy.exp(z) + c`
`julia_sets.py:27: RuntimeWarning: invalid value encountered in exp
  return numpy.exp(z) + c`

alexn11:
Thanks a lot, your remarks are very helpful. I was wondering for the functions that return numpy array, do you know if there is a specific typing convention I should use or should I just not put hinting? There seems to be several solutions to specify array type but the most useful ones use external packages that don't appear in the requirements.txt (such as nptyping).

algobytewise:
I think the type is not "np.array" but "np.ndarray".

fractals/julia_sets.py
```diff
@@ -0,0 +1,139 @@
...
+c_polynomial_1 = -0.4 + 0.6j
+c_polynomial_2 = -0.7269 + 0.1889j
+c_polynomial_2 = -0.1+0.651j 
+c_exponential = -2.0
+nb_iterations = 56
+window_size = 2.0
+nb_pixels = 666
+
+
+def eval_exponential(c: complex, z: numpy.ndarray) -> numpy.ndarray:
```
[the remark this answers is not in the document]
alexn11: (these are standard naming conventions and require no further description)

fractals/julia_sets.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    >>> eval_exponential(0, 0)
+    1.0
+    """
+    return numpy.exp(z) + c
+
+
+def eval_quadratic_polynomial(c: complex, z: numpy.ndarray) -> numpy.ndarray:
```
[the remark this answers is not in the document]
alexn11: (these are standard naming conventions and require no further description)

algobytewise (changes requested):
Thanks for implementing all the changes. I'm not sure that it's a good idea to just hide the RuntimeWarnings. I haven't looked at the code in detail, but I imagine that having an overflow might make the difference whether a pixel is categorized as above or below the escape_radius, resulting in a bug. Maybe that's the reason why the last example is dotted while the example on the website has solid colored sections.

An easy solution might be to change the array-type to 64 bit, see https://stackoverflow.com/questions/7559595/python-runtimewarning-overflow-encountered-in-long-scalars. Maybe reducing the number of iterations might also do the trick. Otherwise you may have to implement a check in the code whether the numbers are getting too big and maybe stop the iteration-process for the concerned values.

I'm also getting the following errors from flake-8 test. I guess some of the comment-lines are too long & need to be separated into several lines.
julia_sets.py:3:89: E501 line too long (388 > 88 characters)
julia_sets.py:6:89: E501 line too long (92 > 88 characters)
julia_sets.py:8:89: E501 line too long (208 > 88 characters)
julia_sets.py:12:1: F401 'typing.NewType' imported but unused
julia_sets.py:12:1: F401 'typing.Union' imported but unused
julia_sets.py:65:89: E501 line too long (108 > 88 characters)

alexn11:
Yes you're right for the dotted graph. I've made a new version that replace nans and infinites with a large value. Reducing the number of iterations would decrease the quality of the output. I was looking for a way to have numpy replace overflows etc with a specific value without raising the warning (while continuing array computations) but couldn't find any. Given the way the loop is implemented, the runtime warnings are unavoidable. I think this is fine as long as we are aware of why these warnings occur.

algobytewise:
Your solution is definitely a progress but I'm not sure that it fully solves the problem. Have a look at the following example:

```python
z_0 = numpy.array([[2, 3]])
z_final = iterate_function(
	eval_quadratic_polynomial,
	0,
	10,
	z_0,
)
```
The final value is `[[          0 -1876701183]]`, which does not make much sense: both values should be really high or infinite, not 0 or negative.

algobytewise:
One solution that comes to my mind is to apply the eval_function not to the array as a whole but to individual values. While iterating you could check whether the value crosses the escape radius and then break the loop for this value. This way you should never get close to overflow.

algobytewise:
I think you can ignore my 2 earlier comments: I entered int-values, not float. The result is correct with float-values (`z_0 = numpy.array([[2., 3.]])`).

algobytewise:
I'm still a little bit concerned about the last example: there is a dotted area between the "stem" and the "petals" e.g. around the coordinates 300 < x < 350 & 300 < y < 350. This area looks different in comparison-image. It might indicate that the bug hasn't fully been solved but there may also be another explanation. I experimented a little with the initial values but it doesn't seem to go away.

alexn11:
Thanks for reviewing the code closely. As you remarked the 1st phenomenon was due to using integer dtype. A solution would be to add a conversion such as .astype('complex64'). This would have the added advantage of ensuring that the values are high precision floating point complex numbers.

Concerning the image of the Julia set of the exponential, I am not sure what you are referring to. There is no stem, the round area is due to numerical limitations. These also appears in the images of Devaney's article. Arnaud Chéristat uses a workaround with shades of grey (using argument information). Implementing this would requires a special "show results" function for the exponential and I'm not sure I know how to do it.

algobytewise (approved):
I think implementing shades of grey wouldn't make much sense since we are just interested in whether the value crosses the escape radius or not. Otherwise, looks good to me.

fractals/julia_sets.py
```diff
@@ -0,0 +1,166 @@
...
```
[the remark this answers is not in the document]
alexn11: (these are standard naming conventions and require no further description)

algobytewise:
One more suggestion: I would hide the warnings so not to confuse the users but add a short comment to the code explaining which warnings arise & how the program handles them.

alexn11:
Hi, what happened with my PR?

alexn11:
> Hi, what happened with my PR?

Only bumping to keep it alive.

fractals/julia_sets.py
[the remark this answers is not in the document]
poyea: Yes, if it's single char the bot will warn you.

poyea:
@alexn11 Please take a look:
```
FAILED fractals/julia_sets.py::julia_sets.eval_quadratic_polynomial
FAILED fractals/julia_sets.py::julia_sets.iterate_function

**Comment by alexn11** [comment]
> @alexn11 Please take a look:
> 
> ```
> FAILED fractals/julia_sets.py::julia_sets.eval_quadratic_polynomial
> FAILED fractals/julia_sets.py::julia_sets.iterate_function
> ```

Thanks for pointing these out. There was some typing issue. Should be resolved now.

File: fractals/julia_sets.py [hunk 4]
```diff
@@ -0,0 +1,214 @@
...
+    nb_iterations = 24
+    escape_radius = 2 * abs(c_cauliflower) + 1
+    z_final = iterate_function(
+        eval_quadratic_polynomial,
+        c_cauliflower,
+        nb_iterations,
+        z_0,
+        infinity=1.1 * escape_radius,
+    )
+    show_results("z²+c", c_cauliflower, escape_radius, z_final)
```
poyea, line 181:
```suggestion
    show_results("z^2+c", c_cauliflower, escape_radius, z_final)
```

fractals/julia_sets.py
```diff
@@ -0,0 +1,214 @@
...
+    nb_iterations = 64
+    escape_radius = 2 * abs(c_polynomial_1) + 1
+    z_final = iterate_function(
+        eval_quadratic_polynomial,
+        c_polynomial_1,
+        nb_iterations,
+        z_0,
+        infinity=1.1 * escape_radius,
+    )
+    show_results("z²+c", c_polynomial_1, escape_radius, z_final)
```
poyea, line 192:
```suggestion
    show_results("z^2+c", c_polynomial_1, escape_radius, z_final)
```

fractals/julia_sets.py
```diff
@@ -0,0 +1,214 @@
...
+    nb_iterations = 161
+    escape_radius = 2 * abs(c_polynomial_2) + 1
+    z_final = iterate_function(
+        eval_quadratic_polynomial,
+        c_polynomial_2,
+        nb_iterations,
+        z_0,
+        infinity=1.1 * escape_radius,
+    )
+    show_results("z²+c", c_polynomial_2, escape_radius, z_final)
```
poyea, line 203:
```suggestion
    show_results("z^2+c", c_polynomial_2, escape_radius, z_final)
```

fractals/julia_sets.py
```diff
@@ -0,0 +1,214 @@
...
+    nb_iterations = 12
+    escape_radius = 10000.0
+    z_final = iterate_function(
+        eval_exponential,
+        c_exponential,
+        nb_iterations,
+        z_0 + 2,
+        infinity=1.0e10,
+    )
+    show_results("exp(z)+c", c_exponential, escape_radius, z_final)
```
poyea, line 214:
```suggestion
    show_results("e^z+c", c_exponential, escape_radius, z_final)
```

fractals/julia_sets.py
```diff
@@ -0,0 +1,214 @@
...
+    the value of escape_radius. Adds the function_label and function_params to
+    the title.
+
+    >>> show_results('80', 0, 1, numpy.array([[0,1,.5],[.4,2,1.1],[.2,1,1.3]]))
+    """
+
+    abs_z_final = (abs(z_final)).transpose()
+    abs_z_final[:, :] = abs_z_final[::-1, :]
+    pyplot.matshow(abs_z_final < escape_radius)
+    pyplot.title(f"Julia set of {function_label}\n c={function_params}")
```
poyea, line 140:
```suggestion
    pyplot.title(f"Julia set of ${function_label}$, $c={function_params}$")
```

fractals/julia_sets.py
```diff
@@ -0,0 +1,214 @@
...
+c_polynomial_1 = -0.4 + 0.6j
+c_polynomial_2 = -0.1 + 0.651j
+c_exponential = -2.0
+nb_iterations = 56
+window_size = 2.0
+nb_pixels = 666
+
+
+def eval_exponential(c_parameter: complex, z_values: numpy.ndarray) -> numpy.ndarray:
+    """
```
poyea, line 40:
```suggestion
    """
    Evaluate $e^z + c$.
```
and a few more test cases?

fractals/julia_sets.py
```diff
@@ -0,0 +1,219 @@
...
+    0j
+    >>> numpy.round(iterate_function(eval_quadratic_polynomial,
+                                     0,
+                                     3,
+                                     numpy.array([0,1,2]))[1])
+    (1+0j)
+    >>> numpy.round(iterate_function(eval_quadratic_polynomial,
+                                     0,
+                                     3,
+                                     numpy.array([0,1,2]))[2])
```
poyea, line 115:
```suggestion
    ... 0,
    ... 3,
    ... numpy.array([0,1,2]))[1])
    (1+0j)
    >>> numpy.round(iterate_function(eval_quadratic_polynomial,
    ... 0,
    ... 3,
    ... numpy.array([0,1,2]))[2])
```

poyea:
@algorithms-keeper review-all

poyea (approved):
Thank you for your pull request!🤩Looks good to me. While we're waiting for the final checks, feel free to improve anything you want.

alexn11:
> Thank you for your pull request!star_struckLooks good to me. While we're waiting for the final checks, feel free to improve anything you want.

Great! Thank you very much for the edits!

PR #4450, opened 2021-05-22
Censor Function

meg-1:
### **Describe your change:**
Put the censor algorithm into a function

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/strings/censor.py
+++ b/strings/censor.py
@@ -1,10 +1,14 @@
 txt = input("Please enter the text from which you would like to censor a certain word: \n")
 word = input("Please enter the word you would like to be censored from the text: \n")
 txt_lst = txt.split()
+
+#here we loop through a list which consists of words from the text we should censor
 for word1 in txt_lst:
   if word1 == word:
     txt_lst[txt_lst.index(word1)] = "*" * len(word1)
 txt = ""
+
+#here we combine all of the split words in txt_lst into one string
 for word1 in txt_lst:
   txt += word1 + " "
 txt = txt[:-1]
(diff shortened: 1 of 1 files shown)


strings/censor.py
```diff
@@ -1,15 +1,14 @@
-txt = input("Please enter the text from which you would like to censor a certain word: \n")
-word = input("Please enter the word you would like to be censored from the text: \n")
-txt_lst = txt.split()
+def censor(txt, word):
```
Jenil-S, line 1: I think you should update this function as
`def censor(txt,word) -> None:`

Jenil-S (changes requested):

strings/censor.py
```diff
@@ -0,0 +1,14 @@
...
+  txt_lst = txt.split()
+  #here we loop through a list which consists of words from the text we should censor
+  for word1 in txt_lst:
+    if word1 == word:
+      txt_lst[txt_lst.index(word1)] = "*" * len(word1)
```
Txbias, line 6: You could make this a little more efficient by using ```for index, word1 in enumerate(txt_lst)```. 
That way you already have the index and don't need to call ```txt_lst.index(word1)```. Instead you just use the variable ```index```.

mrmaxguns:
Hello. It looks like you haven't run `black` on the code. Black can be installed with `pip install black` (assuming you also have `pip` installed). Then, you can run `black your_file.py` (replacing `your_file` with the actual file path). This will automatically format the code according to relevant Python standards.

PR #4452, opened 2021-05-23
[mypy] Fix type annotations/errors for all files in strings

Parminder51:
### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/strings/aho_corasick.py
+++ b/strings/aho_corasick.py
@@ -1,10 +1,15 @@
 from collections import deque
-from typing import Dict, List, Union
+from typing import Union, TypedDict, Deque
 
+class type_of_dict(TypedDict):
+    value: str
+    next_states: list[int]
+    fail_state: int
+    output: list[str]
 
 class Automaton:
-    def __init__(self, keywords: List[str]):
-        self.adlist = list()
+    def __init__(self, keywords: list[str]):
+        self.adlist: list[type_of_dict] = list()
         self.adlist.append(
             {"value": "", "next_states": [], "fail_state": 0, "output": []}
         )
@@ -22,8 +27,10 @@ class Automaton:
     def add_keyword(self, keyword: str) -> None:
         current_state = 0
         for character in keyword:
-            if self.find_next_state(current_state, character):
-                current_state = self.find_next_state(current_state, character)
+            next_state = self.find_next_state(current_state, character)
+            if next_state:
+                assert next_state is not None
+                current_state = next_state
             else:
                 self.adlist.append(
                     {
@@ -38,7 +45,7 @@ class Automaton:
         self.adlist[current_state]["output"].append(keyword)
 
     def set_fail_transitions(self) -> None:
-        q = deque()
+        q: Deque = deque()
         for node in self.adlist[0]["next_states"]:
             q.append(node)
             self.adlist[node]["fail_state"] = 0
@@ -62,13 +69,13 @@ class Automaton:
                     + self.adlist[self.adlist[child]["fail_state"]]["output"]
                 )
 
-    def search_in(self, string: str) -> Dict[str, List[int]]:
+    def search_in(self, string: str) -> dict[str, list[int]]:
         """
         >>> A = Automaton(["what", "hat", "ver", "er"])
         >>> A.search_in("whatever, err ... , wherever")
         {'what': [0], 'hat': [1], 'ver': [5, 25], 'er': [6, 10, 22, 26]}
         """
-        result = dict()  # returns a dict with keywords and list of its occurrences
+        result: dict[str, list[int]] = dict()  # returns a dict with keywords and list of its occurrences
         current_state = 0
         for i in range(len(string)):
             while (
@@ -76,7 +83,9 @@ class Automaton:
                 and current_state != 0
             ):
                 current_state = self.adlist[current_state]["fail_state"]
-            current_state = self.find_next_state(current_state, string[i])
+            next_state = self.find_next_state(current_state, string[i])
+            assert next_state is not None
+            current_state = next_state
             if current_state is None:
                 current_state = 0
             else:
--- a/strings/manacher.py
+++ b/strings/manacher.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 def palindromic_string(input_string: str) -> str:
     """
     >>> palindromic_string('abbbaba')
@@ -20,6 +22,8 @@ def palindromic_string(input_string: str) -> str:
     # if input_string is "aba" than new_input_string become "a|b|a"
     new_input_string = ""
     output_string = ""
+
+    i: Union[int, str]
 
     # append each character + "|" in new_string for range(0, length-1)
     for i in input_string[: len(input_string) - 1]:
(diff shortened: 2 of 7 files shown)


M3T4F34R (approved):

Parminder51:
Sorry for lack of unique commit names

Parminder51:
@dhruvmanila Should be working fine now. Kindly check please.

Parminder51:
@dhruvmanila Sorry for all the errors, it's my first time. I believe it should be working now.

Parminder51:
@dhruvmanila Kindly check plz

strings/aho_corasick.py
```diff
@@ -1,10 +1,17 @@
...
+
+class type_of_dict(TypedDict):
+    value: str
+    next_states: list[int]
+    fail_state: int
+    output: list[str]
 
 
 class Automaton:
-    def __init__(self, keywords: List[str]):
```
QuantumNovice, line 6: Use these (type check time constructs) instead of runtime constructs.

strings/aho_corasick.py
```diff
@@ -62,21 +71,28 @@ def set_fail_transitions(self) -> None:
                     + self.adlist[self.adlist[child]["fail_state"]]["output"]
                 )
 
-    def search_in(self, string: str) -> Dict[str, List[int]]:
```
QuantumNovice, line 65: Use these (type check time constructs) instead of runtime constructs.

strings/aho_corasick.py
```diff
@@ -38,7 +47,7 @@ def add_keyword(self, keyword: str) -> None:
         self.adlist[current_state]["output"].append(keyword)
 
     def set_fail_transitions(self) -> None:
-        q = deque()
+        q: Deque = deque()
```
QuantumNovice, line 50: Deque[dict]

Parminder51: Actually the deque is storing [int], not [dict], so that's why it's like that, and int & str are already present when type checking for Deque

strings/aho_corasick.py
Parminder51: I'm not sure I understand what you're trying to say. Could you please explain it a bit?

QuantumNovice: use `List` instead `list`. `List` is used during check time. Both syntax achieve the same thing. It's neat to use `List` besides.

Parminder51: https://stackoverflow.com/a/62033243

Also in the docs, the generic types from typing have been deprecated for Python3.9

strings/aho_corasick.py
```diff
@@ -1,10 +1,17 @@
 from collections import deque
-from typing import Dict, List, Union
+from typing import Deque, TypedDict, Union
+
+
+class type_of_dict(TypedDict):
```
dhruvmanila, line 5: Could you change the name of the class to something meaningful and also in CamelCase ?

strings/aho_corasick.py
```diff
@@ -22,8 +29,10 @@ def find_next_state(self, current_state: int, char: str) -> Union[int, None]:
     def add_keyword(self, keyword: str) -> None:
         current_state = 0
         for character in keyword:
-            if self.find_next_state(current_state, character):
-                current_state = self.find_next_state(current_state, character)
+            next_state = self.find_next_state(current_state, character)
+            if next_state:
+                assert next_state is not None
```
dhruvmanila, line 34: Can you tell me why this change was necessary? It seems perfectly fine as it was.

If `mypy` was giving any errors then make the below changes:
```suggestion
            if next_state is not None:
```

strings/aho_corasick.py
dhruvmanila: This is not necessary as `mypy` can infer the type.

strings/aho_corasick.py
```diff
@@ -62,21 +71,28 @@ def set_fail_transitions(self) -> None:
...
+    def search_in(self, string: str) -> dict[str, list[int]]:
         """
         >>> A = Automaton(["what", "hat", "ver", "er"])
         >>> A.search_in("whatever, err ... , wherever")
         {'what': [0], 'hat': [1], 'ver': [5, 25], 'er': [6, 10, 22, 26]}
         """
-        result = dict()  # returns a dict with keywords and list of its occurrences
+        result: dict[
+            str, list[int]
+        ] = dict()  # returns a dict with keywords and list of its occurrences
```
dhruvmanila, line 82: Can you put the comment above the assignment statement and reformat the code?

This way the line will be consistent and easy to look.

strings/aho_corasick.py
```diff
@@ -62,21 +71,28 @@ def set_fail_transitions(self) -> None:
...
         for i in range(len(string)):
             while (
                 self.find_next_state(current_state, string[i]) is None
                 and current_state != 0
             ):
                 current_state = self.adlist[current_state]["fail_state"]
-            current_state = self.find_next_state(current_state, string[i])
+            next_state = self.find_next_state(current_state, string[i])
+            if next_state:
+                assert next_state is not None
```
dhruvmanila, line 92: Same as above.

strings/manacher.py
```diff
@@ -21,6 +24,8 @@ def palindromic_string(input_string: str) -> str:
     new_input_string = ""
     output_string = ""
 
+    i: Union[int, str]
```
dhruvmanila, line 27: I think `i` is a loop variable and that's the reason `mypy` must have been giving some error. Instead, you should change the loop variable for other loops to `j`, `k`, etc.

dhruvmanila (changes requested):
Some small changes, otherwise really good work!

strings/aho_corasick.py
Parminder51: Yes sure, I have renamed it to DictionaryOfState, please do tell if it needs to be renamed again😅.

strings/aho_corasick.py
Parminder51: Okay, I have removed all such instances and have restructured the logic as needed

strings/aho_corasick.py
Parminder51: I can move the comment but reformatting is done by black so can't change that

strings/aho_corasick.py
Parminder51: Fixed

strings/manacher.py
Parminder51: I have changed all the loop variables accordingly

strings/aho_corasick.py
Parminder51: I am afraid `mypy` is specifically showing this as an error, and also that instead of Deque which is deprecated, deque can be used, which has been changed as well

strings/aho_corasick.py
Parminder51: Actually I didn't want to change the logic of the statements, that's why I was using `assert`

L3str4nge (approved):

iamrajiv (approved):

PR #4455, opened 2021-05-26
Add catalan_numbers.py

Kommandat:
### **Describe your change:**
* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [x] This pull request is all my own work -- I have not plagiarized

### Additional Notes

Excited to make my first contribution to The Algorithms! Catalan numbers are a sequence in mathematics that represent answers to many combinatorics problems. They can be calculated using a recurrence relation and we can use tabulation (similarly to how we can for the Fibonacci sequence). I thought this best fit in the "dynamic_programming" folder given we're using tabulation to compute the sequence.

All work is my own, but I did take a look at the existing [C++ Catalan number algorithm](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/dynamic_programming/catalan_numbers.cpp) and [Python Fibonacci algorithm ](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/fibonacci.py) for inspiration and best practices. If this is approved, I hope to contribute to Algorithms-Explanation as well.

Please let me know if you have any feedback!

Sources:
[1] https://brilliant.org/wiki/catalan-numbers/
[2] https://en.wikipedia.org/wiki/Catalan_number

### DIFF
--- a/dynamic_programming/catalan_numbers.py
+++ b/dynamic_programming/catalan_numbers.py
@@ -1,0 +1,79 @@
+"""
+Print all the Catalan numbers from 0 to n, n being the user input.
+
+ * The Catalan numbers are a sequence of positive integers that
+ * appear in many counting problems in combinatorics [1]. Such
+ * problems include counting [2]:
+ * - The number of Dyck words of length 2n
+ * - The number well-formed expressions with n pairs of parentheses
+ *   (e.g., `()()` is valid but `())(` is not)
+ * - The number of different ways n + 1 factors can be completely
+ *   parenthesized (e.g., for n = 2, C(n) = 2 and (ab)c and a(bc)
+ *   are the two valid ways to parenthesize.
+ * - The number of full binary trees with n + 1 leaves
+
+ * A Catalan number satisfies the following recurrence relation
+ * which we will use in this algorithm [1].
+ * C(0) = C(1) = 1
+ * C(n) = sum(C(i).C(n-i-1)), from i = 0 to n-1
+
+ * In addition, the n-th Catalan number can be calculated using
+ * the closed form formula below [1]:
+ * C(n) = (1 / (n + 1)) * (2n choose n)
+
+ * Sources:
+ *  [1] https://brilliant.org/wiki/catalan-numbers/
+ *  [2] https://en.wikipedia.org/wiki/Catalan_number
+"""
+
+
+def catalan_numbers(upper_limit: int) -> "list[int]":
+    """
+    Return a list of the Catalan number sequence from 0 through `upper_limit`.
+
+    >>> catalan_numbers(5)
+    [1, 1, 2, 5, 14, 42]
+    >>> catalan_numbers(2)
+    [1, 1, 2]
+    >>> catalan_numbers(-1)
+    Traceback (most recent call last):
+    ValueError: Limit for the Catalan sequence must be ≥ 0
+    """
+    if upper_limit < 0:
+        raise ValueError("Limit for the Catalan sequence must be ≥ 0")
+
+    catalan_list = [0] * (upper_limit + 1)
+
+    # Base case: C(0) = C(1) = 1
+    catalan_list[0] = 1
+    if upper_limit > 0:
+        catalan_list[1] = 1
+
+    # Recurrence relation: C(i) = sum(C(j).C(i-j-1)), from j = 0 to i
+    for i in range(2, upper_limit + 1):
+        for j in range(i):
+            catalan_list[i] += catalan_list[j] * catalan_list[i - j - 1]
+
+    return catalan_list
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


dynamic_programming/catalan_numbers.py
```diff
@@ -0,0 +1,101 @@
...
+ * C(n) = (1 / (n + 1)) * (2n choose n)
+
+ * Sources:
+ *  [1] https://brilliant.org/wiki/catalan-numbers/
+ *  [2] https://en.wikipedia.org/wiki/Catalan_number
+"""
+
+
+class Catalan:
+    def __init__(self, N: int) -> None:
```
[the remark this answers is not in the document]
Kommandat: I added a comment explaining N's function below this line and I don't think a more descriptive variable name is necessary

appgurueu:
I don't like the unneeded OOP.

dynamic_programming/catalan_numbers.py
```diff
@@ -0,0 +1,100 @@
...
+
+            # Recurrence relation: C(i) = sum(C(j).C(i-j-1)), from j = 0 to i
+            for i in range(2, upper_limit + 1):
+                for j in range(i):
+                    self.catalan_numbers[i] += (
+                        self.catalan_numbers[j] * self.catalan_numbers[i - j - 1]
+                    )
+        elif upper_limit == 0:
+            self.catalan_numbers[0] = 1
+        print(self.catalan_numbers)
```
mrmaxguns, line 50: I would avoid print statements in the algorithm itself as they produce unwanted side-effects.

dynamic_programming/catalan_numbers.py
```diff
@@ -0,0 +1,100 @@
...
+        >>> Catalan(5).get(6)
+        [1, 1, 2, 5, 14, 42]
+        Out of bound.
+        >>> Catalan(5).get(-1)
+        [1, 1, 2, 5, 14, 42]
+        []
+        """
+        if sequence_no is not None:
+            if sequence_no < len(self.catalan_numbers):
+                return print(self.catalan_numbers[: sequence_no + 1])
```
mrmaxguns, line 66: This should return the value directly for the same reason described above. This is where the algorithm produces its output, so the output should be directly returned.

dynamic_programming/catalan_numbers.py
```diff
@@ -0,0 +1,100 @@
...
+            else:
+                print("Out of bound.")
```
mrmaxguns, line 68: This is an error, so an exception should be raised here.

dynamic_programming/catalan_numbers.py
```diff
@@ -0,0 +1,100 @@
...
+        else:
+            print("Please specify a value")
```
mrmaxguns, line 70: Again, this should be refactored as an exception.

mrmaxguns (changes requested):
Hello. Thanks for your contribution. I performed a review. If you disagree with the changes I suggested, feel free to explain your disagreement by commenting on the review.

Mostly, I suggested changes to avoid unwanted side-effects. These changes are based on the following contribution guideline:
> have minimal side effects (Ex. print(), plot(), read(), write()).

Also, when the user enters erroneous input, an exception should be raised, instead of it simply being printed (this applies only to the algorithm itself, not in the `if __name__ == "__main__"` block).
> raise Python exceptions (ValueError, etc.) on erroneous input values

Kommandat:
> Hello. Thanks for your contribution. I performed a review. If you disagree with the changes I suggested, feel free to explain your disagreement by commenting on the review.
> 
> Mostly, I suggested changes to avoid unwanted side-effects. These changes are based on the following contribution guideline:
> 
> > have minimal side effects (Ex. print(), plot(), read(), write()).
> 
> Also, when the user enters erroneous input, an exception should be raised, instead of it simply being printed (this applies only to the algorithm itself, not in the `if __name__ == "__main__"` block).
> 
> > raise Python exceptions (ValueError, etc.) on erroneous input values

@mrmaxguns Thanks for the review! I agree with all your comments and made the changes you recommended. I requested a re-review, please take another look when you get a chance.

As an aside, I based much of my code on the [Fibonacci algorithm](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/fibonacci.py) in the same folder thinking an existing algorithm already followed the guidelines but I was mistaken. It would nice to add some examples to the contribution guidelines that demonstrate best practices so others don't do what I did.

Once again, thanks!

mrmaxguns (approved):
Thank you for your contribution! Yes, indeed, some code in this repository does not follow the contribution guidelines, as they may have changed or the code was improperly reviewed. If you notice such code, feel free to open a pull request addressing the changes.

Regarding the code, it looks good. Thanks for applying the suggestions! This should be merged once tests are run and I double check the code for any issues I might not have caught now.

mrmaxguns:
Thank you for your contribution! :100:

PR #4464, opened 2021-05-29
Mistake in maths/average_mode.py fixed.

haningrisha:
### **Describe your change:**

* Mistake in maths/average_mode.py fixed. The previous solution was to return the
value on the first loop iteration, which is not correct, more than that it
used to delete repeating values, so result's array and check array lost
relation between each other

* Type hint added

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/maths/average_mode.py
+++ b/maths/average_mode.py
@@ -1,7 +1,8 @@
 import statistics
+from typing import Any
 
 
-def mode(input_list: list) -> int:  # Defining function "mode."
+def mode(input_list: list) -> Any:  # Defining function "mode."
     """This function returns the mode(Mode as in the measures of
     central tendency) of the input data.
 
@@ -10,21 +11,28 @@ def mode(input_list: list) -> int:  # Defining function "mode."
     >>> input_list = [2, 3, 4, 5, 3, 4, 2, 5, 2, 2, 4, 2, 2, 2]
     >>> mode(input_list)
     2
+    >>> input_list = [3, 4, 5, 3, 4, 2, 5, 2, 2, 4, 4, 2, 2, 2]
+    >>> mode(input_list)
+    2
+    >>> input_list = ["x", "y", "y", "z"]
+    >>> mode(input_list) 
+    'y'
     >>> input_list = [2, 3, 4, 5, 3, 4, 2, 5, 2, 2, 4, 2, 2, 2]
+    >>> mode(input_list) == statistics.mode(input_list)
+    True
+    >>> input_list = ["x", "y", "y", "z"]
     >>> mode(input_list) == statistics.mode(input_list)
     True
     """
     # Copying input_list to check with the index number later.
-    check_list = input_list.copy()
     result = list()  # Empty list to store the counts of elements in input_list
     for x in input_list:
         result.append(input_list.count(x))
     y = max(result)  # Gets the maximum value in the result list.
     # Returns the value with the maximum number of repetitions.
-    return check_list[result.index(y)]
+    return input_list[result.index(y)]
 
 
 if __name__ == "__main__":
-    data = [2, 3, 4, 5, 3, 4, 2, 5, 2, 2, 4, 2, 2, 2]
-    print(mode(data))
-    print(statistics.mode(data))
+    import doctest
+    doctest.testmod()
--- a/maths/average_mode.py
+++ b/maths/average_mode.py
@@ -26,7 +26,7 @@ def mode(input_list: list) -> list:  # Defining function "mode."
     y = max(result)  # Gets the maximum value in the result list.
     # Gets values of modes
     result = [input_list[i] for i, value in enumerate(result) if value == y]
-    result = list(set(result))  # Delete duplicates
+    result = list(set(result))  # Deletes duplicates
     return sorted(result)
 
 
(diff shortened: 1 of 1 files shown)


maths/average_mode.py
```diff
@@ -1,7 +1,7 @@
 import statistics
 
 
-def mode(input_list):  # Defining function "mode."
+def mode(input_list: list) -> int:  # Defining function "mode."
```
mrmaxguns, line 4: The input list could contain any types of values, not just integers. The mode function works perfectly for a list of strings as well (or even a mixed list). We don't really care about the type of the data. We just care how many times the data occurs in the list.

So, the return value of this function should be set to `typing.Any`.

maths/average_mode.py
```diff
@@ -19,10 +19,9 @@ def mode(input_list):  # Defining function "mode."
     result = list()  # Empty list to store the counts of elements in input_list
     for x in input_list:
         result.append(input_list.count(x))
-        input_list.remove(x)
-        y = max(result)  # Gets the maximum value in the result list.
-        # Returns the value with the maximum number of repetitions.
-        return check_list[result.index(y)]
+    y = max(result)  # Gets the maximum value in the result list.
+    # Returns the value with the maximum number of repetitions.
+    return check_list[result.index(y)]
```
mrmaxguns, line 24: You don't really need `check_list` anymore. It was originally made because `input_list` was mutated every iteration. Now that you have removed this mutation, `check_list = input_list.copy()` can be removed. So, this line can instead become:

```suggestion
    return input_list[result.index(y)]
```

maths/average_mode.py
```diff
@@ -1,7 +1,7 @@
 import statistics
 
 
-def mode(input_list):  # Defining function "mode."
+def mode(input_list: list) -> int:  # Defining function "mode."
     """This function returns the mode(Mode as in the measures of
     central tendency) of the input data.
```
mrmaxguns, line 6: Now that a major bug was fixed, please add more tests. The following tests all have the resulting value be at the front of the list. This is the reason the bug was never caught. I suggest adding a test which does the mode of the following list, for example: `["x", "y", "y", "z"]`. This should return `"y"`.

mrmaxguns (changes requested):
Thank you so much for this critical bug fix. The original `mode` function didn't actually return the mode of the input list, and instead always returned the first value of the list (since the `return` was inside the loop, thus happening at the first iteration).

The reason why the bug wasn't caught was because the value that was the mode always happened to be at the front of the list. More varying tests should prevent future mistakes like this.

Here is my review. If you disagree with any of the points, feel free to comment on them.

mrmaxguns:
Pre-commit failed because Black would reformat the file. Make sure to run black with `black your_file.py`. Black can be installed with `pip install black`.

mrmaxguns (approved):

haningrisha:
I was thinking about raising an exception in case the list has several modes(for instance [2,2,1,1]) several modes existence fact may be crucial in some algorithms. Also I can add another function for working with multiple mode cases. 

Example of usage:

```
try:
    m = mode(some list)
    # logic
except SeveralModeException:
    m = multiple_modes(some list)
    # logic with list
```

Or I can return a list even if there's only one mode, but this seem odd. 

What do you think?

mrmaxguns:
> I was thinking about raising an exception in case the list has several modes(for instance [2,2,1,1]) several modes existence fact may be crucial in some algorithms. Also, I can add another function for working with multiple mode cases.
> 
> Example of usage:
> 
> ```
> try:
>     m = mode(some list)
>     # logic
> except SeveralModeException:
>     m = multiple_modes(some list)
>     # logic with list
> ```
> 
> Or I can return a list even if there's only one mode, but this seems odd.
> 
> What do you think?

Personally, I like the one-function approach. A second function that performs almost the same task seems redundant. There could just be one function which always returns a list. That way, the user can check the length by themselves. This also means that the function could return an empty list if the input list is empty.

maths/average_mode.py
```diff
@@ -1,36 +1,33 @@
...
+    ['y']
+    >>> input_list = ["x", "x" , "y", "y", "z"]
+    >>> mode(input_list)
+    ['x', 'y']
     """
-    # Copying input_list to check with the index number later.
     result = list()  # Empty list to store the counts of elements in input_list
     for x in input_list:
         result.append(input_list.count(x))
     y = max(result)  # Gets the maximum value in the result list.
```
mrmaxguns, line 26: This should have a check before this line so that if the result is an empty list, immediately return an empty list. Otherwise, `max` will raise an error saying `ValueError: max() arg is an empty sequence`.

```suggestion
    if not result:
        return []
    y = max(result)  # Gets the maximum value in the result list.
```

maths/average_mode.py
```diff
@@ -1,36 +1,33 @@
...
-    # Returns the value with the maximum number of repetitions.
-    return input_list[result.index(y)]
+    # Gets values of modes
+    result = [input_list[i] for i, value in enumerate(result) if value == y]
+    result = list(set(result))  # Deletes duplicates
```
mrmaxguns, line 29: Instead of creating a list, then converting it to a set, and then back to a list, you could do a set comprehension. Also, sorted takes an iterable, so passing a set to it will work.
```suggestion
    result = {input_list[i] for i, value in enumerate(result) if value == y}
```

mrmaxguns (changes requested):
Thanks for the changes. Here is my review.

mrmaxguns (approved):

mrmaxguns:
Thank you for contributing :tada:

PR #4506, opened 2021-06-12
Add doctest and fix mypy type annotation in bellman ford

mhihasan:
### **Describe your change:**
- Add doctest and 
- fix mypy type annotation in bellman ford

### **Checklist:**

related issue #4052

### DIFF
--- a/graphs/bellman_ford.py
+++ b/graphs/bellman_ford.py
@@ -1,56 +1,73 @@
 from __future__ import annotations
 
 
-def printDist(dist, V):
-    print("Vertex Distance")
-    distances = ("INF" if d == float("inf") else d for d in dist)
-    print("\t".join(f"{i}\t{d}" for i, d in enumerate(distances)))
+def print_distance(distance: list[float], src):
+    print(f"Vertex\tShortest Distance from vertex {src}")
+    for i, d in enumerate(distance):
+        print(f"{i}\t\t{d}")
 
 
-def BellmanFord(graph: list[dict[str, int]], V: int, E: int, src: int) -> int:
+def check_negative_cycle(
+    graph: list[dict[str, int]], distance: list[float], edge_count: int
+):
+    for j in range(edge_count):
+        u, v, w = [graph[j][k] for k in ["src", "dst", "weight"]]
+        if distance[u] != float("inf") and distance[u] + w < distance[v]:
+            return True
+    return False
+
+
+def bellman_ford(
+    graph: list[dict[str, int]], vertex_count: int, edge_count: int, src: int
+) -> list[float]:
     """
     Returns shortest paths from a vertex src to all
     other vertices.
+    >>> edges = [(2, 1, -10), (3, 2, 3), (0, 3, 5), (0, 1, 4)]
+    >>> g = [{"src": s, "dst": d, "weight": w} for s, d, w in edges]
+    >>> bellman_ford(g, 4, 4, 0)
+    [0.0, -2.0, 8.0, 5.0]
+    >>> g = [{"src": s, "dst": d, "weight": w} for s, d, w in edges + [(1, 3, 5)]]
+    >>> bellman_ford(g, 4, 5, 0)
+    Traceback (most recent call last):
+     ...
+    Exception: Negative cycle found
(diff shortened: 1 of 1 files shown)


graphs/bellman_ford.py
```diff
@@ -1,56 +1,53 @@
-from __future__ import annotations
+from typing import Dict, List
```
cclauss, line 1: This is the opposite of what we want.

graphs/bellman_ford.py
```diff
@@ -1,56 +1,53 @@
...
 
-    for i in range(E):
-        graph[i][i] = 0.0
+if __name__ == "__main__":
+    edges = [(2, 1, -10), (3, 2, 3), (0, 3, 5), (0, 1, 4)]
+    g = [{"src": s, "dst": d, "weight": w} for s, d, w in edges]
+    shortest_distance = bellman_ford(g, 4, 4, 0)
+    print_distance(shortest_distance, 0)
 
-    for i in range(E):
```
cclauss, line 48: Why remove user interaction?

graphs/bellman_ford.py
```diff
@@ -1,56 +1,53 @@
...
 
 
-def BellmanFord(graph: list[dict[str, int]], V: int, E: int, src: int) -> int:
+def bellman_ford(
+    graph: List[Dict[str, int]], vertex_count: int, edge_count: int, src: int
+) -> List[float]:
     """
     Returns shortest paths from a vertex src to all
     other vertices.
+    >>> edges = [(2, 1, -10), (3, 2, 3), (0, 3, 5), (0, 1, 4)]
```
cclauss, line 16: Why not just create `test_edges` at global scope and then use them here and also at line 46.

graphs/bellman_ford.py
```diff
@@ -1,56 +1,53 @@
...
+    distance[src] = 0.0
 
-    for i in range(V - 1):
-        for j in range(E):
-            u = graph[j]["src"]
-            v = graph[j]["dst"]
-            w = graph[j]["weight"]
+    for i in range(vertex_count - 1):
+        for j in range(edge_count):
+            u, v, w = graph[j]["src"], graph[j]["dst"], graph[j]["weight"]
```
cclauss, line 31:
```suggestion
            src, dst, weight = graph[j][field] for field in ("src", "dst", "weight")
```

graphs/bellman_ford.py
mhihasan: Do we want to remove type annotation? I think we need to fix mypy annotation, but not to remove.

graphs/bellman_ford.py
mhihasan: If user mistakes any input, he/she will needs to restart again which is cumbersome. Instead, user(obviously a programmer) can provide graph providing as a list of dictionary.

graphs/bellman_ford.py
mhihasan: Will look into it

graphs/bellman_ford.py
cclauss: We want to use `dict, list` not `Dict, List`

mhihasan: I haven't found this requirement at the issue #4052. Is it mentioned elsewhere?

graphs/bellman_ford.py
cclauss: Users trying again and again is OK.  This is how we all deal with Python on a regular basis.

graphs/bellman_ford.py
cclauss: #2464

mhihasan: Ah, ok. Sorry , As I am new here, I couldn't notice that. I will keep that in mind for next PRs.

graphs/bellman_ford.py
mhihasan: Now, `edges` is being used only here. So, it has not been declared globally.

mhihasan:
Are any other change  suggestions left?

cclauss (approved):

PR #4564, opened 2021-07-21
is_contains_number method added

suryapratapsinghsuryavanshi:
### **Describe your change:**

* [x] Add an algorithm?

### **Checklist:**


mateuszz0000 (commented):
What kind of algorithm is that? Can you point some theory documentation for that? 
Why we can't just `return any(i.isdigit() for i in s)` instead of `regexp`?? For me it's not an algorithm.

suryapratapsinghsuryavanshi:
Because `return any(i.isdigit() for I in s)` take O(n) time but any RegEx expression comparison are take O(1) time so I think this is a good way to find a number in any length string.

PR #4699, opened 2021-08-30
Approve functions used as default arguments

cclauss:
### **Describe your change:**

Use linter directive `# noqa: B0008` for these functions because they set a seed value and `BaseBackend` is reused across all calls and `LinearCongruentialGenerator.__init__()` is only called once for each instance.

[flake8-bugbear](https://pypi.org/project/flake8-bugbear) ___B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.___

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/other/linear_congruential_generator.py
+++ b/other/linear_congruential_generator.py
@@ -8,18 +8,18 @@ class LinearCongruentialGenerator:
     A pseudorandom number generator.
     """
 
+    # The default value for **seed** is the result of a function call which is not
+    # normally recommended and causes flake8-bugbear to raise a B008 error. However,
+    # in this case, it is accptable because `LinearCongruentialGenerator.__init__()`
+    # will only be called once per instance and it ensures that each instance will
+    # generate a unique sequence of numbers.
+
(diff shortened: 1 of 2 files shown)


dhruvmanila (changes requested):
> If this is intended, assign the function call to a module-level variable and use that variable as a default value.

Why don't we do what the linter is suggesting us to? I hate adding ignore directives throughout the code. If not, then we should atleast explain in a comment as to why that directive was added.

dhruvmanila (approved):
I think the explanation should be a _comment_ instead of adding in a _docstring_ as it is intended only for developers instead of end-users, otherwise LGTM.

cclauss:
Comment above function definition or below?  If below, before docstring or after?

dhruvmanila:
Above the directive which means above the function definition.

PR #4704, opened 2021-08-31
Add "range update" functionality for segment tree

mrR2D2:
### **Describe your change:**

Update segment tree implementation in `data_structures/binary_tree/segment_tree_other.py` with range update functionality, which updates range of underlying segment tree array and recalculates the tree in O(logN) time.

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/data_structures/binary_tree/segment_tree_other.py
+++ b/data_structures/binary_tree/segment_tree_other.py
@@ -4,17 +4,29 @@ allowing queries to be done later in log(N) time
 function takes 2 values and returns a same type value
 """
 from collections.abc import Sequence
+from functools import reduce
 from queue import Queue
+from typing import TypeVar
+
+T = TypeVar("T", int, float)
 
 
 class SegmentTreeNode:
-    def __init__(self, start, end, val, left=None, right=None):
+    def __init__(
+        self,
+        start: int,
+        end: int,
+        val: T,
+        left: "SegmentTreeNode" = None,
+        right: "SegmentTreeNode" = None,
+    ):
         self.start = start
         self.end = end
         self.val = val
         self.mid = (start + end) // 2
         self.left = left
         self.right = right
+        self.lazy: T = None
 
     def __str__(self):
         return f"val: {self.val}, start: {self.start}, end: {self.end}"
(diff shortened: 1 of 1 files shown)


data_structures/binary_tree/segment_tree_other.py
```diff
@@ -187,7 +220,47 @@ def _update_tree(self, node, i, val):
...
         node.val = self.fn(node.left.val, node.right.val)
 
+    def _update_tree_range(
+        self,
+        node: SegmentTreeNode,
+        i: int,
+        j: int,
+        val: T,
+    ) -> None:
+
```
cclauss, line 230: A function with no doctests?

data_structures/binary_tree/segment_tree_other.py
```diff
@@ -4,17 +4,29 @@
...
 class SegmentTreeNode:
-    def __init__(self, start, end, val, left=None, right=None):
+    def __init__(
+        self,
+        start: int,
+        end: int,
+        val: T,
+        left: "SegmentTreeNode" = None,
+        right: "SegmentTreeNode" = None,
+    ):
```
cclauss, line 22: A function with no doctests?

data_structures/binary_tree/segment_tree_other.py
mrR2D2: It seems, there is no need for doctests for the `__init__` method, since one is quite straightforward - just simple assignments and no complex logic. But if they _are_ required, I can surely add them, it's not a problem.

data_structures/binary_tree/segment_tree_other.py
mrR2D2: @cclauss The `_update_tree_range` is a private function, which isn't supposed to be used on its own. Instead, it is called by the `update_range` method, which is properly tested. So it looks like there is no need to test `_update_tree_range` specifically, but I will do it, if it's really necessary.

cclauss: Private functions can still generate garbage output.  When we want to know what is the root cause of a problem, it is helpful if all the components have tests.

data_structures/binary_tree/segment_tree_other.py
```diff
@@ -4,17 +4,29 @@
...
         self.start = start
         self.end = end
         self.val = val
         self.mid = (start + end) // 2
         self.left = left
         self.right = right
+        self.lazy: T = None
```
cclauss, line 29: Perhaps T should be renamed to `IntOrFloat`.  Given the definition of `T`, it is surprising that we are setting `self.lazy` to `None`.

mrR2D2: Sounds good, I will rename `T`.

data_structures/binary_tree/segment_tree_other.py
mrR2D2: Okay, I will add tests for private methods as well.

mrR2D2: @cclauss Done.

data_structures/binary_tree/segment_tree_other.py
mrR2D2: @cclauss Done.

PR #4847, opened 2021-10-01
crawl_google_results.py update, modularize, documentation and doctest

appledora:
Added large changes to web_programming/crawl_google_results.py with `documentations` and `doctests`. Modularized the script and made it more customizable. Fixed formatting using  `black` and `flake8`. 

* [x] Add an algorithm?
* [x] Documentation change?

### **Checklist:**
* [] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.


web_programming/crawl_google_results.py
```diff
@@ -1,24 +1,39 @@
...
-if __name__ == "__main__":
-    print("Googling.....")
-    url = "https://www.google.com/search?q=" + " ".join(sys.argv[1:])
-    res = requests.get(url, headers={"UserAgent": UserAgent().random})
-    # res.raise_for_status()
-    with open("project1a.html", "wb") as out_file:  # only for knowing the class
-        for data in res.iter_content(10000):
+
+def crawl_and_store_google_search_result(query="potato", filename=""):
+    """This function will crawl google search result for the given query and
```
cclauss, line 7:
```suggestion
    """Crawl google search result for the given query and
```
Active tense.

web_programming/crawl_google_results.py
```diff
@@ -1,24 +1,39 @@
...
+    store it in a file.
+
+    Args:
+        query (str, optional): The search term provided by the user. Defaults
```
cclauss, line 11:
```suggestion
        query: The search term provided by the user. Defaults
```
Document data types with type hints as discussed in CONTRIBUTING.md.

cclauss:
Please undo the changes to requirements.txt.

appledora:
> Please undo the changes to requirements.txt.

Done. 
image

cclauss:
I think that this is a different algorithm than the original so let's keep both.  Please rename this file to `write_google_search_results.py` with an algorithmic function `get_google_search_results()` which returns a list or tuple of search results.  Put the writing of the results to a file in a separate function.  _Crawling_ is traversing each search result to go deeper which this algorithm does not do.  Please do not print in the `get_google_search_results()` function.

appledora:
> I think that this is a different algorithm than the original so let's keep both. Please rename this file to `get_google_search_results.py` with an algorithmic function `get_google_search_results()` which returns a list or tuple of search results. Put the writing of those results to a file in a separate function. _Crawling_ is traversing each search result to go deeper which this algorithm does not do. Please do not print in the `get_google_search_results()` function.

On it!

cclauss:
The problem with this pull request is that the file written to disk (when opened with a text editor) does not contain the results of the search.  When the file is opened with a web browser, the Google search is run again and the search results are shown.  This can be proven by disconnecting from the internet and then reopening the html page in a browser.

appledora:
> The problem with this pull request is that the file written to disk (when opened with a text editor) does not contain the results of the search. When the file is opened with a web browser, the Google search is run again and the search results are shown. This can be proven by disconnecting from the internet and then reopening the html page in a browser.

I also saw that and thought it was bizarre. But as this is what the original script was doing, I thought this was solving some kind of problems for some people? I think I can change it scrape titles, links and preview texts. Should I go for it?

web_programming/get_google_search_results.py
```diff
@@ -1,36 +1,96 @@
...
+    """ "Reads google search result page for the given query and
+    stores all the results in with title, link and descriptions
+    in a list.
     Args:
-        query: The search term provided by the user. Defaults to "potato".
-
-    >>> "hacktober" in get_google_search_results("hacktober")
-    True
-    >>> "potato" in get_google_search_results()
-    True
```
cclauss: Why remove all tests?

appledora: wait, I think I mistakenly removed this while merging. This is my first time trying to contribute to a public repo like this and I am really sorry for all these continuous mistakes. I will fix this. I appreciate your patience, thanks.

web_programming/get_google_search_results.py
```diff
@@ -0,0 +1,126 @@
+import urllib.parse
+
+import requests
+from bs4 import BeautifulSoup
+from requests_html import HTMLSession
```
cclauss, line 5:
```suggestion
from requests import Session
```

web_programming/get_google_search_results.py
```diff
@@ -0,0 +1,126 @@
...
+    >>> len(parse_results("python")) != None
+    True
+    >>> len(parse_results("")) == 0
+    True
+    """
+    query = urllib.parse.quote_plus(query)
+    url = "https://www.google.com/search?q=" + query
+    response = None
+    try:
+        session = HTMLSession()
```
cclauss, line 28:
```suggestion
        session = Session()
```

appledora:
@cclauss , any comments on this one?

appledora:
@poyea I think I fixed the technical and conventional errors on this one. Could you take a look?

appledora:
I am sorry but I believe I am still a little confused about the testing requirements. :| 
What I "think" you ask for is, to only test the `write_google_search_results()` method, without actually calling the `parse_results()` method which sends out the actual `REQUEST` to google. And hence, I could write a mock test method like `test_instagram_user()` as given in this [script](https://github.com/TheAlgorithms/Python/blob/master/web_programming/instagram_crawler.py) .  And by utilizing the doctest library, I will be calling this mock method? 
Am I going the write direction here?

cclauss:
web_programming/get_google_search_results.py:30: error: Name "headers" is not defined

web_programming/get_google_search_results.py
Isskta404: Why

web_programming/get_google_search_results.py
```diff
@@ -0,0 +1,125 @@
...
+    True
+    >>> "test" in write_google_search_results("python", "test")
+    True
+    >>> "test1" in write_google_search_results("", "test1")
+    True
+    >>> "potato" in write_google_search_results("", "")
+    True
+    """
+    if query == "":
+        query = "potato"
```
tianyizheng02, line 105:
```suggestion
    if not query:
        query = "potato"
```

web_programming/get_google_search_results.py
```diff
@@ -0,0 +1,125 @@
...
+    >>> write_google_search_results("python", "") != None
+    True
+    >>> write_google_search_results("", "") != None
+    True
+    >>> "test" in write_google_search_results("python", "test")
+    True
+    >>> "test1" in write_google_search_results("", "test1")
+    True
+    >>> "potato" in write_google_search_results("", "")
+    True
```
tianyizheng02, line 102: Make the doctests more strict by testing the outputs for equality:
```suggestion
    >>> write_google_search_results("python", "test") == "test.txt"
    True
    >>> write_google_search_results("", "tet.html") == "tet.html.txt"
    True
    >>> write_google_search_results("python", "") == "python.txt"
    True
    >>> write_google_search_results("", "test1") == "test1.txt"
    True
    >>> write_google_search_results("", "") == "potato.txt"
    True
```

tianyizheng02 (commented):
Closing this PR because the code no longer works. I tried running this file locally and it wasn't able to get any Google results. I believe this is because the CSS identifiers have changed, so this file is no longer able to find the results in the HTTP response.

PR #4911, opened 2021-10-02
There is an invalid case also in Knapsack. 

abhijitgawai:
Many people ignore it. Please add this invalid case where input is worng (when length of weight and value is not equal)

### **Describe your change:**

 There is invalid case in knapsack. Many people ignore it. I have included that case in knapsack.py
 Invalid case is a case where weight and value array is not of equal length

 Therefore I added  - if len(w)!=len(val): return -1
 
 I have returned -1 because otherwise user will not get to know that input was wrong.
 If I would have returned 0, user would think, there is no space in bag to accept least weight.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/knapsack/knapsack.py
+++ b/knapsack/knapsack.py
@@ -20,6 +20,10 @@ def knapsack(capacity: int, weights: list[int], values: list[int], counter: int)
     which is the limit of the capacity.
     """
 
+    # Invalid Case
+    if len(val)!=len(w):
+    	return -1
+
     # Base Case
     if counter == 0 or capacity == 0:
         return 0


knapsack/knapsack.py
```diff
@@ -20,6 +20,10 @@ def knapsack(capacity: int, weights: list[int], values: list[int], counter: int)
     which is the limit of the capacity.
     """
 
+    # Invalid Case
+    if len(val)!=len(w):
```
PzaThief, line 24: please recheck the variable name.

PzaThief (changes requested):

knapsack/knapsack.py
abhijitgawai: How do i do that now. Can you please suggest me. Should i create new PR or what. Really new here

abhijitgawai:
Hey. I previously didn't knew about branching. So, this time I created a new branch for pull request here https://github.com/TheAlgorithms/Python/pull/4956

Please apologize for any inconvenience

knapsack/knapsack.py
PzaThief: as you can see if knapsack called like knapsack(cap, w, val, c) then knapsack function accept the variables to named capacity, weights, values and counter.
so, there is no val and w now but only use values and weights.
then now you can code like this -> if len(values) != len(weights)

PR #4873, opened 2021-10-01
PasswordGenerator.py

6Infected9:
Creates a Strong Password for fighting against bruteforce attack.

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [x] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #4950, opened 2021-10-03
Added check_strong_password.py

Snimer:
### **Describe your change:**
It checks whether a giving password is strong or not

* [X] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/other/check_strong_password.py
+++ b/other/check_strong_password.py
@@ -1,0 +1,44 @@
+# This Will Check Whether A Giving Password Is Strong Or Not
+# It Follows The Rule that Length Of Password Should Be At Least 8 Characters
+# And At Least 1 Lower, 1 Upper, 1 Number And 1 Special Character
+
+
+import re
+
+
+def strong_password_detector(password: str) -> str:
+    """
+    >>> strong_password_detector('Hwea7$2!')
+    'This is a strong Password'
+
+    >>> strong_password_detector('Sh0r1')
+    'Your Password must be atleast 8 characters long'
+
+    >>> strong_password_detector('Hello123')
+    'Your Password should contain both UPPERCASE and lowercase letters with atleast 1 digit and 1 special character'
+    """
+
+    upper = re.compile(r"[A-Z]")
+    lower = re.compile(r"[a-z]")
+    num = re.compile(r"[0-9]")
+    spec_char = re.compile(r"[!@#$\^&\*\(\):;\'\"<>,\.\?\/|]")
+
+    if re.compile(r"\s").search(password) or len(password) < 8:
+        return "Your Password must be atleast 8 characters long"
+
+    elif (
+        upper.search(password)
+        and lower.search(password)
+        and num.search(password)
+        and spec_char.search(password)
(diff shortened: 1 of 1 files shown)


poyea (commented):
Thanks! I'd suggest modify the current https://github.com/TheAlgorithms/Python/blob/master/other/password_generator.py. As it's more relevant and provides more functionalities. This looks good but if we put these together, it may be better!

poyea:
Related PR #4873

other/check_strong_password.py
```diff
@@ -0,0 +1,47 @@
+# This Will Check Whether A Given Password Is Strong Or Not
+# It Follows The Rule that Length Of Password Should Be At Least 8 Characters
+# And At Least 1 Lower, 1 Upper, 1 Number And 1 Special Character
+
+
+import re
+
+
+def strong_password_detector(password: str) -> str:
```
cclauss, line 9:
```suggestion
def strong_password_detector(password: str, min_length: int = 8) -> str:
```
and make appropriate changes in the function.

other/check_strong_password.py
```diff
@@ -0,0 +1,47 @@
...
+    'This is a strong Password'
+
+    >>> strong_password_detector('Sh0r1')
+    'Your Password must be at least 8 characters long'
+
+    >>> strong_password_detector('Hello123')
+    'Password should contain UPPERCASE, lowercase, numbers, special characters'
+    """
+
+    upper = re.compile(r"[A-Z]")
```
cclauss, line 21: Would [`upper = any(char in string.ascii_uppercase for char in password)`](https://docs.python.org/3/library/string.html#string.ascii_uppercase) be faster or slower?

other/check_strong_password.py
```diff
@@ -0,0 +1,47 @@
...
+    lower = re.compile(r"[a-z]")
+    num = re.compile(r"[0-9]")
+    spec_char = re.compile(r"[!@#$\^&\*\(\):;\'\"<>,\.\?\/|]")
+
+    if re.compile(r"\s").search(password) or len(password) < 8:
+        return "Your Password must be at least 8 characters long"
```
cclauss, line 27: Move these lines before line 22 so that we do not do the work on lines 22-24 if the password is too short.

other/check_strong_password.py
```diff
@@ -0,0 +1,47 @@
...
+
+def strong_password_detector(password: str) -> str:
+    """
+    >>> strong_password_detector('Hwea7$2!')
+    'This is a strong Password'
+
+    >>> strong_password_detector('Sh0r1')
+    'Your Password must be at least 8 characters long'
+
+    >>> strong_password_detector('Hello123')
```
cclauss, line 17: One of the tests should be a long (like 20 char) string.  Also tests for strong_password_detector(0), strong_password_detector(1.3), strong_password_detector(['H', 'w', 'e', 'a', '7', '$', '2', '!'])

cclauss (changes requested):

PR #4956, opened 2021-10-03
There is an invalid case also in Knapsack

abhijitgawai:
Many people ignore it. Please add this invalid case where input is worng (when length of weight and value is not equal)

### **Describe your change:**

There is invalid case in knapsack. Many people ignore it. I have included that case in knapsack.py
Invalid case is a case where weight and value array is not of equal length

Therefore I added - if len(wights)!=len(values): return -1

I have returned -1 because otherwise user will not get to know that input was wrong.
If I would have returned 0, user would think, there is no space in bag to accept least weight.

* [ ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/knapsack/knapsack.py
+++ b/knapsack/knapsack.py
@@ -20,6 +20,10 @@ def knapsack(capacity: int, weights: list[int], values: list[int], counter: int)
     which is the limit of the capacity.
     """
 
+    # Invalid Case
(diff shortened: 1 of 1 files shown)


cclauss:
Closed as invalid for the derogatory comments made about other contributors' work at
#1 (review)

PR #5232, opened 2021-10-11
Python Kinematics

SaucyGames05:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/kinematics/velocity.py
+++ b/kinematics/velocity.py
@@ -18,7 +18,7 @@
 # are more accurate than the one used for the example.
 
 # Typing hints
-from typing import Union
+from typing import Union, Optional
 
 
 # Square root function just so I don't have to import math and it looks cleaner
@@ -69,12 +69,12 @@ class NotEnoughInfo(Exception):
 
 
 def velocity(
-    intial_position: Union[int, float, None] = None,
-    final_position: Union[int, float, None] = None,
-    intial_velocity: Union[int, float, None] = None,
-    acceleration: Union[int, float, None] = None,
-    time_elapsed: Union[int, float, None] = None,
-    retaining_wall: Union[int, None] = None,
+    intial_position: Optional[Union[int, float]] = None,
+    final_position: Optional[Union[int, float]] = None,
+    intial_velocity: Optional[Union[int, float]] = None,
+    acceleration: Optional[Union[int, float]] = None,
+    time_elapsed: Optional[Union[int, float]] = None,
+    retaining_wall: Optional[int] = None,
 ) -> float:
     """
     Find velocity with given intial_velocity, intial_position, final_position,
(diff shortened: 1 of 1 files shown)


cclauss:
Don't do that.  Why are you forcing your reader to guess if `d` is `distance`, or `dimension` or `delta`?  Guessing is not helpful to understanding.

kinematics/acceleration.py
```diff
@@ -0,0 +1,129 @@
...
+# time_elapsed=32.8, retaining_wall=2
+
+# These were calculated by hand so the output for distance is different, but the outputs
+# are more accurate than the one used for the example.
+
+# In all cases distance is final_position-initial_position
+# TOTAL distance is final_position+initial_position
+# If you want to find distance from one point to another assume initial_position=0
+
+# Typing hints
```
cclauss, line 26: We can lose the comment which is redundant.  Please add
```suggestion
from __future__ import annotations

```
so that `pyupgrade` can replace all `Optional` and `Union` with modern syntax.  Same changes please on the other files.

kinematics/acceleration.py
```diff
@@ -0,0 +1,129 @@
...
+
+# Here is an explanation on the "big 4 equations", all of which were used to make this
+# algorithm: https://www.physicsclassroom.com/class/1dkin/Lesson-6/Kinematic-Equations
+
+# Output will always be a float, so if you need to round, set retaining_wall to
+# whatever place you are rounding to
+
+# NOTE: Not all of these equations will give you an exactly same answer, but will be
+# very close. Typically you would only show sig
+# figs when you calculate with these equations.
```
cclauss, line 13:
```suggestion
# very close. Typically you would only show sig figs when you calculate with these
# equations.
```
What is a _sig fig_ for the uninitiated?

kinematics/acceleration.py
```diff
@@ -0,0 +1,129 @@
+# KINEMATICS #
```
cclauss, line 1: From here to line 24 should be a docstring instead of comments so that the content would appear in any generated documentation.

kinematics/acceleration.py
```diff
@@ -0,0 +1,129 @@
...
+    Find acceleration for given initial_position, final_position, initial_velocity,
+    final_velocity, or time_elapsed.
+
+    :param initial_position: int, float
+    :param final_position: int, float
+    :param initial_velocity: int, float
+    :param final_velocity: int, float
+    :param time_elapsed: int, float
+    :param retaining_wall: int
+    :return: float
```
cclauss, line 55: This is repetitive of the info is already stated in the type hints above and modern document generators no longer need data in this format.  A description of the purpose and/or limits of each parameter or return value would be a helpful addition.

kinematics/acceleration.py
```diff
@@ -0,0 +1,129 @@
...
+    initial_position = initial_position
+    time_elapsed = time_elapsed
+    retaining_wall = retaining_wall
+    
+    if not time_elapsed:
+        if None in (initial_velocity, final_velocity, final_position, initial_position):
+            raise NotEnoughInfo("Not enough information to complete calculation.")
+        else:
+            if retaining_wall is not None:
+                if type(retaining_wall) != int:
```
cclauss, line 91:
```suggestion
                if not isinstance(retaining_wall, int):
```
See PEP8.  Please repeat this change below.

cclauss (changes requested):
This is a supercool contribution.  Thanks for doing this.  A few minor points and then we can merge.

cclauss:
In preparation for that denial-of-service attack that is [Hacktoberfest](https://hacktoberfest.com), I am closing all pull requests that do not have passing tests.

PR #5453, opened 2021-10-19
First commit for add_prefix_conversion

manueldilullo:
### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**


poyea:
I think we already have one? : https://github.com/TheAlgorithms/Python/blob/master/conversions/prefix_conversions.py

manueldilullo:
@poyea I was inspired by this https://github.com/TheAlgorithms/Python/blob/master/conversions/prefix_conversions.py, but it works slightly differently. If I understand how it operates, _prefix_conversions.py_ wants a known_prefix as an input to the function, so if we have a number written in his full form we can't convert it. My code works without a known_prefix, it takes the number and write it using a “notation with prefix”. 
Would you suggest I open an issue and fix this directly in _prefix_conversions.py_ instead of creating a new file?

manueldilullo:
> 
> 
> > @poyea I was inspired by this https://github.com/TheAlgorithms/Python/blob/master/conversions/prefix_conversions.py, but it works slightly differently. If I understand how it operates, _prefix_conversions.py_ wants a known_prefix as an input to the function, so if we have a number written in his full form we can't convert it. My code works without a known_prefix, it takes the number and write it using a “notation with prefix”. Would you suggest I open an issue and fix this directly in _prefix_conversions.py_ instead of creating a new file?
> 
> So it is to add a prefix to a number

Yes, it is

poyea (approved):

PR #5647, opened 2021-10-28
Fixes recursion exception in davisb_putnamb_logemann_loveland

spazm:
### Describe your change:

+ Fixes the recursion exception bug in dpll.  Removes a case where recursion is called with unmodified symbols.
+ Fixes recursion branch bug where both left and right actually evaluating right.  Python arrays are copied by value, so both calls received the same modified model.
+ Adds step comments
+ renames vars to better semantic values
+ pythonic clean-up around arrays.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/other/davisb_putnamb_logemannb_loveland.py
+++ b/other/davisb_putnamb_logemannb_loveland.py
@@ -319,11 +319,9 @@ def dpll_algorithm(
     except RecursionError:
         print("raises a RecursionError and is")
         return None, {}
-    P = None
-    if len(pure_symbols) > 0:
-        P, value = pure_symbols[0], assignment[pure_symbols[0]]
 
-    if P:
+    if pure_symbols:
+        P, value = pure_symbols[0], assignment[pure_symbols[0]]
         tmp_model = model
         tmp_model[P] = value
         tmp_symbols = [i for i in symbols]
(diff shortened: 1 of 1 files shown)


other/davisb_putnamb_logemannb_loveland.py
```diff
@@ -304,54 +314,77 @@ def dpll_algorithm(
...
-    print(f"The formula {formula} is", end=" ")
-
-    clauses, symbols = generate_parameters(formula)
-    solution, model = dpll_algorithm(clauses, symbols, {})
-
-    if solution:
-        print(f"satisfiable with the assignment {model}.")
-    else:
-        print("not satisfiable.")
+    # recursion example:
```
L3str4nge, line 356: I would remove this block of code

other/davisb_putnamb_logemannb_loveland.py
```diff
@@ -292,6 +299,9 @@ def dpll_algorithm(
     >>> model
     {'A4': True}
     """
+
+    # 2. Check any False
```
L3str4nge, line 303: 2 and then 1?

L3str4nge (changes requested):

PR #5698, opened 2021-10-30
Added reddit.py to get data from reddit

JDeepD:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/web_programming/reddit.py
+++ b/web_programming/reddit.py
@@ -23,7 +23,7 @@ def get_subreddit_data(
     limit : Number of posts to fetch
     age : ["new", "top", "hot"]
     wanted_data : Get only the required data in the list
-    
+
     >>> pass
     """
     wanted_data = wanted_data or []
--- a/web_programming/reddit.py
+++ b/web_programming/reddit.py
@@ -1,5 +1,3 @@
-from logging import raiseExceptions
-
 import requests
 
 
@@ -34,7 +32,7 @@ def get_data(
     """
     response = requests.get(f"https://reddit.com/r/{sub}/{age}.json?limit={limit}")
     if response.raise_for_status():
-        raiseExceptions
+        raise requests.HTTPError
 
     data = response.json()
     data_dict = {}
--- a/web_programming/reddit.py
+++ b/web_programming/reddit.py
@@ -1,59 +1,52 @@
+from __future__ import annotations
+
 import requests
 
 
-def get_data(
-    sub: str, limit: int = 1, age: str = "new", wanted_data: list = []
+valid_terms = set(
+    """approved_at_utc approved_by author_flair_background_color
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,4 +1,5 @@
 beautifulsoup4
+praw
```
cclauss: Not in alphabetical order.  We are going to need multiple algorithms that use praw to justify its addition to this small list.

JDeepD: Fixed in alphabetical order. Finding a way to do this without praw.

web_programming/reddit.py
```diff
@@ -0,0 +1,45 @@
...
+
+            1: {...},
+            2: {...},
+            ...
+        }
+    """
+    reddit = praw.Reddit(**credentials)
+
+    data: dict = {}
+    for _id, submission in enumerate(reddit.subreddit(subreddit).hot(limit=limit)):
```
cclauss, line 35: What does `type(reddit.subreddit(subreddit).hot(limit=limit))` say?  Is it already a `dict`?

JDeepD: No. Its a generator

`<class 'praw.models.listing.generator.ListingGenerator'>`

cclauss: OK...  So what value does this algorithm add over what `reddit.subreddit().hot()` already delivers?  If the answer is "not much added value" then this is a how-to-use for praw.

JDeepD: @cclauss  I found 2 feasable ways of doing it without `praw` but both of them has some downsights:

1. Scrape reddit using BS4. ( Will be very complicated to scrape data and highly rate limited)
2. Use reddit API directly (Rate limited extremely : like 1 requests per minute)

Which one do you prefer?

I could get the 2nd method working by tomorrow. BS4 might take some time(don't even know if it would be possible)

JDeepD: @cclauss  Rewritten using `requests`

web_programming/reddit.py
```diff
@@ -34,7 +32,7 @@ def get_data(
     """
     response = requests.get(f"https://reddit.com/r/{sub}/{age}.json?limit={limit}")
     if response.raise_for_status():
-        raiseExceptions
+        raise requests.HTTPError
```
JDeepD, line 35: In this case, it is important to raise exception otherwise if the `response` is a 429 (too many requests) , then `response.json()` would make an empty json(and not call an error). This would be problematic because then it would give a `Keyerror` and a false impression that error is caused in Line42 when in reality, the response returned an HTTPError.

JDeepD:
Ping @cclauss

JDeepD:
Please review @cclauss

cclauss (approved):
Thanks for doing this.

PR #5722, opened 2021-10-31
Added a fibonacci algorithm which uses memoization

JDeepD:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/dynamic_programming/fibonacci.py
+++ b/dynamic_programming/fibonacci.py
@@ -5,61 +5,48 @@ sequence problem.
 
 
 class Fibonacci:
-    def __init__(self, N=None):
-        self.fib_array = []
-        if N:
-            N = int(N)
-            self.fib_array.append(0)
-            self.fib_array.append(1)
-            for i in range(2, N + 1):
-                self.fib_array.append(self.fib_array[i - 1] + self.fib_array[i - 2])
-        elif N == 0:
-            self.fib_array.append(0)
-        print(self.fib_array)
+    def __init__(self) -> None:
+        self.sequence = [0, 1]
 
-    def get(self, sequence_no=None):
+    def get(self, index: int) -> list:
         """
-        >>> Fibonacci(5).get(3)
-        [0, 1, 1, 2, 3, 5]
-        [0, 1, 1, 2]
-        >>> Fibonacci(5).get(6)
-        [0, 1, 1, 2, 3, 5]
-        Out of bound.
-        >>> Fibonacci(5).get(-1)
-        [0, 1, 1, 2, 3, 5]
-        []
+        Get the Fibonacci number of `index`. If the number does not exist,
+        calculate all missing numbers leading up to the number of `index`.
+
+        >>> Fibonacci().get(10)
+        [0, 1, 1, 2, 3, 5, 8, 13, 21, 34]
+        >>> Fibonacci().get(5)
+        [0, 1, 1, 2, 3]
         """
-        if sequence_no is not None:
-            if sequence_no < len(self.fib_array):
-                return print(self.fib_array[: sequence_no + 1])
-            else:
-                print("Out of bound.")
-        else:
-            print("Please specify a value")
+        difference = index - (len(self.sequence) - 2)
+        if difference >= 1:
+            for _ in range(difference):
+                self.sequence.append(self.sequence[-1] + self.sequence[-2])
+        return self.sequence[:index]
+
+
+def main():
+    print(
+        "Fibonacci Series Using Dynamic Programming\n",
+        "Enter the index of the Fibonacci number you want to calculate ",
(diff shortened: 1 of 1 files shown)


JDeepD:
I dont know why the build is failing for my PR. It says :

```
Successfully installed types-setuptools-57.4.2 types-toml-0.10.1
strings/check_anagrams.py:32: error: Need type annotation for "count"
Installing missing stub packages:
/opt/hostedtoolcache/Python/3.9.7/x64/bin/python -m pip install types-setuptools types-toml


Found 1 error in 1 file (checked 677 source files)
Error: Process completed with exit code 1.
```

As per error, I think [this](https://github.com/TheAlgorithms/Python/blob/868c2fa0a8e1f51ba32e7622990a9259a8740604/strings/check_anagrams.py#L32) `count` variable should get a type hint.

maths/fibonacci_sequence_memoization.py
```diff
@@ -0,0 +1,29 @@
+cache: dict[int, int] = {}
```
cclauss, line 1: Prefill the cache...
```suggestion
cache: dict[int, int] = {1: 1, 2: 1}
```

maths/fibonacci_sequence_memoization.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    if num < 1:
+        raise ValueError("Enter positive integer")
+    elif num in cache:
+        return cache[num]
+    elif num in [1, 2]:
+        value = 1
+    else:
+        value = fibo(num - 1) + fibo(num - 2)
+
```
cclauss, line 23:
```suggestion
    value = fibo(num - 1) + fibo(num - 2)
```

cclauss (changes requested):

pandrey2003:
@JDeepD btw, why is it chosen to define cache at the module level? Classical memoization technique enables creating an outer and inner function, something like:
```
def fibo() -> Callable:
    cache: dict[int, int] = {1: 1, 2: 1}

    def func_(n: int) -> int:
        nonlocal cache
        ...

    return func_
```

... and then we do something like:
```
fibonacci = fibo()
fibonacci(5)
```

The code in this PR is perfectly workable, but what about showcasing the "classical" ways of solving the problems?

JDeepD:
> @JDeepD btw, why is it chosen to define cache at the module level? Classical memoization technique enables creating an outer and inner function, something like:
> 
> ```
> def fibo() -> Callable:
>     cache: dict[int, int] = {1: 1, 2: 1}
> 
>     def func_(n: int) -> int:
>         nonlocal cache
>         ...
> 
>     return func_
> ```
> 
> ... and then we do something like:
> 
> ```
> fibonacci = fibo()
> fibonacci(5)
> ```
> 
> The code in this PR is perfectly workable, but what about showcasing the "classical" ways of solving the problems?

Yeah sorry, I will change it right away.

JDeepD:
> ```
> fibonacci = fibo()
> fibonacci(5)
> ```

Rather than creating another object, wouldn't it be better if we just pass the `num` from `fibonacci` to `fibo` directly?
Something like:

```
def fibo(num: int) -> int: 
       cache: dict[int, int] = {}
       def fibo(num:int) -> int:
           ...
       return fibo(num) 


```

JDeepD:
Wait I messed something while pushing

pandrey2003:
Hmm, quite frankly, I never saw such an implementation. My checks led me to one classical book about competitive programming ([link](https://cses.fi/book/book.pdf)), where we can see such an implementation of memoization in C++:

```
/* ready[x] indicates whether the value of solve(x) has been calculated,
and if it is, value[x] contains this value.
The constant N has been chosen so that all required values fit in the arrays. */
bool ready[N];
int value[N];

int solve(int x) {
    // the function per se
    if (x < 0) return INF;
    if (x == 0) return 0;
    if (ready[x]) return value[x];
    int best = INF;
    for (auto c : coins) {
       best = min(best, solve(x-c)+1);
    }
    value[x] = best;
    ready[x] = true;
    return best;
}
```

This piece of code is structurally implemented as your PR: cache is defined at the module level. However, this example shows how to write the code in the fastest way (because writing inner and outer functions takes time in programming contests).

However, if we look at some other [example](https://codeburst.io/understanding-memoization-in-3-minutes-2e58daf33a19), we can see we create an outer and an inner function in all cases: because we re-use this code and it should be written well.

JDeepD:
Yeah, I also think putting the `cache`  outside the function is not good practice.Nevertheless, I have somehow managed to mess up the commit(It disappeared from remote smh). I will make a new PR shortly...

pandrey2003:
OK, I'll take a look at it then :)

JDeepD:
Thnx. The new PR is here #5856

PR #5736, opened 2021-11-01
Add search book via ISBN using openlibrary.org API

kellydosocorro:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/web_programming/search_books_by_isbn.py
+++ b/web_programming/search_books_by_isbn.py
@@ -3,14 +3,37 @@ import requests
 URL_BASE = "https://openlibrary.org"
 
 def get_book_by_isbn(isbn: str = "0140328726") -> dict:
+    """
+    Given a isbn code, return a dict.
+
+    :param isbn: string
+    :return: dict
+
+    >>> get_book_by_isbn('0140328726')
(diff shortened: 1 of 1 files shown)


L3str4nge:
I am not sure if we want to have external dependencies like this API. I think it slows down the tests. Is it algorithm actually??

@cclauss @poyea

cclauss:
@L3str4nge The `web_programming` directory is a bit special in that doctests are much less essential because they slow down our continuous integration and the Open Library may change the data which would break our tests.  I will propose a workaround to @kellydosocorro.

On the ___Is it an algorithm___ question in this directory, we have been requesting that there be some [pre- or post-processing of the data](https://github.com/TheAlgorithms/Python/pull/5736/files#diff-6cbe7ec8a82c933741fa7a2ee6744a6bdb0f7cd6f030fcdaa0d498dd47d5b822R48) to add value beyond the raw data provided by the api.  So the entries in this directory  are often more how-to-use-this-api than actual algorithms but I believe that @kellydosocorro has a good submission here.  Suggestions to follow…

cclauss:
Please run psf/black on this code as discussed in CONTRIBUTING.md to fix the failing test.

cclauss (approved):
LGTM.  Thanks for doing this.

poyea (approved):

PR #5739, opened 2021-11-01
[mypy] Fix type annotations in `data_structures/binary_tree/red_black_tree.py`

dylanbuchi:
### Describe your change:

Related Issue: #4052 
* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/binary_tree/red_black_tree.py
+++ b/data_structures/binary_tree/red_black_tree.py
@@ -51,18 +51,13 @@ class RedBlackTree:
         """
         parent = self.parent
         right = self.right
-
         if right is None:
             return self
-
         self.right = right.left
-
         if self.right:
             self.right.parent = self
         self.parent = right
-
         right.left = self
-
         if parent is not None:
             if parent.left == self:
                 parent.left = right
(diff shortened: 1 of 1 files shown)


data_structures/binary_tree/red_black_tree.py
```diff
@@ -150,7 +161,7 @@ def remove(self, label: int) -> RedBlackTree:
                 # it and remove that.
                 value = self.left.get_max()
                 self.label = value
-                self.left.remove(value)
+                self.left.remove(value) if value else None
```
spazm, line 164: This should be a regular if condition because we are not capturing the value. 
```
if value:
    self.left.remove(value)
```


Questions:
1.  Should we be checking for `is None` rather than a false value of `value`?
2. When will `self.left.get_max()` not return a value?

dylanbuchi: > This should be a regular if condition because we are not capturing the value.

@spazm Yes, you're right!

dylanbuchi: > Questions:
> 
> 1. Should we be checking for `is None` rather than a false value of `value`?
> 2. When will `self.left.get_max()` not return a value?

1. Yes, changing to: `if value is not None` because `0` is Falsy and the `value` is an ` int`.
2.  `self.left.get_max()` can return `None` because ` self.label ` is set to `None` by default.

spazm: I like the update. :+1:

cclauss (approved):

PR #5771, opened 2021-11-04
Improve Project Euler problem 203 solution 1

MaximSmolskiy:
### Describe your change:
Improve Project Euler problem 203 solution 1 - the top 9 slowest solution on GitHub Actions logs (under `slowest 10 durations`: `3.82s call     scripts/validate_solutions.py::test_project_euler[problem_203/sol1.py]`):
* Improve solution (locally 1500+ times - from 3+ seconds to ~2 milliseconds)

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -253,6 +253,7 @@
 
 ## Dynamic Programming
   * [Abbreviation](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/abbreviation.py)
+  * [All Construct](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/all_construct.py)
   * [Bitmask](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/bitmask.py)
   * [Catalan Numbers](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/catalan_numbers.py)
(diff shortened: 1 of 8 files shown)


cclauss (changes requested):
These are big changes and changing the test answers at the same time as changing the code worries me.  @fernandobperezm can you please review the proposed changes?

MaximSmolskiy:
> These are big changes and changing the test answers at the same time as changing the code worries me. @fernandobperezm can you please review the proposed changes?

Changing the test answers shouldn't worry you. I change answers for 3 tests:
1 and 2:
```
>>> get_squarefree({1}, [])
set()
>>> get_squarefree({1, 2}, [])
set()
```
I don't understand why the answers are empty sets due to:
```
if len(squared_primes) == 0:
        return set()
```
In my opinion, when we don't provide any primes then we should assume that all numbers are squarefree instead of non-squarefree. But anyway I changed the interface of this function and now it gets only numbers without primes. So, obviously 1 and 2 are squarefree numbers, so I add correct answers.
3:
```
>>> solution(1)
    0
```
The first row of Pascal's triangle contains only one number 1 and it is squarefree number, so the answer should be 1 instead of 0.

PR #5783, opened 2021-11-06
Added: pressure.py

reisgoldmanX:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/physics/pressure.py
+++ b/physics/pressure.py
@@ -1,0 +1,52 @@
+"""
+Description :
+https://en.wikipedia.org/wiki/Pressure
+
+Pressure is the amount of force applied at right angles to the surface of an object per unit area.
+P = F/A
+
+P is the pressure,
+F is the magnitude of the normal force,
+A is the area of the surface on contact.
+"""
+
+
+def solid_pressure(force: float, area: float) -> float:
+    """
+    Return the force applied perpendicular to the surface of an object(Pressure).
+    >>> solid_pressure(10, 5)
+    2.0
+
+    >>> solid_pressure(5.5, 2.1)
+    2.619047619047619
+
+    >>> solid_pressure(0, -3)
+    Traceback (most recent call last):
+    ...
+    ValueError: Force and Area must be > 0
+
+    >>> solid_pressure(-4.3, 2)
+    Traceback (most recent call last):
+    ...
+    ValueError: Force must be > 0
+
+    >>> solid_pressure(4, -2)
+    Traceback (most recent call last):
+    ...
+    ValueError: Area must be > 0
+    """
+    if force <= 0 and area <= 0:
+        raise ValueError("Force and Area must be > 0")
+    elif force <= 0:
+        raise ValueError("Force must be > 0")
+    elif area <= 0:
+        raise ValueError("Area must be > 0")
+    else:
+        return force / area
+
+
+if __name__ == '__main__':
+    import doctest
+
+    # run doctest
+    doctest.testmod()


physics/pressure.py
```diff
@@ -0,0 +1,52 @@
...
+
+    >>> solid_pressure(4, -2)
+    Traceback (most recent call last):
+    ...
+    ValueError: Area must be > 0
+    """
+    if force <= 0 and area <= 0:
+        raise ValueError("Force and Area must be > 0")
+    elif force <= 0:
+        raise ValueError("Force must be > 0")
```
Rohanrbharadwaj, line 41: Why force always > 0 ? It is a vector quantity after all. Only that pressure is scalar and should be the absolute value of the quotient...

physics/pressure.py
```diff
@@ -0,0 +1,52 @@
...
+    elif area <= 0:
+        raise ValueError("Area must be > 0")
+    else:
+        return force / area
+
+
+if __name__ == '__main__':
```
Rohanrbharadwaj, line 48:
```suggestion
if __name__ == "__main__":

```

IdoErel (approved):
Very simple and easy to understand.
Detailed concept & use cases in docs.

Kind regards,
IdoErel

Ralas13 (approved):
Simple and efficient.

physics/pressure.py
```diff
@@ -0,0 +1,53 @@
...
+    ValueError: Area must be > 0
+    """
+    if force <= 0 and area <= 0:
+        raise ValueError("Force and Area must be > 0")
+    elif force <= 0:
+        raise ValueError("Force must be > 0")
+    elif area <= 0:
+        raise ValueError("Area must be > 0")
+    else:
+        return force / area
```
Rohanrbharadwaj, line 45:
```suggestion
    if force <= 0 and area <= 0:
        raise ValueError("Force and Area must be > 0")
    if force <= 0:
        raise ValueError("Force must be > 0")
    if area <= 0:
        raise ValueError("Area must be > 0")
    return float(force) / float(area)
```

physics/pressure.py
```diff
@@ -0,0 +1,94 @@
...
+F is the magnitude of the normal force,
+A is the area of the surface on contact.
+
+- Liquid Pressure
+https://en.wikipedia.org/wiki/Pressure#Liquid_pressure
+P = p*g*h
+P is liquid pressure,
+g is gravity at the surface of overlaying material,
+ρ is density of liquid,
+h is height of liquid column or depth within a substance.
```
Rohanrbharadwaj, line 17:
```suggestion
P = ρ*g*h
P is liquid pressure,
g is gravity at the surface of overlaying material,
ρ is density of liquid,
h is height of liquid column or depth within a substance.
```

Rohanrbharadwaj:
Why can't force, gravity and height be 0 ??

physics/pressure.py
IdoErel: To make these changes, you need to specify the file's encoding to be utf-8. 

Because Not many would notice it until run-time. Even though this encoding may be the default for your IDE, and obvious for you -  it may not be so for others.

Kind regards,
IdoErel

Rohanrbharadwaj: There's ρ in on line 16 already, but it's 'p' on line 13.

IdoErel: Ok, you're right - keeping the docs consistent. Yet this should be addressed anyways.

Even I missed this simple special character in there. This is the kind of stuff that should be resolved ahead of time.

Many thanks,
IdoErel

PR #5786, opened 2021-11-06
Fix `sorts/bucket_sort.py` implementation

drinkertea:
### Describe your change:

Two issues:
* Current implementation is equal to `sorted` function, because all the items goes into the same bucket.
* Buckets count is equal to integer difference between `min` and `max`, the problem illustrated by a new test-case. In general, we don't know the type of data, so `bucket_count` should be set outside. Or can be replaced by some heuristic like `len(my_list) * 2`.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


sorts/bucket_sort.py
```diff
@@ -30,7 +30,7 @@
 from __future__ import annotations
 
 
-def bucket_sort(my_list: list) -> list:
+def bucket_sort(my_list: list, bucket_count: int = 10) -> list:
```
L3str4nge, line 33: why bucket count is set to 10?

drinkertea: Just some default value. It can be 100 or whatever.  
Or we can set it to `None` by default and calculate it by some heuristic if it's not set, ex. `len(my_list) // 2`.

vijaysr4 (approved):

sorts/bucket_sort.py
sidoknowia, line 33: very nit-picky / low-priority - 
Can we change variable name to something more descriptive? Maybe have list_to_sort instead of my_list?

sidoknowia: I suggest that rather this having bucket_count as a parameter, we can have it as part of implementation logic where we calculate bucket count based on type of machine 32 for 32bit, 64 for 64bit. In that way we eliminate user driven errors

sidoknowia (commented):
Great work! I provided few suggestions in comments, please let me know if you have any questions.

tianyizheng02 (approved):
I believe the first issue you mentioned has already been fixed in #6005, so there's now a merge conflict with this PR. However, I still think that making `bucket_count` a function parameter is a good change, so I'll try to fix the merge conflict and merge this PR.

PR #5802, opened 2021-11-10
Update `sorts/bubble_sort.py`

dylanbuchi:
### Describe your change:

- In `sorts/bubble_sort.py`:
    - Add missing type annotations
    - Refactor `bubble_sort` function

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/sorts/bubble_sort.py
+++ b/sorts/bubble_sort.py
@@ -1,4 +1,7 @@
-def bubble_sort(collection):
+from typing import Any
+
+
+def bubble_sort(collection: list[Any]) -> list[Any]:
     """Pure implementation of bubble sort algorithm in Python
 
     :param collection: some mutable ordered collection with heterogeneous
--- a/sorts/bubble_sort.py
+++ b/sorts/bubble_sort.py
@@ -31,9 +31,9 @@ def bubble_sort(collection: list[Any]) -> list[Any]:
     True
     """
     length = len(collection)
-    for i in range(length - 1):
+    for i in reversed(range(length)):
         swapped = False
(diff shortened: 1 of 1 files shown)


sorts/bubble_sort.py
```diff
@@ -28,9 +31,9 @@ def bubble_sort(collection):
     True
     """
     length = len(collection)
-    for i in range(length - 1):
+    for i in reversed(range(length)):
```
L3str4nge, line 34: Is calling `reversed` is more efficient? I don't think so. Did you do some benchmarks?

dylanbuchi: `reversed` has to be a bit better since the Python `heapq.py` module uses it.
It also improves code readability. @L3str4nge 

Source: https://www.python.org/dev/peps/pep-0322/

L3str4nge: I know but it would be good to know if algorithms performance is better too :) Could you perform any benchmarks before and after refactor?

dylanbuchi: Benchmarked with the `timeit` module:

1000 items:

```
Before: 
0.10902250000071945

After: 
0.11019620000297436
```
5000 items:

```
Before: 
2.8898907999973744

After: 
2.844012000001385
```
10000 items:

```
Before: 
12.34897049999563

After: 
11.655540399995516
```
I tested a bit more locally and sometimes refactored one is better and sometimes it's the original.

tianyizheng02 (approved):

PR #5833, opened 2021-11-17
Update merge_insertion_sort.py

### LINKED ISSUE #5774: Merge insertion sort doesn't work
```
>>> merge_insertion_sort([0, 1, 2, 3, 4])
[0, 2, 1, 3, 4]
```

Cc @ulwlu #2211 [`sorts/merge_insertion_sort.py`](../blob/master/sorts/merge_insertion_sort.py)

yellowsto:
Fixes #5774

merge_insertion_sort

Co-Authored-By: AilisOsswald <44617437+AilisOsswald@users.noreply.github.com>

### Describe your change:
We fixed the Issue of Merge-insertion-Sort not sorting correctly.

We have seen that there is already a PR but we would like to present our fix since it is a little bit different: https://github.com/reidstrieker/Python/commit/2ab90dfa03a9b14400eaf82ed52bb518dc2bd835

We have left all Code as it was before, except line 163
original:
 if result[i] == collection[-i]:
corrected: 
 if result[i] == collection[-1]:

because as the flag name states we want to check if the current item in result is the last item of our input list.  (which would be collection[-1] not [-i]

We think that both solutions should work, but the solution with checking (as described above) should perform a tiny bit better. 
Normally we do this, since we know everything left to the current value is smaller than the value we will have to sort in: 
 result = result[: i + 1] + binary_search_insertion(result[i + 1 :], pivot)

If we find the last_odd_item in our result list, we know the same thing like above and we know that there is one additional value, which is already sorted - the last_odd_item. And with that knowlegde we can make the list we have to search through smaller and the list we don't have to look at larger by one each:
 result = result[: i + 1 +1] + binary_search_insertion(result[i + 1 :], pivot)

Keeping this flag in the code should make it a bit quicker, since in some cases the list to be searched through is one element smaller.

Our next step would be writing more tests as suggested in this issue #5774. Should this be done in the same issue or in a new one?

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/sorts/merge_insertion_sort.py
+++ b/sorts/merge_insertion_sort.py
@@ -34,8 +34,8 @@ def merge_insertion_sort(collection: list[int]) -> list[int]:
     Testing with all permutations on range(0,5):
     >>> import itertools
     >>> permutations = list(itertools.permutations([0, 1, 2, 3, 4]))
-    >>> [x for x in permutations if merge_insertion_sort(x) != [0, 1, 2, 3, 4]]
-    []
+    >>> all(merge_insertion_sort(p) == [0, 1, 2, 3, 4] for p in permutations)
+    True
     """
 
     def binary_search_insertion(sorted_list, item):
--- a/sorts/merge_insertion_sort.py
+++ b/sorts/merge_insertion_sort.py
@@ -160,7 +160,7 @@ def merge_insertion_sort(collection: list[int]) -> list[int]:
     """
     is_last_odd_item_inserted_before_this_index = False
     for i in range(len(sorted_list_2d) - 1):
-        if result[i] == collection[-1]:
+        if result[i] == collection[-1] and has_last_odd_item:
             is_last_odd_item_inserted_before_this_index = True
         pivot = sorted_list_2d[i][1]
         # If last_odd_item is inserted before the item's index,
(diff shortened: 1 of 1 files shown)


MaximSmolskiy (changes requested):
Unfortunately, https://github.com/TheAlgorithms/Python/pull/5832 looks more correctly. I tested both solutions on all permutations of `range(n)` up to `n` equal to `10`. https://github.com/TheAlgorithms/Python/pull/5832 passes them all. But this doesn't pass. For example, `merge_insertion_sort([2, 3, 1, 0]) = [0, 2, 1, 3]`, but the correct answer is `[0, 1, 2, 3]`

yellowsto:
Thank you for the hint! We found out why our solution didn't work properly and we have updated it:

Now we check if the length of the input is an odd number, so that the if only triggers if that is the case. 

Furthermore we have tested this on all permutations from range(0,10).

MaximSmolskiy (approved):
Yes, now it passes these tests

Rohanrbharadwaj (approved):

Rohanrbharadwaj:
Looks good, don't mind adding a `doctest.testmod()` towards the end.

yellowsto:
Could any one have a look please? We would appreciate if we could have a feedback :)

poyea (approved):

poyea:
Let's use `all` if you don't mind. Otherwise, I'll merge this later. Thank you for your pull request!🤩

poyea (approved):

PR #5849, opened 2021-11-25
Changed filename, added non-preemptive SJF

sangin-lee:
### Describe your change:
Changed filename of previous [shortest_job_fisrt.py] to [shortest_remaining_time_first.py] because it comes under to
URL : https://en.wikipedia.org/wiki/Shortest_remaining_time

Added new [shortest_job_first.py] which is non-preemptive shortest job first. It comes under to
URL : https://en.wikipedia.org/wiki/Shortest_job_next

+Added Wikipedia URL in each comments

* [x] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/scheduling/shortest_job_first.py
+++ b/scheduling/shortest_job_first.py
@@ -107,8 +107,8 @@ def calculate_average_times(
     for i in range(no_of_processes):
         total_waiting_time = total_waiting_time + waiting_time[i]
         total_turn_around_time = total_turn_around_time + turn_around_time[i]
-    print("Average waiting time = %.5f" % (total_waiting_time / no_of_processes))
-    print("Average turn around time =", total_turn_around_time / no_of_processes)
+    print(f"Average waiting time = {total_waiting_time / no_of_processes:.5f}")
+    print(f"Average turn around time = {total_turn_around_time / no_of_processes:.5f}")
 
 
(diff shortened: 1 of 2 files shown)


sangin-lee:
@cclauss Can you review my work?

sangin-lee:
@cclauss  @dhruvmanila Could you please review my work?

sangin-lee:
Fixed doctest results
ex) 5.0 -> 5.00000

cclauss:
As a user, I would find it very distracting to see `5.00000` instead of `5.0`.  All these extra zeros create unnecessary cognitive load for the reader so I would much prefer the clear and simple `5.0` instead

sangin-lee:
Would it be better to modify codes to print out only first digit of the decimal point?

cclauss:
5.12345 is OK if there are actual values there to the right of the decimal point but 5.00000 must be shortened to 5.0

sangin-lee:
> 5.12345 is OK if there are actual values there to the right of the decimal point but 5.00000 must be shortened to 5.0

Fixed formatting. Thanks for commenting!!

sangin-lee:
@cclauss What should I do to merge this PR?

JiHyunSong:
@cclauss can you follow it?

PR #5865, opened 2021-12-06
Trig functions (Fixes: #2000)

NoahSchiro:
### Describe your change:
I added the 6 trigonometric functions. For sin and cos I used a maclaurin series approximations. For tan, csc, sec, and cot, I used trig identities and my aforementioned sin and cos algorithm. No external library (including the math library that ships with python) was used. I did use ```math``` so that I could use a precise value of pi, but this only served for accurate testing of the algorithms and did not actually factor into the algorithms themselves.

I think the cosine algorithm needs some work, it has a variance around 1E-15 currently but I am unsure of how to improve the accuracy (there are some pythonic limits to how far I can take an infinite series. Namely an infinite series which contains a factorial).

Additionally, I think there should also be the inverse trigonometric functions in this repository but I am not currently aware of an algorithm and it might be better saved for a separate pull request

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


NoahSchiro:
As for the factorial problem: I thought since it is out of the scope of the algorithms that are being focused on here, and only exists as a helper function, doctests was unnecessary.

As for the descriptive parameter name problem: x = input into a function. This is a common practice within mathematics in this context and I believe the x is plenty descriptive within the context of these functions.

wwowowo:
您好，您的邮件已经收到。我会尽快处理，谢谢！

yumengDuang:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #5883, opened 2021-12-25
Added 2 new algorythms (decrypt message, censor)

meg-1:
### Describe your change:

Added 2 new algorythms (decrypt message, censor)

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


IdoErel (changes requested):
The comments are too long.
Keep the comments short and concise.

The code is implementing what's already been implemented in python.

Kind regards,
IdoErel

bayrsaixan:
https://www.easeus.com/backup-software/?utm_source=youtube-heyiran-202107-tb-AsoftClick&utm_medium=referral

meg-1:
> The code is implementing what's already been implemented in python.

Thank you a lot for the feedback! Quick question, you mean both algorithms have been implemented in Python already, right? Or just the censor algorithm?

xcodz-dot (changes requested):
Thank you for trying but it seems that `decrypt` does not do any decryption but only removes the character provided to it and does not do any real world decryption. You might say that it does English deobfuscation but that is also false as well (in case you say so).

here is an example, both of them are equivalent, the latter being already inbuilt.
```python
decrypt("o$ne", "$") == "o$ne".replace("$", "")
```

and for the `censor` function that is also achieved by the same `replace`.

```python
censor("bad word", "bad") == "bad word".replace("bad", "***")
```

Okay, I get it that censor is still reasonable but `decrypt` is unnecessary.

meg-1:
> Thank you for trying but it seems that `decrypt` does not do any decryption but only removes the character provided to it and does not do any real world decryption. You might say that it does English deobfuscation but that is also false as well (in case you say so).
> 
> here is an example, both of them are equivalent, the latter being already inbuilt.
> 
> ```python
> decrypt("o$ne", "$") == "o$ne".replace("$", "")
> ```
> 
> and for the `censor` function that is also achieved by the same `replace`.
> 
> ```python
> censor("bad word", "bad") == "bad word".replace("bad", "***")
> ```
> 
> Okay, I get it that censor is still reasonable but `decrypt` is unnecessary.
 
Thank you so much for the feedback, it's greately appreciated. Would it be okay if I remove the "decrypt" function, and only set a pull request for the censor function? Or does the censor function not hold any value to the repository, and it's not neccessary to commit?

xcodz-dot:
As I am not a maintainer, I can only advise yes but please consider mentioning to any of the maintainers. My advise does not help with the merge decision of any PR and is completely up to maintainer

maths/average_welford.py
```diff
@@ -0,0 +1,21 @@
+def average_welford(values):
```
xcodz-dot, line 1:
```suggestion
def average_welford(values) -> float:
```

xcodz-dot (changes requested):

meg-1:
@Kush1101 , hello! could you please review this pull request??

maths/average_welford.py
```diff
@@ -0,0 +1,21 @@
+def average_welford(values: list) -> float:
+  """
+    >>> average_welford([1, 2, 3, 4, 5])
```
poyea, line 3: Please reformat the code to use 4 spaces as indentation

strings/decrypt.py
```diff
@@ -0,0 +1,38 @@
+def decrypt(txt: str, symbol: str) -> str:
+  """
+  >>> decrypt("o$ne", "$")
+  one
```
poyea, line 4: 4 spaces too.

poyea:
Sometimes the CI is not running - just LMK if that's the case

meg-1:
> Sometimes the CI is not running - just LMK if that's the case

ty, i think that might have been the case. will definitely inform if it happens again!

PR #5917, opened 2022-01-11
Refactor currency_converter.py

MichaelCurrin:
### Describe your change:
 The use of locals in the function is unnecessary. It makes the code harder to follow because you have to check what goes in it, the locals line doesn't save any lines, and the rename of `from_` to `from` has no benefit. 

Also the types can be inferred from constants so no need to say 1.0 is a float.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html). See note at the top.

### DIFF
--- a/web_programming/currency_converter.py
+++ b/web_programming/currency_converter.py
@@ -180,7 +180,7 @@ def convert_currency(
     params = {
         "from": from_,
         "to": to,
-        "amount": amount
+        "amount": amount,
         "api_key": api_key
     }
     res = requests.get(URL_BASE, params=params).json()
--- a/web_programming/currency_converter.py
+++ b/web_programming/currency_converter.py
@@ -12,7 +12,7 @@ TESTING = os.getenv("CI", False)
 API_KEY = os.getenv("AMDOREN_API_KEY", "")
 
 if not API_KEY and not TESTING:
-    raise KeyError("API key must be set as 'AMDOREN_API_KEY' in environment variables")
+    raise KeyError("API key must be provided with 'AMDOREN_API_KEY' environment variable.")
 
 
 # Currency and their description
--- a/web_programming/currency_converter.py
+++ b/web_programming/currency_converter.py
@@ -10,8 +10,9 @@ import requests
 URL_BASE = "https://www.amdoren.com/api/currency.php"
 TESTING = os.getenv("CI", False)
 API_KEY = os.getenv("AMDOREN_API_KEY", "")
+
 if not API_KEY and not TESTING:
-    raise KeyError("Please put your API key in an environment variable.")
+    raise KeyError("API key must be set as AMDOREN_API_KEY")
 
 
 # Currency and their description
@@ -173,12 +174,12 @@ ZMW	Zambian Kwacha
 
 
 def convert_currency(
-    from_: str = "USD", to: str = "INR", amount: float = 1.0, api_key: str = API_KEY
+    from="USD", to="INR", amount=1.0, api_key=API_KEY
 ) -> str:
     """https://www.amdoren.com/currency-api/"""
-    params = locals()
-    params["from"] = params.pop("from_")
+    params = dict(from=from, to=to, amount=amount, api_key=api_key)
     res = requests.get(URL_BASE, params=params).json()
+
     return str(res["amount"]) if res["error"] == 0 else res["error_message"]
 
 
--- a/web_programming/currency_converter.py
+++ b/web_programming/currency_converter.py
@@ -12,7 +12,7 @@ TESTING = os.getenv("CI", False)
 API_KEY = os.getenv("AMDOREN_API_KEY", "")
 
 if not API_KEY and not TESTING:
-    raise KeyError("API key must be set as AMDOREN_API_KEY")
+    raise KeyError("API key must be set as 'AMDOREN_API_KEY' in environment variables")
 
 
 # Currency and their description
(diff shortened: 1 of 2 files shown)


web_programming/currency_converter.py
```diff
@@ -173,12 +174,12 @@
 
 
 def convert_currency(
-    from_: str = "USD", to: str = "INR", amount: float = 1.0, api_key: str = API_KEY
+    from="USD", to="INR", amount=1.0, api_key=API_KEY
```
cclauss, line 177: Why _remove_ type hints?!?

Isn't `from` a reserved word in Python?!?

MichaelCurrin: Oh I misunderstood how Mypy works. And sorry didn't know `from` was reserved. I will fix both.

--- 

I know this works:

```python
x = 123 # inferred as int
x = '456'
# error because of changing type
```

And thought it applied here too

```python
def greet(v=123):
    print(v)


greet(456)

greet("456")
# no error
```
Yet no error is shown :(

Unless I add the explicit type.

```python
def greet(v: int =123):
    print(v)


greet(456)

greet("456")
# error...
```

web_programming/currency_converter.py
```diff
@@ -10,8 +10,9 @@
 URL_BASE = "https://www.amdoren.com/api/currency.php"
 TESTING = os.getenv("CI", False)
 API_KEY = os.getenv("AMDOREN_API_KEY", "")
+
 if not API_KEY and not TESTING:
-    raise KeyError("Please put your API key in an environment variable.")
+    raise KeyError("API key must be set as AMDOREN_API_KEY")
```
MichaelCurrin, line 15:
```suggestion
    raise KeyError("API key must be set as 'AMDOREN_API_KEY' in environment variables")
```

web_programming/currency_converter.py
```diff
@@ -10,8 +10,9 @@
 URL_BASE = "https://www.amdoren.com/api/currency.php"
 TESTING = os.getenv("CI", False)
 API_KEY = os.getenv("AMDOREN_API_KEY", "")
+
 if not API_KEY and not TESTING:
-    raise KeyError("Please put your API key in an environment variable.")
+    raise KeyError("API key must be set as 'AMDOREN_API_KEY' in environment variables")
```
xcodz-dot, line 15: I feel like the message is ambiguous.
```suggestion
    raise KeyError("API key must be provided with 'AMDOREN_API_KEY' environment variable.")
```

web_programming/currency_converter.py
```diff
@@ -176,9 +177,9 @@ def convert_currency(
     from_: str = "USD", to: str = "INR", amount: float = 1.0, api_key: str = API_KEY
 ) -> str:
     """https://www.amdoren.com/currency-api/"""
-    params = locals()
-    params["from"] = params.pop("from_")
+    params = dict(from=from_, to=to, amount=amount, api_key=api_key)
```
xcodz-dot, line 180: This raises a syntax error because `from` is a reserved keyword in Python (?) so its better to use the regular dictionary syntax here.
```suggestion
    params = {
        "from": from_,
        "to": to,
        "amount": amount
        "api_key": api_key
    }
```

xcodz-dot (changes requested):

web_programming/currency_converter.py
MichaelCurrin: Oops, thanks. No problem.

xcodz-dot (approved):

MichaelCurrin:
Tests are failing because of a problem the way the CI is set up. I have the same issue locally on other projects.


---

On a fresh run, there is no cache.

So when Mypy trieds to install types, it doesn't know what to install for

https://github.com/TheAlgorithms/Python/runs/4831403253?check_suite_focus=true

```
Run mypy --ignore-missing-imports --install-types --non-interactive .
error: --install-types failed (no mypy cache directory)
```


---

A flow what does work but is inconvenient is:

```
# generate cache but expect failure, because packages are not installed
mypy --install-types --non-interactive .
# run a 2nd time based on cache from above
mypy --ignore-missing-imports --install-types --non-interactive .
```

What I prefer to is install the types directly in requirements-dev.txt or like this, then run Mypy.

```sh
pip install types-requests
mypy --ignore-missing-imports --non-interactive .
```

This is more manual but less magical (you know exactly what types you install and even the version if you pin it) and lets you run Mypy without the cache or the install flag.

cclauss:
`mkdir --parents --verbose .mypy_cache`

MichaelCurrin:
Thanks that worked locally.

I added a PR for just that

https://github.com/TheAlgorithms/Python/pull/5927

MichaelCurrin:
I am getting an error locally on the type of params being a dict

when it should be `_Params`

Screen Shot 2022-01-17 at 12 55 53 pm

But I can't get this to work as my import it not right. Maybe mean to be from request stubs? I haven't gone this deep into types and don't have capacity at the moment to figure it out

```
    from requests.sessions import _Params

    params = {
        "from": from_,
        "to": to,
        "amount": amount,
        "api_key": api_key
    } # type: _Params
```

muco11m (approved):

cclauss (approved):

MichaelCurrin:
thanks!

PR #6201, opened 2022-06-15
Remove how-to example for support vector machine

epa095:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?
* [x] Delete "how to use a library" example. 

This PR deletes an "algorithm" which purely calls an external library for the computation, and only functions as a how-to. The [contributing guide](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md) states:
>Algorithms in this repo should not be how-to examples for existing Python packages. Instead, they should perform internal calculations or manipulations to convert input values into different output values. Those calculations or manipulations can use data types, classes, or functions of existing Python packages but each algorithm in this repo should add unique value.

The deleted files does nothing of the support-vector machine work, it only shows how to use SVM's from scikit-learn.

### Checklist:


machine_learning/support_vector_machines.py
```diff
@@ -1,58 +0,0 @@
-from sklearn import svm
```
poyea, line 1: Why do we want to remove it?

epa095: One reason is that it is in conflict with the contribution guide, as quoted above. 

Another is that the value of this repo is really **the algorithms**, and that you can read the code and learn how the algorithm works. An example is the [linear regression file](https://github.com/TheAlgorithms/Python/blob/master/machine_learning/linear_regression.py), which actually teaches you how linear regression works. As it is now this repo consists of a mix of actuall algorithm-implementations, which you can read and learn from, and pure usage-examples, which provide much less value. It clutters up the repo, and if we removed them then maybe we could get submissions with the **actuall** algorithms instead.

poyea (approved):

poyea:
There might be other similar files that do similar "how-to-use" thing - so it might be worth to list them in an issue so that we can replace them in the future.

machine_learning/support_vector_machines.py
poyea: After some thoughts, I think we need to discuss this - whether "how-to" examples are acceptable for "machine learning"-type of algorithms. Those algorithms usually consist of building blocks - like SVM, cost functions, neurons or CNN layers - and the use of those blocks might constitute an "algorithm" itself.  However, I would consider for example `add(x, y)` to add two numbers a "how-to" example and removing it should make more sense: there is a `+` operator!

poyea (commented):
We may need to change the guidelines, as we might change the definition of "how-to" examples

machine_learning/support_vector_machines.py
epa095: Since we don't start at the nand-gate, there is always a lower level ;-) E.g. in the linear regression example, you need to know what `np.dot` does, what `transpose` is, etc. I guess deciding exactly the right level is a bit like editing a text book, and you need to have a "feel" for how "far down" you need to explain. 

I don't know if it is possible to agree on some "rule of thumb" for what to keep. But for me I imagine a situation where I let a smart student read the code for some algorithm, and afterwards *explain* to me how that algorithm works. For the linear regression case I can imagine a good student would be able to explain linear regression **the algorithm** after having read the implementation here. For SVM she would not be able to give any better explanation than "SVM's work by importing scikit-learn".

BTW, it is definitely possible to imagine "a machine learning project" as an algorithm here, which uses e.g. SVM, cross validation, train/test/validate etc. But that would then explain "the machine learning process", and should not "pretend" to explain SVM.

poyea: I file an issue #6216 - I'd merge this first because it really doesn't show detailed implementation, but the next steps might be to figure out which files to replace because we might want to keep those contents available

poyea (approved):

PR #6260, opened 2022-07-20
Added leaky rectified linear algorithm

atomicsorcerer:
### Describe your change:
Added the leaky rectified linear algorithm (also known as leaky ReLU). Leaky ReLU is an alternative to normal ReLU because it solves the dying ReLU problem, which is an issue in some neural networks. 

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/maths/leaky_relu.py
+++ b/maths/leaky_relu.py
@@ -12,7 +12,9 @@ https://en.wikipedia.org/wiki/Rectifier_(neural_networks)
 from __future__ import annotations
 
 
-def leaky_relu(vector: float | list[float], slope: float = 0.01) -> float | list[float]:
+def leaky_relu(
+    vector: float | list[float], negative_slope: float = 0.01
+) -> float | list[float]:
     """
     Implements the leaky rectified linear activation function
 
@@ -35,17 +37,17 @@ def leaky_relu(vector: float | list[float], slope: float = 0.01) -> float | list
         raise ValueError(
             "leaky_relu() only accepts floats or a list of floats for vector"
         )
-    if not isinstance(slope, float):
-        raise ValueError("leaky_relu() only accepts a float value for slope")
+    if not isinstance(negative_slope, float):
+        raise ValueError("leaky_relu() only accepts a float value for negative_slope")
 
     if isinstance(vector, float):
         if vector < 0:
-            return vector * slope
+            return vector * negative_slope
         return vector
 
     for index, value in enumerate(vector):
         if value < 0:
-            vector[index] = value * slope
+            vector[index] = value * negative_slope
 
     return vector
 
--- a/maths/leaky_relu.py
+++ b/maths/leaky_relu.py
@@ -1,0 +1,52 @@
+"""
+This algorithm implements the leaky rectified linear (LReLU).
+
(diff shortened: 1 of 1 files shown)


maths/leaky_relu.py
```diff
@@ -0,0 +1,58 @@
...
+
+LReLU is at times used as a substitute to ReLU because it fixes the dying ReLU problem.
+This is done by adding a slight slope to the negative portion of the function.
+The default value for the slope is 0.01.
+The new slope is determined before the network is trained.
+
+Script inspired from its corresponding Wikipedia article
+https://en.wikipedia.org/wiki/Rectifier_(neural_networks)
+"""
+from __future__ import annotations
```
tianyizheng02, line 12: Is the `__future__` import needed here?

maths/leaky_relu.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def leaky_relu(
+    vector: float | list[float], negative_slope: float = 0.01
+) -> float | list[float]:
```
tianyizheng02, line 17:
```suggestion
def leaky_relu(vector: np.ndarray, negative_slope: float = 0.01) -> np.ndarray:
```
I think just type hinting it as `np.ndarray` is fine for this function. Using `numpy` arrays is pretty standard when it comes to NN-related Python programming, and `numpy` functions that take in arrays generally also support scalars as well ("`array_like`")

maths/leaky_relu.py
```diff
@@ -0,0 +1,58 @@
...
+    -0.03
+    >>> leaky_relu(2)
+    Traceback (most recent call last):
+      ...
+    ValueError: leaky_relu() only accepts floats or a list of floats for vector
+    """
+    if isinstance(vector, int):
+        raise ValueError(
+            "leaky_relu() only accepts floats or a list of floats for vector"
+        )
```
tianyizheng02, line 39: Why do we not want to support ints as input? They can all be cast to floats as output

maths/leaky_relu.py
```diff
@@ -0,0 +1,58 @@
...
+    if not isinstance(negative_slope, float):
+        raise ValueError("leaky_relu() only accepts a float value for negative_slope")
```
tianyizheng02, line 41: I think the constraints on the possible range for the negative slope should be clearer. Are we restricting it to a float between 0 and 1? If so, that should be the if-condition instead

maths/leaky_relu.py
```diff
@@ -0,0 +1,58 @@
...
+    if isinstance(vector, float):
+        if vector < 0:
+            return vector * negative_slope
+        return vector
+
+    for index, value in enumerate(vector):
+        if value < 0:
+            vector[index] = value * negative_slope
+
+    return vector
```
tianyizheng02, line 52:
```suggestion
    return np.maximum(vector, negative_slope * vector)
```
`numpy` functions can handle these cases very easily. Also, leaky ReLU is equivalent to $f(x) = \max(x, ax)$ for negative slopes $0 \leq a \leq 1$ (see the Wikipedia article you cited for more info)

tianyizheng02 (changes requested):

tianyizheng02:
Closing in favor of #8962

PR #6298, opened 2022-08-06
Add Max Fenwick Tree

itsamirhn:
### Describe your change:
I created MaxFenwickTreeOneBasedIndexing wich solves this problem:
- Update A[i] = value
- Query Maximum of Range (l, r]

This algorithm use classic fenwick tree with some changes in key ideas.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


data_structures/binary_tree/maximum_fenwick_tree.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    def __init__(self, size: int) -> None:  # Create Fenwick tree with specified size
+        self.size = size
+        self.arr = [0] * (size + 1)
+        self.tree = [0] * (size + 1)
+
+    def update(
+        self, index: int, value: int
+    ) -> None:  # Set index (1-Based) to value in O(lg^2 N)
```
cclauss, line 26: Do not wrap a line just for a comment.  Instead, put the comment on a separate line or in a docstring.

data_structures/binary_tree/maximum_fenwick_tree.py
```diff
@@ -0,0 +1,50 @@
...
+        self.arr[index] = value
+        while index < self.size:
+            self.tree[index] = max(value, self.query(index, index - index & -index))
+            index += index & (-index)
+
+    def query(
+        self, left: int, right: int
+    ) -> int:  # Query maximum value from range (left, right] (1-Based) in O(lg N)
```
cclauss, line 34: How about better variable names instead of comments to explain what the variable names should explain?

data_structures/binary_tree/maximum_fenwick_tree.py
```diff
@@ -0,0 +1,50 @@
...
+        res = 0
+        while left < right:
+            x = right - right & -right
```
cclauss, line 37: What is `x` and we are we forcing the reader to guess what `x` is?

itsamirhn:
@cclauss Let me now if these changes solves the issues that you commented.

cclauss:
Why is one-based indexing and advantage in a language that uses zero-based indexing?

itsamirhn:
The one-based indexing is default for any Fenwick Tree. In all languages we use one-based indexing because of the main theory idea behind Fenwick tree that is `i & (-i)`  will not work for Zero and in every method we have to check for this case.

Even in the previous implementation of Fenwick Tree in this repository ([here](https://github.com/TheAlgorithms/Python/blob/252df0a149502143a14e7283424d40b785dd451c/data_structures/binary_tree/fenwick_tree.py)), we had to used one-based by default.

But there is another tricky way to solve the one-based indexing issue as [Wikipedia](https://en.wikipedia.org/wiki/Fenwick_tree) has used too.

I can add the zero-based indexing too if you want.

itsamirhn:
@cclauss, I changed the tree into 0-based indexing.

cclauss (approved):
Thanks for doing this!

PR #6321, opened 2022-08-12
Add: Two Regex match algorithm (Recursive & DP)

itsamirhn:
### Describe your change:

I added two solution for simple regex match algorithm.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


itsamirhn:
Would anyone review this PR?

dynamic_programming/regex_match.py
```diff
@@ -0,0 +1,106 @@
...
+    >>> recursive_match('abc', 'a.c*d')
+    False
+    >>> recursive_match('aa', '.*')
+    True
+    """
+    if not text and not pattern:
+        return True
+
+    if text and not pattern:
+        return False
```
tianyizheng02, line 37:
```suggestion
    if not pattern:
        return not text
```

dynamic_programming/regex_match.py
```diff
@@ -0,0 +1,106 @@
...
+
+    if not text and pattern and pattern[-1] != "*":
+        return False
+
+    if not text and pattern and pattern[-1] == "*":
+        return recursive_match(text, pattern[:-2])
```
tianyizheng02, line 43:
```suggestion
    if not text:
        return pattern[-1] == "*" and recursive_match(text, pattern[:-2])
```

1) No need to check if `pattern` is non-empty because the previous if-statement already handles it

2) If `pattern[-1] != "*"`, then the above code short-circuits and immediately returns False. Otherwise, it'll proceed to call `recursive_match` as usual.

tianyizheng02 (changes requested):

dynamic_programming/regex_match.py
```diff
@@ -0,0 +1,100 @@
...
+    >>> dp_match('aa', '.*')
+    True
+    """
+    m = len(text)
+    n = len(pattern)
+    dp = [[False for _ in range(n + 1)] for _ in range(m + 1)]
+    dp[0][0] = True
+
+    for i in range(1, m + 1):
+        dp[i][0] = False
```
tianyizheng02, line 78:
```suggestion
```
Although the Medium article includes this for-loop, I think it's unnecessary for this Python implementation because the array `dp` is already initialized with all `False`

tianyizheng02 (changes requested):
One last minor change, but other than that LGTM

dynamic_programming/regex_match.py
itsamirhn: You're right.

tianyizheng02 (approved):

PR #6535, opened 2022-10-02
Removed a security bug on Recaptcha algorithm

thegeekyb0y:
### Describe your change:
There was a security bug which can breach the data as per OWASP top 10 and SANS 25, so I used os module to put the sensitive data (secret key,client key) in environment variable. Either we can do is put the sensitive data in another file or we can do the same thing I did, it depends on the developer.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,8 @@
+version = 1
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
--- a/web_programming/recaptcha_verification.py
+++ b/web_programming/recaptcha_verification.py
@@ -32,6 +32,7 @@ Below a Django function for the views.py file contains a login form for demonstr
 recaptcha verification.
 """
 import requests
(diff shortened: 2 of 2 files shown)


.deepsource.toml
```diff
@@ -0,0 +1,8 @@
+version = 1
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
```
cclauss, line 8: ⛔ means that GitHub expects text files to end in one, and only one carriage return (\n).

.deepsource.toml
```diff
@@ -0,0 +1,8 @@
+version = 1
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
```
cclauss, line 7: I don't think toml files usually have whitespace in front of tags.

cclauss: Let's get rid of deepsource.

PR #6545, opened 2022-10-02
Newton raphson complex

Saksham1970:
### Describe your change:

I noticed the Newton Raphson method could be improved, so I made an altered script for it.

There are currently 2 scripts for newton raphson method in `Python/arithmetic_analysis/ `:
`newton_method.py`
`newton_raphson.py`

There are a **few issues** with them,

-  They both don't support _complex numbers_, while the method itself in theory does.
-  `newton_method.py`: requires user to pass both the mathematical function and its derivative into the function
-  `newton_raphson.py`: uses the wrong precision check; it compares the value of the f(x) with the precision required, rather than the difference in the consecutive x's. This resulted in its values to be wrong as in their test cases they showed pi = `3.1415926536808043` while correct precision check would give a value of `3.141592653589793` which is way closer to actual value of pi.
-  `newton_raphson.py`: also extensively uses `eval()` statements, which are necessary for finding derivatives, but the implementation limits them from only taking the _x as variable_, and it was also used in every loop making the _computation slower_.
-  `newton_raphson.py`: it doesn't check for the division by 0, which is the _biggest_ limitation to using newton method, since the root can't be found for those cases.

I have taken goods from both the files and added more improvements and created a third file for the time being.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


arithmetic_analysis/newton_raphson_new.py
```diff
@@ -0,0 +1,84 @@
...
+    # Find value of pi
+    print(f"The root of sin(x) = 0 is {newton_raphson('sin(x)', 2)}")
+
+    # Find root of polynomial
+    # Find fourth Root of 5
+    print(f"The root of x**4 - 5 = 0 is {newton_raphson('x**4 -5', 0.4 +5j)}")
+
+    # Find value of e
+    print(
+        f"The root of log(y) - 1 = 0 is \
```
cclauss, line 73: PEP8 suggests that we avoid backslashes in Python code because whitespace to the right of the backslash breaks the script that is invisible to the reader.

Saksham1970: I see, what's the correct way of reducing the length of the line to pass the flask8 test?

arithmetic_analysis/newton_raphson_new.py
```diff
@@ -0,0 +1,84 @@
...
+{newton_raphson('log(y) - 1', 2, variable='y')}"
+    )
```
cclauss, line 75:
```suggestion
    print(
        "The root of log(y) - 1 = 0 is "
        f"{newton_raphson('log(y) - 1', 2, variable='y')}"
    )
```
... or ...
```
    e = newton_raphson('log(y) - 1', 2, variable='y')
    print(f"The root of log(y) - 1 = 0 is {e}")
```

Saksham1970: Thank You.

cclauss (approved):

PR #6575, opened 2022-10-03
Added search quadruplet algorithm

Caesarsage:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/searches/search_quadruplet.py
+++ b/searches/search_quadruplet.py
@@ -1,11 +1,11 @@
 def search_quadruplets(sequence: list[int], target: int) -> list[int]:
     """
-    >> > search_quadruplets([4, 1, 2, -1, 1, -3], 1)
+    >>> search_quadruplets([4, 1, 2, -1, 1, -3], 1)
 
     [-3, -1, 1, 4], [-3, 1, 1, 2]
     Explanation: Both the quadruplets add up to the target.
 
-    >> > search_quadruplets([2, 0, -1, 1, -2, 2], 2)
+    >>> search_quadruplets([2, 0, -1, 1, -2, 2], 2)
     [-2, 0, 2, 2], [-1, 0, 1, 2]
     Explanation: Both the quadruplets add up to the target.
 
--- a/searches/search_quadruplet.py
+++ b/searches/search_quadruplet.py
@@ -1,0 +1,64 @@
+
+
+
+def search_quadruplets(sequence, target):
+    """
+    >> > search_quadruplets([4, 1, 2, -1, 1, -3], 1)
+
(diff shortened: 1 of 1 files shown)


cclauss (approved):

searches/search_quadruplet.py
```diff
@@ -0,0 +1,72 @@
+def search_quadruplets(sequence: list[int], target: int) -> list[int]:
```
tianyizheng02, line 1:
```suggestion
def search_quadruplets(sequence: list[int], target: int) -> list[list[int]]:
```
Return type is incorrect

searches/search_quadruplet.py
```diff
@@ -0,0 +1,72 @@
...
+
+    >>> search_quadruplets([2, 0, -1, 1, -2, 2], 2)
+    [-2, 0, 2, 2], [-1, 0, 1, 2]
+    Explanation: Both the quadruplets add up to the target.
+
+    Technique: Two Pointers
+    This problem follows the Two Pointers pattern and
+    shares similarities with Triplet Sum to Zero.
+
+    """
```
tianyizheng02, line 16:
```suggestion
    """
    [Explain the problem here]
    Technique: Two Pointers
    This problem follows the Two Pointers pattern and
    shares similarities with Triplet Sum to Zero.

    >>> search_quadruplets([4, 1, 2, -1, 1, -3], 1)
    [-3, -1, 1, 4], [-3, 1, 1, 2]
    >>> search_quadruplets([2, 0, -1, 1, -2, 2], 2)
    [-2, 0, 2, 2], [-1, 0, 1, 2]
    """
```
Please reformat the docstring and explain the problem.

searches/search_quadruplet.py
tianyizheng02, line 1: What if there are duplicate quadruplets? Ex: `[0, 0, 0, 0, 0]` and `0` as input

tianyizheng02: Note that I ran this file locally and this file breaks on the above example: it returns the empty list, which is obviously  incorrect.

tianyizheng02 (commented):
This solution is incorrect because it breaks when there are repeated numbers in the list. For example,
```py
>>> search_quadruplets([0, 0, 0, 0], 0)
>>> search_quadruplets([0, 0, 0, 0, 0], 0)
>>> search_quadruplets([0, 0, 0, 1], 1)
```
all return an empty list.

tianyizheng02:
Thanks for your contribution, but I'll have to close this PR because your algorithm produces incorrect outputs. We're currently in the process of clearing out our backlog of open PRs, so we unfortunately don't have the time to significantly fix PRs that aren't already correct.

PR #6590, opened 2022-10-03
Bytes to Int

palaniappan-r:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


conversions/bytes_to_int.py
```diff
@@ -0,0 +1,57 @@
...
+>>> bytes_to_int(b'\\x00\\x01',False)
+1
+>>> bytes_to_int('abc',False)
+ERROR - 'str' object cannot be interpreted as an integer
+0
+"""
+
+
+def bytes_to_int(bytes_var: bytearray, signed: bool) -> (int):
+    try:
```
cclauss, line 19: Please move the comments and tests into the function so that if someone copies the function, the documentation and tests come along with it.

The double backslashes seem unnecessary.  See https://docs.python.org/3/library/stdtypes.html#int.from_bytes Both the input and output of this function should match `int.from_bytes(_, "little")` or `int.from_bytes(_, "big")`

palaniappan-r: I've added a byte order parameter. 

I think the double backslashes are necessary as the tests are being typed as part of a string and thus, the single backslash needs an escape sequence, which is a double backlash. Feel free to provide a better solution and correct me if I'm wrong.

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,62 @@
...
+        rslt = 0
+
+        if signed is True:
+            temp = ""
+            for j in binval:
+                flag = j == "0"
+                if flag is True:
+                    temp = str(temp) + "1"
+                else:
+                    temp = str(temp) + "0"
```
cclauss, line 42:
```suggestion
                temp += str(int(flag))
```

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,59 @@
+def bytes_to_int(bytes_var: bytearray, signed: bool, byteorder: str) -> (int):
```
cclauss, line 1: Let's follow the inputs, outputs, and exceptions of `int.from_bytes()` exactly...
```suggestion
def bytes_to_int(bytes: bytes, byteorder: str, *, signed: bool=False) -> int:
```

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,59 @@
...
+
+    """
```
cclauss, line 3:
```suggestion
    r"""
```
This leading `r` will make the docstring a `raw` string and then we should be able to shift to single backslashes.

Then we can test with something like...
```
    >>> tests = ((b"\x00\x10", "big", False), (b'\xfc\x00', "big", True),
    ...          (b'\x00\x01', "big", False), (b'\x00\x10', 'little', False))
    >>> all(bytes_to_int(bytes, byteorder, signed=signed) == int.from_bytes(bytes, byteorder, signed=signed)
    ...     for bytes, byteorder, signed in tests)
    True
    >>> bytes_to_int('abc',False,'big')
        ...
    TypeError: cannot convert 'str' object to bytes
```

cclauss: Extra credit: Exceptions for...
```
>>> bytes_to_int(b'\xfc\x00', 'medium')
>>> bytes_to_int(7, 'big',)
>>> bytes_to_int(7.1, 'little', signed=True)
```

cclauss (changes requested):

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,64 @@
+def bytes_to_int(
+    bytes_var: bytearray,
+    byteorder: str,
+    signed: bool,
+) -> (int):
```
cclauss, line 5: Why the parens around `(int)`?
```
>>> int == (int)
True
>>> int is (int)
True
```

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,64 @@
+def bytes_to_int(
+    bytes_var: bytearray,
```
cclauss, line 2: >>> type(b"\x00\x10")
<class 'bytes'>
```suggestion
    bytes_var: bytes,
```

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,64 @@
...
+    >>> bytes_to_int('abc',False,'big')
+    AttributeError - 'bool' object has no attribute 'lower'
+    0
+    >>> bytes_to_int(7.1, 'little', signed=True)
+    TypeError - 'float' object is not subscriptable
+    0
+
+    """
+
+    try:
```
cclauss, line 26: `try` blocks should be small to make code easier to debug.

Allow your code to raise the exceptions.  If raising exceptions on bad input data is good enough for the Python Standard Library, then it is good enough for `bytes_to_int()`.

conversions/bytes_to_int.py
palaniappan-r: Edited now

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,64 @@
...
+    True
+    >>> bytes_to_int('abc',False,'big')
+    AttributeError - 'bool' object has no attribute 'lower'
+    0
+    >>> bytes_to_int(7.1, 'little', signed=True)
+    TypeError - 'float' object is not subscriptable
+    0
+
+    """
+
```
cclauss, line 25:
```suggestion
    if byteorder not in ("little", "big"):
        raise ValueError(byteorder must be either 'little' or 'big')
```

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,65 @@
...
+    >>> bytes_to_int('abc',False,'big')
+    AttributeError - 'bool' object has no attribute 'lower'
+    0
+    >>> bytes_to_int(7.1, 'little', signed=True)
+    TypeError - 'float' object is not subscriptable
+    0
+
+    """
+    if byteorder not in ("little", "big"):
+        raise ValueError(byteorder must be either 'little' or 'big')
```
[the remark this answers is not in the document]
palaniappan-r: Now a ValueError exception will be raised, which will again be caught by the Except block and then an error message will be displayed.

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,67 @@
...
+    TypeError - 'float' object is not subscriptable
+    0
+
+    """
+
+    try:
+        if byteorder not in ("little", "big"):
+            raise ValueError(" byteorder must be either 'little' or 'big' ")
+
+        byteorder = byteorder.lower()
```
cclauss, line 30: Fix first, test second.
```suggestion
        byteorder = byteorder.lower()
        if byteorder not in ("little", "big"):
            raise ValueError(" byteorder must be either 'little' or 'big' ")
```

palaniappan-r: Yes, my bad

conversions/bytes_to_int.py
```diff
@@ -0,0 +1,67 @@
...
+    This script first converts a bytearray,
+    to its corresponding signed/unsigned binary number,
+    and then converts the binary number to an integer.
+
+    >>> tests = ((b"\x00\x10", "big", False), (b'\xfc\x00', "big", True),
+    ...          (b'\x00\x01', "big", False), (b'\x00\x10', 'little', False))
+    >>> all(bytes_to_int(a,b,c) == int.from_bytes(a,b,signed=c)
+    ...     for a,b,c in tests)
+    True
+    >>> bytes_to_int('abc',False,'big')
```
cclauss, line 17:
```suggestion
    >>> bytes_to_init(b"", "medium")
    Traceback (most recent call last):
        ...
    ValueError: byteorder must be either 'little' or 'big'
    >>> bytes_to_int('abc',False,'big')
```

palaniappan-r:
Hi, I committed the change, can you please add a hacktoberfest label to the
commit?

PR #6618, opened 2022-10-03
Created new file PrefixToInfix.py

itzyash01:
Python code to convert prefix to infix

Algorithm:
Read the Prefix expression from right to left
If operand, then push it onto the Stack
If operator, then pop two operands from the Stack 
Concatenate the two operands and the operator between them, and create a string 
string = (operand1 + operator + operand2) 
Push the resultant string back to Stack
Repeat the above steps until the end of Prefix expression.
At the end stack will have only resultant string

### DIFF
--- a/data_structures/stacks/PrefixToInfix.py
+++ b/data_structures/stacks/PrefixToInfix.py
@@ -1,24 +1,26 @@
 def prefixToInfix(prefix):
-	stack = []
-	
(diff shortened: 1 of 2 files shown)


itzyash01 (commented):
I have made required changes.

cclauss:
Please read CONTRIBUTING.md before opening another pull request.

PR #6625, opened 2022-10-03
Adding inductive reactance calculation

RobisLV:
Adding new file ind_reactance.py for calculating inductive reactance, frequency or inductance.


cclauss (approved):

cclauss (approved):
Used [`math.pi`](https://docs.python.org/3/library/math.html#math.pi)

PR #6652, opened 2022-10-04
Feature best match finder

gitwikc:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### Best Match Finder Algorithm

Implemented the best match finder algorithm as discussed in the MITOCW [video lecture](https://youtu.be/5RSMLgy06Ew?list=PL6MpDZWD2gTF3mz26HSufmsIO-COKKb5j&t=1623) for the course 6.042 by Prof. Tom Leighton. The link above leads to the timestamp where he starts explaining how the algorithm works, using an example.

The problem is as follows:
- We are given two groups with same number of elements
- Each element has a *preference order*, ranking the elements in the descending order in which the element wants to pair with them, *i.e. 1st element is most preferred and last element is least preferred*
- We are given the task to find out such a matching of the elements such that there are no rogue couples
- *Rogue couple* is a set of two elements from different groups that have a higher preference for each other than the element they have currently been paired up with.
> **EXAMPLE**
> Let `g`, `f` be two sets whose elements are to be matched
> Let `(g1, f1)` and `(g2, f2)` be matched but 
> - `g2` has higher preference for `f1` than `f2` 
> - `f1` has higher preference for `g2` than `g1`
> In such a case, `g2` and `f1` are a **rogue couple**

The best matching is achieved when there are no rogue couples. Also for any two groups with the specifications as given above, there exists a best match. **Proof in video lecture**

### DIFF
--- a/other/best_match_finder.py
+++ b/other/best_match_finder.py
@@ -1,6 +1,89 @@
+from typing import List, Dict, Tuple
+
+
+def best_match_found(group_possible_matches: Dict[object, List[object]]) -> bool:
+    """
+    Best match corresponds to the case where each element in group2
+    has exactly ONE possible match
+
+    >>> best_match_found({0: ['A'], 1: ['B'], 2: ['C']})
+    >>> True
+    """
+    return all(
+        [
+            len(possible_matches) == 1
+            for possible_matches in group_possible_matches.values()
+        ]
+    )
+
+
+def perform_matching(
+    group1: Dict[object, List[object]],
+    group2: Dict[object, List[object]],
+    group1_checklist: Dict[object, List[object]],
+    group2_possible_matches: Dict[object, List[object]]
+) -> List[Tuple[object]]:
+    """
+    Performs one round of matching elements in group 1 to group 2
+    1. All group 1 elements try to match with their remaining best
+    top preference in group 2
+    2. group 2 elements then keep only their top preference out of the
+    possible matches from group 1 (from step 1) and reject the remaining
+    elements from group 1
+    3. The rejected elements cross out the group 2 element that rejected them,
+    from their checklist
+    4. We go back to step 1 and continue till each element in group 2 has
+    exactly ONE matching element in group 1
+
+    Then we return the match as a list of tuples in the form:
+    [(group2_elemt, group1_elemt), ...]
+    """
+    # Each group1 element tries to match with their available top match in group2
+    for group1_element, checklist in group1_checklist.items():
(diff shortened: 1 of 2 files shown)


other/best_match_finder.py
```diff
@@ -0,0 +1,129 @@
...
+    :param group2: Preferences of elements of group2 as a dict of the form
+    {<group2_elemt>: [...<List of group1_elemts in descending order of preference>]}
+    """
+    group1_checklist: dict[object, list[object]] = dict(
+        zip(group1.keys(), list(group1.values()).copy())
+    )
+    group2_possible_matches: dict[object, list[object]] = dict(
+        zip(group2.keys(), [[]] * 5)
+    )
+
```
CaedenPH, line 28: You should typehint the return type of this function

CaedenPH (changes requested):
You should typehint the function returns and potentially move the functions outside of the function and pass arguments. May make code more understandable

other/best_match_finder.py
gitwikc: Could you help me on how to typehint this one? This one is actually an assignment.

gitwikc:
Added type hints, also created the test file and took out the nested functions

other/best_match_finder.py
gitwikc: Yes, I added the typehints for all functions now and took the nested functions out of the `get_best_match` function for more readability

gitwikc:
I have written the code by import from the `typing` module in python

```
from typing import List, Dict, Tuple
```

But this bot keeps changing all the occurences of `List, Dict, Tuple` to `list, dict, tuple` respectively resulting in an error for unused imports. And using `list, dict, tuple` also gives an error for `types cannot be indexed`. What should I do now?

rohan472000 (changes requested):
There are a few issues with the code:

1. The imports `typing.Dict`, `typing.List`, and `typing.Tuple` are unused and can be removed.
2. In `test_best_match_finder.py`, the line `if len(best_match) == True:` should be changed to `if len(best_match):` to follow PEP 8 guidelines.
3. In `test_best_match_finder.py`, `Dict` and `List` are undefined names. You should import them from `typing` or remove them altogether, depending on whether or not they are being used.
4. In `best_match_finder.py`, `zip` is being passed to the `list` function, which is not compatible. You can change it to `list(matches)` instead.

tianyizheng02:
Closing due to lack of response to code review

PR #6657, opened 2022-10-04
Create rainwatertrapping.py

YashGupta1510:
### Describe your change:
Added a solution to the Rain Water Trapping problem using the two pointer approach as I couldn't find this solution or any other algorithm with the two pointer approach

### DIFF
--- a/other/rainwatertrapping.py
+++ b/other/rainwatertrapping.py
@@ -1,31 +1,30 @@
 def maxTrappedWater(arr, n):
-	size = n - 1
-	prev = arr[0]
-	prev_index = 0
-	water = 0
-	temp = 0
-	for i in range(1, size + 1):
-		if (arr[i] >= prev):
-			prev = arr[i]
-			prev_index = i
-			temp = 0
-		else:
-			water += prev - arr[i]
-			temp += prev - arr[i]
-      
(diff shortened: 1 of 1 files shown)


other/rainwatertrapping.py
```diff
@@ -0,0 +1,30 @@
...
+            if arr[i] >= prev:
+                prev = arr[i]
+            else:
+                water += prev - arr[i]
+    return water
+
+
+if __name__ == "__main__":
+    arr = [0, 6, 0, 4, 1, 3, 2, 6, 1, 2, 1, 3]
+    N = len(arr)
```
CaedenPH, line 29:
```suggestion
```

other/rainwatertrapping.py
```diff
@@ -0,0 +1,30 @@
+def maxTrappedWater(arr, n):
```
CaedenPH, line 1:
```suggestion
def maxTrappedWater(arr):
    n = len(arr)
```
You should get the length inside, not pass it as a parameter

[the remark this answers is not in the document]
YashGupta1510: integer

cclauss:
Please read CONTRIBUTING.md before submitting another pull request.

PR #6698, opened 2022-10-05
Added wordbreakfile

samyak1999:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
+def wordbreak(word: str, word_dict: list) -> bool:
```
CaedenPH, line 1: Can this be renamed to `word_break` please

dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
+def wordbreak(word: str, word_dict: list) -> bool:
+    """
+    word ---> Input word
+    word_dict ---> Input dictionary
+    >>> wordbreak('deepImpact',['deep','Im','pact'])
+    True
+    >>> wordbreak('mangoinmango',['mango','in'])
```
CaedenPH, line 7:
```suggestion
    >>> wordbreak("mangoinmango", ["mango", "in"])
```

dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
+def wordbreak(word: str, word_dict: list) -> bool:
+    """
+    word ---> Input word
+    word_dict ---> Input dictionary
+    >>> wordbreak('deepImpact',['deep','Im','pact'])
```
CaedenPH, line 5:
```suggestion
    >>> wordbreak("deepImpact", ["deep", "Im", "pact"])

```

dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
...
+    # overall solution to the complete problem
+    for i in range(len(word)):
+        for j in start:
+            if word[j : i + 1] in dict_set:
+                start.append(i + 1)  # start of next word in dict
+                break  # - note [1]
+    return start[-1] == len(word)
+
+
+print(wordbreak("deepImpact", ["deep", "Im", "pact"]))
```
CaedenPH, line 25: Can this be moved into an `if __name__ == "__main__"` check please

dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    dict_set = set(word_dict)
+    start = [0]
+    # start[0] is the equivalent of a base-case
+    # from the recursive solution and start[-1] is the
+    # overall solution to the complete problem
+    for i in range(len(word)):
+        for j in start:
+            if word[j : i + 1] in dict_set:
+                start.append(i + 1)  # start of next word in dict
```
CaedenPH, line 20: I don't like the idea of appending to the list that is being currently iterated through. Should we copy the start list when iterating over it?

dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
+def wordbreak(word: str, word_dict: list) -> bool:
+    """
+    word ---> Input word
+    word_dict ---> Input dictionary
```
CaedenPH, line 4:
```suggestion
```
Doesn't provide the reader with any information they don't already know

CaedenPH (changes requested):
@samyak1999 
Please resolve my changes accordingly and also link to some documentation describing what this algorithm does. It's a bit unclear (to me at least) precisely what this does

tianyizheng02:
> Please resolve my changes accordingly and also link to some documentation describing what this algorithm does. It's a bit unclear (to me at least) precisely what this does

This file just implements a solution to [this LeetCode problem](https://leetcode.com/problems/word-break/), so that problem should be linked as a reference.

> This algorithm is also already implemented here (https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/word_break.py) - Not sure what to do about that.

I think it should be OK to allow this PR's implementation as well because it uses a different DP method. The existing implementation uses a top-down DP approach, while this one uses a bottom-up approach. We should have both implementations in the same file for ease of comparison.

dynamic_programming/word_break.py
tianyizheng02, line 1: Please add a link to the LeetCode page where this problem comes from: https://leetcode.com/problems/word-break/

Also, please add a description of the problem and the algorithm.

dynamic_programming/word_break.py
tianyizheng02, line 1:
```suggestion
def wordbreak(word: str, word_dict: set[str]) -> bool:
```
There's no need to make `word_dict` a list and then immediately convert it to a set. Just because LeetCode set up the problem with `word_dict` as a list doesn't mean _we_ have to.

dynamic_programming/word_break.py
```diff
@@ -0,0 +1,25 @@
...
+    dict_set = set(word_dict)
+    start = [0]
+    # start[0] is the equivalent of a base-case
+    # from the recursive solution and start[-1] is the
+    # overall solution to the complete problem
+    for i in range(len(word)):
+        for j in start:
+            if word[j : i + 1] in dict_set:
+                start.append(i + 1)  # start of next word in dict
+                break  # - note [1]
```
tianyizheng02, line 21: What does this comment mean?

dynamic_programming/word_break.py
tianyizheng02, line 20:
```suggestion
    for i in range(1, len(word) + 1):
        for j in start:
            if word[j : i] in dict_set:
                start.append(i)  # start of next word in dict
```
No need to add 1 to the index every time, just shift the range of the for-loop.

tianyizheng02: I don't think appending to `start` while iterating over it should pose any issues because we immediately break out of that for-loop when we append. This bottom-up DP approach normally involves instantiating the whole array at the beginning, so appending to the list like this usually never happens. I'm guessing that @samyak1999 chose to implement it using a dynamic array in order to minimize the number of indices that need to be iterated over.

tianyizheng02 (changes requested):

tianyizheng02:
Closing due to no response to code reviews. If we want someone to implement this bottom-up algorithm, we can wait for a new PR to do so instead of waiting for a code fix that will never come.

PR #6736, opened 2022-10-06
AdemolaAri: An Algorithm to sort a list with multiple datatypes in-place #6734

AdemolaAri:
An Algorithm to sort a list with multiple datatypes in-place placing numbers on the left side and strings on the right.

### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/binary_tree/binary_search_tree.py
+++ b/data_structures/binary_tree/binary_search_tree.py
@@ -155,6 +155,46 @@ class BinarySearchTree:
         self.inorder(arr, node)  # append all values to list using inorder traversal
         return arr[k - 1]
 
+    def __get_leftmost_descendant(self, node: Node):
+        """
+        Keeps traversing left until it finds the leftmost node
+        Returns the value of the that node
+        """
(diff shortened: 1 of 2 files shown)


rohan472000 (commented):
One improvement that could be made to this implementation of a binary search tree is to add a size attribute to the BinarySearchTree class that keeps track of the number of nodes in the tree. This would allow for constant-time access to the size of the tree, which could be useful in certain applications.

Another potential improvement would be to modify the search method to return None instead of raising an exception if the tree is empty or the value is not found. This would make the method more consistent with other methods in the class that return None in similar situations.

One issue with the implementation is that the remove method does not handle the case where the node to be removed has two children correctly. The current implementation finds the maximum value of the left subtree and replaces the value of the node to be removed with that value, but this can cause problems if the node being removed has other nodes that depend on its position in the tree. A better approach would be to replace the node with its successor (i.e., the smallest node in the right subtree), which preserves the order of the remaining nodes in the tree.

PR #6759, opened 2022-10-06
Add find shortest word in a sentence

happiestbee:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x]  This pull request is all my own work -- I have not plagiarized.

### DIFF
--- a/strings/shortest_word.py
+++ b/strings/shortest_word.py
@@ -18,9 +18,7 @@ def shortest_word(sentence: str) -> str:
     # Split the sentence into a list of words
     # Sort by length and take the first element
     word_list = sentence.split()
-    for word in word_list:
-        if len(word) == min(len(word) for word in word_list):
-            return word
+    return min(word_list, key=len)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


strings/shortest_word.py
```diff
@@ -0,0 +1,28 @@
...
+    >>> shortest_word("Black Stack")
+    'Black'
+    >>> shortest_word("funny bunny rabbit")
+    'funny'
+    """
+
+    # Split the sentence into a list of words
+    # Sort by length and take the first element
+    word_list = sentence.split()
+    word_list.sort(key=len)
```
Cjkjvfnby, line 21: Sorting requires N*logN complexity,  you could get the same result just in N by iterating over the sentence and finding the min length.   min(word_list , key=len) can do the trick.

Cjkjvfnby (changes requested):

strings/shortest_word.py
```diff
@@ -0,0 +1,29 @@
...
+    >>> shortest_word("funny bunny rabbit")
+    'funny'
+    """
+
+    # Split the sentence into a list of words
+    # Sort by length and take the first element
+    word_list = sentence.split()
+    for word in word_list:
+        if len(word) == min(len(word) for word in word_list):
+            return word
```
Cjkjvfnby, line 23: `for word` is O(n), `min` is O(n) and it's nested, so we get O(n^2) complexity in total. 


You could use min with `key` argument to do the job in O(n).

```suggestion
    return min(word_list, key=len)
```

PR #6794, opened 2022-10-07
Added Coulomb's law program

Ujjawal179:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


CaedenPH (changes requested):
Personally, I think instead of passing the value `0`, every parameter should be optional and the parameter that is `None` should be calculated. This prevents miscalculation of a parameter that is actually supposed to be `0`.

In addition, why return a dictionary with the missing parameter when the user knows what they wanted calculated. Instead, you can just return the value.

physics/coulombs_law.py
```diff
@@ -0,0 +1,86 @@
...
+    raise ValueError("One and only one argument must be 0")
+
+
+if __name__ == "__main__":
+    print("Enter the magnitude of following  values and enter 0 in unknown value.")
+    force = float(input("Enter the magnitude of force:"))
+    charge_1 = float(input("Enter the magnitude of charge_1:"))
+    charge_2 = float(input("Enter the magnitude of charge_2:"))
+    distance = float(input("Enter the magnitude of distance:"))
+    print(coulombs_law(force, charge_1, charge_2, distance))
```
ZeroDayOwl, line 86: It seems this file uses the codes from the already existing [Coulomb's Law](https://github.com/TheAlgorithms/Python/blob/master/electronics/coulombs_law.py) in the repository to implement the same.

Please refer to CONTRIBUTING.md and DIRECTORY.md to avoid identical implementation of the same algorithm.

ZeroDayOwl (changes requested):

ZeroDayOwl:
@cclauss you can close this PR. It duplicates the same algorithm.

cclauss:
Duplicate? https://github.com/TheAlgorithms/Python/blob/master/electronics/coulombs_law.py

Issue #4133, opened 2021-01-20
 Why is the checkout version 1?


ryuichi1208:
https://github.com/TheAlgorithms/Python/blob/master/.github/workflows/directory_writer.yml#L9

Why is the checkout version 1? The latest version is 2.

https://github.com/actions/checkout/releases

dhruval-tech:
Assign this issue to me, I've already implemented .

PR #6852, opened 2022-10-09
Bump actions/checkout to v3 (#4133)

madhead:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?
* [x] CI scripts fix

### Checklist:

### DIFF
--- a/.github/workflows/directory_writer.yml
+++ b/.github/workflows/directory_writer.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1  # v1, NOT v2 or v3
+      - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
           python-version: 3.x


cclauss (changes requested):
The comment is on this line because accepting this change would break the `Update DIRECTORY.md` job step.  V1 allows repo maintainers to write changes back to a pull request while later versions do not.

The best solution would be to move `Update DIRECTORY.md` to [pre-commit](https://pre-commit.com) as a _local_ job and once we are confident that works, we can remove the job step here and upgrade the Action.

madhead:
I see. I tested it in my fork and it pushed the change, but I didn't see similar changes pushed by the bot.

What do you mean by "allows repo maintainers to write changes back to a pull request while later versions do not". Do you want this workflow to update the PRs automatically with the updated MD file. I mean, when someone pushes a new algorithm, do you want this workflow to update (add commits) to the same PR?

cclauss:
> I didn't see similar changes pushed by the bot.

Are you a **maintainer** on this repo?

https://github.com/TheAlgorithms/Python/blob/master/scripts/build_directory_md.py Writes a new file.

>  when someone pushes a new algorithm, do you want this workflow to update (add commits) to the same PR?

When a **project maintainer** makes a modification, a new markdown file is generated and added to the same PR.

Pre-commit was designed to do this so the best solution would be to move `build_directory_md.py` to pre-commit.

dhruvmanila:
I don't think pre-commit is the right place to put it as most of the contributors won't install it. It's possible to do the auto-commit on PR or maybe after merging the PR. Let me take a closer look at it as I've seen it done on other projects.

This PR can be closed as of now as this requires a bit of discussion into how we want to do this.

cclauss:
> most of the contributors won't install it

They do not have to because we run https://results.pre-commit.ci/repo/github/63476337 on every pull request.  Have you noticed that if you submit a PR that is not black formatted or isort sorted not pyupgrade compatible, pre-commit.ci is laying a commit on top that fixes all these issues.  This is saving us tons of grief.

dhruvmanila:
Yes, I'm aware of that but does it work on local hooks? Another alternative would be to keep this script in a separate repo which will define the hook and thus allowing us to use it across all our repositories. I don't know if this works or not, don't have details regarding how the script works, but I guess if the structure is similar then it should work.

cclauss:
I will submit a solution in a few hours unless @madhead wants to give it a whirl.

PR #7141, opened 2022-10-14
Add algorithm for Casimir Effect

ZeroDayOwl:
### Describe your change:
I implemented an algorithm to find out the magnitudes of either Force of attraction, distance or surface area of any of the two parallel plates in vacuum between which there is an existence of Casimir effect, using the general equation for Casimir effect, provided the values of the any of the two parameters are non-zero numbers.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/physics/casimir_effect.py
+++ b/physics/casimir_effect.py
@@ -1,38 +1,38 @@
 """
-Title : Finding the value of magnitude of either the Casimir force, the surface area 
-of one of the plates or distance between the plates provided that the other 
+Title : Finding the value of magnitude of either the Casimir force, the surface area
+of one of the plates or distance between the plates provided that the other
 two parameters are given.
 
-Description : In quantum field theory, the Casimir effect is a physical force 
-acting on the macroscopic boundaries of a confined space which arises from the 
-quantum fluctuations of the field. It is a physical force exerted between separate 
-objects, which is due to neither charge, gravity, nor the exchange of particles, 
-but instead is due to resonance of all-pervasive energy fields in the intervening 
-space between the objects. Since the strength of the force falls off rapidly with 
-distance it is only measurable when the distance between the objects is extremely 
-small. On a submicron scale, this force becomes so strong that it becomes the 
+Description : In quantum field theory, the Casimir effect is a physical force
+acting on the macroscopic boundaries of a confined space which arises from the
+quantum fluctuations of the field. It is a physical force exerted between separate
+objects, which is due to neither charge, gravity, nor the exchange of particles,
+but instead is due to resonance of all-pervasive energy fields in the intervening
+space between the objects. Since the strength of the force falls off rapidly with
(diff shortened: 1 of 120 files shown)


physics/casimir_effect.py
```diff
@@ -0,0 +1,121 @@
...
+# Pi and the function
+REDUCED_PLANCK_CONSTANT = 1.054571817e-34  # unit of ℏ : J * s
+
+SPEED_OF_LIGHT = 3e8  # unit of c : m * s^-1
+
+PI = 3.14
+
+
+def casimir_force(force: float, area: float, distance: float) -> dict[str, float]:
+
```
CaedenPH, line 50:
```suggestion
```

CaedenPH (commented):
Instead of passing `0`, how about all arguments are `None`, and whichever argument isnt passed is evaluated. Otherwise, something may break if a legitimate value of a parameter is `0`

In addition, instead of returning a dictionary of the missing value, you can just return the missing value

physics/casimir_effect.py
ZeroDayOwl: @CaedenPH, thank you for your suggestions. Technically if the legitimate value of a parameter is zero, the value of the other two parameters in the equation become `0` as well. That is why the argument which is to be evaluated by the algorithm can be `0` here. Another reason of using this is to explicitly let the user know which arguments are passed on to the function.

Moreoever, I wanted to return the data in key:value pair, so I used the dictionary to return the value.

ZeroDayOwl:
Hey @cclauss could you kindly take a look at the PR?

physics/casimir_effect.py
```diff
@@ -0,0 +1,121 @@
...
+from __future__ import annotations
+
+# Define the Reduced Planck Constant ℏ (H bar), speed of light C, value of
+# Pi and the function
+REDUCED_PLANCK_CONSTANT = 1.054571817e-34  # unit of ℏ : J * s
+
+SPEED_OF_LIGHT = 3e8  # unit of c : m * s^-1
+
+PI = 3.14
```
cclauss, line 46: `from math import pi`

cclauss (approved):
Nice!

physics/casimir_effect.py
CaedenPH: @cclauss 
I'm willing to say that returning in a key:value is better, but what is your opinion?
To me it seems like doing
```py
def add(a, b):
    return {"sum": a+b}
```

physics/casimir_effect.py
ZeroDayOwl: @cclauss Thanks for the suggestion! I used the module, and also updated the doc tests accordingly!

ZeroDayOwl:
@cclauss Thanks for the suggestion! I used the module, and also updated the doc tests accordingly!

Issue #7020, opened 2022-10-11
Sum of binary tree nodes


CaedenPH:
### Feature description

```py
    3
   / \
  9   20
     /  \
    15   7
```
The output would be `3+9+20+15+7`

### Would you like to work on this feature?

- [X] Yes, I want to work on this feature!

nfarhaan:
Hello,
Can you assign this issue to me?

Akashhpatil:
I would like to work on this issue!

PremSundarB:
Hey I would love to work on this issue !

UTKARSHMOHAN2003:
SIE ASSIGN ME

AbrarAhmed3:
I would like to work on this issue!
Can you assign this issue to me?

aryan-1311:
Hi, I would like to work on this issue.
Assign it to me.

imnisi:
Hii @CaedenPH! Could you assign this issue to me? I would love to solve this issue. 
Thank You

huangderek02:
Hii @CaedenPH! Could you assign this issue to me? I would love to solve this issue.
Thank You

Shubham74500:
Hii sir could you assign this issue to me.

drashan14:
Hello Sir/Mam,
I would like to do this job could you assign me this task.

anuragshuklajec:
Hii @CaedenPH! Could you assign this issue to me ?

SaurabhBudhwani26:
Hi @CaedenPH , I would love to work on this issue, can you please assign me this?

DeiseCristiane:
Olá, quero contribuir com esse projeto.

shyambaral:
hlo can i do that

ashutosh-gupta2111:
Hello Sir,
Can You assign this issue to me?

Sanjana20304:
I would like to work on this. Please assign this to me

PR #7162, opened 2022-10-14
feat: Binary tree node sum (#7020)

CaedenPH:
### Describe your change:
Implements #7020 
Algorithm to get the sum of all tree nodes

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/binary_tree/binary_tree_node_sum.py
+++ b/data_structures/binary_tree/binary_tree_node_sum.py
@@ -14,8 +14,7 @@ from __future__ import annotations
 
 class Node:
     """
-    A Node has value variable and pointers
-    to Nodes to its left and right.
+    A Node has a value variable and pointers to Nodes to its left and right.
     """
 
     def __init__(self, value: int) -> None:
--- a/data_structures/binary_tree/binary_tree_node_sum.py
+++ b/data_structures/binary_tree/binary_tree_node_sum.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+
+class Node:
+    """
+    A Node has value variable and pointers
+    to Nodes to its left and right.
+    """
+
+    def __init__(self, value: int) -> None:
+        self.value = value
+        self.left: Node | None = None
+        self.right: Node | None = None
+
+
+class BinaryTreeNodeSum:
+    r"""
+    The below tree looks like this
+        10
+       /  \
+      5   -3
+     /    / \
+    12   8  0
+
+    >>> tree = Node(10)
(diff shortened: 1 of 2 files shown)


CaedenPH:
@cclauss pinging for review

data_structures/binary_tree/binary_tree_node_sum.py
```diff
@@ -0,0 +1,75 @@
...
+    >>> tree.right.left = Node(8)
+    >>> tree.right.right = Node(0)
+    >>> BinaryTreeNodeSum(tree).node_sum()
+    32
+    """
+
+    def __init__(self, tree: Node) -> None:
+        self.tree = tree
+
+    def depth_first_search(self, node: Node | None) -> int:
```
cclauss, line 61: from collections.abc import Iterator

Extra credit: `def __iter__(self) -> Iterator[int]`

Then you could replace `node_sum()` with `sum(binary_tree_node_sum)`.

cclauss (approved):
Let me know if you want to try the extra credit bit or not.

CaedenPH:
@cclauss 
Is this what you meant?

dhruvmanila:
> Is this what you meant?

No, that's not what he meant. He's asking you to implement the iteration protocol for a custom Python object. It comprises of two methods `__iter__` to which returns an iterator and `__next__` which returns the next object.

These methods would be made available to the `Node` class which then would allow us to do:
```python
sum(node.value for node in tree)
```

If you need a bit more understanding, I could help you out with an example.

A class for doing a sum seems redundant. Alternatively, you could just create a function instead.

cclauss:
`list(binary_tree_node_sum)` would return `[10, 5, 12, -3, 8, 0]`

`sum(binary_tree_node_sum)` would return `32`

 I would just define `__iter__` and not define `__next__` but whatever works for you.  I would use `yield` more than return.  There are lots of examples in this repo that I linked to above.

CaedenPH:
> `list(binary_tree_node_sum)` would return `[10, 5, 12, -3, 8, 0]`
> 
> 
> 
> `sum(binary_tree_node_sum)` would return `32`.
> 
> 
> 
> `max()`, `min()`, `mean()` would all just work out-of-the-box.
> 
> 
> 
>  I would just define `__iter__` and not define `__next__` but whatever works for you.  I would use `yield` more than `return`.  There are lots of examples in this repo that I linked to above.

Okay, I understand. However, perhaps all of this should be implemented in the basic binary tree or some other more general file instead of binary tree path sum

PR #7191, opened 2022-10-14
refactor: use abs() and inplace operation

Luk-ESC:
n = -n if n < 0 else n --> n = abs(n)
n = n // 10 --> n //= 10

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/maths/sum_of_digits.py
+++ b/maths/sum_of_digits.py
@@ -14,11 +14,11 @@ def sum_of_digits(n: int) -> int:
     >>> sum_of_digits(0)
     0
     """
-    n = -n if n < 0 else n
+    n = abs(n)
     res = 0
     while n > 0:
         res += n % 10
-        n = n // 10
+        n //= 10
     return res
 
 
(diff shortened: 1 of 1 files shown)


cclauss:
Did the `timeit()` benchmarks improve or not?

dhruvmanila (commented):
These changes are fine, are there similar code elsewhere which can be refactored as well?

cclauss:
I would like to avoid this issue https://github.com/TheAlgorithms/Python/pull/7190#issuecomment-1279545707. I think we are ok on this pr.

I still want to make sure that the timeit test show that the proposed solution is faster than the current one.

dhruvmanila:
> I would like to avoid this issue [#7190 (comment)](https://github.com/TheAlgorithms/Python/pull/7190#issuecomment-1279545707). I think we are ok on this pr.

I do understand where you're coming from, but this is not an optimization, rather a refactor.

> I still want to make sure that the timeit test show that the proposed solution is faster than the current one.

There won't be any difference.

cclauss:
> There won't be any difference.

Prove it.

CaedenPH (approved):
@cclauss Looks good

cclauss (approved):

PR #7197, opened 2022-10-14
types: Update binary search tree typehints

CaedenPH:
### Describe your change:
Update the typehints in `data_structures/binary_tree/binary_tree_search.py`

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/binary_tree/binary_search_tree.py
+++ b/data_structures/binary_tree/binary_search_tree.py
@@ -45,8 +45,7 @@ class BinarySearchTree:
     def is_right(self, node: Node) -> bool:
         if node.parent and node.parent.right:
             return node == node.parent.right
-        else:
-            return False
+        return False
 
     def empty(self) -> bool:
         return self.root is None
@@ -61,7 +60,7 @@ class BinarySearchTree:
         else:  # Tree is not empty
             parent_node = self.root  # from root
             if parent_node is None:
-                return
+                return None
             while True:  # While we don't get to a leaf
                 if value < parent_node.value:  # We go left
                     if parent_node.left is None:
--- a/data_structures/binary_tree/binary_search_tree.py
+++ b/data_structures/binary_tree/binary_search_tree.py
@@ -2,15 +2,20 @@
 A binary search Tree
 """
 
+from __future__ import annotations
+
+from collections.abc import Iterable
+from typing import Any
+
 
 class Node:
-    def __init__(self, value, parent):
+    def __init__(self, value: int | None = None):
         self.value = value
-        self.parent = parent  # Added in order to delete a node easier
-        self.left = None
-        self.right = None
+        self.parent: Node | None = None  # Added in order to delete a node easier
+        self.left: Node | None = None
+        self.right: Node | None = None
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         from pprint import pformat
 
         if self.left is None and self.right is None:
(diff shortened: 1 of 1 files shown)


data_structures/binary_tree/binary_search_tree.py
```diff
@@ -69,12 +79,12 @@ def __insert(self, value):
                         parent_node = parent_node.right
             new_node.parent = parent_node
 
-    def insert(self, *values):
+    def insert(self, *values) -> BinarySearchTree:
```
dhruvmanila, line 82: Postponed evaluation of annotations has been put on hold indefinitely, so we should not be incorporating them into the codebase until the PEP is finalized: https://docs.python.org/3.11/whatsnew/3.11.html#pep-563-may-not-be-the-future. Please update this and all the similar changes.

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -2,15 +2,20 @@
 A binary search Tree
 """
 
+from __future__ import annotations
+
```
dhruvmanila, line 6:
```suggestion
```
This should not be needed as we use latest Python for testing and writing code which as of now is `3.10.*`

dhruvmanila (changes requested):

data_structures/binary_tree/binary_search_tree.py
dhruvmanila: Actually, this might still be required, please verify.

dhruvmanila: Yaa, this is not needed.

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -69,12 +77,11 @@ def __insert(self, value):
                         parent_node = parent_node.right
             new_node.parent = parent_node
 
-    def insert(self, *values):
+    def insert(self, *values) -> None:
```
dhruvmanila, line 80: Oh, this is not what I meant, sorry if I was not clear. When you're returning the type, it's not actually defined, so it will raise an exception when you run the code. To solve this problem, the linked PEP was proposed to postpone the evaluation of annotations like this.

The solution for now is to quote it:
```suggestion
    def insert(self, *values) -> "BinarySearchTree":
```

Update the return statement and doctest as well.

dhruvmanila (changes requested):

data_structures/binary_tree/binary_search_tree.py
CaedenPH: I understand what you were saying, instead I realised that the fact that the `insert` statement was returning `self` was questionable, so I removed the return statement and made it return `None`

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -37,23 +40,28 @@ def __reassign_nodes(self, node, new_children):
...
-            self.root = new_children
+            self.root = None
 
-    def is_right(self, node):
-        return node == node.parent.right
+    def is_right(self, node: Node) -> bool:
+        if node.parent and node.parent.right:
+            return node == node.parent.right
+        else:
+            return False
```
dhruvmanila, line 49:
```suggestion
        return False
```

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -37,23 +40,28 @@ def __reassign_nodes(self, node, new_children):
...
         Insert a new node in Binary Search Tree with value label
         """
-        new_node = Node(value, None)  # create a new Node
+        new_node = Node(value)  # create a new Node
         if self.empty():  # if Tree is empty
             self.root = new_node  # set its root
         else:  # Tree is not empty
             parent_node = self.root  # from root
+            if parent_node is None:
+                return
```
dhruvmanila, line 64:
```suggestion
                return None
```

dhruvmanila (changes requested):

data_structures/binary_tree/binary_search_tree.py
dhruvmanila: > I am aware that the class is not defined at runtime, which is why I imported annotations to automatically encase the import in quotes

Yes, but this feature has been put on hold indefinitely, so we should not incorporate it. But not returning makes more sense, thanks!

dhruvmanila (approved):

CaedenPH:
@cclauss Do you have any additional comments?

cclauss:
if @dhruvmanila is satisfied then that is all that is required to merge.

Issue #7233, opened 2022-10-15
Add flake8-comprehensions to pre-commit


CaedenPH:
### Feature description

Add `flake8-comprehensions` to `pre-commit` and fix errors

dhruvmanila:
closed by #7235 

Please make sure to link the issue, if existing, in the PR itself, so that it gets closed automatically.

PR #7235, opened 2022-10-15
Add Flake8 comprehensions to pre-commit

CaedenPH:
### Describe your change:
Implements (#7233)

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/hashes/enigma_machine.py
+++ b/hashes/enigma_machine.py
@@ -1,8 +1,8 @@
 alphabets = [chr(i) for i in range(32, 126)]
-gear_one = [i for i in range(len(alphabets))]
-gear_two = [i for i in range(len(alphabets))]
-gear_three = [i for i in range(len(alphabets))]
-reflector = [i for i in reversed(range(len(alphabets)))]
+gear_one = range(len(alphabets))
+gear_two = range(len(alphabets))
+gear_three = range(len(alphabets))
+reflector = reversed(range(len(alphabets)))
 code = []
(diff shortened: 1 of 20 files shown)


realstealthninja (approved):

hashes/enigma_machine.py
```diff
@@ -1,8 +1,8 @@
 alphabets = [chr(i) for i in range(32, 126)]
-gear_one = [i for i in range(len(alphabets))]
-gear_two = [i for i in range(len(alphabets))]
-gear_three = [i for i in range(len(alphabets))]
-reflector = [i for i in reversed(range(len(alphabets)))]
+gear_one = range(len(alphabets))
+gear_two = range(len(alphabets))
+gear_three = range(len(alphabets))
+reflector = reversed(range(len(alphabets)))
```
dhruvmanila, line 5: The old code is constructing a list, but the updated code is giving us a `range` object. Is this ok?

dhruvmanila (approved):
One small doubt, otherwise LGTM! 🎉

hashes/enigma_machine.py
realstealthninja: no because range returns a list

CaedenPH: Range returns a `range` object, which does not have the same methods as a list. Now that I realise, yes, I believe this is a problem. I will encase these range objects in the list function

PR #7239, opened 2022-10-15
simple fixes

itsdiy0:
### Describe your change:
1. fixed a bug in the input of the function which was not  letting the code run
2. removed an input out of 2 inputs in the main function (the input was for the getting number of elements in the array and the second one was for the array itself, I removed the first one so now it just takes the array itself and calculates the length of array, more efficient and not hard coded)

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -22,5 +22,4 @@ def max_subarray(nums: list) -> int:
 
 if __name__ == "__main__":
     array = list(map(int, input("\nEnter the numbers : ").strip().split()))
-    n = len(array)
-    print(max_subarray(array[:n]))
+    print(max_subarray(array))
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -1,4 +1,4 @@
-def max_subarray(nums: list) -> int:
+def max_subarray(nums: list[int]) -> int:
     """
     Returns the subarray with maximum sum
     >>> max_subarray([1,2,3,4,-2])
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -1,4 +1,4 @@
-def max_subarray(nums: list[int]) -> int:
+def max_subarray(nums: list) -> int:
     """
     Returns the subarray with maximum sum
     >>> max_subarray([1,2,3,4,-2])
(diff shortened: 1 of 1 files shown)


realstealthninja (approved):

other/maximum_subarray.py
```diff
@@ -21,6 +21,6 @@ def max_subarray(nums: list[int]) -> int:
 
 
 if __name__ == "__main__":
-    n = int(input("Enter number of elements : ").strip())
-    array = list(map(int, input("\nEnter the numbers : ").strip().split()))[:n]
-    print(max_subarray(array))
+    array = list(map(int, input("\nEnter the numbers : ").strip().split()))
+    n = len(array)
+    print(max_subarray(array[:n]))
```
CaedenPH, line 26: Whats the point of doing `array[:n]` or even creating variable `n`? All you're doing is saying "index this list up to all its elements"

other/maximum_subarray.py
```diff
@@ -1,4 +1,4 @@
-def max_subarray(nums: list[int]) -> int:
+def max_subarray(nums: list) -> int:
```
CaedenPH, line 1: What is the point of this change?

itsdiy0: its getting type error with that and not running, so I just removed : 
  File "maxsubbarray.py", line 1, in <module>
    def max_subarray(nums: list[int]) -> int:
TypeError: 'type' object is not subscriptable

other/maximum_subarray.py
itsdiy0: yea, You are right, n is not need at all

other/maximum_subarray.py
CaedenPH: > its getting Typeerror with that and not running, so I just removed (try it) : Traceback (most recent call last): File "maxsubbarray.py", line 1, in def max_subarray(nums: list[int]) -> int: TypeError: 'type' object is not subscriptable

That is a problem with your python version. `list` being used for typehinting is only available on python versions after python `3.10`

itsdiy0: oh ok, I just removed the n in my commit then, I kept the type-hinting

PR #7244, opened 2022-10-15
Add area of kite and hexagon function

mahiuddin-dev:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/maths/area.py
+++ b/maths/area.py
@@ -458,6 +458,52 @@ def area_rhombus(diagonal_1: float, diagonal_2: float) -> float:
     return 1 / 2 * diagonal_1 * diagonal_2
 
 
+def area_kite(diagonal_1: float, diagonal_2: float) -> float:
+    """
+    Calculate the area of a kite.
+    >>> area_kite(5, 10)
+    25.0
+    >>> area_kite(2.5, 3.5)
+    4.375
+    >>> area_kite(0, 0)
+    0.0
+    >>> area_kite(-1, -2)
+    Traceback (most recent call last):
+        ...
+    ValueError: area_kite() only accepts non-negative values
+    >>> area_kite(1, -2)
+    Traceback (most recent call last):
+        ...
+    ValueError: area_kite() only accepts non-negative values
+    >>> area_kite(-1, 2)
+    Traceback (most recent call last):
+        ...
+    ValueError: area_kite() only accepts non-negative values
+    """
+    if diagonal_1 < 0 or diagonal_2 < 0:
+        raise ValueError("area_kite() only accepts non-negative values")
+    return 1 / 2 * diagonal_1 * diagonal_2
+
+
+def area_hexagon(side: float) -> float:
+    """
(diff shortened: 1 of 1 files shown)


maths/area.py
```diff
@@ -458,6 +458,52 @@ def area_rhombus(diagonal_1: float, diagonal_2: float) -> float:
...
+        ...
+    ValueError: area_kite() only accepts non-negative values
+    >>> area_kite(-1, 2)
+    Traceback (most recent call last):
+        ...
+    ValueError: area_kite() only accepts non-negative values
+    """
+    if diagonal_1 < 0 or diagonal_2 < 0:
+        raise ValueError("area_kite() only accepts non-negative values")
+    return 1 / 2 * diagonal_1 * diagonal_2
```
realstealthninja, line 485:
```suggestion
    return (diagonal_1 * diagonal_2) * 0.5
```
multiplication is always faster than division!

maths/area.py
```diff
@@ -458,6 +458,52 @@ def area_rhombus(diagonal_1: float, diagonal_2: float) -> float:
...
+    >>> area_hexagon(0)
+    0.0
+    >>> area_hexagon(-1)
+    Traceback (most recent call last):
+        ...
+    ValueError: area_hexagon() only accepts non-negative values
+    """
+    if side < 0:
+        raise ValueError("area_hexagon() only accepts non-negative values")
+    return (3 * sqrt(3) * (side * side)) / 2
```
realstealthninja, line 504:
```suggestion
    return (3 * sqrt(3) * (side * side))  * 0.5

```
same here multiplication is always faster than division!

maths/area.py
```diff
@@ -458,6 +458,52 @@ def area_rhombus(diagonal_1: float, diagonal_2: float) -> float:
     return 1 / 2 * diagonal_1 * diagonal_2
 
 
+def area_kite(diagonal_1: float, diagonal_2: float) -> float:
```
realstealthninja, line 461: @cclauss  need you opinion on something this function is quite similar to the one above like exactly the same just with different tests and a different name are similar functions allowed?

realstealthninja (changes requested):

maths/area.py
realstealthninja, line 504: question: is pow(side, 2 ) or side ** 2 better than side * side?

maths/area.py
cclauss: I think that if the function above will also work for kites then we should add a comment to that function and avoid a duplicate function.

realstealthninja: I completely agree.

maths/area.py
cclauss: * https://docs.python.org/3/library/timeit.html
* https://github.com/TheAlgorithms/Python/search?q=timeit

realstealthninja (changes requested):

PR #7392, opened 2022-10-18
added games folder

AADITYAG2002:


games/pong.py
```diff
@@ -0,0 +1,133 @@
...
+        if pressed_keys[K_s]:
+            self.rect.move_ip(0, 1)
+
+        if self.rect.top <= 0:
+            self.rect.top = 0
+        if self.rect.bottom >= screen_height:
+            self.rect.bottom = screen_height
+
+
+class Ai_Paddle(pygame.sprite.Sprite):
```
CaedenPH, line 31:
```suggestion
class AiPaddle(pygame.sprite.Sprite):
```

CaedenPH (commented):
The point of classes is that you can instantiate multiple instances of the same class. 
This means that you don't have to create identical classes for the user and for the ai, instead two instances of the same class. Once you have you have these instances you can edit certain properties if necessary.

@cclauss What is your opinion on a games folder? It seems to me like it's veering slightly off of the algorithms topic, and that games can't really be testing that easily with doctests, or otherwise.

CaedenPH (changes requested):
Meant to request changes

cclauss:
> @cclauss What is your opinion on a games folder?

Games are not algorithms.  There are plenty of repos for games but this is not one of them.

PR #7565, opened 2022-10-23
ci: Consolidate requirements into `pyproject.toml`

CaedenPH:
### Describe your change:
Consolidate the requirements into `pyproject.toml`
https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


CaedenPH:
@cclauss Never used pyproject.toml, so expecting lots of changes

cclauss:
We could also move our codespell, mypy, and pyupgrade args from .pre-commit into pyproject.toml.

Please run https://pypi.org/project/validate-pyproject to validate the file and let's [add that tool to pre-commit](https://github.com/internetarchive/openlibrary/blob/master/.pre-commit-config.yaml#L57-L60).

cclauss:
We might run into trouble consolidating our pytest args because our GitHub Actions run it [two](https://github.com/TheAlgorithms/Python/blob/master/.github/workflows/build.yml#L25) different [ways](https://github.com/TheAlgorithms/Python/blob/master/.github/workflows/project_euler.yml#L25).  So that might be difficult to express.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[project]
+name = "python-algorithms"
+description = "All algorithms implemented in Python"
+version = "1.0.0"
+
+authors = [
+    { name = "Christian Clauss", email = "cclauss@me.com" },
+    { name = "Dhruv Manilawala", email = "dhruvmanila@gmail.com" },
+    { name = "John law", email = "dhruvmanila@gmail.com" }
```
cclauss, line 9:
```suggestion
    { name = "John Law", email = "dhruvmanila@gmail.com" }
```

CaedenPH:
> We could also move our codespell, mypy, and pyupgrade args from .pre-commit into pyproject.toml.
> 
> Please run https://pypi.org/project/validate-pyproject to validate the file and let's [add that tool to pre-commit](https://github.com/internetarchive/openlibrary/blob/master/.pre-commit-config.yaml#L57-L60).

How would this be done?

pyproject.toml
```diff
@@ -6,7 +6,7 @@ version = "1.0.0"
 authors = [
     { name = "Christian Clauss", email = "cclauss@me.com" },
     { name = "Dhruv Manilawala", email = "dhruvmanila@gmail.com" },
-    { name = "John law", email = "dhruvmanila@gmail.com" }
+    { name = "John Law", email = "dhruvmanila@gmail.com" }
```
cclauss, line 9: The email address is wrong.

CaedenPH: > The email address is wrong.

What is the correct email address?
I took it from his github bio

cclauss: @poyea Do you want to be listed here?  With your email address?

dhruvmanila:
The only benefit we get in using `pyproject.toml` is to have a central place for configuration. We don't need the project declaraction, build system, dependencies.

Build system is used to provide a backend like poetry, flit or setuptools to actually build the Python package (wheels). For our use case, `requirements.txt` is actually the best option instead of declaring it in `pyproject.toml`. Otherwise, users will be required to install using `pip install .` which installs the dependencies along with the *package* (remember this is not a package) and the egg info. Project declaration is not useful as it's used to provide metadata information for PyPi. I guess it was added to declare the dependencies which as mentioned not useful. Adding the authors field is a bit misleading as this is a work of all the contributors.

cclauss:
Let's use `maintainers` rather than `authors`... https://peps.python.org/pep-0621/#authors-maintainers

Would we be able to use `pip install .` or probably better `pip install --editable .` in our GitHub Actions?

Could `pip install git+https://github.com/TheAlgorithms/Python.git` be made to work?

dhruvmanila:
As I said, this won't work as `TheAlgorithms/Python` is not a "Python package", you cannot install it as it's not even structured like a standard package. I don't really see why you're wasting time in this.

cclauss:
OK...  So now pip seems to be installing all of our dependencies and dev-dependencies.  It is pytest that seems broken.

cclauss:
Now it seems to be working end-to-end without the need for `requirements.txt`!!

pyproject.toml
poyea: I'm good with that

cclauss:
Please rebase.

CaedenPH:
> Please rebase.

Done

dhruvmanila:
I don't really recommend defining dependencies in `pyproject.toml` file as this is not a Python package. It doesn't make sense to define them there. We can use `pyproject.toml` to define all the tool configs though, but we should be using `requirements.txt` to define the dependencies.

cclauss:
I fail to understand the logic.  `pip` knows how to install project dependencies from two different file types.  Why does it matter whether this is a package or not?

dhruvmanila (changes requested):
I guess I must've misunderstood something, if it works then it's fine @cclauss

pyproject.toml
```diff
@@ -1,3 +1,52 @@
+[build-system]
+requires = ["setuptools ~= 65.5", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "python-algorithms"
+description = "All algorithms implemented in Python"
+version = "1.0.0"
+
+authors = [
```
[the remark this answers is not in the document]
cclauss: https://peps.python.org/pep-0621#drop-maintainers-to-unify-with-authors

.pre-commit-config.yaml
```diff
@@ -7,7 +7,10 @@ repos:
       - id: end-of-file-fixer
         types: [python]
       - id: trailing-whitespace
-      - id: requirements-txt-fixer
+        exclude: |
+          (?x)^(
+            data_structures/heap/binomial_heap.py
+          )$
```
dhruvmanila, line 13: This got removed here: https://github.com/TheAlgorithms/Python/commit/57ccabbaeb0f32165271e3a218bc9c6dcfc21823. I don't know how it got added back here?

dhruvmanila (changes requested):

dhruvmanila (approved):
Thanks a lot! 🎉

poyea (approved):
Thank you!

pyproject.toml
```diff
@@ -1,3 +1,53 @@
...
+    "numpy",
+    "opencv-python",
+    "pandas",
+    "pillow",
+    "projectq",
+    "qiskit",
+    "requests",
+    "rich",
+    "scikit-fuzzy",
+    "sklearn",
```
cclauss, line 31:
```suggestion
    "scikit-learn",
```
https://pypi.org/project/scikit-learn/

pyproject.toml
```diff
@@ -1,3 +1,53 @@
...
+    "statsmodels",
+    "sympy",
+    "tensorflow",
```
cclauss, line 34:
```suggestion
    # "tensorflow",  # Not yet compatible with Python 3.11
```

PR #7589, opened 2022-10-24
Update count_number_of_one_bits.py

JatinR05:
removed the modulo operator as it is very time consuming in comparison to the and operator

### Describe your change:

* [ ] Add an algorithm?
* [X] Documentation change?

### Checklist:

### DIFF
--- a/bit_manipulation/count_number_of_one_bits.py
+++ b/bit_manipulation/count_number_of_one_bits.py
@@ -59,10 +59,12 @@ def get_set_bits_count_using_modulo_operator(number: int) -> int:
         number = number >> 1
     return result
 
+
 def benchmark() -> None:
     """
     Benchmark code for comparing 2 functions, with different length int values.
     """
+
     def do_benchmark(number: int) -> None:
         setup = "import __main__ as z"
         print(f"Benchmark when {number = }:")
@@ -80,11 +82,6 @@ def benchmark() -> None:
         do_benchmark(number)
         print()
 
-
-
-
-
-    
 
 if __name__ == "__main__":
     import doctest
--- a/bit_manipulation/count_number_of_one_bits.py
+++ b/bit_manipulation/count_number_of_one_bits.py
@@ -59,84 +59,32 @@ def get_set_bits_count_using_modulo_operator(number: int) -> int:
         number = number >> 1
     return result
 
-
 def benchmark() -> None:
     """
-    Benchmark code for comparing 2 functions,
-    with 3 different length int values.
+    Benchmark code for comparing 2 functions, with different length int values.
     """
-    print("\nFor 25 = :")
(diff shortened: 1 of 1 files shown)


cclauss:
Please add a [timeit](https://docs.python.org/3/library/timeit.html#timeit.timeit) (or similar) benchmark to prove that the proposed code is faster than the original code.
There are lots of examples in this repo.  https://github.com/TheAlgorithms/Python/search?q=timeit

bit_manipulation/count_number_of_one_bits.py
```diff
@@ -1,19 +1,51 @@
...
+    >>> get_set_bits_count_using_brian_kernighans_algorithm(0)
     0
-    >>> get_set_bits_count(256)
+    >>> get_set_bits_count_using_brian_kernighans_algorithm(256)
     1
-    >>> get_set_bits_count(-1)
+    >>> get_set_bits_count_using_brian_kernighans_algorithm(-1)
+    Traceback (most recent call last):
+        ...
+    ValueError: the value of input must be positive
```
cclauss, line 22: Zero is not positive.
```suggestion
    ValueError: the value of input must not be negative
```

JatinR05:
I have the code as suggested

JatinR05:
I have updated the code as suggested.

cclauss:
Please run this on your own machine.  The doctest fails.

JatinR05:
I am extremely sorry for the silly mistakes. Please could you check it now

cclauss:
Would it be more self-documenting to change the variable `number` to `bit_length` throughout the script?

Let's make `benchmark()` more readable with something like this.
```
def benchmark() -> None:
    """
    Benchmark code for comparing 2 functions, with different length int values.
    """
    def do_benchmark(number: int) -> None:
        print(f"Benchmark when {number = }:")
        print(f"{get_set_bits_count_using_modulo_operator(number) = }")
        timing = timeit(
            "z.get_set_bits_count_using_modulo_operator(25)",
            setup="import __main__ as z",
        )
        print(f"timeit() runs in {timing} seconds")
        print(f"{get_set_bits_count_using_brian_kernighans_algorithm(number) = }")
        timing = timeit(
            "z.get_set_bits_count_using_brian_kernighans_algorithm(25)",
            setup="import __main__ as z",
        )
        print(f"timeit() runs in {timing} seconds")

    for number in (25, 37, 58, 0):
        do_benchmark(number)
        print()
```

JatinR05:
I guess the variable name "number" is more appropriate as it's representing the number which the user will be providing as the input and not the bit_length.
Regarding the benchmark updates, it's really a great idea. I'll surely update it.

JatinR05:
Thank you.

PR #7608, opened 2022-10-25
Added function that checks if a string is an isogram

M3talM0nk3y:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


strings/check_isogram.py
```diff
@@ -0,0 +1,28 @@
+"""
+wiki: https://en.wikipedia.org/wiki/Heterogram_(literature)#Isograms
+"""
+
+
+def check_isogram(string: str) -> bool:
```
cclauss, line 6: This is a slightly more self-documenting name for a function that returns a bool.
```suggestion
def is_isogram(string: str) -> bool:
```

Also, please add a test for `true123`.  Should that return True or raise a ValueError?

strings/check_isogram.py
```diff
@@ -0,0 +1,28 @@
...
+    >>> check_isogram('allowance')
+    False
+    """
+    letters = sorted(string)
+
+    for idx, letter in enumerate(letters):
+        if letter == letters[idx - 1]:
+            return False
+
+    return True
```
cclauss, line 21:
```suggestion
    return len(letters) == len(set(letters))
```

strings/check_isogram.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+if __name__ == "__main__":
+    input_str = input("Enter a string ").strip().lower()
```
cclauss, line 25: The `str.lower()` should happen ___in___ the function.

strings/check_isogram.py
```diff
@@ -0,0 +1,28 @@
+"""
+wiki: https://en.wikipedia.org/wiki/Heterogram_(literature)#Isograms
+"""
+
+
+def check_isogram(string: str) -> bool:
+    """
+    An isogram is a word in which no letter is repeated.
+    Examples of isograms are uncopyrightable and ambidextrously.
+    >>> check_isogram('uncopyrightable')
```
cclauss, line 10:
```suggestion
    >>> check_isogram('Uncopyrightable')
```

strings/check_isogram.py
M3talM0nk3y: Renamed function and added test for when a string contains 1 or more digits. The latter now raises a ValueError. @cclauss

strings/check_isogram.py
M3talM0nk3y: @cclauss Made suggested changes.

strings/check_isogram.py
M3talM0nk3y: @cclauss Lowercasing string inside function.

M3talM0nk3y:
@cclauss Thanks for your feedback. :)

strings/is_isogram.py
```diff
@@ -0,0 +1,32 @@
...
+    >>> is_isogram('allowance')
+    False
+    >>> is_isogram('copy1')
+    Traceback (most recent call last):
+     ...
+    ValueError: String must only contain alphabetic characters.
+    """
+    import re
+
+    if bool(re.search(r"\d", string)):
```
ChrisO345, line 21:
```suggestion
    if not all(x.isalpha() for x in string):
```
This reduces the need to import `re` and catches things other non alphabetic characters such as symbols.

ChrisO345 (changes requested):

ChrisO345 (approved):

PR #7658, opened 2022-10-26
Add srilankan phone number validator.py

sinsankio:
A program to confirm if a phone number is a valid srilankan phone number or not

### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/strings/srilankan_phone_validator.py
+++ b/strings/srilankan_phone_validator.py
@@ -1,6 +1,6 @@
 import re
 
-def srilanka_phone_validator(phone: str) -> bool:
+def is_sri_lankan_phone_number(phone: str) -> bool:
     """
     Determine whether the string is a valid sri lankan phone number or not
     :param phone:
@@ -21,14 +21,11 @@ def srilanka_phone_validator(phone: str) -> bool:
     False
     """
 
-    pattern = re.compile(r'^(?:0|94|\+94|0094)?(?:(11|21|23|24|25|26|27|31|32|33|34|35|36|37|38|41|45|47|51|52|54|55|57|63|65|66|67|81|91)(0|2|3|4|5|7|9)|7(0|1|2|4|5|6|7|8)\d)\d{6}$')
-    matched = re.search(pattern, phone)
-
-    if matched:
-        return True
-    return False
+    pattern = re.compile(r'^(?:0|94|\+94|0{2}94)?(?:(1{2}|21|23|24|25|26|27|31|32|3{2}|34|35|36|37|38|41|45|47|51|52|54|5{2}|57|63|65|6{2}|67|81|91)(0|2|3|4|5|7|9)|7(0|1|2|4|5|6|7|8)\d)\d{6}$')    
+    
+    return bool(re.search(pattern, phone))
 
 if __name__ == '__main__':
-    phone = "0957651234"
+    phone = "+94475682163"
 
-    print(srilanka_phone_validator(phone))
+    print(is_sri_lankan_phone_number(phone))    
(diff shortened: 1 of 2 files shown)


strings/srilankan_phone_validator.py
```diff
@@ -0,0 +1,38 @@
+import re
+
+
+def srilanka_phone_validator(phone: str) -> bool:
+    """
+    Determine whether the string is a valid sri lankan phone number or not
+    :param phone:
+    :return: Boolean
```
CaedenPH, line 8:
```suggestion

```
This tells us nothing that we don't already know

strings/srilankan_phone_validator.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    pattern = re.compile(
+        r"^(?:0|94|\+94|0094)?(?:(11|21|23|24|25|26|27|31|32|33|34|35|36|37|38|41|45|47|51|52|54|55|57|63|65|66|67|81|91)(0|2|3|4|5|7|9)|7(0|1|2|4|5|6|7|8)\d)\d{6}$"
+    )
+    matched = re.search(pattern, phone)
+
+    if matched:
+        return True
+    return False
```
CaedenPH, line 32:
```suggestion
    return bool(re.search(pattern, phone))
```

CaedenPH:
@cclauss @dhruvmanila @poyea What are your opinions on this?

It seems that the fundamental aspect here is the pattern matching, which I personally think could be a good addition to this repository. 

Perhaps we could habe a `match` or `patterns` directory with regex/pattern matching code inside it, such as this one.

In addition, @sinsankio, perhaps you could include some more countries within this file. Maybe a regex dictionary of countries. because just sri lanka seems rather specific

strings/srilankan_phone_validator.py
```diff
@@ -0,0 +1,38 @@
...
+    >>> srilanka_phone_validator("+94-3283048")
+    False
+    >>> srilanka_phone_validator("07779209245")
+    False
+    >>> srilanka_phone_validator("0957651234")
+    False
+    """
+
+    pattern = re.compile(
+        r"^(?:0|94|\+94|0094)?(?:(11|21|23|24|25|26|27|31|32|33|34|35|36|37|38|41|45|47|51|52|54|55|57|63|65|66|67|81|91)(0|2|3|4|5|7|9)|7(0|1|2|4|5|6|7|8)\d)\d{6}$"
```
CaedenPH, line 26: This is causing the `pre-commit` to error because it is so long. Perhaps there is a better, shorter pattern

strings/srilankan_phone_validator.py
```diff
@@ -0,0 +1,38 @@
+import re
+
+
+def srilanka_phone_validator(phone: str) -> bool:
```
CaedenPH, line 4:
```suggestion
def is_sri_lanka_phone_number(phone: str) -> bool:
```

cclauss:
> Perhaps we could habe a match or patterns directory with regex/pattern matching code inside it, such as this one.

Regex operates on strings so this is in the correct directory.  If we create a separate directory for regex then we get a ton of duplicate algorithms (one with regex and one without) to review.  I think we only need one `is_sri_lanka_phone_number()` in this repo.  ;-)

cclauss:
The answer was:
```
    r"^(?:0|94|\+94|0094)?(?:(11|21|23|24|25|26|27|31|32|33|34|35|36|37|38|41|45|47|"
    r"51|52|54|55|57|63|65|66|67|81|91)(0|2|3|4|5|7|9)|7(0|1|2|4|5|6|7|8)\d)\d{6}$"

PR #7721, opened 2022-10-27
Add matrix_shortest_path_using_depth_first_search.py

sinsankio:
### Describe your change:

finding the shortest path of a non symmetrical matrix (m x n) between given start and end points including consideration of obstacles using depth first search approach.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/path_tracing/matrix_shortest_path_using_depth_first_search.py
+++ b/path_tracing/matrix_shortest_path_using_depth_first_search.py
@@ -10,14 +10,14 @@ For manual testing run:
 python3 matrix_shortest_path_using_depth_first_search.py
 """
 
-import math
+import sys
 
 OBSTACLE_POS = -1
 VALID_POS = 1
 START_POS = "s"
 END_POS = "e"
 VISITED_POS = None
-current_shortest_path_distance = math.inf
+current_shortest_path_distance = sys.maxsize
 
 
 def find_shortest_path_using_dfs(
@@ -35,6 +35,10 @@ def find_shortest_path_using_dfs(
     :param col: currently traversing column index of the matrix in a given point of time
     :param distance: calculated distance through out a particular path of the matrix
     :return: None
+
+    Example:
+    >>> find_shortest_path_using_dfs([['s', 1, 1], [1, -1, 1], [-1, 1, 'e']], 0, 0, 0)
+    >>> find_shortest_path_using_dfs([[1, 1], [1, -1, 's', -1], [1, -1, 1]], 1, 2, 0)
     """
     global current_shortest_path_distance
 
@@ -76,14 +80,14 @@ def helper(matrix: list) -> int:
     >>> helper([[-1, -1, 1, 1, 's'], [-1, 1, 1, -1], [1, 1, -1], [1, -1], ['e']])
     8
     >>> helper([['s', 1, 1], [1, 1, -1], [1, -1, 'e']])
-    inf
+    -1
     >>> helper([[1, 1, -1, 1], [1, -1, 's', -1], [1, 1, -1, 1]])
-    inf
+    -1
     """
     global current_shortest_path_distance
 
     found = False
-    current_shortest_path_distance = math.inf
+    current_shortest_path_distance = sys.maxsize
 
     for i in range(len(matrix)):
         if not found:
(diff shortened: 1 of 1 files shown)


path_tracing/matrix_shortest_path_using_depth_first_search.py
```diff
@@ -0,0 +1,108 @@
...
+    Examples:
+    >>> find_shortest_path_using_dfs([['s', 1, 1], [1, -1, 1], [-1, 1, 'e']], 0, 0, 0)
+    None
+    >>> find_shortest_path_using_dfs([[-1, 1], ['s', 1], [-1, 1], [1, 'e']], 1, 0, 0)
+    None
+    >>> find_shortest_path_using_dfs([[1, 1, 1], [1, 1,'s'], [1, -1, 'e']], 1, 2, 0)
+    None
+    >>> find_shortest_path_using)dfs([[1], [-1, 's', 1], [-1, -1, 'e', -1]], 2, 1, 0)
+    """
+    global current_shortest_path_distance
```
CaedenPH, line 36: Using `global` is strongly recommended against. Perhaps return `current_shortest_path_distance`

path_tracing/matrix_shortest_path_using_depth_first_search.py
```diff
@@ -0,0 +1,108 @@
...
+) -> None:
+    """
+    Examples:
+    >>> find_shortest_path_using_dfs([['s', 1, 1], [1, -1, 1], [-1, 1, 'e']], 0, 0, 0)
+    None
+    >>> find_shortest_path_using_dfs([[-1, 1], ['s', 1], [-1, 1], [1, 'e']], 1, 0, 0)
+    None
+    >>> find_shortest_path_using_dfs([[1, 1, 1], [1, 1,'s'], [1, -1, 'e']], 1, 2, 0)
+    None
+    >>> find_shortest_path_using)dfs([[1], [-1, 's', 1], [-1, -1, 'e', -1]], 2, 1, 0)
```
CaedenPH, line 34:
```suggestion
    >>> find_shortest_path_using_dfs([[1], [-1, 's', 1], [-1, -1, 'e', -1]], 2, 1, 0)
```

CaedenPH, line 34: There is no point of these doctests if they return `None`. All this doctest is really doing is checking for errors, which only needs to be done once.

path_tracing/matrix_shortest_path_using_depth_first_search.py
```diff
@@ -0,0 +1,108 @@
...
+OBSTACLE_POS = -1
+VALID_POS = 1
+START_POS = "s"
+END_POS = "e"
+VISITED_POS = None
+current_shortest_path_distance = math.inf
+
+
+def find_shortest_path_using_dfs(
+    matrix: list, row: int, col: int, distance: int
```
CaedenPH, line 24: `list` of what?

path_tracing/matrix_shortest_path_using_depth_first_search.py
```diff
@@ -0,0 +1,108 @@
...
+matrix = [
+    [START_POS, VALID_POS, OBSTACLE_POS],
+    [VALID_POS, VALID_POS, VALID_POS],
+    [END_POS, OBSTACLE_POS, VALID_POS],
+    [OBSTACLE_POS, OBSTACLE_POS, OBSTACLE_POS],
+]
+
+if __name__ == "__main__":
+    shortest_path_distance = helper(matrix)
+
```
CaedenPH, line 107:
```suggestion
```

path_tracing/matrix_shortest_path_using_depth_first_search.py
```diff
@@ -0,0 +1,108 @@
...
+
+matrix = [
+    [START_POS, VALID_POS, OBSTACLE_POS],
+    [VALID_POS, VALID_POS, VALID_POS],
+    [END_POS, OBSTACLE_POS, VALID_POS],
+    [OBSTACLE_POS, OBSTACLE_POS, OBSTACLE_POS],
+]
+
+if __name__ == "__main__":
```
CaedenPH, line 105:
```suggestion
if __name__ == "__main__":
    matrix = [
        [START_POS, VALID_POS, OBSTACLE_POS],
        [VALID_POS, VALID_POS, VALID_POS],
        [END_POS, OBSTACLE_POS, VALID_POS],
        [OBSTACLE_POS, OBSTACLE_POS, OBSTACLE_POS],
    ]

```
It is only being used inside the `if`, so that is where it will go

tianyizheng02 (commented):
1) Like @CaedenPH said, you shouldn't be using global variables, but it would take an unnecessary amount of work to refactor this file to not use them.

2) Why create a new directory `path_tracing`? We want to avoid creating new directories unless an algorithm truly doesn't fall under any of the existing directories.

3) I take issue with the premise of this PR. DFS is an inherently really poor choice of algorithm here because DFS explores as far as possible before backtracking, meaning it needs to traverse every possible path to find the shortest one. If you wish to open a new PR to implement this using BFS instead, I think it'd be much more appropriate.

4) I kind of feel that this file is more a specific _use case_ of an algorithm (namely DFS) rather than purely implementing an algorithm itself. In any case, I don't think this is the right algorithm for the job anyway.

PR #7757, opened 2022-10-27
Update maximum_subarray.py

pronoym99:
1. Rectify documentation to indicate the correct output: function doesn't return the subarray, but rather returns a sum.
2. Make the function more Pythonic and thereby avoid:
    - any unnecessary ```if``` statements.
    - usage of non - optimal ```in range(len())``` constructs.
4. Make function annotation generic i.e. can accept any sequence ```nums```.
5. Raise value error when the input sequence ```nums``` is empty.

### Describe your change:

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -1,20 +1,24 @@
-def max_subarray(nums: list[int]) -> int:
-    """
-    Returns the subarray with maximum sum
-    >>> max_subarray([1,2,3,4,-2])
+from collections.abc import Sequence
+
+
+def max_subarray_sum(nums: Sequence[int]) -> int:
+    """Returns the maximum possible sum amongst all non - empty subarrays
+
+    Raises:
+      ValueError: when nums is empty.
+
+    >>> max_subarray_sum([1,2,3,4,-2])
     10
-    >>> max_subarray([-2,1,-3,4,-1,2,1,-5,4])
+    >>> max_subarray_sum([-2,1,-3,4,-1,2,1,-5,4])
     6
     """
+    if not nums:
+        raise ValueError("Input sequence should not be empty")
 
     curr_max = ans = nums[0]
 
-    for i in range(1, len(nums)):
-        if curr_max >= 0:
-            curr_max = curr_max + nums[i]
-        else:
-            curr_max = nums[i]
-
+    for num in nums[1:]:
+        curr_max = max(curr_max + num, num)
         ans = max(curr_max, ans)
 
     return ans
@@ -23,4 +27,4 @@ def max_subarray(nums: list[int]) -> int:
 if __name__ == "__main__":
     n = int(input("Enter number of elements : ").strip())
     array = list(map(int, input("\nEnter the numbers : ").strip().split()))[:n]
-    print(max_subarray(array))
+    print(max_subarray_sum(array))
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -23,7 +23,7 @@ def max_subarray_sum(nums: Sequence[int]) -> int:
         curr_max = max(curr_max + num, num)
         ans = max(curr_max, ans)
 
-    return int(ans)
+    return ans
 
 
 if __name__ == "__main__":
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -1,4 +1,4 @@
-from collections.abc import Sequence
+from typing import Sequence
 
 
 def max_subarray_sum(nums: Sequence[int]) -> int:
--- a/other/maximum_subarray.py
+++ b/other/maximum_subarray.py
@@ -1,4 +1,4 @@
-from typing import Sequence
+from collections.abc import Sequence
 
 
 def max_subarray_sum(nums: Sequence[int]) -> int:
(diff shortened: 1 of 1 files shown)


pronoym99:
I would suggest you add a hacktoberfest label to this PR.

cclauss:
maximum_subarray solves one problem and maximum_subarray_sum solves a different one.  It sounds like a bad idea to me to destroy an existing algorithm in order to solve a different problem.

pronoym99:
I think you are mistaken: `maximum_subarray` was **incorrectly documented** as returning an array as opposed to a sum. I have rectified the same. It was **still** solving the `maximum_subarray_sum` problem **only**. I haven't made any changes in the actual logic as a result. If you would like to have a completely different file for the same, I shall do so.

other/maximum_subarray.py
```diff
@@ -1,20 +1,24 @@
-def max_subarray(nums: list[int]) -> int:
-    """
-    Returns the subarray with maximum sum
-    >>> max_subarray([1,2,3,4,-2])
+from collections.abc import Sequence
+
+
+def max_subarray_sum(nums: Sequence[int]) -> int:
+    """Returns the maximum possible sum amongst all non - empty subarrays
```
Cjkjvfnby, line 5: Could you please add a period to the end and change `returns` to `return`.

[pep](https://peps.python.org/pep-0257/)
>  The docstring is a phrase ending in a period.  It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

other/maximum_subarray.py
```diff
@@ -1,20 +1,24 @@
...
 
     curr_max = ans = nums[0]
 
-    for i in range(1, len(nums)):
-        if curr_max >= 0:
-            curr_max = curr_max + nums[i]
-        else:
-            curr_max = nums[i]
-
+    for num in nums[1:]:
```
Cjkjvfnby, line 20: `nums[1:]`  creates a copy, so this is O(n) extra memory.  You could use https://docs.python.org/3/library/itertools.html#itertools.islice

other/maximum_subarray.py
CaedenPH: Good point - but in the future you could use the 
`suggestion` tag so that the user can easily commit the changes you are proposing

pronoym99:
1. Followed conventions as given in [pep-0257](https://peps.python.org/pep-0257/)
2. Initialised the ```current_maximum``` and ```ans``` to negative infinity and returned an answer cast to an integer. This makes the code constant in memory and avoids using an additional module i.e. ```itertools``` for such a simple implementation.

other/maximum_subarray.py
pronoym99: Sure. Would take care next time.

Cjkjvfnby:
Personally, I don't like using inf.  It adds a bit of noise with float/int types.  Maybe get back to the simple range? 

```python
    for i in range(1, len(nums)):
        num = nums[i]
```

(not a blocker for merge)

other/maximum_subarray.py
```diff
@@ -1,26 +1,30 @@
-def max_subarray(nums: list[int]) -> int:
-    """
-    Returns the subarray with maximum sum
-    >>> max_subarray([1,2,3,4,-2])
+from collections.abc import Sequence
```
Cjkjvfnby, line 1:
```suggestion
from typing import Sequence
```

Before Python 3.8 (or 3.9) this would be an exception.

pronoym99:
> Personally, I don't like using inf. It adds a bit of noise with float/int types. Maybe get back to the simple range?
> 
> ```python
>     for i in range(1, len(nums)):
>         num = nums[i]
> ```
> 
> (not a blocker for merge)

True that even I don't like the additional noise and what I did seems like _patchwork_. So have done the requested change. Do note that this change would lead to a constant access cost for every ```nums[i]```.

other/maximum_subarray.py
pronoym99: I have resolved this in the latest set of commits.

other/maximum_subarray.py
pronoym99: Committing this suggestion is creating an issue whilst running the test associated with ```pyupgrade```.

Cjkjvfnby: Looks like after they got the support of [] in Python 3.9, typing became the less favorable place to import them.

Cjkjvfnby:
Thank you. 

Looks like we could remove int() now and after it's good for merge.

other/maximum_subarray.py
pronoym99: Yes. Now, the pre-commit hooks following the failed ```pyupgrade``` test are designed in a way that it makes the suggestion null and void.

pronoym99:
Now we are good to go for the merge. Don't forget to add the hacktoberfest label to this PR  :^)

Cjkjvfnby (approved):

pronoym99:
Can anyone of you @Cjkjvfnby, @CaedenPH or @cclauss finally merge this PR into the main branch ?

cclauss (approved):

PR #7764, opened 2022-10-27
Change shebang of validation script to python, because python3 is not installed on Windows

Cjkjvfnby:
### Describe your change:

When I run `pre-commit run --all files` I got an error. 

```
Validate filenames.......................................................Failed
- hook id: validate-filenames
- exit code: 9009
```

This error happens because the python3 executable is not found. 

I have a Windows machine and just installed the latest Python via the official installer. Python3 alias is not included.

It was a time when using python2 and python3 aliases was important, but now python2 is not used anymore so it should not be a problem.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/scripts/validate_filenames.py
+++ b/scripts/validate_filenames.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 import os
 
 try:


cclauss:
Does Windows use shebang files?!?

On my Mac... `python --version`
```
zsh: command not found: python
```
This can be fixed by adding a venv but that is the default behavior.

Cjkjvfnby:
> Does Windows use shebang files?!?

It uses the associated program by an extension.  When Python is installed, `py` is associated with it.   
`py` is a switcher between installed Pythons, and it respects the shebang.  https://docs.python.org/3/using/windows.html#python-launcher-for-windows

> On my Mac... python --version

On my Mac I use pyenv, I have both python and python3.  But not sure that my installation could be considered typical.

Docker image python:3.9.10 also supports `python` 


I'll check if I could fix it on my Win machine with py.ini.  If it will work I'll post the solution here and close the PR.  Other users will be able to find this problem by searching.

cclauss:
Funny because, I use `py` on my Mac and earlier today I opened and then closed https://github.com/brettcannon/python-launcher/issues/235

Cjkjvfnby:
In my case, it's `py` from cPython.   According to docs and my experiments `python3` and `/usr/bin/env python` are supported and `/usr/bin/env python3` is not.  I Will resubmit the issue there. 

I haven't found any workaround yet.

PR #7765, opened 2022-10-27
Update breadth_first_search_2.py

Cjkjvfnby:
### Describe your change:
- Change Queue to deque. https://docs.python.org/3.11/library/queue.html?highlight=queue#module-queue
  > [collections.deque](https://docs.python.org/3.11/library/collections.html#collections.deque) is an alternative implementation of unbounded queues with fast atomic [append()](https://docs.python.org/3.11/library/collections.html#collections.deque.append) and [popleft()](https://docs.python.org/3.11/library/collections.html#collections.deque.popleft) operations that do not require locking and also support indexing. 
- Change return type from `set` to `list`.  Order of traversal is important for BFS, set is not good data structure to represent it. 
- Changed test to test order of returned values, old test just checks that we iterate over all nodes, even DFS traversal could pass it.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/graphs/breadth_first_search_2.py
+++ b/graphs/breadth_first_search_2.py
@@ -30,7 +30,7 @@ G = {
 
 def breadth_first_search(graph: dict, start: str) -> list[str]:
     """
-    Implementation of BFS using queue.Queue.
+    Implementation of breadth first search using queue.Queue.
 
     >>> ''.join(breadth_first_search(G, 'A'))
     'ABCDEF'
@@ -51,7 +51,7 @@ def breadth_first_search(graph: dict, start: str) -> list[str]:
 
 def breadth_first_search_with_deque(graph: dict, start: str) -> list[str]:
     """
-    Implementation of BFS using collection.queue.
+    Implementation of breadth first search using collection.queue.
 
     >>> ''.join(breadth_first_search_with_deque(G, 'A'))
     'ABCDEF'
--- a/graphs/breadth_first_search_2.py
+++ b/graphs/breadth_first_search_2.py
@@ -30,6 +30,8 @@ G = {
 
 def breadth_first_search(graph: dict, start: str) -> list[str]:
     """
+    Implementation of BFS using queue.Queue.
+
     >>> ''.join(breadth_first_search(G, 'A'))
     'ABCDEF'
     """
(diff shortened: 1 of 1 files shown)


cclauss:
Let's keep both the existing function and the new function in the same file and let's be sure that they are both tested with the same doctests so we can compare them side-by-side.

Let's also [add a timeit benchmark](https://github.com/TheAlgorithms/Python/blob/master/bit_manipulation/count_number_of_one_bits.py#L63) so visitors can understand the performance differences between the two implementations.  This repo has lots of timeit benchmarks... https://github.com/TheAlgorithms/Python/search?q=timeit

Cjkjvfnby:
> keep both the existing function and the new function in the same file and let's be sure that they are both tested with the same tests and test data so visitors can compare them side-by-side.

I still need to modify the existing function, because it returns traversed elements in the unordered collection.

Both functions use exactly the same approach, just different data structure that implements a similar API.

The not thread-safe deque beats thread-safe Queue in a single-threaded environment.
```
breadth_first_search                finished 10000 runs in 0.22710 seconds
breadth_first_search_with_deque     finished 10000 runs in 0.01708 seconds
```

cclauss:
> I still need to modify the existing function, because it returns traversed elements in the unordered collection.

I do not understand this.  Why should the new function force the old function to change?  I would think that traversing elements in an unordered collection was the original problem.

The benchmark looks awesome.  Please put the benchmark results in the comments so that the reader has a sense of with is faster and by how much.

cclauss:
Maintaining the breadth-first order is vital but the wiki definition says nothing about eliminating duplicate values.

Having only a single implementation is not an option.

Cjkjvfnby:
> Maintaining the breadth-first order is vital but the wiki definition says nothing about eliminating duplicate values.

Yes, because these algorithm does not suppose to have any duplication during traversal.  Using set as a return value does not allow testing this. 

> Having only a single implementation is not an option.

Do not worry about it, there are at least 5 files related to it in the repository.

These two function does not bring any diversity to the implementation, they are pretty much the same.  I could keep both implementations, but using Queue here is a bad example.

cclauss (approved):

PR #7811, opened 2022-10-28
Create maximum_subsequence.py

pronoym99:
### Describe your change:
Add an algorithm for the ```maximum_subsequence_sum``` in an array inline with the existing ```maximum_subarray``` in the [other](https://github.com/TheAlgorithms/Python/tree/master/other) directory.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] This pull request is all my work -- I have not plagiarized.


other/maximum_subsequence.py
```diff
@@ -0,0 +1,31 @@
...
+    Raises:
+      ValueError: when nums is empty.
+
+    >>> max_subsequence_sum([1,2,3,4,-2])
+    10
+    >>> max_subsequence_sum([-2, -3, -1, -4, -6])
+    -1
+    """
+    if not nums:
+        raise ValueError("Input sequence should not be empty")
```
CaedenPH, line 16: This is kind of implied, but if you want to have this check, create a doctest that fails with it
ie 
```py
>>> max_subsequence_sum()
    ....
ValueError: Input sequence should not be empty
```

other/maximum_subsequence.py
```diff
@@ -0,0 +1,31 @@
...
+
+    ans = nums[0]
+    nums_len = len(nums)
+
+    for i in range(1, nums_len):
```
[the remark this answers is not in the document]
pronoym99: ```len(nums)``` is computed for every ```i``` while we know that the length will remain constant throughout. Saving it in a variable speeds up the loop by roughly $33$ percent in large-sized ```nums```.

other/maximum_subsequence.py
pronoym99: Ok, I will do so.

pronoym99:
@CaedenPH Please have a look at the updated code. It is pending for a long time. I have also added appropriate doctests as requested.

ChrisO345 (approved):

PR #7821, opened 2022-10-28
Format docs

Cjkjvfnby:
### Describe your change:
Fix some warnings that would be raised by adding adding `flake8-docstrings` to the pre-commit hooks.

This PR is just to show the features of that tool.

There are two groups of checks.

#### Documentation
This looks like a good match with the repo guidelines. For classes and functions, there is an easy trick to use the leading underscore for names, then classes/functions won't be public, so you could document only what you need.

- D100 Missing docstring in public module
- D101 Missing docstring in public class
- D104 Missing docstring in public package

#### Docstring formating
It might be a bit confusing for people since the summary does not have a lot of sense in the context of this repo.  It will have the same words as in the file and function names and algorithm description is in the module docstring. Ignoring that line is not an option since there is a lot of code with doctests. 

- D205 1 blank line required between summary line and description
- D400 First line should end with a period
- D401 First line should be in imperative mood; try rephrasing
- D403 First word of the first line should be properly capitalized
 

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:


physics/kinetic_energy.py
```diff
@@ -1,5 +1,6 @@
 """
-Find the kinetic energy of an object, give its mass and velocity
+Find the kinetic energy of an object, give its mass and velocity.
```
CaedenPH, line 2:
```suggestion
Find the kinetic energy of an object, given its mass and velocity.
```

sorts/merge_sort.py
```diff
@@ -10,7 +11,9 @@
 
 
 def merge_sort(collection: list) -> list:
-    """Pure implementation of the merge sort algorithm in Python
+    """
+    Pure implementation of the merge sort algorithm in Python.
+
```
CaedenPH, line 16: Or maybe this just doesn't need to be said twice
```suggestion
```

CaedenPH:
Perhaps what you could normalize would be all the different types of docstring params/returns, ie 
```py
"""
:param a: The first integer
:param b: The second integer
:return: The output
"""
````
and 
```py
"""
Parameters
------------
a: :class:`int`
    The first integer
b: :class:`int`
    The second integer

Returns
--------
:class:`int`
    The output
```
and 
```py
@params 
@returns 
```
etc...
However maybe it's just better to leave it to persona preference

Cjkjvfnby (commented):
...

Cjkjvfnby:
>  However maybe it's just better to leave it to persona preference

Since we don't build documentation for that code we could leave it as is.

80% of these params/returns could be eliminated by annotations and good variable names.

cclauss (approved):

cclauss:
These all seems fine except maybe the one word sentence.  ;-)

I think we ___should___ start to auto-generate docs using `Sphinx` plus `sphinx-autodoc-typehints` which I have [used on other projects](https://archive.org/developers/internetarchive/api.html#archivesession-objects) with great success.  `sphinx-autodoc-typehints` does a great job of turning type hints into docs.

https://github.com/jjjake/internetarchive/pull/524#issuecomment-1123713962

PR #7831, opened 2022-10-29
Create collision_between_rectangles.py

0shuvo0:
### Describe your change:
Added algorithm to check collision between two rectangles

[x] Add an algorithm

### DIFF
--- a/maths/collision_between_rectangles.py
+++ b/maths/collision_between_rectangles.py
@@ -1,0 +1,29 @@
+rect1 = {
+    "x": 10,
+    "y": 10,
+    "height": 30,
+    "width": 50
+}
+
+rect2 = {
+    "x": 20,
+    "y": 30,
+    "height": 40,
+    "width": 30
+}
+
+def check_collision(rect1, rect2) -> bool:
+    """
+    Check if two rectangle are colliding/overlaping
+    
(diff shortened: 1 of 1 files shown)


maths/collision_between_rectangles.py
```diff
@@ -0,0 +1,27 @@
+rect1 = {"x": 10, "y": 10, "height": 30, "width": 50}
+
+rect2 = {"x": 20, "y": 30, "height": 40, "width": 30}
```
CaedenPH, line 3: These are not used within the actual code, so it should go within the `if __name__ == "__main__"` as test variables

maths/collision_between_rectangles.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def check_collision(rect1: dict, rect2: dict) -> bool:
+    """
+    Check if two rectangle are colliding/overlapping
+
+    >>> check_collision(rect1, rect2)
```
CaedenPH, line 10: Test variables should go inside the doctest, not outside

tianyizheng02 (changes requested):
Why dicts for the rectangles? Why not a class? It'd make creating and using the rectangles a lot easier. This feels like it should be part of a much larger shape library than just a single standalone function.

PR #6782, opened 2022-10-06
Created equivalent_resistance under Electronics

Gmuslow:
### Describe your change:
Hi all, 
I added the equivalent_resistance package to the electronics folder!
This is my first time contributing to open source so please let me know if you have any suggestions!

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [x] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #7838, opened 2022-10-29
Create calculate_resistence_in_parallel.py

0shuvo0:
### Describe your change:
Add an algorithm to calculate the total resistance of resistors used in parallel

[x] Add an algorithm

### DIFF
--- a/electronics/calculate_resistence_in_parallel.py
+++ b/electronics/calculate_resistence_in_parallel.py
@@ -3,7 +3,7 @@
 def calculate_resistence_in_parallel(resistors_in_parallel: list) -> float:
     """
     Calculate total resistance of resistors used in parallel
-    >>> calculate_resistence_in_parallel(resistors_in_parallel)
+    >>> calculate_resistence_in_parallel([200, 470, 220])
     85.67
     """
     total = 0
--- a/electronics/calculate_resistence_in_parallel.py
+++ b/electronics/calculate_resistence_in_parallel.py
@@ -1,0 +1,17 @@
+resistors_in_parallel = [200, 470, 220]
+
+def calculate_resistence_in_parallel(resistors_in_parallel: list) -> float:
+    """
+    Calculate total resistance of resistors used in parallel
+    >>> calculate_resistence_in_parallel(resistors_in_parallel)
+    85.67
+    """
+    total = 0
+    for val in resistors_in_parallel:
+        total = total + (1 / val)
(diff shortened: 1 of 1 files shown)


electronics/calculate_resistence_in_parallel.py
```diff
@@ -0,0 +1,19 @@
+resistors_in_parallel = [200, 470, 220]
```
CaedenPH, line 1: Because this is a test variable, put it inside the docstring
```suggestion
```

electronics/calculate_resistence_in_parallel.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def calculate_resistence_in_parallel(resistors_in_parallel: list) -> float:
+    """
+    Calculate total resistance of resistors used in parallel
+    >>> calculate_resistence_in_parallel(resistors_in_parallel)
```
CaedenPH, line 7:
```suggestion
    >>> calculate_resistence_in_parallel([200, 470, 220])
```

electronics/calculate_resistence_in_parallel.py
```diff
@@ -0,0 +1,19 @@
...
+    85.67
+    """
+    total = 0
+    for val in resistors_in_parallel:
+        total = total + (1 / val)
+    return round(1 / total, 2)
```
CaedenPH, line 13:
```suggestion
    return 1 / total
```
No point rounding the output

electronics/calculate_resistence_in_parallel.py
```diff
@@ -0,0 +1,19 @@
...
+
+def calculate_resistence_in_parallel(resistors_in_parallel: list) -> float:
+    """
+    Calculate total resistance of resistors used in parallel
+    >>> calculate_resistence_in_parallel(resistors_in_parallel)
+    85.67
+    """
+    total = 0
+    for val in resistors_in_parallel:
+        total = total + (1 / val)
```
CaedenPH, line 12:
```suggestion
    total = sum(1 / val for val in resistors_in_parallel)
```

electronics/calculate_resistence_in_parallel.py
0shuvo0: if we don't round it we will get crazy result like 85.66694283347141

CaedenPH:
@cclauss #6782 implemented this

PR #7874, opened 2022-10-30
Updated `url`, `identifier`, etc...

hakiKhuva:
- added user agent because sometimes yahoo finance does not sent stock prices
- `in` subdomain URL not working updated that
- using `content` property instead of `text`
- changed identifier from `class` to `dataset` and removed specific tag
- added `ZOMATO.NS` stock to check for Indian stocks price

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/web_programming/current_stock_price.py
+++ b/web_programming/current_stock_price.py
@@ -1,20 +1,26 @@
 import requests
 from bs4 import BeautifulSoup
+from fake_useragent import UserAgent
 
-USER_AGENT = (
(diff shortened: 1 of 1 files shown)


CaedenPH (changes requested):
Instead of setting the `USER_AGENT`, you an use the `fake_useragent` module in the `requirementrs.txt`
https://pypi.org/project/fake-useragent/

hakiKhuva:
> Instead of setting the `USER_AGENT`, you an use the `fake_useragent` module in the `requirementrs.txt` https://pypi.org/project/fake-useragent/

Instead using module user can easily copy and paste user agent of the browser user is using.

PR #7901, opened 2022-10-31
Modified 'pascal_triangle.py' program

TechFreak107:
### Describe your change: Added a new optimized function 'generate_pascal_triangle_optimized' to generate Pascal's Triangle to the existing 'pascal_triangle.py' file which reduces the computations done for each row by half. It reduces the redundant computations in generating each row. Also, added type check functionalities, which raise 'ValueError' and 'TypeError', to the existing 'generate_pascal_triangle' function. Also, added 3 additional doctests for the existing 'generate_pascal_triangle' function'.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


cclauss:
Cool!  Please add a [timeit](https://docs.python.org/3/library/timeit.html#timeit.timeit) (or similar) **benchmark** to prove that the proposed code is faster than the original code.

There are lots of examples in this repo. https://github.com/TheAlgorithms/Python/search?q=timeit but https://github.com/TheAlgorithms/Python/blob/master/bit_manipulation/count_number_of_one_bits.py#L63 is probably one of the best.

other/pascal_triangle.py
```diff
@@ -90,6 +111,66 @@ def calculate_current_element(
...
+    result: list[list[int]] = [[1]]
+
+    for row_index in range(1, num_rows):
+        temp_row = [0] + result[-1] + [0]
+        row_length = row_index + 1
+
+        # It calculates the number of distinct elements in a row
+        distinct_elements = (
+            row_length // 2 if row_length % 2 == 0 else row_length // 2 + 1
+        )
```
cclauss, line 160: Leverage the [Python builtin functions](https://docs.python.org/3/library/functions.html):
```suggestion
        distinct_elements = sum(divmod(row_length, 2))
```

other/pascal_triangle.py
```diff
@@ -90,6 +111,66 @@ def calculate_current_element(
...
+
+        row_first_half = [
+            temp_row[x - 1] + temp_row[x] for x in range(1, distinct_elements + 1)
+        ]
```
cclauss, line 164:
```suggestion
        row_first_half = [item + temp_row[i + 1] for i, item in enumerate(temp_row)]
```

cclauss:
The benchmarks show that sometimes the _optimized_ version is slower and other times it is faster.

PR #7105, opened 2022-10-13
Add flake8-builtins to pre-commit and fix errors

CaedenPH:
### Describe your change:
Implements #7104

* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [x] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

PR #7949, opened 2022-11-02
Flake8: Drop ignore of issue A003

cclauss:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


CaedenPH (changes requested):
Personally, I prefer attributes such as `next` and `prev` to `next_` and `prev_`
Especially when it comes to `set` and `put`, because they are different concepts.
I don't see a problem with globally ignoring `A003`

poyea (approved):

dhruvmanila:
I agree with @CaedenPH on this and was [discussed earlier](https://github.com/TheAlgorithms/Python/pull/7105) as well. `A003` is a false positive and is supposedly going to be disabled by default in the plugin itself. I suggest to revert this change.

cclauss:
@gilforcada Your sense of this pull request?  Does `A003` generate false positives?  Should we revert?

dhruvmanila:
@cclauss there's no such user

gforcada:
If my opinion is asked, here is is given 😄 : I very much prefer to keep `A003` enabled on the plugin itself, and let each user decide if it's useful or not. For my projects I very much prefer not to use builtins even on `A003` cases, that my personal preference. Fortunately `flake8` has a powerful configuration system and you probably are already configuring some bits of it anyway. I will not get offended if someone ignores `A003` 😃

PR #8029, opened 2022-12-14
Add autoclave cipher

VictorRS27:
### Describe your change:

I'm adding the autoclave or autokey cipher.
Is similar to the vigenerè cipher, but using the own plaintext as the key.
The code implements an encrypt and a decrypt function.
For more info:
https://en.wikipedia.org/wiki/Autokey_cipher

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


ciphers/autoclave.py
```diff
@@ -0,0 +1,111 @@
...
+    Function that encrypt a given plaintext (string)
+    and key (string), returning the encrypted ciphertext
+
+    @params
+    plaintext - a normal text to be encrypted (string)
+    key - a small text or word to start the replacing (sFtring)
+
+    @return
+    A string with the ciphertext
+
```
CaedenPH, line 12:
```suggestion
```
Typically, we want the parameter names to be self-documenting so this isn't really necessary

ciphers/autoclave.py
```diff
@@ -0,0 +1,111 @@
...
+    'jsqqs avvwo'
+    """
+    if type(plaintext) != str:
+        raise TypeError("plaintext must be a string")
+    if type(key) != str:
+        raise TypeError("key must be a string")
+
+    if plaintext == "":
+        raise ValueError("plaintext is empty")
+    if key == "":
```
CaedenPH, line 23:
```suggestion
    if not plaintext:
        raise ValueError("plaintext is empty")
    if not key:
```

ciphers/autoclave.py
```diff
@@ -0,0 +1,111 @@
...
+        raise ValueError("key is empty")
+
+    key = key + plaintext
```
CaedenPH, line 26:
```suggestion
    key += plaintext
```

ciphers/autoclave.py
```diff
@@ -0,0 +1,111 @@
+def encrypt(plaintext, key) -> str:
```
CaedenPH, line 1:
```suggestion
def encrypt(plaintext, key) -> str:
```
These parameters need to be typehinted

ciphers/autoclave.py
```diff
@@ -0,0 +1,111 @@
...
+operation = int(input("Type 1 to encrypt or 2 to decrypt:"))
+if operation == 1:
+    plaintext = str(input("Typeplaintext to be encrypted:\n"))
+    key = str(input("Type the key:\n"))
+    print(encrypt(plaintext, key))
+elif operation == 2:
+    ciphertext = str(input("Type the ciphertext to be decrypted:\n"))
+    key = str(input("Type the key:\n"))
+    print(decrypt(ciphertext, key))
+decrypt("jsqqs avvwo", "coffee")
```
CaedenPH, line 111: please move this into an `if __name__ == "__main__":` block

CaedenPH (changes requested):
Please link some documentation at the start of the file and include more doctests within each function

ciphers/autoclave.py
cclauss, line 1: Clave is Latin for key so the more popular name for this algorithm is autokey.
```suggestion
”””
https://en.wikipedia.org/wiki/Autokey_cipher
> An autokey cipher (also known as the autoclave cipher) is a cipher that incorporates the message (the plaintext into the key. The key is generated from the message in some automated fashion, sometimes by selecting certain letters from the text or, more commonly, by adding a short primer key to the front of the message.
”””


def encrypt(plaintext, key) -> str:
```

ciphers/autoclave.py
```diff
@@ -0,0 +1,111 @@
...
+    plaintext - a normal text to be encrypted (string)
+    key - a small text or word to start the replacing (sFtring)
+
+    @return
+    A string with the ciphertext
+
+    >>> encrypt("hello world", "coffee")
+    'jsqqs avvwo'
+    """
+    if type(plaintext) != str:
```
cclauss, line 16: PEP8: Use `isinstance(plaintext, str)` instead of directly comparing types.

cclauss (approved):

cclauss:
@priya-sundaram-dev Several years ago, when I merged this pull request, I disabled five unrelated algorithms.  I have no idea why I disabled them, but perhaps they were slowing down our CI tests.  Can you please create a pull request that re-enables them and also re-enables neural_network/perceptron.py.DISABLED, as discussed in:
* #15206

priya-sundaram-dev:
Done — opened #15208. It re-enables the four ML examples plus `neural_network/perceptron.py`.

The reason they wouldn't come back as-is: they'd bit-rotted against scikit-learn. `plot_confusion_matrix` and the Boston dataset were both removed in scikit-learn 1.2, so `--doctest-modules` couldn't even import the files. I swapped in `ConfusionMatrixDisplay.from_estimator` and `load_diabetes`, dropped the artificial `time.sleep()`s, and for the perceptron I seeded a local RNG + capped the training loop so it can't spin forever on non-separable data. All doctests + ruff pass locally.

PR #8050, opened 2022-12-27
Create brightness_augmentation.py

Erfaniaa:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/computer_vision/brightness_augmentation.py
+++ b/computer_vision/brightness_augmentation.py
@@ -19,7 +19,6 @@ def main() -> None:
     Save images in output dir.
     """
     img_paths = get_dataset(IMAGE_DIR)
-    print("Processing...")
     new_images, paths = update_images(img_paths, BRIGHTNESS_PERCENT)
 
     for index, image in enumerate(new_images):
--- a/computer_vision/brightness_augmentation.py
+++ b/computer_vision/brightness_augmentation.py
@@ -6,7 +6,6 @@ from string import ascii_lowercase, digits
 import cv2
 import numpy as np
 
-
 # Params
 IMAGE_DIR = ""
 OUTPUT_DIR = ""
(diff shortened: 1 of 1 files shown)


computer_vision/brightness_augmentation.py
```diff
@@ -0,0 +1,88 @@
...
+    32
+    """
+    assert number_char > 1, "The number of character should greater than 1"
+    letter_code = ascii_lowercase + digits
+    return "".join(random.choice(letter_code) for _ in range(number_char))
+
+
+if __name__ == "__main__":
+    main()
+    print("DONE ✅")
```
realstealthninja, line 88:
```suggestion
```
Unnecessary print statement

computer_vision/brightness_augmentation.py
```diff
@@ -0,0 +1,88 @@
...
+import random
+from string import ascii_lowercase, digits
+
+import cv2
+import numpy as np
+
+# Params
+IMAGE_DIR = ""
+OUTPUT_DIR = ""
+BRIGHTNESS_PERCENT = 50  # (between 0 and 100)
```
realstealthninja, line 12: Wouldn't it be better to add these directly to the functions?

computer_vision/brightness_augmentation.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            hsv[:, :, 2] = np.where(hsv[:, :, 2] < value, 0, hsv[:, :, 2] - value)
+        hsv[hsv > 255] = 255
+        hsv[hsv < 0] = 0
+        img = cv2.cvtColor(hsv, cv2.COLOR_HSV2BGR)
+        new_imgs_list.append(img)
+    return new_imgs_list, path_list
+
+
+def random_chars(number_char: int = 32) -> str:
```
realstealthninja, line 74: Requires cases where number of characters are negative and 0. Function seems pointless as it is used just for creating a temporary name for a file

computer_vision/brightness_augmentation.py
```diff
@@ -0,0 +1,88 @@
+import glob
```
realstealthninja, line 1: Requires links to sources

realstealthninja (changes requested):

computer_vision/brightness_augmentation.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def main() -> None:
+    """
+    Get images list from input dir.
+    Update new images.
+    Save images in output dir.
+    """
+    img_paths = get_dataset(IMAGE_DIR)
+    print("Processing...")
```
realstealthninja, line 22:
```suggestion
```
Unnecessary print statement

realstealthninja (changes requested):

computer_vision/brightness_augmentation.py
Erfaniaa: I think this way is easier for testing by the user.

computer_vision/brightness_augmentation.py
Erfaniaa: It's just used for creating a temporary name for the files. I think it's a good way because if you run the script several times, it will generate many different images with new unique names, so you would have a larger dataset. The new images are generated randomly and they are different after each run.

computer_vision/brightness_augmentation.py
Erfaniaa: Would you please explain more?

computer_vision/brightness_augmentation.py
realstealthninja: ah i see.

computer_vision/brightness_augmentation.py
realstealthninja: wouldn't it be better for the user to just reference the function. as most of the algorithms in this repository seems to a way for users to incorporate algorithms into their projects

PR #8065, opened 2023-01-01
Revamp `md5.py`

tianyizheng02:
### Describe your change:

Revamped `hashes/md5.py`:
- Added type hints to all functions
- Added doctests to all functions
- Greatly expanded documentation for all functions
- Added input validation to some functions as needed
- Added reference to the Wikipedia article that the implementation is based on
- Renamed some variables and functions to clarify their usage
- Minor refactoring for clarity

There's a lot in this PR, so let me know if anything else should be changed or if any existing changes should be reverted

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:


hashes/md5.py
```diff
@@ -1,127 +1,329 @@
...
+been cracked, so it shouldn't be used for security purposes.
 
-def rearrange(bit_string_32):
-    """[summary]
-    Regroups the given binary string.
+For more info, see https://en.wikipedia.org/wiki/MD5
+"""
+
+from collections.abc import Generator
+from math import sin
```
cclauss, line 15: Please use https://docs.python.org/3/library/struct.html to do the endian conversations or at least use `struct` in the doctests for  `to_little_endian()`.

tianyizheng02: Admittedly I'm not that familiar with the `struct` module, but I'm not sure how to make it work with the `to_little_endian()` function.

To me, the problem is that the original function `rearrange()` doesn't _actually_ convert strings to little-endian. Instead, it treats 32-char string inputs as if they were _32-bit bit strings_, with each char as a single "bit". It then restructures the input in an "little-endian fashion": the 8 least significant "bits" come first, followed by the 8 next least significant "bits", etc. Thus it looks little-endian if you squint hard enough.

Since the inputs to `rearrange()`/`to_little_endian()` are being restructured in units far larger than a byte, I'm not sure if the `struct` module would work here, unless I'm misunderstanding how the module works.

cclauss:
The ruff errors should go away if you rebase.

cclauss:
For me, there are two things that I see missing...
1. Tests comparing against [`hashlib.md5(msg).hexdigest()`](https://docs.python.org/3/library/hashlib.html#hashlib.hash.hexdigest)
2. Hashes are usually calculated from messages that are `bytes`, not `str` so `md5_me()` should take `bytes`, not `str`.
```
    >>> import hashlib
    >>> from strings import ascii_letters
    >>> msgs = (b"", ascii_letters.encode("utf-8"), "Üñîçø∂é".encode("utf-8"),
    ...         b"The quick brown fox jumps over the lazy dog.")
    >>> all(md5_me(bytes_msg) == hashlib.md5(bytes_msg).hexdigest() for bytes_msg in msgs)
    True
```

hashes/md5.py
```diff
@@ -111,26 +111,27 @@ def preprocess(message: str) -> str:
     Returns:
         processed bit string padded to a multiple of 512 chars
 
-    >>> preprocess("a") == "01100001" + "1" + ("0" * 439) + "00001000" + ("0" * 56)
+    >>> preprocess(b"a") == b"01100001" + b"1" + (b"0" * 439) + b"00001000" + \
```
cclauss, line 114: Please use () instead of \ -- See PEP8

cclauss (approved):
Nice!!

PR #8102, opened 2023-01-26
Update permutations.py

sundaram2021:
Earlier in the permutation function it is creating an infinite recursion which is causing an error 

instead of that I used backtracking to get the output

*[x]  Fixed a bug or typo in an existing algorithm in permutations.py?

### DIFF
--- a/data_structures/arrays/permutations.py
+++ b/data_structures/arrays/permutations.py
@@ -4,12 +4,27 @@ from typing import List
 def permute(nums: list[int]) -> list[list[int]]:
     """
     Return all permutations.
-
     >>> from itertools import permutations
     >>> numbers= [1,2,3]
     >>> all(list(nums) in permute(numbers) for nums in permutations(numbers))
     True
     """
+    result = []
+    if len(nums) == 1:
+        return [nums.copy()]
+    for _ in range(len(nums)):
+        n = nums.pop(0)
+        permutations = permute(nums)
+        for perm in permutations:
+            perm.append(n)
+        result.extend(permutations)
+        nums.append(n)
+    return result
+
+
+
+
+def permute2(nums: List[int]) -> List[List[int]]:
 
     def backtrack(first=0):
         if first == n:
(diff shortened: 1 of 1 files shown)


CaedenPH (changes requested):
Instead of deleting the old code, perhaps create an alternate function using the backtracking method, while keeping the original function.

sundaram2021:
I have created a new function named as permute2 with the backtracking method

data_structures/arrays/permutations.py
```diff
@@ -1,7 +1,9 @@
+from typing import List
+
```
rohan472000, line 2: remove `from typing import List` as it is unused , creating pre-commit error.

rohan472000 (changes requested):

data_structures/arrays/permutations.py
```diff
@@ -20,7 +19,27 @@ def permute(nums: list[int]) -> list[list[int]]:
     return result
 
 
+def permute2(nums: list[int]) -> list[list[int]]:
+    def backtrack(first=0):
```
ChrisO345, line 23:
```suggestion
	 """
    Return all permutations.
    >>> from itertools import permutations
    >>> numbers= [1,2,3]
    >>> all(list(nums) in permute2(numbers) for nums in permutations(numbers))
    True
    """
    def backtrack(first=0):
```

ChrisO345 (dismissed):

data_structures/arrays/permutations.py
sundaram2021: added that

data_structures/arrays/permutations.py
```diff
@@ -20,7 +19,32 @@ def permute(nums: list[int]) -> list[list[int]]:
     return result
 
 
+def permute2(nums):
+    """
+    Return all permutations of the given list.
+
+    >>> permute2([1, 2, 3])
+     [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 2, 1], [3, 1, 2]]
```
ChrisO345, line 27:
```suggestion
    [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 2, 1], [3, 1, 2]]
```

ChrisO345 (changes requested):
This needs to be inline

data_structures/arrays/permutations.py
sundaram2021: can i commit your suggestion

ChrisO345 (approved):

PR #8125, opened 2023-02-07
Solving the `Top k most frequent words` problem using a max-heap

aparibocci:
### Describe your change:
This PR aims to add an algorithm to identify the top `k` most frequent strings given a provided string list of elements.
To do this, the algorithm is using a max-heap implementation already existing in this repository (a generic type was introduced to allow the usage).

Time complexity is `O(n)`, where `n` is the number of words:
- `O(n)` for building the max-heap
- `k*O(logn)` for extracting the `k` most frequent strings

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/maths/pi_generator.py
+++ b/maths/pi_generator.py
@@ -1,0 +1,94 @@
+def calculate_pi(limit: int) -> str:
+    """
+    https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80
+    Leibniz Formula for Pi
+
+    The Leibniz formula is the special case arctan 1 = 1/4 Pi .
+    Leibniz's formula converges extremely slowly: it exhibits sublinear convergence.
+
+    Convergence (https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80#Convergence)
+
+    We cannot try to prove against an interrupted, uncompleted generation.
+    https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80#Unusual_behaviour
+    The errors can in fact be predicted;
+    but those calculations also approach infinity for accuracy.
+
+    Our output will always be a string since we can defintely store all digits in there.
+    For simplicity' sake, let's just compare against known values and since our outpit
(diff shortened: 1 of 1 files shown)


CaedenPH (approved):
Very well written algorithm, great stuff!

aparibocci:
Hello @cclauss, I'm tagging you as I saw your involvement in several PRs.
Do you know who can I ask to for a review? Is there anything I should still change in the PR?
Thank you!

cclauss:
All of this seems like it could be done in a few lines with only a `collections.Counter` so what are the advantages of this approach?

aparibocci:
> All of this seems like it could be done in just a few lines using only a `collections.Counter` so what are the advantages of this approach?

Thanks for the feedback.
I had a look to `collections.Counter` and noticed that, indeed, `most_common` solves exactly this problem: https://github.com/python/cpython/blob/3.11/Lib/collections/__init__.py#L608

This is using `heapq.nlargest` behind the scenes: https://github.com/python/cpython/blob/3.11/Lib/heapq.py#L523

There is already a `Heap` class in this repository, I imagined it would be useful to show a typical usage of heaps (i.e. finding out order statistics). Do you think I should close this PR entirely, or do you have other suggestions?

Thank you

cclauss:
```python
from collections import Counter
def top_k_frequent_words(words, k_value):
    return [x[0] for x in Counter(words).most_common(k_value)]
```

cclauss:
Please put the text of our last two messages in the file’s docstring and then we can merge this.   That will show why this work was done and that there also that the Python standard library provides a far more straightforward  way to solve the same problem.

aparibocci:
> Please put the text of our last two messages into the file’s docstring and then we can merge this. That will show why this work was done and that there also that the Python standard library provides a far more straightforward way to solve the same problem.

Thanks for the feedback.
I have updated the docstring mentioning the (preferable) Python standard library solution.

cclauss (approved):
Nice!

PR #8143, opened 2023-02-16
Bisection Method

agathipanagiotatou:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


arithmetic_analysis/bisection_method.py
```diff
@@ -0,0 +1,77 @@
+from sympy import diff
+from sympy.abc import x
+
```
rohan472000, line 3: I've checked your code and it seems to be free of syntax errors. However, I did notice that you have used the `eval()` function to convert the string function to an algebraic expression. While this approach works fine for simple expressions, it can be dangerous and lead to security vulnerabilities if the input string is not properly sanitized. It is generally recommended to use more secure methods like the `ast` module to parse and evaluate user input.

tianyizheng02 (commented):
Thanks for your contribution, but unfortunately I must close this PR for the following reasons:

1) I agree with @rohan472000: we should avoid using `eval` whenever possible.

2) We actually already have not one, but _two_ implementations of the bisection method: [maths/bisection.py](https://github.com/TheAlgorithms/Python/blob/f66568e981edf5e384fe28a357daee3e13f16de9/maths/bisection.py) and [arithmetic_analysis/bisection.py](https://github.com/TheAlgorithms/Python/blob/f66568e981edf5e384fe28a357daee3e13f16de9/arithmetic_analysis/bisection.py). Instead of contributing a new implementation, I recommend improving upon the existing implementations instead (or consolidating them into a single implementation).

PR #8154, opened 2023-02-21
Make some ruff fixes

cclauss:
### Describe your change:

`ruff --select=RET501,RET502,RET503,SIM101,SIM102,SIM108,SIM109,SIM110,SIM117,SIM118,SIM201,SIM300,PD002,RUF005 --fix .`

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/audio_filters/iir_filter.py
+++ b/audio_filters/iir_filter.py
@@ -47,7 +47,7 @@ class IIRFilter:
             >>> filt.set_coefficients(a_coeffs, b_coeffs)
         """
         if len(a_coeffs) < self.order:
-            a_coeffs = [1.0] + a_coeffs
+            a_coeffs = [1.0, *a_coeffs]
 
         if len(a_coeffs) != self.order + 1:
             raise ValueError(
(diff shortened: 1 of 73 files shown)


dhruvmanila (approved):
Looks good. There's just one thing I noticed which is that some places have explicit `return None` while others implicit `return`. I think we should adopt one or the other. What do you think?

cclauss:
Those rules cover different situations.  Sometimes `return None` should be added and other times `return None` should be removed.  So let's follow the modifications that the rules recommend.
* https://beta.ruff.rs/docs/rules/#flake8-return-ret

Code | Name | Description | Fix?
-- | -- | -- | --
RET501 | unnecessary-return-none | Do not explicitly return None in function if it is the only possible return value | 🛠
RET502 | implicit-return-value | Do not implicitly return None in function able to return non-None value | 🛠
RET503 | implicit-return | Missing explicit return at the end of function able to return non-None value | 🛠

dhruvmanila:
The rules exists, but it depends on the project, the maintainers and the audience through which the choice of rules need to be made. Just because a rule exist, doesn't mean it's a necessary thing to do.

This is what I think but it's upto you. If you think it's fine, then ok, but I think consistency matters.

Issue #5315, opened 2021-10-15
Singly_Linked_List: extra O(n) time complexity while getting length of the list


jssparas:
While getting the length of the linked list, code will take O(n) of unnecessary time as it executes - 
`len(tuple(iter(self)))`
This also takes up O(n) of extra space as it is creating a tuple first.

This can be easily avoided by maintaining an instance variable `len`, increment/decrement it while inserting/deleting elements.

PR #8183, opened 2023-03-15
change space complexity of linked list's __len__ from O(n) to O(1)

amirsoroush:
### Describe your change:

Following #5315 and #5320, I was convinced that it's better to calculate `__len__` each time on demand. But current implementation has "*space*" complexity problem when dealing with a huge linked list. That temporary `tuple` is unnecessary.  A one-line change using built-in `sum()` and a generator expression would solve it without breaking existing codes.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/linked_list/singly_linked_list.py
+++ b/data_structures/linked_list/singly_linked_list.py
@@ -72,7 +72,7 @@ class LinkedList:
         >>> len(linked_list)
         0
         """
-        return len(tuple(iter(self)))
+        return sum(1 for _ in self)
 
     def __repr__(self) -> str:
         """
--- a/data_structures/linked_list/circular_linked_list.py
+++ b/data_structures/linked_list/circular_linked_list.py
@@ -24,7 +24,7 @@ class CircularLinkedList:
                 break
 
     def __len__(self) -> int:
-        return len(tuple(iter(self)))
+        return sum(1 for _ in self)
 
     def __repr__(self):
         return "->".join(str(item) for item in iter(self))
--- a/data_structures/linked_list/doubly_linked_list.py
+++ b/data_structures/linked_list/doubly_linked_list.py
@@ -51,7 +51,7 @@ class DoublyLinkedList:
         >>> len(linked_list) == 5
         True
         """
-        return len(tuple(iter(self)))
+        return sum(1 for _ in self)
 
     def insert_at_head(self, data):
         self.insert_at_nth(0, data)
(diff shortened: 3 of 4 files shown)


data_structures/linked_list/singly_linked_list.py
```diff
@@ -72,7 +72,7 @@ def __len__(self) -> int:
         >>> len(linked_list)
         0
         """
-        return len(tuple(iter(self)))
+        return sum(1 for _ in self)
```
Cjkjvfnby, line 75: The sum will call iter(self) inside,  so this code is still O(n).

Cjkjvfnby: For the classical linked list, the best length is O(n).

PS. `collections.deque` is not a real linked list,  it's a wrapper around it, so it basically counts how many append/delete methods were called.    https://github.com/python/cpython/blob/main/Modules/_collectionsmodule.c#L196



https://wiki.python.org/moin/TimeComplexity

amirsoroush: Are you talking about "Time" complexity? This line would fix the "Space" complexity as it only fetches one item at a time and adds 1 to the final result as opposed to `tuple()` which loads all the items to the memory.

cclauss: Fun exercise... Go to https://pyodide.org/en/latest/console.html to get a current Python repl running on WASM.

Paste in the following code and hit return.
```
>>> from timeit import timeit
setup="from itertools import product; from string import ascii_letters"
timeit("sum(1 for _ in product(ascii_letters, repeat=4))", number=10, setup=setup)
timeit("len(tuple(product(ascii_letters, repeat=4)))", number=10, setup=setup)

5.0610000000000355
4.121999999999957
```
`sum()` is slower than `len()` for 7,311,616 items.
---
Refresh the webpage to clear out any clutter in memory...

Paste in the following code and hit return.
```
>>> from timeit import timeit
setup="from itertools import product; from string import ascii_letters"
timeit("sum(1 for _ in product(ascii_letters, repeat=5))", number=1, setup=setup)
timeit("len(tuple(product(ascii_letters, repeat=5)))", number=1, setup=setup)
```
26.686000000000035
  ...
MemoryError
```
`sum()` delivers an answer for 380,204,032 items but `len()` raises a `MemoryError`.

cclauss (approved):

PR #8545, opened 2023-03-22
Update in the workflow

muzakkirsaifi123:
### Describe your change:
* [x] More user interactive
* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] Change in one file.

### DIFF
--- a/scheduling/round_robin.py
+++ b/scheduling/round_robin.py
@@ -11,7 +11,6 @@ from shutil import get_terminal_size
 from threading import Thread
 
 
-
 class Loading:
     def __init__(self, desc="processing..", points="Done..", timeout=0.2):
         self.desc = desc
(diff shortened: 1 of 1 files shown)


muzakkirsaifi123:
I have made a change in the round-robin algorithm. I have added a loader to it. This means when a user run that algorithm then it will show a loading logo which makes the more interactive. Before this, It is quite strange when the user runs the code and nothing happen. But now the user will see a live-loading logo.

scheduling/round_robin.py
```diff
@@ -4,8 +4,76 @@
 https://en.wikipedia.org/wiki/Round-robin_scheduling
 """
 from __future__ import annotations
-
 from statistics import mean
+import time
+from itertools import cycle as cy
+from shutil import get_terminal_size as gts
+from threading import Thread as th
```
cclauss, line 11: Please do not replace standard names with cryptic names.
```suggestion
from itertools import cycle
from shutil import get_terminal_size
from threading import Thread

```

PR #8600, opened 2023-04-01
Remove extra `len` calls in doubly-linked-list's methods

amirsoroush:
### Describe your change:

I reduced the number of calls to `__len__` by storing the result into a local variable as it's currently a costly method( O(N) ).

To elaborate more: Suppose we insert an item at the tail like:
```python
l = DoublyLinkedList()
l.insert_at_tail("foo")
l.insert_at_tail("boo")   # <------- Here
```
The second call to `insert_at_tail` causes `__len__` to be called three times:
1. inside `self.insert_at_nth(len(self), data)` (because of delegating to `insert_at_nth`)
2. one at `if not 0 <= index <= len(self)` inside `insert_at_nth`
3. one at `elif index == len(self)` inside `insert_at_nth`

Same thing happens in `delete_at_nth`.

---

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/linked_list/doubly_linked_list.py
+++ b/data_structures/linked_list/doubly_linked_list.py
@@ -81,7 +81,9 @@ class DoublyLinkedList:
             ....
         IndexError: list index out of range
         """
-        if not 0 <= index <= len(self):
+        length = len(self)
+
+        if not 0 <= index <= length:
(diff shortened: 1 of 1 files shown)


dhruvmanila (approved):
Makes sense, thanks!

PR #8605, opened 2023-04-01
Reduce the complexity of linear_algebra/src/polynom_for_points.py

MaximSmolskiy:
### Describe your change:

Reduce the complexity of linear_algebra/src/polynom_for_points.py from 17 to 13

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/linear_algebra/src/polynom_for_points.py
+++ b/linear_algebra/src/polynom_for_points.py
@@ -43,29 +43,13 @@ def points_to_polynomial(coordinates: list[list[int]]) -> str:
 
     x = len(coordinates)
 
-    count_of_line = 0
-    matrix: list[list[float]] = []
     # put the x and x to the power values in a matrix
-    while count_of_line < x:
-        count_in_line = 0
-        a = coordinates[count_of_line][0]
-        count_line: list[float] = []
-        while count_in_line < x:
-            count_line.append(a ** (x - (count_in_line + 1)))
-            count_in_line += 1
-        matrix.append(count_line)
-        count_of_line += 1
+    matrix: list[list[float]] = [[coordinates[count_of_line][0] ** (x - (count_in_line + 1)) for count_in_line in range(x)] for count_of_line in range(x)]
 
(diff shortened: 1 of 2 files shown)


linear_algebra/src/polynom_for_points.py
```diff
@@ -43,29 +43,19 @@ def points_to_polynomial(coordinates: list[list[int]]) -> str:
...
-    while count_of_line < x:
-        vector.append(coordinates[count_of_line][1])
-        count_of_line += 1
+    vector: list[float] = [coordinates[count_of_line][1] for count_of_line in range(x)]
 
-    count = 0
-
-    while count < x:
+    for count in range(x):
         zahlen = 0
```
tianyizheng02, line 59: 1) Could we refactor this inner while-loop to be a for-loop too? It'd probably require changing the control flow slightly, but it should be fine as long as we preserve the program logic. As a starting point, it seems to me that the reason why it's not trivial to convert this while-loop to a for-loop is because of this if-statement changing the value of `zahlen` inside the loop:
    ```py
    if count == zahlen:
        zahlen += 1
    ```
    I think this condition is only there to ensure that `zahlen != count`. If my understanding is correct, then this would simply translate to a `continue` statement in a for-loop:
    ```py
    for zahlen in range(x):
        if zahlen == count:
            continue
        ...
    ```
    From here, it should be pretty easy to convert this inner while-loop to a for-loop.
2) Let's change the variable names to be in English. Nothing against German, but these names would be confusing when the rest of the repo is entirely in English.
`zahlen` -> `number` or `num`
`bruch` -> `fraction` or `frac`

tianyizheng02 (changes requested):

linear_algebra/src/polynom_for_points.py
MaximSmolskiy: Done

tianyizheng02 (approved):

PR #8607, opened 2023-04-01
Add Project Euler problem 79 solution 1

ishabahmed:
### Describe your change:

Added Project Euler problem 79 solution 1

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


project_euler/problem_079/sol1.py
```diff
@@ -0,0 +1,53 @@
...
+import itertools
+import os
+
+
+def solution() -> int:
+    """
+    Returns the shortest possible secret passcode of unknown length.
+
+    >>> solution()
+    73162890
```
dhruvmanila, line 27: As mentioned in the [README](https://github.com/TheAlgorithms/Python/tree/master/project_euler#coding-style):

> There should not be a doctest for testing the answer as that is done by our GitHub Actions build ...

```suggestion
```

Instead, have some test cases. Now as this requires reading a file to get the input, you could separate out the core part of the solution (which is everything after the file is read) and take in a parameter `logins`:

```python
# This function could be tested with variety of inputs.
def find_secret_passcode(logins):
	...


def solution():
	# Read the keylog.txt file and fill the `logins` array
    logins = [...]

	return find_secret_passcode(logins)
```

dhruvmanila (changes requested):
It's better to keep the filename same as that of problem statement: `keylog.txt`. We already know the problem number from the directory.

project_euler/problem_079/sol1.py
ishabahmed: Good idea. I will implement this.

project_euler/problem_079/sol1.py
```diff
@@ -0,0 +1,53 @@
...
+                < permutation.index(login[2])
+            ):
+                satisfied = False
+                break
+
+        if satisfied:
+            return int("".join(permutation))
+            break
+
+    return 0
```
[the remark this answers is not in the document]
ishabahmed: Given the fixed input from keylog.txt, the code should hypothetically never reach that line. However, it was necessary to include it to ensure that mypy detects a reachable return statement. Following the suggestion to split the core solution into a separate function, raising an exception makes sense, so I will implement that.

project_euler/problem_079/sol1.py
```diff
@@ -0,0 +1,76 @@
...
+
+    return find_secret_passcode(logins)
+
+
+if __name__ == "__main__":
+    import doctest
+
+    doctest.testmod()
+
+    print(solution())
```
dhruvmanila, line 76: `doctest` are tested directly by the CI. If you want to run them locally, I would recommend to use the command:
```console
python -m doctest -v sol1.py
```

```suggestion
    print(f"{solution() = }")
```

dhruvmanila (approved):

dhruvmanila (approved):

PR #8617, opened 2023-04-07
Queue implementation using two Stacks

amirsoroush:
# Queue using two Stacks:

LQJBQ

Here is the implementation of Queue using two stacks. It has O(1) time complexity for en-queuing and *Amortized* O(1) for de-queuing. The worst case time complexity for de-queuing is O(n) though.  The rationale behind this and its use-cases are mentioned [here](https://stackoverflow.com/a/2050402/13944524) and [here](https://stackoverflow.com/a/7395431/13944524).

I also included *doctest* for methods. It's formatted using "Black" and checked by "Mypy".

---

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/queue/queue_on_two_stacks.py
+++ b/data_structures/queue/queue_on_two_stacks.py
@@ -1,13 +1,13 @@
 """Queue implementation using two stacks"""
 
 from collections.abc import Iterable
-from typing import Generic, Optional, TypeVar
+from typing import Generic, TypeVar
 
 _T = TypeVar("_T")
 
 
 class Queue(Generic[_T]):
-    def __init__(self, iterable: Optional[Iterable[_T]] = None) -> None:
+    def __init__(self, iterable: Iterable[_T] | None = None) -> None:
         """
         >>> queue1 = Queue()
         >>> str(queue1)
--- a/data_structures/queue/queue_on_two_stacks.py
+++ b/data_structures/queue/queue_on_two_stacks.py
@@ -1,0 +1,123 @@
+"""Queue implementation using two stacks"""
+
+from typing import Any
+
+
+class Queue:
+    def __init__(self) -> None:
+        self._stack1: list[Any] = []
+        self._stack2: list[Any] = []
+
+    def put(self, item: Any) -> None:
+        """
+        Put `item` into the Queue
+
+        >>> queue = Queue()
+        >>> queue.put(10)
+        >>> queue.put(20)
+        >>> len(queue) == 2
+        True
+        >>> str(queue)
+        '<[10, 20]>'
+        """
+
+        self._stack1.append(item)
+
+    def get(self) -> Any:
+        """
+        Get `item` from the Queue
+
+        >>> queue = Queue()
+        >>> for i in (10, 20, 30):
+        ...     queue.put(i)
+        >>> len(queue)
+        3
+        >>> queue.get()
+        10
+        >>> queue.get()
+        20
+        >>> len(queue) == 1
+        True
+        >>> queue.get()
+        30
+        >>> queue.get()
+        Traceback (most recent call last):
+            ...
+        IndexError: Queue is empty
+        """
+
+        # To reduce number of attribtue look-ups in `while` loop.
+        stack1_pop = self._stack1.pop
+        stack2_append = self._stack2.append
+
+        if not self._stack2:
+            while self._stack1:
+                stack2_append(stack1_pop())
+
+        if not self._stack2:
+            raise IndexError("Queue is empty")
+        return self._stack2.pop()
+
+    def size(self) -> int:
+        """
+        Returns the length of the Queue
+
+        >>> queue = Queue()
+        >>> queue.size()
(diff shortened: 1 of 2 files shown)


amirsoroush:
Apparently ruff is not happy with my commit. It says:
```none
Error: computer_vision/lenet_pytorch.py:9:8: PLR0402 Use `from torch import nn` in lieu of alias
Error: Process completed with exit code 1.
```
Unfortunately I have no idea about it. I didn't import anything from `torch`.

CaedenPH:
> Apparently ruff is not happy with my commit. It says:
> 
> ```
> Error: computer_vision/lenet_pytorch.py:9:8: PLR0402 Use `from torch import nn` in lieu of alias
> Error: Process completed with exit code 1.
> ```
> 
> Unfortunately I have no idea about it. I didn't import anything from `torch`.

Seems like you didn't touch that file so someone must've merged a commit that failed ruff

amirsoroush:
@CaedenPH I guess the failure is because of the merged commit which is currently requested to be reverted [here](https://github.com/TheAlgorithms/Python/pull/8621).

data_structures/queue/queue_on_two_stacks.py
```diff
@@ -0,0 +1,123 @@
+"""Queue implementation using two stacks"""
+
+from typing import Any
+
+
+class Queue:
```
cclauss, line 6: Please name the class something else so it does not conflict with `queue.Queue`

data_structures/queue/queue_on_two_stacks.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(self) -> None:
```
cclauss, line 7: Please make an `.__init__()` that accepts an iterable of initial values.

cclauss (changes requested):

data_structures/queue/queue_on_two_stacks.py
amirsoroush: @cclauss  Sure. Do you have any naming convention in mind so that I make it consistent with other queue implementations like [this PR](https://github.com/TheAlgorithms/Python/pull/8608)? Which one is more suitable? `Queue_TwoStacks`, `QueueByStacks` or `QueueTwoStacks`.

cclauss: No underscore in ClassNames --> https://beta.ruff.rs/docs/rules/#pep8-naming-n

You can run `ruff --select=N path/to/my/python/file.py` to ensure all is OK.

data_structures/queue/queue_on_two_stacks.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        >>> queue = Queue()
+        >>> str(queue)
+        '<[]>'
+        >>> queue.put(10)
+        >>> str(queue)
+        '<[10]>'
+        >>> queue.put(20)
+        >>> str(queue)
+        '<[10, 20]>'
```
cclauss, line 105: https://docs.python.org/3/reference/datamodel.html#object.\_\_repr\_\_

The output should be the same as the command to create a new instance with these contents.

```suggestion
        'Queue(10, 20)'
```

data_structures/queue/queue_on_two_stacks.py
```diff
@@ -0,0 +1,138 @@
...
+        2
+        >>> queue.get()
+        10
+        >>> queue.size() == 1
+        True
+        """
+
+        return len(self)
+
+    def __str__(self) -> str:
```
cclauss, line 95: Let's define a `__repr__()` instead.

data_structures/queue/queue_on_two_stacks.py
amirsoroush: @cclauss  Yes totally agree. It's True for `__repr__`. I'm gonna do that next and change the tests.  One more question `Queue` doesn't accept `*args` but an iterable so the representation can't be `Queue(10, 20, 30)`. Is it okay if I wrap them in bracket? like `Queue([10, 20, 30])`.

cclauss:
For readability, please move the "magic" functions like `.__len__()` and .__str__()` before `.get()` and `.put()`

cclauss (approved):
LGTM

A few changes:
* If you can use something immediately without naming it then do so to keep doctests short
* Test the result, not the result == expected is True
* Use repr() instead of str() in tests unless essential.
* Favor tuples over lists when possible because they take less RAM.
* We do not need a .size()` method if we have defined len.

amirsoroush:
@cclauss Thank you for your advice and also for introducing me to `ruff` linter.

For `.size()`, "result == expected is True", "implementing `__str__`", I wanna say that I wrote them because they've been implemented in other files. I tried to be consistent with those but I'm totally agree with you. I'll also fix those in the future.

cclauss:
Thanks for your contribution.  Our codebase is not (yet) perfect so please submit pull requests for any existing algorithms that can be improved.

Issue #8622, opened 2023-04-07
Refactor code so it's readable


Starmania:
### Feature description

I was exploring the git when I was reading this file: https://github.com/TheAlgorithms/Python/blob/f32d611689dc72bda67f1c4636ab1599c60d27a4/ciphers/mixed_keyword_cypher.py

The variable name are really sort, and it make difficulties to read... I think it could be cool if someone take the time to describe a little bit more the code...

tianyizheng02:
The final print statement isn't enclosed in any `__main__` statement either, so that should be done too

yanvoi:
My refactoring could be better but I hope it's gonna cut it.

PR #8626, opened 2023-04-08
Improve readability of ciphers/mixed_keyword_cypher.py

yanvoi:
fixes: #8622 
### Describe your change:
I've tried my best with refactoring the code - it still might be too little but that's how much I've managed to do with the time I had.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/ciphers/mixed_keyword_cypher.py
+++ b/ciphers/mixed_keyword_cypher.py
@@ -1,7 +1,9 @@
 from string import ascii_uppercase
 
 
-def mixed_keyword(keyword: str, plaintext: str, alphabet: str = ascii_uppercase) -> str:
+def mixed_keyword(
+    keyword: str, plaintext: str, verbose: bool = True, alphabet: str = ascii_uppercase
+) -> str:
     """
     For keyword: hello
 
@@ -14,11 +16,14 @@ def mixed_keyword(keyword: str, plaintext: str, alphabet: str = ascii_uppercase)
     Y Z
     and map vertically
 
-    >>> mixed_keyword("college", "UNIVERSITY")  # doctest: +NORMALIZE_WHITESPACE
+    >>> mixed_keyword("college", "UNIVERSITY", True)  # doctest: +NORMALIZE_WHITESPACE
     {'A': 'C', 'B': 'A', 'C': 'I', 'D': 'P', 'E': 'U', 'F': 'Z', 'G': 'O', 'H': 'B',
      'I': 'J', 'J': 'Q', 'K': 'V', 'L': 'L', 'M': 'D', 'N': 'K', 'O': 'R', 'P': 'W',
      'Q': 'E', 'R': 'F', 'S': 'M', 'T': 'S', 'U': 'X', 'V': 'G', 'W': 'H', 'X': 'N',
      'Y': 'T', 'Z': 'Y'}
+    'XKJGUFMJST'
+
+    >>> mixed_keyword("college", "UNIVERSITY", False)  # doctest: +NORMALIZE_WHITESPACE
     'XKJGUFMJST'
     """
     keyword = keyword.upper()
@@ -26,6 +31,7 @@ def mixed_keyword(keyword: str, plaintext: str, alphabet: str = ascii_uppercase)
     alphabet_set = set(alphabet)
 
     # create a list of unique characters in the keyword
+    # their order matters, it determines how we will map plaintext characters to the ciphertext
     unique_chars = []
     for char in keyword:
         if char in alphabet_set and char not in unique_chars:
@@ -58,11 +64,12 @@ def mixed_keyword(keyword: str, plaintext: str, alphabet: str = ascii_uppercase)
             mapping[alphabet[letter_index]] = row[column]
             letter_index += 1
 
-    print(mapping)
+    if verbose:
+        print(mapping)
     # create the encrypted text by mapping the plaintext to the modified alphabet
     return "".join(mapping[char] if char in mapping else char for char in plaintext)
 
 
 if __name__ == "__main__":
     # example use
-    print(mixed_keyword("college", "UNIVERSITY", ascii_uppercase))
+    print(mixed_keyword("college", "UNIVERSITY"))
--- a/ciphers/mixed_keyword_cypher.py
+++ b/ciphers/mixed_keyword_cypher.py
@@ -61,6 +61,7 @@ def mixed_keyword(keyword: str = "college", plaintext: str = "UNIVERSITY") -> st
                 break
             k += 1
 
+    print(mapping)
     # create the encrypted text by mapping the plaintext to the modified alphabet
     return "".join(mapping[char] for char in plaintext)
 
--- a/ciphers/mixed_keyword_cypher.py
+++ b/ciphers/mixed_keyword_cypher.py
@@ -1,7 +1,7 @@
 from string import ascii_uppercase
 
 
-def mixed_keyword(keyword: str = "college", plaintext: str = "UNIVERSITY") -> str:
+def mixed_keyword(keyword: str = "college", plaintext: str = "UNIVERSITY", alphabet: str = ascii_uppercase) -> str:
     """
     For keyword: hello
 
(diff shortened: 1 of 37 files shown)


yanvoi:
well then I guess the original code doesn't pass the tests either

ciphers/mixed_keyword_cypher.py
```diff
@@ -19,50 +21,49 @@ def mixed_keyword(key: str = "college", pt: str = "UNIVERSITY") -> str:
...
-    # print(temp)
-    r = int(26 / 4)
-    # print(r)
+    keyword = keyword.upper()
+    plaintext = plaintext.upper()
+
+    unique_chars = []
+    for char in keyword:
+        if char not in unique_chars:
+            unique_chars.append(char)
```
cclauss, line 30: Please make this a `set`.

ciphers/mixed_keyword_cypher.py
```diff
@@ -19,50 +21,49 @@ def mixed_keyword(key: str = "college", pt: str = "UNIVERSITY") -> str:
...
+
+    rows = int(26 / 4)
+    # k is an index variable
     k = 0
-    for _ in range(r):
-        s = []
-        for _ in range(len_temp):
-            s.append(temp[k])
+    modified_alphabet = []
```
cclauss, line 43: Please do all this with `set`

cclauss:
On your local machine, please run `python3 -m doctest -v ciphers/mixed_keyword_cypher.py`
because pytest is failing on:
```
=================================== FAILURES ===================================
_____________ [doctest] ciphers.mixed_keyword_cypher.mixed_keyword _____________
008     H E L O
009     A B C D
010     F G I J
011     K M N P
012     Q R S T
013     U V W X
014     Y Z
015     and map vertically
016 
017     >>> mixed_keyword("college", "UNIVERSITY")  # doctest: +NORMALIZE_WHITESPACE
Expected:
    {'A': 'C', 'B': 'A', 'C': 'I', 'D': 'P', 'E': 'U', 'F': 'Z', 'G': 'O', 'H': 'B',
     'I': 'J', 'J': 'Q', 'K': 'V', 'L': 'L', 'M': 'D', 'N': 'K', 'O': 'R', 'P': 'W',
     'Q': 'E', 'R': 'F', 'S': 'M', 'T': 'S', 'U': 'X', 'V': 'G', 'W': 'H', 'X': 'N',
     'Y': 'T', 'Z': 'Y'}
    'XKJGUFMJST'
Got:
    'XKJGUFMJST'
```

ciphers/mixed_keyword_cypher.py
yanvoi: It cannot be done with sets due to the nature of the algorithm. The order of those unique characters matters.

yanvoi: I believe it cannot be done with sets because the order of these characters in the keyword matters - it determines how we will map plaintext characters to the cyphertext.

yanvoi:
My PR failing the test was indeed my fault, I took the liberty of getting rid of print() from the body of the function which caused it to fail the test. My bad. Although I still believe that the printing of the mapping dictionary shouldn't be included in the code.

ciphers/mixed_keyword_cypher.py
yanvoi: As this is a basic substitution cypher, it is bounded by the number of different letters in the alphabet we're using. Using sets in this case wouldn't change the performance almost at all.

yanvoi:
It turned out that whoever has written this algorithm hard-coded the number of rows into it.

tianyizheng02 (changes requested):

ciphers/mixed_keyword_cypher.py
```diff
@@ -19,50 +21,48 @@ def mixed_keyword(key: str = "college", pt: str = "UNIVERSITY") -> str:
...
-    for j in range(len_temp):
-        for m in modalpha:
-            if not len(m) - 1 >= j:
-                break
-            d[alpha[k]] = m[j]
-            if not k < 25:
+    keyword = keyword.upper()
+    plaintext = plaintext.upper()
+    alphabet = list(ascii_uppercase)
+    alphabet_set = set(alphabet)
```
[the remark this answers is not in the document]
yanvoi: I can of course change that but I wanted to make it easier for people who would like to use other alphabets (e.g. cyrillic) for their ciphers. So it is your call - should I change it or not?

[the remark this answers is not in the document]
tianyizheng02: If people want to use other alphabets such as Cyrillic, then they would have to go into the source code and change `ascii_uppercase` to a different variable. That doesn't really address the issue that `ascii_uppercase` (or whatever the user replaces it with) still doesn't need to be converted to a list.

If you want it to be easier to switch to a different alphabet, then perhaps you could have the function take in the alphabet string (`ascii_uppercase`, etc) as an input parameter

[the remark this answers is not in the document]
yanvoi: Fair point - I've done both - Now there's no conversion to the list and a different alphabet can be passed to the function. Thank you for the review!

tianyizheng02 (approved):

tianyizheng02:
> The tests seem to be failing due to code outside of the file I've been editing.

Other contributors have been experiencing the same issue with ruff... try running `ruff check --fix .` to see if ruff can automatically fix any of the linting issues

yanvoi:
Sadly it resulted in even more, worse issues

tianyizheng02:
@cclauss How is it that @rohan472000's changes in their own branch are making their way into this PR?

tianyizheng02:
Oh my goddddd I knew this was gonna happen...

@yanvoi The ruff error is once again in another file that you didn't edit, but I already fixed it in #8743. Once that PR gets merged, updating your branch on GitHub should finally fix the issue

yanvoi:
@tianyizheng02 No worries, I'll update when #8743 gets merged

tianyizheng02:
@yanvoi The ruff error was fixed in #8746 instead, so you should be good to update your branch

yanvoi:
@tianyizheng02 All done, thanks again!

ciphers/mixed_keyword_cypher.py
tianyizheng02: > I believe it cannot be done with sets because the order of these characters in the keyword matters - it determines how we will map plaintext characters to the cyphertext.

Then I think it would help if this was mentioned in the comments, so that future maintainers know that the char ordeer matters

ciphers/mixed_keyword_cypher.py
```diff
@@ -19,51 +21,48 @@ def mixed_keyword(key: str = "college", pt: str = "UNIVERSITY") -> str:
...
-    cypher = ""
-    for i in pt:
-        cypher += d[i]
-    return cypher
+
+            # map current letter to letter in modified alphabet
+            mapping[alphabet[letter_index]] = row[column]
+            letter_index += 1
+
+    print(mapping)
```
tianyizheng02, line 61: Ideally the algorithm shouldn't have side effects such as printing, but I understand that printing the mapping is helpful for debugging purposes. Perhaps it'd be better to have a "verbose" function parameter that allows users to enable/disable printing the mapping

tianyizheng02 (changes requested):

ciphers/mixed_keyword_cypher.py
yanvoi: Done - I've also added a test case that includes setting the verbose parameter to False

ciphers/mixed_keyword_cypher.py
```diff
@@ -1,7 +1,11 @@
-def mixed_keyword(key: str = "college", pt: str = "UNIVERSITY") -> str:
-    """
+from string import ascii_uppercase
+
 
-    For key:hello
+def mixed_keyword(
+    keyword: str, plaintext: str, verbose: bool = True, alphabet: str = ascii_uppercase
```
tianyizheng02, line 5:
```suggestion
    keyword: str, plaintext: str, verbose: bool = False, alphabet: str = ascii_uppercase
```

tianyizheng02 (changes requested):

tianyizheng02 (approved):

cclauss (approved):

Starmania:
Thanks you for the edit yanvoi !

PR #8662, opened 2023-04-16
Project Euler problem148

saitejamanchi:
### Describe your change:

Added solution for Problem 148 in Project Euler.

* [X] Add an algorithm?
* [ ] Documentation change?

### Checklist:


project_euler/problem_148/sol1.py
```diff
@@ -0,0 +1,77 @@
...
+We only need to generate half of the values.
+We then count the values which are not divisible by 7.
+We only store the remainders(when divided by 7) in the list to reduce memory usage.
+
+Note: In the original problem, we need to calculate for 10^9 rows
+      but we took 10^4 rows here by default.
+"""
+
+
+def solution(pascal_row_count: int = 10**4) -> int:
```
rohan472000, line 31: try this:

`def solution(pascal_row_count: int = 10**9) -> int:`

saitejamanchi: Hey,

I've tried with 10**9 but one of the tests failed due to timeout as it was taking more than 6 hours to run.

rohan472000: It's not uncommon for brute force solutions to take a long time or run into timeout issues when dealing with very large inputs. In this case, the input value of 10**9 is extremely large, and the current implementation of the "solution" function does not have an efficient algorithm to handle such large inputs.

rohan472000 (changes requested):
kindly go through below code, it should fix your timeout issue.

"""def solution(pascal_row_count: int = 10**9) -> int:
    """
    To evaluate the solution, use solution()
    >>> solution(3)
    6
    >>> solution(10)
    40
    >>> solution(100)
    2361
    """
    # Helper function to compute the number of odd binomial coefficients in a row
    def odd_binomials(n: int) -> int:
        res = 0
        while n > 0:
            res += n % 2
            n //= 2
        return res

    # Compute the sum of odd binomial coefficients modulo 7 for each row up to pascal_row_count
    result = 0
    for k in range(pascal_row_count):
        result += odd_binomials(k) % 7

    return result
"""

project_euler/problem_148/sol1.py
```diff
@@ -0,0 +1,77 @@
...
+We then count the values which are not divisible by 7.
+We only store the remainders(when divided by 7) in the list to reduce memory usage.
+
+Note: In the original problem, we need to calculate for 10^9 rows
+      but we took 10^4 rows here by default.
+"""
+
+
+def solution(pascal_row_count: int = 10**9) -> int:
+    """
```
rohan472000, line 32: use below codes for better efficiency which will solve the timeout issue:-

    

      def solution(pascal_row_count: int = 10**9) -> int:
          """
          To evaluate the solution, use solution()
          >>> solution(3)
          6
          >>> solution(10)
          40
          >>> solution(100)
          2361
          """
      
          # Helper function to compute the number of odd binomial coefficients in a row
          def odd_binomials(n: int) -> int:
              res = 0
              while n > 0:
                  res += n % 2
                  n //= 2
              return res
      
          # Compute the sum of odd binomial coefficients modulo 7 for each row up to pascal_row_count
          result = 0
          for k in range(pascal_row_count):
              result += odd_binomials(k) % 7
      
          return result

rohan472000 (changes requested):

project_euler/problem_148/sol1.py
rohan472000: also add docs in `odd_binomials`, you can put above comment of that function inside as a doc.

rohan472000:
use `ruff .` then `ruff . --fix` to fix ruff issues.

project_euler/problem_148/sol1.py
```diff
@@ -0,0 +1,52 @@
...
+      1          5         10         10          5          1
+1          6         15         20         15          6          1
+However, if we check the first one hundred rows, we will find that
+only 2361 of the 5050 entries are not divisible by 7.
+Find the number of entries which are not divisible by 7
+in the first one billion (109) rows of Pascal's triangle.
+"""
+
+
+def get_num_binomials(row_num: int) -> int:
```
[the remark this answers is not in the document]
rohan472000: @saitejamanchi add doctests.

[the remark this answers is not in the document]
saitejamanchi: @rohan472000 Sure, will add.

Thanks a lot for the help!!

rohan472000:
kindly fix ruff issue before committing...by `ruff .` then `ruff , --fix`

saitejamanchi:
Hi @dhruvmanila ,

Could you please look into this pull request?

PR #8664, opened 2023-04-17
Create apparent_power.py

rohan472000:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/electronics/apparent_power.py
+++ b/electronics/apparent_power.py
@@ -5,13 +5,6 @@ def apparent_power(real_power: float, reactive_power: float) -> float:
     """
     Calculate apparent power from real power and reactive power.
 
-    Args:
-        real_power (float): Real power in watts.
-        reactive_power (float): Reactive power in vars.
-
-    Returns:
-        float: Apparent power in volt-amperes (VA).
-
     Examples:-
     >>> calculate_apparent_power(100, 50)
     111.80339887498948
--- a/electronics/apparent_power.py
+++ b/electronics/apparent_power.py
@@ -26,9 +26,7 @@ def apparent_power(
     current_rect = cmath.rect(current, current_angle_rad)
 
     # Calculate apparent power
-    apparent_power = voltage_rect * current_rect
-
-    return apparent_power
+    return voltage_rect * current_rect
 
 
 if __name__ == "__main__":
--- a/electronics/apparent_power.py
+++ b/electronics/apparent_power.py
@@ -1,0 +1,30 @@
+import math
+
+def calculate_apparent_power(real_power: float, reactive_power: float) -> float:
+    """
+    Calculate apparent power from real power and reactive power.
+
+    Args:
+        real_power (float): Real power in watts.
+        reactive_power (float): Reactive power in vars.
+
+    Returns:
+        float: Apparent power in volt-amperes (VA).
+
+    Examples:-
+    >>> calculate_apparent_power(100, 50)
+    111.80339887498948
+    >>> calculate_apparent_power(0, 0)
+    0.0
+    >>> calculate_apparent_power('100', 50)
+    Traceback (most recent call last):
+    ...
+    ValueError: real_power and reactive_power must be valid float values.
+    """
+    apparent_power = math.sqrt(real_power ** 2 + reactive_power ** 2)
+    return apparent_power
+
+
+if __name__ == "__main__":
+    import doctest
+    doctest.testmod()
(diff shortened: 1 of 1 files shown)


electronics/apparent_power.py
```diff
@@ -0,0 +1,30 @@
+import math
+
+
+def calculate_apparent_power(real_power: float, reactive_power: float) -> float:
```
cclauss, line 4:
```suggestion
def apparent_power(real_power: float, reactive_power: float) -> float:
```

electronics/apparent_power.py
```diff
@@ -0,0 +1,30 @@
...
+    Examples:-
+    >>> calculate_apparent_power(100, 50)
+    111.80339887498948
+    >>> calculate_apparent_power(100, -50)
+    111.80339887498948
+    >>> calculate_apparent_power(0, 0)
+    0.0
+    """
+    apparent_power = math.sqrt(real_power**2 + reactive_power**2)
+    return apparent_power
```
cclauss, line 24:
```suggestion
    return math.sqrt(real_power**2 + reactive_power**2)
```

electronics/apparent_power.py
```diff
@@ -0,0 +1,30 @@
...
+def calculate_apparent_power(real_power: float, reactive_power: float) -> float:
+    """
+    Calculate apparent power from real power and reactive power.
+
+    Args:
+        real_power (float): Real power in watts.
+        reactive_power (float): Reactive power in vars.
+
+    Returns:
+        float: Apparent power in volt-amperes (VA).
```
cclauss, line 13: The problem with repeating the data type (float) from the function signature above in the dockstring is that users will get very confused if one changes but the other does not.  In general, we encourage that the data type should be in the function signature and the docstring should not contain the data type.

rohan472000: Yeah.. removed that.

cclauss (changes requested):

EngineerMarek:
Shouldn't we call this something like abs_apparent_power? This is only the absolute value of apparent power calculated from the pythagorean theorem, but it does not include the phase. Though I don't know how you guys would like to implement complex numbers for that or if there is already an implementation/a standart for that

cclauss:
Or should we `raise ValueError("real_power must not be negative")`

rohan472000:
> Shouldn't we call this something like abs_apparent_power? This is only the absolute value of apparent power calculated from the pythagorean theorem, but it does not include the phase. Though I don't know how you guys would like to implement complex numbers for that or if there is already an implementation/a standart for that

I tried to implement that, have a look.

electronics/apparent_power.py
```diff
@@ -1,25 +1,34 @@
...
+    # Convert voltage and current to rectangular form
+    voltage_rect = cmath.rect(voltage, voltage_angle_rad)
+    current_rect = cmath.rect(current, current_angle_rad)
+
+    # Calculate apparent power
+    apparent_power = voltage_rect * current_rect
 
+    return apparent_power
 
+# Doctests
```
cclauss: This comment just repeats what the code just below it says so it can be dropped.

electronics/apparent_power.py
```diff
@@ -1,25 +1,34 @@
...
+    current_angle_rad = math.radians(current_angle)
+
+    # Convert voltage and current to rectangular form
+    voltage_rect = cmath.rect(voltage, voltage_angle_rad)
+    current_rect = cmath.rect(current, current_angle_rad)
+
+    # Calculate apparent power
+    apparent_power = voltage_rect * current_rect
 
+    return apparent_power
```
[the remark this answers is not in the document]
rohan472000: Done

cclauss (approved):
Awesome!!  Thanks for your contribution.

cclauss:
The one thing that I forgot to ask for is a URL to a Wikipedia article or some other source that explains the algorithm and where/when it should be used.  This can be a great help to visitors to our repo.  If you create a pull request that adds https://en.wikipedia.org/wiki/AC_power#Apparent_power or something similar to the docstring then I will gladly review that.

rohan472000:
> The one thing that I forgot to ask for is a URL to a Wikipedia article or some other source that explains the algorithm and where/when it should be used. This can be a great help to visitors to our repo. If you create a pull request that adds https://en.wikipedia.org/wiki/AC_power#Apparent_power or something similar to the docstring then I will gladly review it.

sure...done that, please have a look.

PR #8666, opened 2023-04-17
Create maths/pi_generator.py

JulianStiebler:
### Describe your change:

* [X] Add an algorithm?
* [ ] Documentation change?

### Checklist:


JulianStiebler:
Kay all tests seem fine. Do i need to add a unittest myself?

maths/pi_generator.py
```diff
@@ -0,0 +1,80 @@
...
+            q *= k
+            t *= l
+            l += 2
+            k += 1
+            n = nn
+            r = nr
+    return result
+
+
+def main() -> None:
```
rohan472000, line 71: you can remove main method block, instead of that keep below only -
```
if __name__ == "__main__":
    import doctest

    doctest.testmod()
```

maths/pi_generator.py
```diff
@@ -0,0 +1,80 @@
...
+
+    To further proof, since we cannot predict errors or interrupt an infinite generation
+    or interrupt any alternating series, here some more tests.
+
+    >>> calculate_pi(50)
+    '3.14159265358979323846264338327950288419716939937510'
+
+    >>> calculate_pi(80)
+    '3.14159265358979323846264338327950288419716939937510582097494459230781640628620899'
+    """
```
rohan472000, line 28: for viewer understanding, you can add one more line -

''In the Leibniz formula for calculating pi, the variables q, r, t, k, n, and l are used for the iteration process''

rohan472000 (changes requested):

maths/pi_generator.py
cclauss: We can keep the `main()` block and ask @dhruvmanila to tame @algorithms-keeper so it does not complain about `main()` needing tests.

maths/pi_generator.py
```diff
@@ -0,0 +1,80 @@
...
+    The Leibniz formula is the special case arctan 1 = 1/4 Pi .
+    Leibniz's formula converges extremely slowly: it exhibits sublinear convergence.
+
+    Convergence (https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80#Convergence)
+
+    We cannot try to prove against an interrupted, uncompleted generation.
+    https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80#Unusual_behaviour
+    The errors can in fact be predicted;
+    but those calculations also approach infinity for accuracy.
+    For simplicity' sake, let's just compare against known values.
```
cclauss, line 15: This is why god invented [`math.isclose()`](https://docs.python.org/3/library/math.html#math.isclose).

maths/pi_generator.py
```diff
@@ -0,0 +1,80 @@
...
+    '3.141592653589793'
+
+    To further proof, since we cannot predict errors or interrupt an infinite generation
+    or interrupt any alternating series, here some more tests.
+
+    >>> calculate_pi(50)
+    '3.14159265358979323846264338327950288419716939937510'
+
+    >>> calculate_pi(80)
+    '3.14159265358979323846264338327950288419716939937510582097494459230781640628620899'
```
cclauss, line 27: Please add at least two tests:
1. calculate_pi(x) == [math.pi](https://docs.python.org/3/library/math.html#math.pi)  # for some value of x
2. Combines the above with `math.isclose()`

maths/pi_generator.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Since the math.pi-constant is only 16 digits long, here some test with preknown values:
+
+    >>> calculate_pi(50)
+    '3.14159265358979323846264338327950288419716939937510'
+    >>> calculate_pi(80)
+    '3.14159265358979323846264338327950288419716939937510582097494459230781640628620899'
+
+    In the Leibniz formula for calculating pi, the variables q, r, t, k, n, and l are used for the iteration process.
+    """  # noqa: E501
```
cclauss, line 41: Please fold long lines like everyone else does.
```suggestion
    """
```

JulianStiebler: Oops! I didn't realized i did a quickfix, wasn't in there before aswell! ^^

maths/pi_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            l += 2
+            k += 1
+            n = nn
+            r = nr
+    return result
+
+
+def main() -> None:
+    pi_digits = calculate_pi(50)
+    print(pi_digits)
```
cclauss, line 88:
```suggestion
    print(f"{calculate_pi(50) = }")
```

JulianStiebler: done! should i remove the main-statement now or do we just keep ignoring algorithms-keeper

cclauss (approved):
Let's land this one!  The only other thing that I could think of doing would be to see how low the `tolerances` on [`math.isclose()`](https://docs.python.org/3/library/math.html#math.isclose) could be set and still pass the tests but that is clearly not vital.

Thanks for your contribution!

JulianStiebler:
Trying:
    math.isclose(float(calculate_pi(15)), math.pi, abs_tol = 0)
Expecting:
    True
ok

Trying:
    math.isclose(float(calculate_pi(50)), math.pi, abs_tol = 0.1)
Expecting:
    True
ok

Thanks for your review!

cclauss:
My sense is that when `abs_tol = 0`, the function goes with its default of 9 decimal places.

Perhaps if `abs_tol` was set to 0.00000000000000000000000001 then tests would start to fail.

JulianStiebler:
```
>>>math.isclose(float(calculate_pi(100)), math.pi, abs_tol = 0.00000000000000000000000000000000000001)
True

Trying:
    math.isclose(float(calculate_pi(50)), math.pi, abs_tol = 0.00000000000000000000000000000000000001)
Expecting:
    True
ok
```

Since we didnt interrupted the series but rather finished it with a limit it should always match pi to the exact digit

cclauss:
Nice work!  Thanks.

PR #8693, opened 2023-04-26
BST and RSA doctest

isidroas:
### Describe your change:

* [ ] Add an algorithm?
* [X] Documentation change?

### Checklist:


ankit-gautam23 (commented):
LGTM

isidroas:
Hi @rohan472000, thanks for the review.

1. I just added https://github.com/TheAlgorithms/Python/pull/8693/commits/4a13d5d118191f40fc443ea6a5fd121fcff3b268 which raises an error when trying to delete a non-existent element in a Binary Search Tree. As for the `generate_key` function, it only has the size argument and I don't see any interesting potential error.
2. I discovered that the secrets module [doesn't have  have `seed` method](https://stackoverflow.com/a/65432128/16926605). That would be important for doctests. Also, I thought that the main goal this project is not to use it in production code, but to teach algorithms in a way that it is even easier than reading Wikipedia or a book.

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -20,6 +77,12 @@ def __repr__(self) -> str:
             return str(self.value)
         return pformat({f"{self.value}": (self.left, self.right)}, indent=1)
 
+    @property
+    def is_right(self) -> bool:
+        if self.parent and self.parent.right:
+            return self == self.parent.right
+        return False
```
tianyizheng02, line 84: I think this function can be simplified like this:

```suggestion
        return self.parent and self is self.parent.right
```

1) `==` checks if two variables have the same value, while `is` checks if they refer to the same object, so I think `is` is more appropriate here.

2) We don't need to check if `self.parent.right` is not None: if it _is_ None, then `self is self.parent.right` would fail and the function would return False anyway.

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -35,18 +98,13 @@ def __reassign_nodes(self, node: Node, new_children: Node | None) -> None:
         if new_children is not None:  # reset its kids
             new_children.parent = node.parent
         if node.parent is not None:  # reset its parent
-            if self.is_right(node):  # If it is the right children
+            if node.is_right:  # If it is the right children
```
tianyizheng02, line 101:
```suggestion
            if node.is_right:  # If it is the right child
```

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -1,5 +1,62 @@
...
+14
+>>> t.get_min().value
+1
+>>> t.empty()
+False
+>>> for i in testlist:
+...     t.remove(i)
+>>> t.empty()
+True
 """
```
tianyizheng02, line 60: I don't know if `doctest.testmod` will run these doctests since they're not contained in any function or class. @cclauss Where should these examples go?

tianyizheng02 (changes requested):

data_structures/binary_tree/binary_search_tree.py
```diff
@@ -20,6 +77,10 @@ def __repr__(self) -> str:
             return str(self.value)
         return pformat({f"{self.value}": (self.left, self.right)}, indent=1)
 
+    @property
+    def is_right(self):
+        return self.parent and self is self.parent.right
```
tianyizheng02, line 82: I think this should fix the return type error while keeping the return type annotation
```suggestion
    def is_right(self) -> bool:
        return self.parent is not None and self is self.parent.right
```

tianyizheng02 (changes requested):

tianyizheng02 (approved):

PR #8707, opened 2023-05-03
Optimize and improve swap nodes on linked list code

MarouaneBouf:
This commit refactors the linked list code to improve its structure and readability. The code has been split into separate classes for the node and linked list, making it easier to understand and modify. The methods have been well-documented with comments to explain their functionality.

### Describe your change:

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/linked_list/swap_nodes.py
+++ b/data_structures/linked_list/swap_nodes.py
@@ -51,7 +51,6 @@ class LinkedList:
         """
         A method that swaps the nodes with data values
         node_data_1 and node_data_2 in the linked list.
-
         If either node is not found, the method simply returns.
         """
         # if the two data values are the same, no swapping is needed
--- a/data_structures/linked_list/swap_nodes.py
+++ b/data_structures/linked_list/swap_nodes.py
@@ -51,7 +51,7 @@ class LinkedList:
         """
         A method that swaps the nodes with data values
         node_data_1 and node_data_2 in the linked list.
-        
+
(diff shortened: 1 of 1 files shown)


rohan472000 (changes requested):
1. Add docstrings to methods: It's a good practice to provide docstrings for your methods to explain their functionality. You've already added docstrings for the class and the `print_list` method. Consider adding docstrings for other methods like `push` and `swap_nodes` as well.

2. Use more descriptive variable names: Variable names like `node_data_1` and `node_data_2` could be more descriptive. Consider using names like `data1` and `data2` instead.

3. Handle the case when nodes are not found: Currently, if either of the nodes with the specified data values is not found, the `swap_nodes` method simply returns without any indication. You might want to consider raising an exception or returning a specific value to indicate that the nodes were not found.

tianyizheng02 (commented):
> This commit refactors the linked list code to improve its structure and readability. The code has been split into separate classes for the node and linked list, making it easier to understand and modify.

I don't know what you mean by this, because this file already had separate classes for the Node and LinkedList. You didn't add that.

> The methods have been well-documented with comments to explain their functionality.

To be honest, I don't think most of these comments add much if any value. The comments mostly just talk through what the code is doing line by line, but a reader would've gotten that same information from just reading the code itself. I think it's better to _not_ have comments that are _this_ detailed in order to avoid clutter.

PR #8708, opened 2023-05-03
Added a new Python script to send message via telegram

NiyazUrazaev:
### Describe your change:

Added a new Python script to send message via telegram

* [X] Add a new feature?
* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/web_programming/telegram_message.py
+++ b/web_programming/telegram_message.py
@@ -5,6 +5,7 @@ import requests
 
 BOT_TOKEN = os.getenv("YOUR_BOT_TOKEN", "")
 
+
 def send_message_to_telegram(token: str, chat_id: str, message_text: str):
     url = f"https://api.telegram.org/bot{token}/sendMessage"
     data = {"chat_id": chat_id, "text": message_text}
--- a/web_programming/telegram_message.py
+++ b/web_programming/telegram_message.py
@@ -1,0 +1,18 @@
+# Created by NiyazUrazaev on 03/05/23
+
+import requests
+
+
+def send_message_to_telegram(token: str, chat_id: str, message_text: str):
+    url = f"https://api.telegram.org/bot{token}/sendMessage"
+    data = {"chat_id": chat_id, "text": message_text}
+    response = requests.post(url, data=data)
+    if not response.ok:
+        print("Failed to send a message in Telegram:", response.text)
+
+
+if __name__ == "__main__":
+    token = "<YOUR BOT TOKEN>"
+    chat_id = "<YOUR CHANNEL ID>"
+    message_text = "<YOUR MESSAGE BODY>"
+    send_message_to_telegram(token, chat_id, message_text)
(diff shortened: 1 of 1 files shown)


CaedenPH (changes requested):
In my opinion, this is not an algorithm, rather a script interacting with a specific api

cclauss:
We have more leniency on the "algorithms" in `web_programming` because they are difficult to write tests for and can be broken if the destination /backend changes.

web_programming/telegram_message.py
```diff
@@ -0,0 +1,18 @@
...
+
+import requests
+
+
+def send_message_to_telegram(token: str, chat_id: str, message_text: str):
+    url = f"https://api.telegram.org/bot{token}/sendMessage"
+    data = {"chat_id": chat_id, "text": message_text}
+    response = requests.post(url, data=data)
+    if not response.ok:
+        print("Failed to send a message in Telegram:", response.text)
```
cclauss, line 11: Please raise an Exception instead.  See CONTRIBUTING.md because algorithmic functions should not print().

web_programming/telegram_message.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+if __name__ == "__main__":
+    token = "<YOUR BOT TOKEN>"
```
cclauss, line 15: It is unsafe to hardcode a token into a Python script so please get it from an environment variable.  See other examples in this directory.

cclauss (changes requested):

web_programming/telegram_message.py
NiyazUrazaev: Thanks, had been fixed

web_programming/telegram_message.py
NiyazUrazaev: Thanks, had been fixed

tianyizheng02:
> We have more leniency on the "algorithms" in `web_programming` because they are difficult to write tests for and can be broken if the destination /backend changes.

@cclauss I think @CaedenPH's point is that this PR isn't substantial enough to qualify as an algorithm in the first place—not that the code itself is low-quality.

Personally, I agree with @CaedenPH here: there's very little that distinguishes this code as "an algorithm for sending Telegram messages" rather than "a generic script for making HTTP POST requests" (for comparison, see the contributor's other PR, #8793). It also doesn't meet [our criteria](https://github.com/TheAlgorithms/Python/blob/4f2a346c277076ce1d69578ef52a9766e5040176/CONTRIBUTING.md) for an algorithm anyway:
> An Algorithm is one or more functions (or classes) that:
> - take one or more inputs,
> - perform some internal calculations or data manipulations,
> - return one or more outputs,
> - have minimal side effects (Ex. `print()`, `plot()`, `read()`, `write()`).

I'm not sure if what the code does can be considered "calculations" or "data manipulations". It doesn't really return any values since it's an HTTP POST request, and it consists almost entirely of side effects.

PR #8719, opened 2023-05-09
Fix #8715 Update the remove func of binary search tree.

### LINKED ISSUE #8715: Removing the root node in binary_search_tree.py removes the whole tree.
### Repository commit

73105145090f0ce972f6fa29cc5d71f012dd8c92

Python 3.10.6

astroid==2.15.4
...

### Expected behavior

https://github.com/TheAlgorithms/Python/blob/master/data_structures/binary_tree/binary_search_tree.py

Removing the root node should not delete the whole tree, it should behave just like removing any other nodes. It works fine if we delete other nodes but I guess because of the [this line](https://github.com/TheAlgorithms/Python/blob/73105145090f0ce972f6fa29cc5d71f012dd8c92/data_structures/binary_tree/binary_search_tree.py#L43), it has problem when we remove the root node. There should be other checks as well.

### Actual behavior

```python
>>> t = BinarySearchTree()
>>> t.insert(1)
>>> t.insert(2)
>>> t.insert(3)
>>> t.remove(1)
>>> str(t)
'None'
>>> 
```

yabea:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/binary_tree/binary_search_tree.py
+++ b/data_structures/binary_tree/binary_search_tree.py
@@ -40,7 +40,7 @@ class BinarySearchTree:
             else:
                 node.parent.left = new_children
         else:
-            self.root = None
+            self.root = new_children
 
     def is_right(self, node: Node) -> bool:
         if node.parent and node.parent.right:
--- a/maths/euclidean_distance.py
+++ b/maths/euclidean_distance.py
@@ -1,12 +1,11 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
-from typing import Union
 
 import numpy as np
 
-Vector = Union[Iterable[float], Iterable[int], np.ndarray]
-VectorOut = Union[np.float64, int, float]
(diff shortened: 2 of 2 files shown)


rohan472000 (changes requested):
To resolve `ruff` issue, you need to modify the type annotations in lines 8 and 9 of the "maths/euclidean_distance.py" file. Instead of the current syntax, you should use the `X | Y` syntax.

rohan472000:
Actually this error has been seen in a couple of PRs, even when there is no code written in that context, @cclauss plz look into it.

yabea:
@rohan472000 I've fixed ruff error, but in fact, I also error when I pre-commit locally, maybe someone needs to check it out

tianyizheng02:
@cclauss This should be closed now that #8752 has been merged and the issue has been fixed

PR #8721, opened 2023-05-11
The input can be either comma or space separated value -- Improved user experience

Prade7:
### Describe your change:

The code change was done in binary_search.py file and it will allow user to give input as comma separated or space separated value as input .

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/searches/binary_search.py
+++ b/searches/binary_search.py
@@ -298,11 +298,15 @@ def binary_search_by_recursion(
 
 
 if __name__ == "__main__":
-    user_input = input("Enter numbers separated by comma:\n").strip()
-    collection = sorted(int(item) for item in user_input.split(","))
+    user_input = input("Enter numbers separated by either comma or space \n").strip()
+    if("," in user_input):
+        collection = sorted(map(int,user_input.split(",")))
+    else:
+        collection=sorted(map(int,user_input.split(" ")))
     target = int(input("Enter a single number to be found in the list:\n"))
     result = binary_search(collection, target)
     if result is None:
         print(f"{target} was not found in {collection}.")
     else:
         print(f"{target} was found at position {result} in {collection}.")
+   
--- a/searches/binary_search.py
+++ b/searches/binary_search.py
@@ -299,14 +299,13 @@ def binary_search_by_recursion(
 
 if __name__ == "__main__":
     user_input = input("Enter numbers separated by either comma or space \n").strip()
-    if("," in user_input):
-        collection = sorted(map(int,user_input.split(",")))
+    if "," in user_input:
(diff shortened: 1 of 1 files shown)


Prade7:
I didn't change any other file . I don't know why it is failing

searches/binary_search.py
```diff
@@ -298,8 +298,11 @@ def binary_search_by_recursion(
 
 
 if __name__ == "__main__":
-    user_input = input("Enter numbers separated by comma:\n").strip()
-    collection = sorted(int(item) for item in user_input.split(","))
+    user_input = input("Enter numbers separated by either comma or space \n").strip()
+    if "," in user_input:
```
tianyizheng02, line 302: This would break if there were both commas and spaces in the input, i.e. `1,000 1,001 1,002`

searches/binary_search.py
```diff
@@ -298,8 +298,11 @@ def binary_search_by_recursion(
...
+        collection = sorted(map(int, user_input.split(",")))
```
tianyizheng02, line 303:
```suggestion
        collection = sorted(int(num) for num in user_input.split(","))
```

searches/binary_search.py
```diff
@@ -298,8 +298,11 @@ def binary_search_by_recursion(
...
+    else:
+        collection = sorted(map(int, user_input.split(" ")))
```
tianyizheng02, line 305:
```suggestion
        collection = sorted(int(num) for num in user_input.split(" "))
```

tianyizheng02, line 305:
```suggestion
        collection = sorted(int(num) for num in user_input.split(" "))
```

tianyizheng02 (changes requested):
The current change will parse the input incorrectly if there are commas in the numbers themselves, so that case needs to be handled

tianyizheng02:
> I didn't change any other file . I don't know why it is failing

It's failing because `ruff` found errors in some other files that was edited by a previous contributor. Try running `ruff check --fix .` to see if that can automatically fix the errors.

searches/binary_search.py
tianyizheng02: Just believing that the user won't provide input in a certain way generally isn't a good way to deal with user input. If it's something that a user might reasonably do then we should still handle it somehow. At the very least, you could just prohibit the user from using both commas and spaces at the same time.

tianyizheng02:
I'm closing this PR because, as I've pointed out already, it doesn't properly handle both commas and spaces. Unfortunately, we're currently trying to clear the massive backlog of open PRs, and we don't have the time to completely rework this PR.

PR #8741, opened 2023-05-16
Updated and solved the issue in circular_linked_list_list.py

PranavYewale:
Added a length attribute to the CircularLinkedList class to track the length of the linked list.

Updated the __len__ method to return the length attribute instead of calculating the length using a loop. This change improves the space complexity of the __len__ method to O(1).

Updated the insert_tail and insert_head methods to update the length attribute directly instead of calling the len function. This change ensures that the length is accurately maintained.

Updated the insert_nth method to include a check for the index range and raise an IndexError if the index is out of range. This provides better error handling.

Added a delete_tail method that deletes the last node of the linked list. This method calls the delete_nth method with the index set to self.length - 1 and the update_length flag set to True. This ensures that the length is correctly updated when the last node is deleted.

Updated the delete_nth method to include an update_length parameter. When this parameter is set to True, the method will update the length attribute after deleting the node.


amirsoroush:
Although I'm agree that keeping track of the length would give us O(1) time complexity in those operations, this idea was rejected in the past. You may want to look at the comments [here](https://github.com/TheAlgorithms/Python/issues/5315) and [here](https://github.com/TheAlgorithms/Python/pull/5320).

rohan472000 (changes requested):
Look in `ruff` issue, to fix that use `ruff . --fix` before pushing.

tianyizheng02 (commented):
Ignoring the arguments for and against having a length attribute, this PR breaks things in other significant ways: removing _all_ imports, removing a required class, removing the test function, and removing the `__main__` block.

PR #8755, opened 2023-05-20
Mish activation function

Mitra-babu:
### Describe your change:
Mish activation function is added

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


neural_network/activation_functions/mish_activation.py
```diff
@@ -0,0 +1,48 @@
...
+        array([ 0.86509839,  8.99955208, 10.99992663, -1.93211787])
+
+
+        >>> mish_activation(np.array([8,2,-0.98,13]))
+        array([14.9999982 ,  2.94395896, -2.28214659, 25.        ])
+
+
+    """
+    soft_plus = np.log(np.exp(vector) + 1)
+    return vector * (2 / (1 + np.exp(-2 * soft_plus))) - 1
```
tianyizheng02, line 42: Given that we already have an implementation of the $\tanh$ function in `maths/tanh.py`, why not use it and simplify the code?
```py
from maths.tanh import tangent_hyperbolic as tanh

...

return vector * tanh(soft_plus)
```

tianyizheng02 (changes requested):

neural_network/activation_functions/mish_activation.py
Mitra-babu: Thank you for the suggestion. we can use it

Issue #8770, opened 2023-05-26
i found another typo and problem greedy_best_first


JadeKim042386:
### 1. Typo

`self.target.pos_y` and `self.target.pos_x` are reversed.

https://github.com/TheAlgorithms/Python/blob/a17791d022bdc942c8badabc52307c354069a7ae/graphs/greedy_best_first.py#L145-L154

parameter of Node is below
https://github.com/TheAlgorithms/Python/blob/a17791d022bdc942c8badabc52307c354069a7ae/graphs/greedy_best_first.py#L38-L46

The reason why it worked well is that `self.target.pos_y`(6) and `self.target.pos_x`(6) are the same.

### 2. Class equality

`__eq__` is not implement in `Node` class.
https://github.com/TheAlgorithms/Python/blob/6f21f76696ff6657bff6fc2239315a1650924190/graphs/greedy_best_first.py#L20-L56

so the below `child_node not in self.open_nodes` or ` child_node not in self.open_nodes` did't work.
https://github.com/TheAlgorithms/Python/blob/6f21f76696ff6657bff6fc2239315a1650924190/graphs/greedy_best_first.py#L105-L114

### 3. node in self.open_nodes is always better_node

How can a previous path cost more than the next one? A node that has already entered self.open_nodes is always less expensive than the next node.
So I checked that it works well even if I simply modify it as below.
```python
if child_node not in self.open_nodes:
    self.open_nodes.append(child_node)
# delete else
```

## Additional information

I think `sort()` and `pop(0)` are unnecessary. I think it's rather to increase the complexity of time.

PR #8771, opened 2023-05-26
fix graphs/greedy_best_first

JadeKim042386:
### Describe your change:

#8770 

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


tianyizheng02:
There doesn't seem to have been any tests for the individual class methods. We definitely wanna avoid letting such code mistakes go unnoticed/untested, so could you add some tests to the file? Preferably testing various example inputs for each method in order to avoid having a bug pass through due to some oversight in the examples.

JadeKim042386:
i added some tests and found other problems.
i renewed #8770, so please check it.

JadeKim042386:
😱 i.....i made a mistake. i pushed it wrong

PR #8774, opened 2023-05-27
Update README.md

Navraj88:
fixed the ci badge


CaedenPH (changes requested):
This pull request does not fix anything

It changes the image from
pre-commit
to 
pre-commit

Which would result in two of the same badges as we already have a ci passing

tianyizheng02 (commented):
Same as what @CaedenPH said, this would just lead to us having two of the same badge

PR #8778, opened 2023-05-27
adding a max_sectors_of_circle algorithm

meg-1:
### Describe your change:
adding a **max_sectors_of_circle** algorithm, which returns the maximum amount of sectors a circle can be divided by if cut 'num_cuts' times

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


maths/max_sectors_of_circle.py
```diff
@@ -0,0 +1,21 @@
...
+    265.375
+    >>> max_sectors_of_circle(-222)
+    -1
+    """
+
+    return ((num_cuts + 2 + num_cuts**2) * (1 / 2)) if num_cuts >= 0 else -1
+
+
+if __name__ == "__main__":
+    __import__("doctest").testmod()
```
amirsoroush, line 21:
```suggestion
    import doctest
    
    doctest.testmod()
```

rohan472000:
@meg-1 , While your solution does provide the expected output and solves the problem, we strive to maintain a repository that focuses on comprehensive algorithms and functions rather than individual one-liners.

Considering the nature of circle-related calculations, I agree that consolidating multiple circle-related questions and algorithms into a dedicated class or module would be a more organized approach. This approach would allow us to cover various aspects of circles, such as sectors, arcs, angles, and segments, while providing detailed explanations and examples.

By adopting this approach, we can ensure that the repository contains meaningful and informative algorithms that cater to a wider range of circle-related concepts. It will also facilitate easier navigation and exploration of circle-related topics for users who are interested in learning about different aspects of circles.

maths/max_sectors_of_circle.py
```diff
@@ -0,0 +1,23 @@
...
+    >>> max_sectors_of_circle(54)
+    1486.0
+    >>> max_sectors_of_circle(7)
+    29.0
+    >>> max_sectors_of_circle(22.5)
+    265.375
+    >>> max_sectors_of_circle(-222)
+    -1
+    """
+
```
rohan472000, line 16: Consider adding a check for the input type of `num_cuts` to ensure it is a numeric value. You can use the `isinstance` function to check if `num_cuts` is a float or an integer. This will help prevent potential type errors when the function is used.

rohan472000 (changes requested):

meg-1:
> @meg-1 , While your solution does provide the expected output and solves the problem, we strive to maintain a repository that focuses on comprehensive algorithms and functions rather than individual one-liners.
> 
> Considering the nature of circle-related calculations, I agree that consolidating multiple circle-related questions and algorithms into a dedicated class or module would be a more organized approach. This approach would allow us to cover various aspects of circles, such as sectors, arcs, angles, and segments, while providing detailed explanations and examples.
> 
> By adopting this approach, we can ensure that the repository contains meaningful and informative algorithms that cater to a wider range of circle-related concepts. It will also facilitate easier navigation and exploration of circle-related topics for users who are interested in learning about different aspects of circles.

Thank you for your time, detailed answer and explanation! I absolutely agree, it would be much more convenient for users to be able to access the information in a collected and structured way. Considering this, do i understand correctly, you are suggesting this example be rewritten with the usage of classes? Or rather, are you suggesting that it'd be a good idea to collect all algorithms in this repo, connected with calculations of various aspects of circles, and to push them all in this commit as a module? Thank you for your answer in advance!

rohan472000:
yes , You got it right, You should find out all possible or potential question around circles or any polygon(if interested to do), make a class for that polygon under which u will keep different methods to solve those problems.

rohan472000:
@cclauss is it good to do like above , making class for a polygon to solve its related problems?

maths/max_sectors_of_circle.py
meg-1: Thank you for your suggestion! Already implemented and added tests for this :)

maths/max_sectors_of_circle.py
```diff
@@ -0,0 +1,28 @@
...
+    >>> max_sectors_of_circle(-222)
+    -1
+    >>> max_sectors_of_circle("-222")
+    Traceback (most recent call last):
+    TypeError: num_cuts must be a numeric value.
+    """
+
+    if not isinstance(num_cuts, (int, float)):
+        raise TypeError("num_cuts must be a numeric value.")
+    return ((num_cuts + 2 + num_cuts**2) * (1 / 2)) if num_cuts >= 0 else -1
```
cclauss, line 22: Would this produce identical results?  It should be faster than doing the calculation each time.
```suggestion
    return ((num_cuts + 2 + num_cuts**2) * .5) if num_cuts >= 0 else -1
```

cclauss (changes requested):
@dhruvmanila Your review on the failing tests?

maths/max_sectors_of_circle.py
meg-1: Most definitely! Thank you very much for the review. Committed the suggestion just now :))

PR #8787, opened 2023-05-30
Updated postfix_evaluation.py to support Unary operators

### LINKED ISSUE #8724: Duplicate solutions for postfix notation evaluation.
### Feature description

Hi,
I noticed there are two files for evaluating postfix notation which are:

1. [...stacks/evaluate_postfix_notations.py](https://github.com/TheAlgorithms/Python/blob/master/data_structures/stacks/evaluate_postfix_notations.py)
2. [...stacks/postfix_evaluation.py](https://github.com/TheAlgorithms/Python/blob/master/data_structures/stacks/postfix_evaluation.py)

The latter has bunch of print statements explaining the steps which is pretty useful for educational purposes but not in production. (It can be changed to have a `verbose=True` parameter for prints for example.)

Is having these two files intentional or we should combine them?

### LINKED ISSUE #8754: Need to support for unary operator in postfix evaluation.
### Feature description

As mentioned in #8724, we have two files for evaluating postfix expressions. But currently both of them only work with binary operators and will raise an exception when an unary operator is given.

example:
```python
>>> res = evaluate_postfix(["2", "-", "3", "+"])
  File "<stdin>", line 1, in <module>
  File ".../stacks/evaluate_postfix_notations.py", line 31, in evaluate_postfix
    b, a = stack.pop(), stack.pop()
IndexError: pop from empty list
>>> 
```
`2-3+` is a postfix notation for `-2+3` infix and should result in `1`.

arijitde92:
Fixes #8754
Fixes #8724

### Describe your change:
* [ ] Add an algorithm?
* [ ] Documentation change?
Fixes #8754 and #8724

### Checklist:

Updated postfix_evaluation.py to support Unary operators and floating point numbers. Fixes #8754.
Also merged evaluate_postfix_notations.py and postfix_evaluation.py into postfix_evaluation.py which fixes #8724.


CaedenPH:
@arijitde92 Put Fixes #8754 and #8724 in the pull request description

arijitde92:
Hi @CaedenPH , @amirsoroush , Please review my PR and suggest changes if any.  postfix_evaluation.py now works with unary operators and floating point numbers.
Plesee see example output below-

```
Enter a Postfix Expression (space separated): 2 1 + 3 *
Do you want to see stack contents while evaluating? [y/N]: 
Result =  9
Do you want to enter another expression? [y/N]: y
Enter a Postfix Expression (space separated): 4 13 5 / +
Do you want to see stack contents while evaluating? [y/N]: y
 Symbol  |    Action    | Stack
-----------------------------------
       4 | push(4)      | [4]
      13 | push(13)     | [4, 13]
       5 | push(5)      | [4, 13, 5]
         | pop(5)       | [4, 13]
         | pop(13)      | [4]
       / | push(13/5)   | [4, 2.6]
         | pop(2.6)     | [4]
         | pop(4)       | []
       + | push(4+2.6)  | [6.6]
Result =  6.6
Do you want to enter another expression? [y/N]: y
Enter a Postfix Expression (space separated): 2 - 3 +
Do you want to see stack contents while evaluating? [y/N]: y
 Symbol  |    Action    | Stack
----------------------------------
       2 | push(2)      | [2]
         | pop(2)       | [-2]
       - | push(-2)     | [-2]
       3 | push(3)      | [-2, 3]
         | pop(3)       | [-2]
         | pop(-2)      | []
       + | push(-2+3)   | [1]
Result =  1
Do you want to enter another expression? [y/N]: 

Process finished with exit code 0

```

cclauss:
Seems like a lot more code to solve the same problem.  Is it faster?

umairefiroz:
Hi Christian,

Yes you're right!
There really are various vaguely available set of libraries in order fix
this issue via simple set of codes.

Let me know if you require any additional information in order to resolve
this issue.

If need any help feel free to contact me. 😀

Thanks and best regards,
Umaire Firoz
⌨️/🐍

arijitde92:
> Seems like a lot more code to solve the same problem. Is it faster?
Hi @cclauss ,
I haved added extra functions and extra code to handle edge cases. For example-
`get_number` is used to simultaneously check whether the string is a number or not. In the current version it is checked by [.isdigit()](https://github.com/TheAlgorithms/Python/blob/3a9e5fa5ecea0df54ed3ffdcb74f46171199f552/data_structures/stacks/postfix_evaluation.py#L39) but floating point numbers cannot be detected by isdigit. The current postfix evaluation can only handle integer numbers.
`is_operator` is simply a helper function such that the program does'nt crash when it gets some other characters apart from numbers and mathematical operators (+,-,*,/,^) and prompts appropriate error message.
The `evaluate()` method is basically the [solve()](https://github.com/TheAlgorithms/Python/blob/3a9e5fa5ecea0df54ed3ffdcb74f46171199f552/data_structures/stacks/postfix_evaluation.py#L23) method of the current implementation with additional if else block to handle wrong input and unary operators. I have not changed the postfix evaluation logic.
`is_yes` is just a helper function as I have made the program a little bit more interactive in the main method by promtping for more expressions and asking choice for the user. If you want I can remove this part.

I am new to open source contribution. Please guide me on what else I can do to make the code shorter or faster.
Thanks.

arijitde92:
Hi @CaedenPH @cclauss @darkstar @amirsoroush, is there anything else I can do to improve the code? Could you please review the PR and let me know if any changes are required?

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+    Returns
+    -------
+    bool
+        True if data is an operator else False.
+    """
+    return data in ["-", "+", "*", "^", "/"]
 
 
-def solve(post_fix):
+def evaluate(post_fix: list, verbose: bool = False) -> int | float | str | None:
```
tianyizheng02, line 77:
```suggestion
def evaluate(post_fix: list[str], verbose: bool = False) -> int | float | str | None:
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+        "*": lambda p, q: p * q,
+        "/": lambda p, q: p / q,
+        "+": lambda p, q: p + q,
+        "-": lambda p, q: p - q,
     }  # operators & their respective operation
+    if len(post_fix) > 0:
+        if verbose:
+            # print table header
+            print("Symbol".center(8), "Action".center(12), "Stack", sep=" | ")
+            print("-" * (30 + len(post_fix)))
```
tianyizheng02, line 115: Negate the if-statement to return early. That way you can avoid nesting the code as much.
```suggestion
    if len(post_fix) == 0:
        return 0
    if verbose:
        # print table header
        print("Symbol".center(8), "Action".center(12), "Stack", sep=" | ")
        print("-" * (30 + len(post_fix)))
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+                    print(
+                        str(x).rjust(8),
+                        ("push(" + str(x) + ")").ljust(12),
+                        stack,
+                        sep=" | ",
+                    )
+            elif is_operator(x):
+                # If only 1 value is inside stack and + or - is encountered
+                # then this is unary + or - case
+                if x in ["-", "+"] and len(stack) < 2:
```
tianyizheng02, line 131:
```suggestion
                if x in {"-", "+"} and len(stack) < 2:
```
`{"-", "+"}` should also be refactored into a constant for readability

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+                    b = stack.pop()  # pop stack
+                    if x == "-":
+                        stack.append(-b)  # negate b and push again into stack
+                    else:  # when x is unary +
+                        stack.append(b)
```
tianyizheng02, line 136:
```suggestion
                    if x == "-":
                        b *= -1  # negate b
                    stack.append(b)
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+                            sep=" | ",
+                        )
+                        print(
+                            str(x).rjust(8),
+                            ("push(" + str(x) + str(b) + ")").ljust(12),
+                            stack,
+                            sep=" | ",
+                        )
+                else:
+                    b = stack.pop()  # pop stack
```
tianyizheng02, line 152: You can use continue statements to avoid nesting code inside else blocks:
```suggestion
                        )
                    continue
                
                # x is a binary operator
                b = stack.pop()  # pop stack
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+                        # output in tabular format
+                        print(
+                            str(x).rjust(8),
+                            ("push(" + str(a) + str(x) + str(b) + ")").ljust(12),
+                            stack,
+                            sep=" | ",
+                        )
+            else:
+                print(f"{x} is neither a number, nor a valid operator")
+                break
```
tianyizheng02, line 183:
```suggestion
            else:
                msg = f"{x} is neither a number nor a valid operator"
                raise ValueError(msg)
```
Just raise an error if the input is invalid

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+        # If everything executed correctly,the stack will contain
```
tianyizheng02, line 184:
```suggestion
        # If everything executed correctly, the stack will contain
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+        # only one element which is the result
+        result = stack[0] if len(stack) == 1 else None
+        return result
```
tianyizheng02, line 187: If `len(stack) != 1` by the end of the computation, then it means that the program couldn't process the whole string, likely because it's not a valid postfix string. Thus, it should be fine to just raise an error:
```suggestion
        if len(stack) != 1:
            raise ArithmeticError("Input is not a valid postfix expression")
        return stack[0]
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
-                sep=" | ",
-            )
-
-    return int(stack[0])
+    Returns
+    -------
+    bool
+        True if Yes, otherwise False
+    """
+    return val in ["Y", "y"]
```
tianyizheng02, line 206:
```suggestion
    return val in {"Y", "y"}
```
Use a set

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
+    # Creating a loop so that user can evaluate postfix expression multiple times
+    while loop:
+        expression = input("Enter a Postfix Expression (space separated): ").split(" ")
+        choice = input("Do you want to see stack contents while evaluating? [y/N]: ")
+        display = is_yes(choice)
+        output = evaluate(expression, display)
+        if output is not None:
+            print("Result = ", output)
+        choice = input("Do you want to enter another expression? [y/N]: ")
+        loop = is_yes(choice)
```
tianyizheng02, line 220: You can just use an infinite loop with a break to avoid the need for an extra helper variable:
```suggestion
    # Creating a loop so that user can evaluate postfix expression multiple times
    while True:
        expression = input("Enter a Postfix Expression (space separated): ").split(" ")
        choice = input("Do you want to see stack contents while evaluating? [y/N]: ")
        display = is_yes(choice)
        output = evaluate(expression, display)
        print("Result = ", output)
        choice = input("Do you want to enter another expression? [y/N]: ")
        if is_yes(choice):
             break
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,197 @@
...
     }  # operators & their respective operation
+    if len(post_fix) > 0:
+        if verbose:
+            # print table header
+            print("Symbol".center(8), "Action".center(12), "Stack", sep=" | ")
+            print("-" * (30 + len(post_fix)))
+        for x in post_fix:
+            is_number, x = get_number(x)
+            if is_number:  # if x is a number (integer, float)
+                stack.append(x)  # append x to stack
```
tianyizheng02, line 119: Wouldn't it be easier to just convert all the numbers in the input beforehand instead of converting them one by one?

tianyizheng02 (changes requested):
General feedback: pay attention to where you can invert conditions or use early breaks/returns in order to simplify the control flow and avoid heavily nested code

data_structures/stacks/postfix_evaluation.py
arijitde92: Yes I have changed the code. See d6874ad24fa963de962d2bb700411d2a35f4f2b5. Now it first converts all the numbers (which also serves as a check for valid postfix expression), then begins the actual postfix evaluation.

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,211 @@
...
-    # print table header
-    print("Symbol".center(8), "Action".center(12), "Stack", sep=" | ")
-    print("-" * (30 + len(post_fix)))
-
+    if len(post_fix) == 0:
+        return 0
+    # Checking the list to find out whether the postfix expression is valid
     for x in post_fix:
-        if x.isdigit():  # if x in digit
+        valid_expression.append(get_number(x))
```
tianyizheng02, line 124:
```py
opr = {
    "^": lambda p, q: p**q,
    "*": lambda p, q: p * q,
    "/": lambda p, q: p / q,
    "+": lambda p, q: p + q,
    "-": lambda p, q: p - q,
}  # operators & their respective operation

if len(post_fix) == 0:
    return 0

# Checking the list to find out whether the postfix expression is valid
valid_expression = [get_number(token) for token in post_fix]
```

1) A list comprehension makes this step really easy
2) I recommend changing the name of `get_number` to something more accurate (see previous comment)

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,211 @@
...
-                ("push(" + a + x + b + ")").ljust(12),
-                ",".join(stack),
+                str(x).rjust(8),
+                ("push(" + str(a) + str(x) + str(b) + ")").ljust(12),
+                stack,
                 sep=" | ",
             )
+        # else:
+        #     msg = f"{x} is neither a number nor a valid operator"
+        #     raise ValueError(msg)
```
tianyizheng02, line 195:
```suggestion
```
Commented-out code

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,211 @@
...
+    >>> evaluate(["2", "1", "+", "3", "*"])
+    9
+    >>> evaluate(["4", "13", "5", "/", "+"])
+    6.6
+    >>> evaluate(["2", "-", "3", "+"])
+    1
+    >>> evaluate(["-4", "5", "*", "6", "-"])
+    -26
+    >>> evaluate([])
+    0
```
tianyizheng02, line 94: Can we get some doctests that test _invalid_ postfix expressions?

tianyizheng02 (changes requested):

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,211 @@
...
+        try:
+            return float(data)
+        except ValueError:
+            if is_operator(data):
+                return data
+            msg = f"{data} is neither a number nor a valid operator"
+            raise ValueError(msg)
+
+
+def is_operator(data: str | int | float) -> bool:
```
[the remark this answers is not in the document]
arijitde92: Hi @tianyizheng02, `is_operator` can have float inputs apart from strings as it takes each token from the `valid expression` list. The `valid_expression` list is made using the `parse_token` function which returns strings if it's an operator, or a float if it's a number as shown below-
image
.
Hence it is necessary to give the float type along with str.

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,211 @@
...
+
+    Returns
+    -------
+    bool
+        True if data is an operator else False.
+    """
+    return data in BINARY_OP_SYMBOLS
+
+
+def evaluate(post_fix: list[str], verbose: bool = False) -> int | float | str | None:
```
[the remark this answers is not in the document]
arijitde92: Yes, this would never return a string or None. But mypy test is failing because apparently it thinks the stack[0] (which is being returned) is made from both float and string as x is appended to the stack and x can be both float and string. See snippet below-
image
.
The mypy test passes when I cast `stack[0]` to float before returning. Shall I procced by casting it to float then?

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,211 @@
...
+    # Checking the list to find out whether the postfix expression is valid
     for x in post_fix:
-        if x.isdigit():  # if x in digit
+        valid_expression.append(get_number(x))
+    if verbose:
+        # print table header
+        print("Symbol".center(8), "Action".center(12), "Stack", sep=" | ")
+        print("-" * (30 + len(post_fix)))
+    for x in valid_expression:
+        if not is_operator(x):
```
cclauss, line 130: We do not need `is_operator()` if we use:
```suggestion
        if x not in opr:  # if x is not an operator
```

data_structures/stacks/postfix_evaluation.py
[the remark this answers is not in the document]
tianyizheng02: Yeah, I think casting to float before returning would be best.

data_structures/stacks/postfix_evaluation.py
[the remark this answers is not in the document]
tianyizheng02: I see, that makes sense. In that case, the function header would be
```suggestion
def is_operator(token: str | float) -> bool:
```
1) No `int` because I believe ruff now enforces `int | float` -> `float`
2) Let's rename the variable to `token` to make it clearer that the input is a token (which is allowed to be a number)

tianyizheng02 (changes requested):

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,210 @@
...
+    try:
+        return int(data)
+    except ValueError:
+        try:
+            return float(data)
+        except ValueError:
+            if is_operator(data):
+                return data
+            msg = f"{data} is neither a number nor a valid operator"
+            raise ValueError(msg)
```
tianyizheng02, line 61:
```suggestion
    if is_operator(data):
        return data
    try:
        return int(data)
    except ValueError:
        try:
            return float(data)
        except ValueError:
            msg = f"{data} is neither a number nor a valid operator"
            raise ValueError(msg)
```
We can un-nest the operator check

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,210 @@
...
+
+
+def is_operator(token: str | float) -> bool:
+    """
+    Checks whether a given input is one of the valid operators or not.
+    Valid operators being '-', '+', '*', '^' and '/'.
+
+    Parameters
+    ----------
+    token : str
```
tianyizheng02, line 71:
```suggestion
    token : str or float
```

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,210 @@
...
+    ----------
+    data : str
+        The data which needs to be converted to the appropriate number
+
+    Returns
+    -------
+    bool,  int or float
+        Returns a tuple of (a, b) where 'a' is True if data is indeed a number (integer
+        or numeric) and 'b' is either an integer of a floating point number.
+        If 'a' is False, then b is 'data'
```
tianyizheng02, line 50: This needs to be updated
```suggestion
    int, float, or str
        Returns an int or float if `data` is a number and returns `data` if it's an operator
```

tianyizheng02 (changes requested):
Some minor tweaks, but otherwise LGTM

data_structures/stacks/postfix_evaluation.py
arijitde92: I have added an invalid postfix expression that raises ArithmeticError in the latest commit.

data_structures/stacks/postfix_evaluation.py
arijitde92: I have removed the int check and now it returns only float if number else a string (if operator).

data_structures/stacks/postfix_evaluation.py
arijitde92: I have updated it specifying it returns float if `token` is a number and returns `token`  if it's an operator.

data_structures/stacks/postfix_evaluation.py
tianyizheng02: Sorry, just realized that I added this suggestion to the wrong line—I meant for it to be for the `is_operator` function

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,205 @@
...
+
+def parse_token(token: str | float) -> float | str:
+    """
+    Converts the given data to appropriate number if it is indeed a number, else returns
+    the data as it is with a False flag. This function also serves as a check of whether
+    the input is a number or not.
+
+    Parameters
+    ----------
+    token : str or float
```
tianyizheng02, line 42:
```suggestion
def parse_token(token: str) -> float | str:
    """
    Converts the given data to appropriate number if it is indeed a number, else returns
    the data as it is with a False flag. This function also serves as a check of whether
    the input is a number or not.

    Parameters
    ----------
    token : str
```
I meant for [this suggestion](https://github.com/TheAlgorithms/Python/pull/8787#discussion_r1302655963) to be for `is_operator`, not this function. I believe this function is only used when initially parsing the input list, so all tokens passed into this function should be strings.

data_structures/stacks/postfix_evaluation.py
tianyizheng02: Each element in the token list `valid_expression` is passed into this function to check if it's an operator—those elements may be numbers.

tianyizheng02 (changes requested):

data_structures/stacks/postfix_evaluation.py
arijitde92: Yes understood. I have made the necessary changes.

data_structures/stacks/postfix_evaluation.py
```diff
@@ -17,52 +24,205 @@
...
+    """
+    return token in BINARY_OP_SYMBOLS
+
+
+def evaluate(post_fix: list[str], verbose: bool = False) -> float:
+    """
+    Function that evaluates postfix expression using a stack.
+    >>> evaluate(["2", "1", "+", "3", "*"])
+    9.0
+    >>> evaluate(["4", "13", "5", "/", "+"])
```
cclauss, line 82: Please add doctests for:
* [ ] evaluate(["0"])
* [ ] evaluate(["-0"])
* [ ] evaluate(["1"])
* [ ] evaluate(["-1"])
* [ ] evaluate(["-1.1"])
* [ ] evaluate(["2", "1.9", "+", "3", "*"])
* [ ] evaluate(["2", "-1.9", "+", "3", "*"])
* [ ] evaluate(["2", "--1.9", "+", "3", "*"])

cclauss (approved):

PR #8796, opened 2023-06-02
add disjoint_set using list instead of node

JadeKim042386:
### Describe your change:

* [X] Add an algorithm?
* [ ] Documentation change?

### Checklist:


data_structures/disjoint_set/disjoint_set_2.py
```diff
@@ -0,0 +1,91 @@
+n = 9
```
rohan472000, line 1: No docs is provided for this PR, you should look into other merged PRs for docs syntax.

data_structures/disjoint_set/disjoint_set_2.py
```diff
@@ -0,0 +1,91 @@
...
+
+parent = list(range(n))
+
+rank = [0] * n
+
+
+def union_set(x: int, y: int) -> None:
```
rohan472000, line 8: The `union_set` function has a logic error. It checks if x is equal to y before performing any operations, but it should be checking if `find_set(x)` is equal to `find_set(y)`. Replace `if x == y:` with `if find_set(x) == find_set(y):`.

data_structures/disjoint_set/disjoint_set_2.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Union of two sets.
+    set with bigger rank should be parent, so that the
+    disjoint set tree will be more flat.
+    """
```
rohan472000, line 13: Add doctests for every function.

data_structures/disjoint_set/disjoint_set_2.py
```diff
@@ -0,0 +1,91 @@
...
+            else:
+                assert find_set(node0) == find_set(node1)
+
+    # compressed path
+    print("parent:", parent)
+    print("rank:", rank)
+
+
+if __name__ == "__main__":
+    test_disjoint_set()
```
rohan472000, line 91: add this in main - `doctest.testmod()` after adding doctests in every function

rohan472000 (changes requested):

data_structures/disjoint_set/disjoint_set_2.py
madhavgaba: +1

888lilili: 11111

tianyizheng02 (commented):
I don't feel that this adds much value over the existing implementations. [data_structures/disjoint_set/alternate_disjoint_set.py](https://github.com/TheAlgorithms/Python/blob/9d86d4edaa754af06e0da9cac4a717f3765db7f4/data_structures/disjoint_set/alternate_disjoint_set.py) already implements disjoint set using lists, and this implementation specifically is a near-copy of [data_structures/disjoint_set/disjoint_set.py](https://github.com/TheAlgorithms/Python/blob/9d86d4edaa754af06e0da9cac4a717f3765db7f4/data_structures/disjoint_set/disjoint_set.py):
```py
# This PR
def union_set(x: int, y: int) -> None:
    """
    Union of two sets.
    set with bigger rank should be parent, so that the
    disjoint set tree will be more flat.
    """
    x, y = find_set(x), find_set(y)
    if x == y:
        return

    if rank[x] > rank[y]:
        parent[y] = x
    else:
        parent[x] = y
        if rank[x] == rank[y]:
            rank[x] += 1  # Also, I think this might be a typo in the PR
```
```py
# data_structures/disjoint_set/disjoint_set.py
def union_set(x: Node, y: Node) -> None:
    """
    Union of two sets.
    set with bigger rank should be parent, so that the
    disjoint set tree will be more flat.
    """
    x, y = find_set(x), find_set(y)
    if x == y:
        return

    elif x.rank > y.rank:
        y.parent = x
    else:
        x.parent = y
        if x.rank == y.rank:
            y.rank += 1
```
@cclauss Do you think this change from Node to list adds enough value to consider merging this PR?

JadeKim042386:
@tianyizheng02 
As you said, it's not worth adding a new one when i look back at the comments after some time. If you think other reviewers are unnecessary, you may do close this pr.

PR #8798, opened 2023-06-03
Update max_sub_array.py

Abdul-16:
from __future__ import annotations line has been removed since it's not necessary in Python 3.x.


cclauss:
Please reread the docs... https://docs.python.org/3/library/__future__.html

PR #8800, opened 2023-06-04
tt

minjia828:
/wsp sum

### DIFF
--- a/server.py
+++ b/server.py
@@ -1,4 +1,5 @@
 # all imports - including #s
+#add newline by minjia 0604
 import socket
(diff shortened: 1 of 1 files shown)


server.py
```diff
@@ -40,4 +40,4 @@ def func4client():
     s.accept()
```
minjia828, line 41: ok

minjia828 (commented):
i am ok

tianyizheng02 (commented):
What?

Where is this file even from? What does it even do?

PR #8803, opened 2023-06-07
Speed of sound

ShivaDahal99:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


physics/speed_of_sound.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+
+Source : https://en.wikipedia.org/wiki/Speed_of_sound
+"""
+
+
+class SpeedOfSound:
+    @staticmethod
+    def fluid(fluid_density: float, bulk_modulus: float) -> float:
```
cclauss, line 29: We don't really use the class so it seems like unnecessary overhead.
```suggestion
def speed_of_sound_in_a_fluid(density: float, bulk_modulus: float) -> float:
```

physics/speed_of_sound.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+
+        if fluid_density <= 0:
+            raise ValueError("Impossible fluid density")
+        if bulk_modulus <= 0:
+            raise ValueError("Impossible bulk modulus")
+
+        speed_of_sound = (bulk_modulus / fluid_density) ** 0.5
+
+        return speed_of_sound
```
cclauss, line 50: Remove more unnecessary overhead...  The function name describes what we are returning to the caller.
```suggestion
        return (bulk_modulus / fluid_density) ** 0.5
```

cclauss (changes requested):
Nice!  Simple yet powerful.

CaedenPH (commented):
You have created an algorithm function that determines the speed of sound *in a fluid*, would it be possible to implement the same algorithm for the speed of sound in solids, vacuums, sea level static conditions, etc

PR #8824, opened 2023-06-16
Gradient boosting classifier

SannketNikam:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


CaedenPH (changes requested):
Please put this code inside functions

tianyizheng02 (changes requested):
From the [contributing guidelines](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md):
> Algorithms in this repo should not be how-to examples for existing Python packages. Instead, they should perform internal calculations or manipulations to convert input values into different output values. Those calculations or manipulations can use data types, classes, or functions of existing Python packages but each algorithm in this repo should add unique value.

This PR just uses sklearn's existing GradientBoostingClassifier class and doesn't implement any part of the actual gradient boosting classification algorithm. Please implement the algorithm yourself. You can use packages such as numpy to help with calculations, but you shouldn't use packages like sklearn that do all the hard work for you.

Issue #8820, opened 2023-06-14
Create optimization folder


imengus:
### Feature description

I have implemented the simplex algorithm in Python, but I can’t find the appropriate category to place it in.

The simplex algorithm finds the best value of a function, while restricted by inequality constraints. It enters the linear programming sub-field of mathematical optimisation. 

Other linear programming algorithms include the:
- Criss-cross,
- Ellipsoid,
- Projective, and
- Input sparsity time 
algorithms.

I considered putting simplex in `other`, `math`, and `machine learning`:

For `other`, mathematical optimisation is an organised discipline and has many algorithms which could be implemented in Python, deserving its own folder.

For `math`, it is too applied to fit in with the pure math algorithms there.

For `machine learning`, while simplex finds use in ML, so it does in operations research, engineering and physics. It would be false to say simplex is an ML algorithm.

CaedenPH:
@cclauss @dhruvmanila 
What are your opinions on the creation of this directory?

PR #8825, opened 2023-06-16
Simplex algorithm 

imengus:
### Describe your change:
Fixes: #8820
Implements the simplex algorithm for solving linear programs.
Adds `linear_programming` folder.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


linear_programming/simplex.py
```diff
@@ -0,0 +1,327 @@
...
+        ... [1,0,-0.125,0.375,1]
+        ... ])
+        >>> t = Tableau(tableau, 2)
+        >>> t.interpret_tableau(tableau, ["x1", "x2", "s1", "s2", "RHS"])
+        {'P': 5.0, 'x1': 1.0, 'x2': 1.0}
+        """
+        output_dict = {}
+
+        # P = RHS of final tableau
+        output_dict["P"] = abs(tableau[0, -1])
```
CaedenPH, line 301:
```suggestion
        # P = RHS of final tableau
        output_dict = {"P":  abs(tableau[0, -1])}
```

linear_programming/simplex.py
```diff
@@ -0,0 +1,327 @@
...
+        ['x1', 'x2', 's1', 's2', 's3', 'a1', 'RHS']
+        """
+        # decision | slack | artificial
+        string_starts = ["x", "s", "a"]
+        titles = []
+        for i in range(3):
+            for j in range(args[i]):
+                titles.append(string_starts[i] + str(j + 1))
+        titles.append("RHS")
+        return titles
```
CaedenPH, line 88: Much more consise
```suggestion
        return [string_starts[i] + str(j + 1) for j in range(args[i]) for i in range(3), "RHS"]
```

CaedenPH (changes requested):
1) Reformat all your doctests so that the `np.ndarray` utilises the full line length, as shown in one suggestion
2) In what scenarioes would you enter a different `tableau` into `pivot`, `find_pivot`, and other functions like this? In the doctests you just us `t.tableau` but if you refactor this to not accept a tableau input you can create more concise doctests such as `Tableau(np.array([[-2,-1,0,0,0], [3,1,1,0,6], [1,2,0,1,7.]]), 2).find_pivot()` without having to assign the variable `t`

linear_programming/simplex.py
imengus: This list comprehension doesn't work. For readability's sake, I'll stick with the for loops

imengus:
@cclauss here is the simplex algorithm I mentioned in the issue. Please could you review?

linear_programming/simplex.py
```diff
@@ -0,0 +1,318 @@
...
+            # If there are no more negative values in objective row
+            if self.stop_iter:
+                # Delete artificial variable columns and rows. Update attributes
+                self.tableau = self.change_stage(self.tableau)
+
+                # Pivot again
+                continue
+
+            self.tableau = self.pivot(self.tableau, row_idx, col_idx)
+            iter_num += 1
```
tianyizheng02, line 274:
```suggestion
        for _ in range(100):
            # Completion of each stage removes an objective. If both stages
            # are complete, then no objectives are left
            if not self.objectives:
                self.col_titles = self.generate_col_titles(
                    self.n_vars, self.n_slack, self.n_art_vars
                )

                # Find the values of each variable at optimal solution
                return self.interpret_tableau(self.tableau, self.col_titles)

            row_idx, col_idx = self.find_pivot(self.tableau)

            # If there are no more negative values in objective row
            if self.stop_iter:
                # Delete artificial variable columns and rows. Update attributes
                self.tableau = self.change_stage(self.tableau)
            else:
                self.tableau = self.pivot(self.tableau, row_idx, col_idx)
```
For-loops are more efficient than while-loops in Python, and they're easier to read in this case, IMO. Also,  the value of `iter_num` is never actually used

tianyizheng02 (changes requested):

linear_programming/simplex.py
imengus: Yes, agreed and thank you

cclauss (approved):

PR #8834, opened 2023-06-23
added floating discount rate support

Mike-gag:
The existing algorithm provided the ability to calculate the present value of a series of cash flows with a fixed discount rate, i enhanced it so it works with a floating discount rate. 
The basic functionallity has not been altered, the algorithm can be used as it was before without any problems.
Added doctests for new cases i introduced, also added new Errors and edge case tests.

- [x] This PR only changes one algorithm file. To ease review, please open separate PRs for separate algorithms.
- [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
- [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
- [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.


Mike-gag:
@cclauss i'd really appreciate some feedback

financial/present_value.py
```diff
@@ -1,38 +1,69 @@
...
 2. An array of cash flows, with the index of the cash flow being the associated year
 
 Note: This algorithm assumes that cash flows are paid at the end of the specified year
 """
 
 
-def present_value(discount_rate: float, cash_flows: list[float]) -> float:
+def present_value(discount_rates: list[tuple], cash_flows: list[float]) -> float:
     """
     >>> present_value(0.13, [10, 20.70, -293, 297])
```
cclauss, line 21: I am confused as to why line 21 delivers `4.69` in the old code and `4.15` in the new.
Will it do the same if `0.13` is changed to `[0.13]` on line 21.
I would like us to keep the old tests (with modifications if required) to prove that the old and new are equivelent.

financial/present_value.py
```diff
@@ -1,38 +1,69 @@
...
-    >>> present_value(0.07, [109129.39, 30923.23, 15098.93, 29734,39])
-    175519.15
-    >>> present_value(-1, [109129.39, 30923.23, 15098.93, 29734,39])
+    4.15
+    >>> present_value([(0.13, 4)], [10, 20.70, -293, 297])
+    4.15
+    >>> present_value([(0.03, 4), (0.05, 1)], [6, 6, 6, 106, 6])
+    115.85
+    >>> present_value([(0.05, 1), (0.07, 1), (0.02, 1), (0.04, 1)], \
+    [32.21, 324.45, 524.9, 12])
```
cclauss, line 28:
```suggestion
    >>> present_value([(0.05, 1), (0.07, 1), (0.02, 1), (0.04, 1)],
    ... [32.21, 324.45, 524.9, 12])
```

financial/present_value.py
Mike-gag: - I am confused as to why line 21 delivers 4.69 in the old code and 4.15 in the new.
The formula produces 4.15, i have no idea why it was 4.69 in the first place. You can cross check with the formula found [here](https://en.wikipedia.org/wiki/Present_value). Another test came up with wrong results, fixed it as well.

-  Will it do the same if 0.13 is changed to [0.13] on line 21?
no it will produce an error. The idea is if there is a fixed rate, the user need only provide a float number of the rate, and if its a floating rate, the user provides a list of tuples.There is no instance where a variable of type `[float]` is provided.

Mike-gag: @cclauss the `enumerate()` function starts counting from `0` where `i` in 
`cash_flow / ((1 + discount_rate) ** i) for i, cash_flow in enumerate(cash_flows)` should start counting from 1, not 0, hence the computational error in the previous iteration. My implementation produces the correct result.

cclauss: What does `enumerate(cash_flows, 1)` do?

cclauss: Should
`def present_value(discount_rates: list[tuple]` be changed to
`def present_value(discount_rates: float | list[tuple]`?

Mike-gag: > Should `def present_value(discount_rates: list[tuple]` be changed to `def present_value(discount_rates: float | list[tuple]`?

Yes that's better, changing it rn

Mike-gag: > What does `enumerate(cash_flows, 1)` do?

It fixes the problem but i think that's beside the point. The test cases in place had different results than my tests because the previous implementation calculated wrong results. My implementation provides the correct results which is why there were differences between the tests.

Mike-gag: @cclauss do i resolve the conversation or do i leave this up to reviewers? I'd appreciate a review at this time

PR #8835, opened 2023-06-24
Provide a different implemention of the graham scan algorithm

ValentinKlinger:
### Describe your change:

In the former version, segments are
not always represented by only their two endpoints, as it should be, see e.g. the tests of existing graham_scan function :

        >>> graham_scan([(0, 0), (1, 1), (2, 2), (3, 3), (-1, 2)])
        [(0, 0), (1, 1), (2, 2), (3, 3), (-1, 2)]

we should have instead (if we follow wikpedia pseudo-code):

        >>> graham_scan([(0, 0), (1, 1), (2, 2), (3, 3), (-1, 2)])
        [(0, 0), (3, 3), (-1, 2)]

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/other/graham_scan.py
+++ b/other/graham_scan.py
@@ -1,172 +1,91 @@
 """
-This is a pure Python implementation of the merge-insertion sort algorithm
-Source: https://en.wikipedia.org/wiki/Graham_scan
+An implementation of Graham's scan algorithm for finding
+the convex hull of a set of points.
+pseudo-code from https://en.wikipedia.org/wiki/Graham_scan :
 
-For doctests run following command:
-python3 -m doctest -v graham_scan.py
+let points be the list of points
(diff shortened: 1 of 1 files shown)


CaedenPH (changes requested):
Please make a new pull request creating a second implementation either in the same file or in a new file. There are no reasons why your implementation should get rid of someone else's implementation.

kkploo112:
Your project has helped me solve a problem that I've been struggling with for a long time. Thank you!

ValentinKlinger:
> Please make a new pull request creating a second implementation either in the same file or in a new file. There are no reasons why your implementation should get rid of someone else's implementation.

Thanks for your feedback, I'll do that and make a new pull request

PR #8836, opened 2023-06-24
Remove duplicate implementation of Collatz sequence

### LINKED ISSUE #8061: Remove duplicate Collatz sequence implementation
### Feature description

I think one of [`3n_plus_1.py`](https://github.com/TheAlgorithms/Python/blob/master/maths/3n_plus_1.py) and [`collatz_sequence.py`](https://github.com/TheAlgorithms/Python/blob/master/maths/collatz_sequence.py) should be removed, as they're nearly identical implementations of the same algorithm for computing Collatz sequences.

I'm not sure which of the two should be deleted, as I'm not sure which of "the Collatz conjecture" and "the 3n + 1 problem" is the more well-known name for the problem.

tianyizheng02:
### Describe your change:

Fixes #8061

- Removed `maths/3n_plus_1.py` in favor of `maths/collatz_sequence.py` because the latter is the more well-known name for the problem
- Added tests from `maths/3n_plus_1.py` to `maths/collatz_sequence.py`
- Converted the `collatz_sequence` function to save memory
- Expanded explanation of the Collatz conjecture

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:


maths/collatz_sequence.py
```diff
@@ -1,43 +1,64 @@
...
-    sequence = [n]
+    yield n
     while n != 1:
-        n = 3 * n + 1 if n & 1 else n // 2
-        sequence.append(n)
-    return sequence
+        if n % 2 == 0:
+            n //= 2
+        else:
+            n = 3 * n + 1
```
cclauss, line 53: # OPTIONAL:
```suggestion
        div, mod = divmod(n, 2)
        n = div if mod else 3 * n + 1
```

cclauss (changes requested):
Looking cool!  Do these changes effect performance?

tianyizheng02:
> Looking cool! Do these changes effect performance?

@cclauss I don't think performance is really affected in most cases since the algorithm itself is basically unchanged. However, I think there'd be slightly better performance if one were to iterate over the algorithm's output:
```py
for num in collatz_sequence(n): 
    # do something with num
```
since having a generator means not needing to create and store the entire sequence in memory. However, I haven't timed this generator implementation against the original list implementation, so I don't know how much of a difference it makes.

cclauss (approved):
Nice!

PR #8856, opened 2023-07-08
Update prime_check.py

cvane238:
### Describe your change:

Add a condition to check if the last digit of the number is 0, 2, 4, 5, 6, or 8, indicating divisibility by 2 or 5. This optimization improves the function's performance for some cases as it allows early termination of non-primes with `False`.

### Checklist:

### DIFF
--- a/maths/prime_check.py
+++ b/maths/prime_check.py
@@ -40,7 +40,7 @@ def is_prime(number: int) -> bool:
 
     # Check if number is divisible by 2 or 5
     last_digit = number % 10
-    elif last_digit not in [1, 3, 7, 9]:
+    if last_digit not in [1, 3, 7, 9]:
          return False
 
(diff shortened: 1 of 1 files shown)


maths/prime_check.py
```diff
@@ -37,6 +37,12 @@ def is_prime(number: int) -> bool:
     if 1 < number < 4:
         # 2 and 3 are primes
         return True
+
+    # Check if number is divisible by 2 or 5
+    last_digit = number % 10
+    if last_digit not in [1, 3, 7, 9]:
+        return False
```
tianyizheng02, line 44: 5 is divisible by 5, and 5 is prime.

tianyizheng02 (commented):
This change is breaking other Python files that rely on this function, so unfortunately I think it's best that we don't merge this.

PR #8868, opened 2023-07-15
adding a split workload algorithm

meg-1:
### Describe your change:

This algorithm is a case of the partition problem. it accepts a multiset ( list ) of positive integers, distributes them, and returns a tuple, containing two lists, with minimal difference between their sums. This algorithm can be used to find the optimal even distribution of task timings.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


scheduling/split_workload.py
```diff
@@ -0,0 +1,80 @@
+from math import fabs
+
+
+def split_list(timings: list) -> tuple:
+    """
```
rohan472000, line 5: where is the code for this method?

scheduling/split_workload.py
```diff
@@ -0,0 +1,80 @@
...
+    >>> split_list(["12.5", "10", "11", "9"])
+    ([10, 11], [12.5, 9], 0.5)
+    >>> split_list(["twelve", "ten", "eleven", "nine"])
+    ([0], [0, 0, 0], 0)
+
+    """
+
+    result = None
+
+    def split_workload(arr: list) -> tuple:
```
rohan472000, line 29: what will this function `split_workload` do? provide description.

tianyizheng02: It appears that the inner function `split_workload` is the actual algorithm, while the outer function `split_list` merely validates the input list before running `split_workload`.

@meg-1 Instead of handling different input types in an outer function and then always calling an inner function, I think the code structure would be a lot simpler if you wrote the code all in one function:
```py
def split_workload(timings: list) -> tuple:    # Make type hints more specific
    # Validate input at the top, performing any necessary type checks and conversions
    # (currently what split_list is doing)
    
    # The actual algorithm body (currently what split_workload is doing):
    if len(arr) == 0:
        ...
    if len(arr) == 1:
        ...
    n = len(arr)
    smallest_diff = float("inf")
    ...
    ...
    return result
```

scheduling/split_workload.py
```diff
@@ -0,0 +1,80 @@
...
+                    timings[val] = fabs(float(current_element))
+                else:
+                    timings[val] = abs(int(float(current_element)))
+            else:
+                timings[val] = 0
+        result = split_workload(timings)
+
+    except ValueError:
+        timings = [c * -1 if c < 0 else c for c in timings]
+        result = split_workload(timings)
```
tianyizheng02, line 74: If the timings must be non-negative, then I don't think it's right to automatically convert negative timings. Strictly speaking, the output wouldn't be "correct" since the algorithm would actually be working a different input than the one provided. I think it's better to just raise a `ValueError` in this case.

However, if one were to convert the timings like this, it'd be easier to just use `abs` anyway:
```py
timings = [abs(c) for c in timings]
```

scheduling/split_workload.py
tianyizheng02, line 1: This algorithm is a brute-force search over (nearly) all $2^n$ possible partitions, which is incredibly inefficient. It's fine to include brute-force algorithms like this for educational purposes, but you should at least make that point clear by explicitly stating how the algorithm works and providing the asymptotic runtime ($O(2^n n)$ in this case I think?)

tianyizheng02 (changes requested):

tianyizheng02:
> This algorithm is a case of the partition problem.

I think it's more accurate to say that this algorithm _is_ a solution to the partition problem rather than a special case of the partition problem. Because of this, I'm not sure if it's accurate to categorize this as a scheduling algorithm, since the problem and the algorithm are for more than just scheduling.

scheduling/split_workload.py
meg-1: Thank you very much for your detailed and structured code review! I appreciate your time and have implemented all suggestions i can :))

meg-1:
https://github.com/TheAlgorithms/Python/pull/8868#discussion_r1264594132 regarding this, first off thank you for your recommendations very much! The only question i have is that while implementing this one, I tried using: 
`def split_list(timings: List[Union[int, float, str]]) -> Tuple[List[Union[int, float]]]:`
(which, as i know, is one of the newest notations) as a type hint, however the Ruff tests don't pass. How would you recommend I go around this?
I have investigated this issue: https://github.com/astral-sh/ruff/issues/4843
As I understand, this is strictly a Ruff issue.

sorts/split_workload.py
```diff
@@ -0,0 +1,72 @@
...
+                timings[i] = abs(int(float(current_element)))
+        elif isinstance(current_element, (int, float)):
+            timings[i] = abs(current_element)
+        else:
+            raise ValueError("only numbers please")
+
+    if len(timings) == 0:
+        return ([], [], 0)
+    elif len(timings) == 1:
+        return ([timings[0]], [], 1)
```
CaedenPH, line 50: If it returns the third index as the difference, should this not be
```suggestion
        return ([timings[0]], [], timings[0])
```

CaedenPH, line 50: If it returns the third index as the difference, should this not be
```suggestion
        return ([timings[0]], [], timings[0])
```

sorts/split_workload.py
```diff
@@ -0,0 +1,72 @@
...
+    else:
```
CaedenPH, line 51: This else isn't necessary because you return

CaedenPH, line 51: This else isn't necessary because you return

CaedenPH, line 51: This else isn't necessary because you return

CaedenPH, line 51: This else isn't necessary because you return

CaedenPH, line 51: This else isn't necessary because you return

sorts/split_workload.py
CaedenPH, line 50: If it returns the third index as the difference, should this not be
```suggestion
        return ([timings[0]], [], timings[0])
```

sorts/split_workload.py
```diff
@@ -0,0 +1,72 @@
...
+        ):
+            is_current_elem_int = float(current_element).is_integer()
+            if not is_current_elem_int:
+                timings[i] = abs(float(current_element))
+            else:
+                timings[i] = abs(int(float(current_element)))
+        elif isinstance(current_element, (int, float)):
+            timings[i] = abs(current_element)
+        else:
+            raise ValueError("only numbers please")
```
CaedenPH, line 45:
```suggestion
            raise ValueError("Timings must be a list of numbers")
```

sorts/split_workload.py
```diff
@@ -0,0 +1,72 @@
...
+
+def split_list(timings: List[Union[int, float, str]]) -> Tuple[List[Union[int, float]]]:
+    """
+
+    This algorithm is a brute-force search over (nearly) all 2^n
+    possible partitions and was created for educational purposes.
+    the asymptotic runtime of this code is: O(n * 2^n)
+
+    this is a case of the partition problem.
+    it accepts a multiset ( list ) of positive integers,
```
CaedenPH, line 12: It says here that the list must be of positive, but there is a working test including a list of negative numbers

CaedenPH (changes requested):
Because there are many different paths in the function, could you add some more doctests for each path, for example empty lists

meg-1:
> Because there are many different paths in the function, could you add some more doctests for each path, for example empty lists

Thank you very much for your detailed code review! All suggestions have been worked through and implemented :))

PR #8869, opened 2023-07-16
Update newton_raphson.py

### LINKED ISSUE #8863: Refactor arithmetic_analysis/newton_raphson to not use eval

https://github.com/TheAlgorithms/Python/blob/master/arithmetic_analysis/newton_raphson.py#L28
The newton raphson function uses eval, which is flagged by ruff here 
https://github.com/TheAlgorithms/Python/actions/runs/5515973175/jobs/10056851168?pr=8861

### Additional information

```
Error: arithmetic_analysis/newton_raphson.py:28:3[5](https://github.com/TheAlgorithms/Python/actions/runs/5515973175/jobs/10056851168?pr=8861#step:4:6): S307 Use of possibly insecure function; consider using `ast.literal_eval`
Error: arithmetic_analysis/newton_raphson.py:28:57: S307 Use of possibly insecure function; consider using `ast.literal_eval`
Error: arithmetic_analysis/newton_raphson.py:30:1[6](https://github.com/TheAlgorithms/Python/actions/runs/5515973175/jobs/10056851168?pr=8861#step:4:7): S307 Use of possibly insecure function; consider using `ast.literal_eval
```

rohan472000:
### Describe your change:

Fixes https://github.com/TheAlgorithms/Python/issues/8863

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #8863".

### DIFF
--- a/arithmetic_analysis/newton_raphson.py
+++ b/arithmetic_analysis/newton_raphson.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 from sympy import diff, symbols, sympify
 
 
-def newton_raphson(func: str, a: float, precision: float = 10**-10) -> float:
+def newton_raphson(func: str, start_point: float, precision: float = 10**-10) -> float:
     """Finds root from the point 'a' onwards by Newton-Raphson method
     >>> newton_raphson("sin(x)", 2)
     3.1415926536808043
@@ -18,22 +18,25 @@ def newton_raphson(func: str, a: float, precision: float = 10**-10) -> float:
     >>> newton_raphson("log(x)- 1", 2)
     2.718281828458938
     """
-    x = a
-    symbol = symbols("x")
+    x = start_point
+    symbol = symbols('x')
+
     # expressions to be represented symbolically and manipulated algebraically
-    exp = sympify(func)
-    # calculate the derivative value at the current x value
-    exp_diff = diff(exp, symbol)
-    maximum_iterations = 100
+    expression = sympify(func)
 
-    for _ in range(maximum_iterations):
-        val = exp.subs(symbol, x)
-        diff_val = exp_diff.subs(symbol, x)
+    # calculates the derivative value at the current x value
+    derivative = diff(expression, symbol)
 
-        if abs(val) < precision:
+    max_iterations = 100
+
+    for _ in range(max_iterations):
+        function_value = expression.subs(symbol, x)
+        derivative_value = derivative.subs(symbol, x)
+
+        if abs(function_value) < precision:
             return float(x)
 
-        x -= val / diff_val
+        x = x - (function_value / derivative_value)
 
     return float(x)
 
--- a/arithmetic_analysis/newton_raphson.py
+++ b/arithmetic_analysis/newton_raphson.py
@@ -7,7 +7,9 @@ from __future__ import annotations
 from sympy import diff, symbols, sympify
 
 
-def newton_raphson(func: str, start_point: float, precision: float = 10**-10) -> float:
+def newton_raphson(
+    func: str, start_point: float, precision: float = 10**-10
+) -> float:
     """Finds root from the point 'a' onwards by Newton-Raphson method
     >>> newton_raphson("sin(x)", 2)
     3.1415926536808043
@@ -19,7 +21,7 @@ def newton_raphson(func: str, start_point: float, precision: float = 10**-10) ->
     2.718281828458938
     """
     x = start_point
-    symbol = symbols('x')
+    symbol = symbols("x")
 
     # expressions to be represented symbolically and manipulated algebraically
     expression = sympify(func)
--- a/arithmetic_analysis/newton_raphson.py
+++ b/arithmetic_analysis/newton_raphson.py
@@ -21,7 +21,8 @@ def newton_raphson(
     >>> newton_raphson("log(x)- 1", 2)
     2.718281828458938
     >>> from scipy.optimize import newton
-    >>> all(newton_raphson("log(x)- 1", 2) == newton("log(x)- 1", 2) for precision in 10, 100, 1000, 10000))
+    >>> all(newton_raphson("log(x)- 1", 2) == newton("log(x)- 1", 2)
+    ...     for precision in 10, 100, 1000, 10000))
     True
     >>> newton_raphson("log(x)- 1", 2, 0)
     Traceback (most recent call last):
--- a/arithmetic_analysis/newton_raphson.py
+++ b/arithmetic_analysis/newton_raphson.py
@@ -22,7 +22,7 @@ def newton_raphson(
     2.718281828458938
     >>> from scipy.optimize import newton
     >>> all(newton_raphson("log(x)- 1", 2) == newton("log(x)- 1", 2)
-    ...     for precision in 10, 100, 1000, 10000))
+    ...     for precision in (10, 100, 1000, 10000))
     True
     >>> newton_raphson("log(x)- 1", 2, 0)
     Traceback (most recent call last):
--- a/arithmetic_analysis/newton_raphson.py
+++ b/arithmetic_analysis/newton_raphson.py
@@ -51,7 +51,7 @@ def newton_raphson(
         if abs(function_value) < precision:
             return float(x)
 
-        x = x - (function_value / derivative_value)
+        x -= function_value / derivative_value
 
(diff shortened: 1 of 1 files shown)


rohan472000:
@cclauss , plz review this.

arithmetic_analysis/newton_raphson.py
```diff
@@ -4,34 +4,41 @@
...
-from math import *  # noqa: F403
+from sympy import diff, symbols, sympify
 
-from sympy import diff
 
-
-def newton_raphson(
-    func: str, a: float | Decimal, precision: float = 10**-10
-) -> float:
+def newton_raphson(func: str, a: float, precision: float = 10**-10) -> float:
```
cclauss, line 10: Can we come up with a more self-documenting name than `a`?

arithmetic_analysis/newton_raphson.py
```diff
@@ -4,34 +4,41 @@
...
-        )
-        # This number dictates the accuracy of the answer
-        if abs(eval(func)) < precision:  # noqa: S307
+    symbol = symbols("x")
+    exp = sympify(
+        func
+    )  # expressions to be represented symbolically and manipulated algebraically
+    exp_diff = diff(
+        exp, symbol
+    )  # calculates the derivative value at the current x value
```
cclauss, line 28: It is very distracting for the reader when code is wrapped just to make room for a comment. 
```suggestion
    # expressions to be represented symbolically and manipulated algebraically
    exp = sympify(func)
    # calculate the derivative value at the current x value
    exp_diff = diff(exp, symbol)
```

arithmetic_analysis/newton_raphson.py
```diff
@@ -4,34 +4,41 @@
...
+    maximum_iterations = 100
+
+    for _ in range(maximum_iterations):
+        val = exp.subs(symbol, x)
+        diff_val = exp_diff.subs(symbol, x)
```
cclauss, line 33: These variable names are weak.  Why can't words be used so that humans can read the code like sentences?

arithmetic_analysis/newton_raphson.py
```diff
@@ -4,34 +4,41 @@
...
+
+        if abs(val) < precision:
             return float(x)
 
+        x = x - (val / diff_val)
```
cclauss, line 38:
```suggestion
        x -= val / diff_val
```

arithmetic_analysis/newton_raphson.py
```diff
@@ -4,34 +4,41 @@
...
     >>> newton_raphson("sin(x)", 2)
     3.1415926536808043
     >>> newton_raphson("x**2 - 5*x +2", 0.4)
-    0.4384471871911695
+    0.4384471871911696
     >>> newton_raphson("x**2 - 5", 0.1)
     2.23606797749979
     >>> newton_raphson("log(x)- 1", 2)
     2.718281828458938
     """
```
cclauss, line 20: Please add a test with 0 precision and -1 precision.
Tests should also validate that this algorithm generates the same results as `scipy.optimize.newton()`.

cclauss (changes requested):

arithmetic_analysis/newton_raphson.py
rohan472000: well that's done by `ruff`, I usually don't change anything after executing `ruff` otherwise it throws errors.

arithmetic_analysis/newton_raphson.py
rohan472000: a negative precision value doesn't make sense in the context of the Newton-Raphson method since it implies an infinite level of accuracy....

cclauss: # SO RAISE A ValueError EXCEPTION.

rohan472000:
while executing this test `print(f"The root of sin(x) = 0 with precision 0 is {newton_raphson('sin(x)', 2, precision=0)}")` it took too much time and still not executed...so cancelled that.

cclauss:
> why we should make a test case with different precisions??

Because the caller might choose to change the default.  Why are there no tests that change the default precision?

arithmetic_analysis/newton_raphson.py
cclauss:
```suggestion
    >>> from scipy.optimize import newton
    >>> all(newton_raphson("log(x)- 1", 2) == newton("log(x)- 1", 2) for precision in 10, 100, 1000, 10000))
    True
    >>> newton_raphson("log(x)- 1", 2, 0)
    Traceback (most recent call last):
        ...
    ValueError: precision must be greater than zero
    >>> newton_raphson("log(x)- 1", 2, -1)
    Traceback (most recent call last):
        ...
    ValueError: precision must be greater than zero
    """
```

cclauss (changes requested):

arithmetic_analysis/newton_raphson.py
cclauss: How does `scipy.optimize.newton()` deal with a precision of 0 and -1?

arithmetic_analysis/newton_raphson.py
```diff
@@ -1,37 +1,60 @@
...
+    max_iterations = 100
+
+    for _ in range(max_iterations):
+        function_value = expression.subs(symbol, x)
+        derivative_value = derivative.subs(symbol, x)
+
+        if abs(function_value) < precision:
             return float(x)
 
+        x = x - (function_value / derivative_value)
```
cclauss, line 54:
```suggestion
        x -= function_value / derivative_value
```

arithmetic_analysis/newton_raphson.py
```diff
@@ -1,37 +1,60 @@
...
+    ValueError: precision must be greater than zero
     """
-    x = a
-    while True:
-        x = Decimal(x) - (
-            Decimal(eval(func)) / Decimal(eval(str(diff(func))))  # noqa: S307
-        )
-        # This number dictates the accuracy of the answer
-        if abs(eval(func)) < precision:  # noqa: S307
+    x = start_point
```
cclauss, line 36: if precision <= 0 then raise a ValueError with the message that matches the test above.

cclauss:
How does `scipy.optimize.newton()` deal with a `precision` of 0 and -1?

cclauss:
@QuantumNovice your review of these changes please.

rohan472000:
> How does `scipy.optimize.newton()` deal with a `precision` of 0 and -1?

Screenshot (228)

for -1 it is throwing error.

QuantumNovice:
https://github.com/TheAlgorithms/Python/blob/27b1bbacb69b6370c3dd217e5a2dc4835971b579/arithmetic_analysis/newton_raphson.py#L36
This should `precision <= 0` as python will set the float to zero which is a valid answer.
image

The `newton` method implemented in `scipy` is entirely a numerical method. 
https://github.com/scipy/scipy/blob/cfe80116aaa145061246b8aec0e98248fbefb835/scipy/optimize/_zeros_py.py#L105

QuantumNovice:
https://github.com/TheAlgorithms/Python/blob/27b1bbacb69b6370c3dd217e5a2dc4835971b579/arithmetic_analysis/newton_raphson.py#L48
This should be an input variable as it does not generate correct solution for arbitrary function.

cclauss:
@QuantumNovice What would be a valid test to replace
```
    >>> all(newton_raphson("log(x)- 1", 2) == newton("log(x)- 1", 2)
    ...     for precision in (10, 100, 1000, 10000))
    True
````

**Comment by cclauss** [comment]
Why does this test not pass?
```
>>> from math import log
>>> from scipy.optimize import newton
>>>
>>> def log_x_minus_one(x):
...     return log(x) - 1
...
>>> a = newton_raphson("log(x) - 1", 2, precision=1)
>>> b = newton(log_x_minus_one, 2, tol=1)
>>> assert a == b, f"{a = } != {b = }"
```

tianyizheng02:
Closing due to lack of fixes in response to code review. We need to clear out our backlog of PRs as much as possible before Hacktoberfest, and this PR still has failing tests.

PR #8885, opened 2023-07-22
Optimal coefficients, fixes #8847

### LINKED ISSUE #8847: `machine_learning/linear_regression.py` doesn't give optimal coefficients
### Feature description

Related to issue #8844

```py

X = data[:, 0].reshape(-1, 1)
y = data[:, 1]

print(f"{sse = }")

plt.legend()
plt.show()
```

```
...
Resultant Feature vector: 
-9.34325
1.53067
sse = 253795.17406773588
mse = 253.79517406773587
half mse = 126.89758703386794
```

linear_regression_plot

liakdimi1:
### Describe your change:
FIxes #8847
I found the optimal iterations and learning rate for achieving the closest coefficients
Στιγμιότυπο οθόνης (266)
* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/linear_regression.py
+++ b/machine_learning/linear_regression.py
@@ -9,7 +9,6 @@ Rating). We try to best fit a line through dataset and estimate the parameters.
 """
 import numpy as np
 import requests
-import matplotlib.pyplot as plt
 
 
 def collect_dataset():
@@ -99,7 +98,7 @@ def mean_absolute_error(predicted_y, original_y):
 def main():
     """Driver function"""
     data = collect_dataset()
-    
+
     len_data = data.shape[0]
     data_x = np.c_[np.ones(len_data), data[:, :-1]].astype(float)
     data_y = data[:, -1].astype(float)
@@ -109,6 +108,7 @@ def main():
     print("Resultant Feature vector : ")
     for i in range(0, len_result):
         print(f"{theta[0, i]:.5f}")
+
 
 if __name__ == "__main__":
     main()
--- a/machine_learning/linear_regression.py
+++ b/machine_learning/linear_regression.py
@@ -9,6 +9,7 @@ Rating). We try to best fit a line through dataset and estimate the parameters.
 """
 import numpy as np
 import requests
+import matplotlib.pyplot as plt
 
 
 def collect_dataset():
@@ -41,7 +42,6 @@ def run_steep_gradient_descent(data_x, data_y, len_data, alpha, theta):
                        curr_features - alpha_ * gradient(w.r.t. feature)
     """
     n = len_data
-
     prod = np.dot(theta, data_x.transpose())
     prod -= data_y.transpose()
     sum_grad = np.dot(prod, data_x)
@@ -70,8 +70,8 @@ def run_linear_regression(data_x, data_y):
     :param data_y  : contains the output (result vector)
     :return        : feature for line of best fit (Feature vector)
     """
-    iterations = 100000
-    alpha = 0.0001550
+    iterations = 250000
+    alpha = 0.000326
 
     no_features = data_x.shape[1]
     len_data = data_x.shape[0] - 1
@@ -99,7 +99,7 @@ def mean_absolute_error(predicted_y, original_y):
 def main():
     """Driver function"""
     data = collect_dataset()
-
+    
     len_data = data.shape[0]
     data_x = np.c_[np.ones(len_data), data[:, :-1]].astype(float)
     data_y = data[:, -1].astype(float)
@@ -109,7 +109,6 @@ def main():
     print("Resultant Feature vector : ")
     for i in range(0, len_result):
         print(f"{theta[0, i]:.5f}")
-
 
 if __name__ == "__main__":
     main()
--- a/machine_learning/tempCodeRunnerFile.py
+++ b/machine_learning/tempCodeRunnerFile.py
@@ -1,0 +1,1 @@
+plt.axline(xy1=(0, reg.intercept_), slope=reg.coef_[0], color="blue", label="Sklearn")
--- a/machine_learning/tempCodeRunnerFile.py
+++ b/machine_learning/tempCodeRunnerFile.py
@@ -1,1 +1,0 @@
-plt.axline(xy1=(0, reg.intercept_), slope=reg.coef_[0], color="blue", label="Sklearn")


machine_learning/linear_regression.py
```diff
@@ -70,8 +69,8 @@ def run_linear_regression(data_x, data_y):
     :param data_y  : contains the output (result vector)
     :return        : feature for line of best fit (Feature vector)
     """
-    iterations = 100000
-    alpha = 0.0001550
+    iterations = 250000
+    alpha = 0.000326
```
tianyizheng02, line 73: This won't give the optimal coefficients. Like I said in my original issue, gradient descent is an approximation algorithm _by definition,_ so tweaking the parameters will only give you a better approximation. My point with the original issue is that we should add a _different_ algorithm, one that is a _direct method_ rather than an _iterative method._

tianyizheng02:
> If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

Please add the issue number to the _description_ of your PR. You'll know it worked when you see the issue appear in this section on the right-hand side of this PR:

screenshot

machine_learning/linear_regression.py
tianyizheng02: This also doesn't fix the fact that the code is still calculating SSE incorrectly.

liakdimi1: can you elaborate on what do you mean by direct method?

tianyizheng02: A [direct method](https://en.wikipedia.org/wiki/Iterative_method) gives an exact solution. By contrast, an iterative method (such as gradient descent) gives increasingly accurate approximations. With an iterative method, you get _closer and closer_ to the exact solution but you may never actually reach that exact solution.

For example, let's say you want to find the roots of a quadratic polynomial $ax^2 + bx + c$. You could use Newton's method to approximate the roots, refining your solutions to be more and more accurate—this is an iterative method. Alternatively, you could just use the quadratic formula to just get the exact solutions—this is a direct method.

liakdimi1: also, explain with facts why the SSE is calculated incorrectly because the code does the predictions finds the deviation and takes the squares of the deviations, and sums the errors, which is the definition of SSE.

tianyizheng02: I already explained this with my code sample in the original issue, but sure, let's go over the math for the SSE function:
https://github.com/TheAlgorithms/Python/blob/9e08c7726dee5b18585a76e54c71922ca96c0b3a/machine_learning/linear_regression.py#L52-L64

Likes 60–62 are fine because they calculate the SSE exactly as we expect:
$$\mathrm{SSE} = \sum_{i = 0}^{N} (\mathbf{x}_i^{\top} \boldsymbol{\theta} - \mathbf{y}_i)^2$$

However, for some reason, the function then divides the SSE by `2 * len_data`. Dividing the SSE (_sum_ of squared errors) by the number of datapoints gives you the MSE (_mean_ squared error):
$$\mathrm{MSE} = \frac{1}{N} \sum_{i = 0}^{N} (\mathbf{x}_i^{\top} \boldsymbol{\theta} - \mathbf{y}_i)^2$$

Naturally, it follows that the "SSE" function is actually calculating half of the MSE because it's dividing the SSE by $2N$.

liakdimi1: i appreciate your response

liakdimi1:
As far as the calculation of the SSE this is what i found :

when it comes to the calculation of the Sum of Squared Errors (SSE), the division by 2 * len_data is a significant factor. This division is related to the derivative of the cost function during optimization, particularly when using gradient descent. By dividing the SSE by 2 * len_data, the derivative of the cost function (SSE) with respect to the model parameters (theta) becomes simpler and results in a cleaner expression.

Additionally, it is worth noting that the cost function commonly used in linear regression is the Mean Squared Error (MSE), which is calculated as SSE divided by 2 * len_data:

MSE = SSE / (2 * len_data)

When taking the derivative of MSE with respect to the model parameters (theta), you get:

d(MSE)/d(theta) = (1/len_data) * (theta.dot(data_x.transpose()) - data_y.transpose()).dot(data_x)

Here, the division by 2 in the original SSE equation cancels out during the differentiation process, and the term 1/len_data helps scale the gradients appropriately to handle datasets of different sizes.

Although the division by 2 * len_data doesn't directly impact the optimization process, it offers a more interpretable representation of the derivative of the cost function, which can be advantageous in certain contexts.

Whether to include the division by 2 * len_data in the cost function is a matter of preference. Some implementations choose to omit this division and adjust the learning rate accordingly to compensate for the scale change. However, to ensure consistency throughout the optimization process and related calculations, it's essential to make a deliberate choice and adhere to it.

So i don't believe your speculation is correct and the implementation of the SSE stands that's from me.

tianyizheng02:
> This division is related to the derivative of the cost function during optimization, particularly when using gradient descent. By dividing the SSE by 2 * len_data, the derivative of the cost function (SSE) with respect to the model parameters (theta) becomes simpler and results in a cleaner expression.

It's perfectly fine to scale the loss function to make the math cleaner, but again the problem here is that this scaled version is no longer the SSE. SSE has a precise mathematical definition, so why should we call the scaled version the SSE when it's clearly not? Why not just call it `scaled_sum_of_squared_errors` or `half_mean_squared_error`?

> Mean Squared Error (MSE), which is calculated as SSE divided by 2 * len_data

This is just factually wrong. MSE is defined to be SSE divided by `len_data`, _with no division by 2._ I pointed this out already in [my previous comment](https://github.com/TheAlgorithms/Python/pull/8885#discussion_r1271330609), and you can easily verify this fact with [Wikipedia](https://en.wikipedia.org/wiki/Mean_squared_error) or your search engine of choice:

$$\mathrm{MSE} = \frac{1}{n} \sum_{i = 1}^{n} (Y_i - \hat{Y}_i)^2$$

While I _have_ seen half of the MSE used in the context of ML (because, as you pointed out, it can make the math cleaner), we should definitely avoid misleading people with a different definition because this repo is for educational purposes.

Anyway, you missed the whole point of my issue. My whole point was that the output of the function `sum_of_square_error` is _not_ the SSE. This is still absolutely true, and I've already demonstrated it to you using both code and math. Plus, the fact that you've been referring to the implementation's loss function as the MSE shows that you also acknowledge that it's _not_ the SSE. If you prefer MSE (or half of the MSE) because it's more interpretable or more computationally elegant, then that's perfectly fine—I personally couldn't care less as to whether this implementation uses SSE or MSE. However, in order to not be incorrect or misleading, the function name should still be changed to reflect what it actually calculates.

tianyizheng02:
Closing this PR since, as I've stated already, it doesn't actually resolve the linked issue

PR #8905, opened 2023-07-29
Fix greyscale computation and inverted coords

colinleroy:
### Describe your change:

### Checklist:

### The change:
Hello there,

I've used your algorithm to re-implement it in C for a side-project of mine, and noticed two little bugs in it:

1) the luminance calculation is incorrect - according to Internet, it's either (0.2126*R + 0.7152*G + 0.0722*B) or (0.299*R + 0.587*G + 0.114*B). Here, we had a mix of both (and the greyscale value could never reach 255.

2) the current_error setting used inverted coordinates.

Hope this helps!

### DIFF
--- a/digital_image_processing/dithering/burkes.py
+++ b/digital_image_processing/dithering/burkes.py
@@ -45,6 +45,11 @@ class Burkes:
         >>> Burkes.get_greyscale(255, 255, 255)
         255.0
         """
+        """
+        Formula from https://en.wikipedia.org/wiki/HSL_and_HSV
+        cf Lightness section, and Fig 13c.
+        We use the first of four possible.
+        """
         return 0.114 * blue + 0.587 * green + 0.299 * red
 
     def process(self) -> None:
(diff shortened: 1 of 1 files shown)


CaedenPH (changes requested):
Could you add some doctests to prove that it will reach 255 and works correctly please

colinleroy:
> Could you add some doctests to prove that it will reach 255 and works correctly please

Good idea, thanks. done :)

CaedenPH (approved):
LGTM!

tianyizheng02 (approved):

digital_image_processing/dithering/burkes.py
```diff
@@ -39,20 +39,24 @@ def __init__(self, input_img, threshold: int):
...
         >>> Burkes.get_greyscale(3, 4, 5)
-        3.753
+        4.185
+        >>> Burkes.get_greyscale(0, 0, 0)
+        0.0
+        >>> Burkes.get_greyscale(255, 255, 255)
+        255.0
         """
-        return 0.114 * blue + 0.587 * green + 0.2126 * red
+        return 0.114 * blue + 0.587 * green + 0.299 * red
```
tianyizheng02, line 48: Could you add a source for these numbers so that we can avoid mistakes like this in the future?

tianyizheng02 (changes requested):

digital_image_processing/dithering/burkes.py
colinleroy: Good idea, done :)

tianyizheng02 (approved):

PR #8915, opened 2023-08-01
added a greedy method

HN026:
### Describe your change:

* [ . ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/greedy_methods/interval_scheduling.py
+++ b/greedy_methods/interval_scheduling.py
@@ -15,6 +15,7 @@ def interval_scheduling(intervals):
 
     return result
 
+
 # Example usage:
 # List of intervals represented as tuples (start_time, end_time)
 intervals = [(1, 4), (3, 6), (2, 8), (5, 9), (7, 10)]
--- a/greedy_methods/interval_scheduling.py
+++ b/greedy_methods/interval_scheduling.py
@@ -1,0 +1,26 @@
+# Interval Scheduling: https://en.wikipedia.org/wiki/Interval_scheduling
+
+
+def interval_scheduling(intervals):
+    # Sort intervals based on their end time in ascending order
+    sorted_intervals = sorted(intervals, key=lambda x: x[1])
+
+    # Initialize the result list with the first interval (earliest end time)
+    result = [sorted_intervals[0]]
+
+    for interval in sorted_intervals[1:]:
+        # Check if the start time of the current interval is after the end time of the last interval in the result
+        if interval[0] >= result[-1][1]:
+            result.append(interval)
+
+    return result
+
+# Example usage:
+# List of intervals represented as tuples (start_time, end_time)
+intervals = [(1, 4), (3, 6), (2, 8), (5, 9), (7, 10)]
+
+# Find the maximum non-overlapping intervals
+max_intervals = interval_scheduling(intervals)
+
+print("Maximum non-overlapping intervals:")
+print(max_intervals)


greedy_methods/interval_scheduling.py
```diff
@@ -0,0 +1,27 @@
+# Interval Scheduling: https://en.wikipedia.org/wiki/Interval_scheduling
```
tianyizheng02, line 1: "Interval scheduling" is the name of the problem, not the algorithm. What is the name of the algorithm?

tianyizheng02, line 1: Also, please provide an explanation of the algorithm: what problem it solves (interval scheduling, in this case), how it works, its time complexity, etc.

greedy_methods/interval_scheduling.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def interval_scheduling(intervals):
```
tianyizheng02, line 4: Rename this function to match the name of the _algorithm,_ not the problem.

tianyizheng02, line 4: Please add type hints

greedy_methods/interval_scheduling.py
```diff
@@ -0,0 +1,27 @@
...
+
+    # Initialize the result list with the first interval (earliest end time)
+    result = [sorted_intervals[0]]
+
+    for interval in sorted_intervals[1:]:
+        # Check if the start time of the current interval is after the end time of the last interval in the result
+        if interval[0] >= result[-1][1]:
+            result.append(interval)
+
+    return result
```
tianyizheng02, line 16:
```suggestion
    # Sort intervals based on their end time
    sorted_intervals = sorted(intervals, key=lambda x: x[1])

    selected = [sorted_intervals[0]]
    for interval in sorted_intervals[1:]:
        # Add current interval if it doesn't overlap with the last selected interval
        if interval[0] >= result[-1][1]:
            selected.append(interval)

    return selected
```
1) Let's be less verbose in the comments, explain things at a relatively high level and avoid explaining every line of code
2) `result` isn't a very descriptive name, so I suggest changing it to something like `selected`

greedy_methods/interval_scheduling.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Example usage:
+# List of intervals represented as tuples (start_time, end_time)
+intervals = [(1, 4), (3, 6), (2, 8), (5, 9), (7, 10)]
+
+# Find the maximum non-overlapping intervals
+max_intervals = interval_scheduling(intervals)
+
+print("Maximum non-overlapping intervals:")
+print(max_intervals)
```
tianyizheng02, line 27: Put demo code inside a `__main__` block.

greedy_methods/interval_scheduling.py
tianyizheng02, line 4: Please add doctests

tianyizheng02 (changes requested):
This algorithm is actually called [Earliest Deadline First Scheduling](https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling), so please change the name of this PR, your file, and your function to reflect this.

We also already have a [scheduling/](https://github.com/TheAlgorithms/Python/tree/c9a7234a954dd280dc8192ae77a564e647d013d4/scheduling) directory, so please put this file in there.

tianyizheng02:
@dhruvmanila This PR doesn't have any tests. How come the algorithms-keeper bot didn't point it out?

greedy_methods/interval_scheduling.py
HN026: sure I will make the changes and psh my PR again

greedy_methods/interval_scheduling.py
cclauss: If this is going to land, it should be with the other scheduling algorithms...
https://github.com/TheAlgorithms/Python/tree/master/scheduling

tianyizheng02:
Closing due to inactivity. There are no type hints or doctests, and the demo code should be within a `__main__` block. This has been pointed out and is still unfixed.

PR #8932, opened 2023-08-08
improvements to jump_search.py

amirsoroush:
### Describe your change:

Here are a few minor improvements to the current implementation of [jump_search.py](https://github.com/TheAlgorithms/Python/blob/master/searches/jump_search.py) algorithm.

1. `int(math.floor(math.sqrt(n)))` is the same thing as `int(math.sqrt(n))` and it can be calculated once.
2. Better naming for variables.
3. Add type hints which indicates that this function can also work with other sequence types as long as they are comparable.
4. `prev = prev + 1` --> `prev += 1`

### Checklist:


tianyizheng02 (approved):
Nice improvements

PR #8933, opened 2023-08-08
I have simplified the input so it will be more efficinet and also removed the for loop which decrease the complexity

samarth64:
### Describe your change:

* [ x] Add an algorithm?
* [ x] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [ x] This pull request is all my own work -- I have not plagiarized.
* [x ] I know that pull requests will not be merged if they fail the automated tests.
* [x ] All new Python files are placed inside an existing directory.
* [x ] All filenames are in all lowercase characters with no spaces or dashes.
* [x ] All functions and variable names follow Python naming conventions.
* [x ] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x ] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [ x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.

### DIFF
--- a/searches/binary_search.py
+++ b/searches/binary_search.py
@@ -298,8 +298,8 @@ def binary_search_by_recursion(
 
 
 if __name__ == "__main__":
-    user_input = input("Enter numbers separated by comma:\n").strip()
-    collection = sorted(int(item) for item in user_input.split(","))
+    user_input = input("Enter numbers separated by space\n")
+    collection = list(map(int,user_input.split()))
     target = int(input("Enter a single number to be found in the list:\n"))
     result = binary_search(collection, target)
     if result is None:
--- a/searches/binary_search.py
+++ b/searches/binary_search.py
@@ -299,7 +299,7 @@ def binary_search_by_recursion(
 
 if __name__ == "__main__":
     user_input = input("Enter numbers separated by space\n")
-    collection = list(map(int,user_input.split()))
+    collection = list(map(int, user_input.split()))
     target = int(input("Enter a single number to be found in the list:\n"))
     result = binary_search(collection, target)
     if result is None:
--- a/searches/linear_search.py
+++ b/searches/linear_search.py
@@ -29,7 +29,7 @@ def linear_search(sequence: list, target: int) -> int:
     """
     for index, item in enumerate(sequence):
         if item == target:
-            return index
+            return index+1 # The final output should not be based on indexing as it starts from 0
     return -1
 
 
@@ -66,8 +66,8 @@ def rec_linear_search(sequence: list, low: int, high: int, target: int) -> int:
 
 
 if __name__ == "__main__":
-    user_input = input("Enter numbers separated by comma:\n").strip()
-    sequence = [int(item.strip()) for item in user_input.split(",")]
+    user_input = input("Enter numbers separated by space\n")
+    sequence = list(map(int,user_input.split()))
 
     target = int(input("Enter a single number to be found in the list:\n").strip())
     result = linear_search(sequence, target)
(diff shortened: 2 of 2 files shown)


searches/binary_search.py
```diff
@@ -299,7 +299,7 @@ def binary_search_by_recursion(
 
 if __name__ == "__main__":
     user_input = input("Enter numbers separated by space\n")
-    collection = list(map(int,user_input.split()))
+    collection = list(map(int, user_input.split()))
```
tianyizheng02, line 302:
```suggestion
    collection = [int(num) for num in user_input.split()]
```

searches/linear_search.py
```diff
@@ -67,7 +69,7 @@ def rec_linear_search(sequence: list, low: int, high: int, target: int) -> int:
 
 if __name__ == "__main__":
     user_input = input("Enter numbers separated by space\n")
-    sequence = list(map(int,user_input.split()))
+    sequence = list(map(int, user_input.split()))
```
tianyizheng02, line 72:
```suggestion
    sequence = [int(num) for num in user_input.split()]
```

searches/binary_search.py
```diff
@@ -298,8 +298,8 @@ def binary_search_by_recursion(
 
 
 if __name__ == "__main__":
-    user_input = input("Enter numbers separated by comma:\n").strip()
-    collection = sorted(int(item) for item in user_input.split(","))
+    user_input = input("Enter numbers separated by space\n")
```
tianyizheng02, line 301:
```suggestion
    user_input = input("Enter numbers separated by space\n").strip()
```
Why get rid of the `.strip()`? It'd help remove any trailing whitespace in the input.

searches/linear_search.py
```diff
@@ -66,8 +68,8 @@ def rec_linear_search(sequence: list, low: int, high: int, target: int) -> int:
 
 
 if __name__ == "__main__":
-    user_input = input("Enter numbers separated by comma:\n").strip()
-    sequence = [int(item.strip()) for item in user_input.split(",")]
+    user_input = input("Enter numbers separated by space\n")
```
tianyizheng02, line 71:
```suggestion
    user_input = input("Enter numbers separated by space\n").strip()
```
Same here, why get rid of the `.strip()`?

searches/linear_search.py
```diff
@@ -29,7 +29,9 @@ def linear_search(sequence: list, target: int) -> int:
     """
     for index, item in enumerate(sequence):
         if item == target:
-            return index
+            return (
+                index + 1
+            )  # The final output should not be based on indexing as it starts from 0
```
tianyizheng02, line 34:
```suggestion
            return index
```
Why change it to `index + 1`? That breaks the function because all outputs would be off by 1. I'm not sure what you mean when you say that the final output _shouldn't_ be based on the index.

tianyizheng02 (changes requested):
One of your changes breaks a function (see my other comments). Also, I don't see how your PR title relates to any of these changes:
> I have simplified the input so it will be more efficinet [sic] and also removed the for loop which decrease the complexity

I wouldn't say that having the user input be separated by spaces rather than commas is "simpler", and you never removed any for-loops.

searches/binary_search.py
```diff
@@ -298,8 +298,8 @@ def binary_search_by_recursion(
...
+    collection = list(map(int, user_input.split()))
```
tianyizheng02, line 302: Your edit changes the behavior of this line. Before, `collection` would be always sorted, but your change makes it so that it's no longer sorted.

Also, this should use a list comprehension rather than `map`.

tianyizheng02 (changes requested):

tianyizheng02:
Closing due to inactivity. This PR breaks `linear_search.py` and hasn't been fixed following code review.

PR #8938, opened 2023-08-09
jacobi_iteration_method.py the use of vector operations, which reduces the calculation time by dozens of times

quant12345:
### Describe your change:

This is my first pull request, if I did something wrong please let me know.
Replaced loops in jacobi_iteration_method function with vector operations. That gives a reduction in the time for calculating the algorithm.

I am attaching a file test75.py with performance tests (you need the perfplot library).

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

jacobi

[test75.py.tar.gz](https://github.com/TheAlgorithms/Python/files/12302089/test75.py.tar.gz)


quant12345 (commented):
How long does the status of awaiting reviews?

arithmetic_analysis/jacobi_iteration_method.py
```diff
@@ -115,6 +115,20 @@ def jacobi_iteration_method(
...
+    """
+    denom - a list of values along the diagonal
+    val - values of the last column of the table array
+
+    masks - boolean mask of all strings without diagonal
+    elements array coefficient_matrix
+
+    ttt - coefficient_matrix array values without diagonal elements
+    ind - column indexes for each row without diagonal elements
+    arr - list obtained by column indexes from the list init_val
```
tianyizheng02, line 127: There shouldn't be such a long comment in the middle of the algorithm. Instead of writing a block comment explaining what each variable does, why not just make the variable names clearer? You could also just add comments to individual lines.

arithmetic_analysis/jacobi_iteration_method.py
```diff
@@ -115,6 +115,20 @@ def jacobi_iteration_method(
...
+
+    the code below uses vectorized operations based on
+    the previous algorithm on loopss:
```
tianyizheng02, line 130: Why comment out the old code? Why not just delete it?

tianyizheng02 (commented):
The change mostly looks good to me, but could we get a performance plot with larger input sizes? Based on your current plot, it looks like your new algorithm has runtimes growing at a faster rate than the old algorithm, and my concern is that your algorithm will become slower than the old one once inputs get sufficiently large.

quant12345:
> The change mostly looks good to me, but could we get a performance plot with larger input sizes? Based on your current plot, it looks like your new algorithm has runtimes growing at a faster rate than the old algorithm, and my concern is that your algorithm will become slower than the old one once inputs get sufficiently large.

Attached in the archived file 'test75.py.tar.gz', file 'test75.py'. It contains the 'jacobi_test' function, which calculates according to the new algorithm. The arrays 'coefficient', 'constant', 'init_val' and the number 'n' (ascending) are generated. In the first iteration 2 x 2, the last iteration we have an array of 256 x 256(). 
To check the performance, the perfplot library is used, in which, in the 'equality_check' function, both received arrays are checked for equality.

The size of the 'coefficient' array is printed on each iteration:

After n > 256 the array becomes so much larger and np.inf and -np.inf appear, which leads to nan numbers, both in the new and old calculation.

arithmetic_analysis/jacobi_iteration_method.py
quant12345: Attached in the archived file 'test75.py.tar.gz', file 'test75.py'. It contains the 'jacobi_test' function, which calculates according to the new algorithm. The arrays 'coefficient', 'constant', 'init_val' and the number 'n' (ascending) are generated. In the first iteration 2 x 2, the last iteration we have an array of 256 x 256( 'coefficient').
To check the performance, the perfplot library is used, in which, in the 'equality_check' function, both received arrays are checked for equality.

The size of the 'coefficient' array is printed on each iteration:

After n > 256 the array becomes so much larger and np.inf and -np.inf appear, which leads to nan numbers, both in the new and old calculation.

quant12345:
> The change mostly looks good to me, but could we get a performance plot with larger input sizes? Based on your current plot, it looks like your new algorithm has runtimes growing at a faster rate than the old algorithm, and my concern is that your algorithm will become slower than the old one once inputs get sufficiently large.

Attached in the archived file 'test75.py.tar.gz', file 'test75.py'. It contains the 'jacobi_test' function, which calculates according to the new algorithm. The arrays 'coefficient', 'constant', 'init_val' and the number 'n' (ascending) are generated. In the first iteration 2 x 2, the last iteration we have an array of 256 x 256( 'coefficient').
To check the performance, the perfplot library is used, in which, in the 'equality_check' function, both received arrays are checked for equality.

The size of the 'coefficient' array is printed on each iteration:

After n > 256 the array becomes so much larger and np.inf and -np.inf appear, which leads to nan numbers, both in the new and old calculation.

arithmetic_analysis/jacobi_iteration_method.py
quant12345: I thought that based on the old example with loops, it is more clear how the algorithm is made. If not, can I delete it?

arithmetic_analysis/jacobi_iteration_method.py
quant12345: Corrected.
Double - At the very bottom, I wrote about performance tests with a large array (answered your message).

arithmetic_analysis/jacobi_iteration_method.py
quant12345: Just in case, I am attaching the jacobi_simple_two.py file, in which you can run the algorithms once by specifying the array size n = 256 # array size 'coefficient'. I am attaching data for array sizes of 256 and 500 and the difference only increases 54 and 56 times faster.
```
n = 256 # array size 'coefficient'

time_old 0:00:14.445305
time_new 0:00:00.264163
difference in time 54.68469389300262
True

n = 500 # array size 'coefficient'

time_old 0:01:53.208590
time_new 0:00:01.997707
difference in time 56.66922360714822
```
[jacobi_simple_two.py.tar.gz](https://github.com/TheAlgorithms/Python/files/12329834/jacobi_simple_two.py.tar.gz)

quant12345:
@tianyizheng02 fixed what you asked. Also added answers on the speed of calculations of the old and new algorithms.

quant12345 (commented):
I'm trying to run an empty commit, because the status is: tests are failing, all three checks were successful (maybe it helps?).

quant12345:
It is not clear that my pull request is not needed, I seem to have done everything that was asked?

rohan472000:
@tianyizheng02 , it looks good to me, if you also think the same, then better to proceed with some actions on this PR.

arithmetic_analysis/jacobi_iteration_method.py
```diff
@@ -130,8 +131,45 @@ def jacobi_iteration_method(
...
+    """
+    masks = ~np.eye(coefficient_matrix.shape[0], dtype=bool)
+    """
+    no_diag - coefficient_matrix array values without diagonal elements
+    """
+    no_diag = coefficient_matrix[masks].reshape(-1, rows - 1)
+    """
+    Here we get 'i_col' - these are the column numbers, for each row
+    without diagonal elements, except for the last column.
+    """
```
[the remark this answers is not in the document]
cclauss: Please use real variable names: `denom` -> `denominator`, `no_diag` --> `no_diagonals` (some readers could assume `no_diagnostics`).  Is there a better variable name than `temp`?  Using real names will make the code more self-documenting and will mean that fewer comments will be required to make things clear to your reader.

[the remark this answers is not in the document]
quant12345: @cclauss  changed variable names. But temp is a multifaceted task, I made this name not too long, if everything is described in its name it will turn out to be a completely cumbersome name. Replaced comments with single-row ones.

Why did I use multi-line comments in the previous version: since not everything fit, and I didn’t want to comment several lines with single-line comments.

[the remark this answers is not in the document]
cclauss:
```python
# There is nothing wrong with really long comments if they are appropriately wrapped at
# 88 characters per line.
```

cclauss (approved):

PR #8940, opened 2023-08-09
add an octal_to_binary converter

### LINKED ISSUE #8921: Feat:Octal To Binary Converter
### Feature description

Octal To Binary Converter is missing in conversion section.
So want to add octal to binary converter

RayaneQH:
### Describe your change:

* [X] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [X] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #8921".

### DIFF
--- a/conversions/octal_to_binary.py
+++ b/conversions/octal_to_binary.py
@@ -1,0 +1,49 @@
+"""
+Convert a octal value to its binary equivalent
+>>> octal_to_binary("17")
+'1111'
+>>> octal_to_binary("52523")
+'101010101010011'
+>>> octal_to_binary("532.51")
+'101011010.101001'
+"""
+
+
+from operator import index
+
+
+def octal_to_binary(octal_string: str) -> str:
+    if not all(char in "01234567." for char in octal_string):
+        raise ValueError("Non-octal value was passed to the function")
+    if not octal_string:
+        raise ValueError("Empty string was passed to the function")
+    binary_string = ""
+    dict_of_octal_to_binary = {
+        "0": "000",
+        "1": "001",
+        "2": "010",
+        "3": "011",
+        "4": "100",
+        "5": "101",
+        "6": "110",
+        "7": "111",
+    }
+    indexl = 0
+    for i in octal_string:
+        if i == ".":
+            binary_string += "."
+            continue
+        if binary_string == "":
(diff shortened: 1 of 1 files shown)


conversions/octal_to_binary.py
```diff
@@ -0,0 +1,49 @@
+"""
+Convert a octal value to its binary equivalent
+>>> octal_to_binary("17")
+'1111'
+>>> octal_to_binary("52523")
+'101010101010011'
+>>> octal_to_binary("532.51")
+'101011010.101001'
```
tianyizheng02, line 8: These doctests should go inside a docstring within the function, not at the top of the file.

conversions/octal_to_binary.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+
+from operator import index
```
tianyizheng02, line 12: This import is never used, and it's causing pre-commit to fail.

conversions/octal_to_binary.py
```diff
@@ -0,0 +1,49 @@
...
+    indexl = 0
+    for i in octal_string:
+        if i == ".":
+            binary_string += "."
+            continue
+        if binary_string == "":
+            for char in dict_of_octal_to_binary[i]:
+                if char != "0" or indexl >= 1:
+                    binary_string += char
+                    indexl += 1
```
tianyizheng02, line 40: Instead of having a special case for handling 0's at the start of the binary string, why not just append to the binary string as normal and handle the starting 0's at the very end?

conversions/octal_to_binary.py
```diff
@@ -0,0 +1,49 @@
...
+        "1": "001",
+        "2": "010",
+        "3": "011",
+        "4": "100",
+        "5": "101",
+        "6": "110",
+        "7": "111",
+    }
+    indexl = 0
+    for i in octal_string:
```
tianyizheng02, line 32: Can we do this with a comprehension instead of manually appending to a string?

tianyizheng02 (changes requested):

tianyizheng02:
Closing in favor of #8949

PR #8942, opened 2023-08-10
updated the URL and HTML tags for scrapping yahoo finance

Suman2023:
#### The existing Yahoo URL for scrapping the stock price is no longer maintained from August 26th, 2021 (attached the screenshot below). Moreover there is a change in HTML tags for the current price field. I have added the new URL and have updated the HTML tags required to fetch the current price of the stock.
Screenshot 2023-08-10 203953

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:
- [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
- [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
- [ ] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
- [ ] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
- [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

### DIFF
--- a/web_programming/current_stock_price.py
+++ b/web_programming/current_stock_price.py
@@ -11,9 +11,7 @@ def stock_price(symbol: str = "AAPL") -> str:
     if specific_fin_streamer_tag:
         text = specific_fin_streamer_tag.get_text()
         return text
-    else:
-        print("No <fin-streamer> tag with the specified data-test attribute found.")
-    return "Not Found"
+    return "No <fin-streamer> tag with the specified data-test attribute found."
 
 # Search for the symbol at https://finance.yahoo.com/lookup
 if __name__ == "__main__":
--- a/web_programming/current_stock_price.py
+++ b/web_programming/current_stock_price.py
@@ -3,12 +3,19 @@ from bs4 import BeautifulSoup
 
 
 def stock_price(symbol: str = "AAPL") -> str:
-    url = f"https://in.finance.yahoo.com/quote/{symbol}?s={symbol}"
-    soup = BeautifulSoup(requests.get(url).text, "html.parser")
-    class_ = "My(6px) Pos(r) smartphone_Mt(6px)"
-    return soup.find("div", class_=class_).find("span").text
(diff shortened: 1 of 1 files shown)


Suman2023:
This is my first contribution will really appreciate some feedback

web_programming/current_stock_price.py
```diff
@@ -3,12 +3,20 @@
...
+    yahoo_finance_source = requests.get(url, headers={"USER-AGENT": "Mozilla/5.0"}).text
+    soup = BeautifulSoup(yahoo_finance_source, "html.parser")
+    specific_fin_streamer_tag = soup.find("fin-streamer", {"data-test": "qsp-price"})
 
+    if specific_fin_streamer_tag:
+        text = specific_fin_streamer_tag.get_text()
+        return text
+    else:
+        print("No <fin-streamer> tag with the specified data-test attribute found.")
+    return "Not Found"
```
tianyizheng02, line 16:
```suggestion
    if specific_fin_streamer_tag:
        text = specific_fin_streamer_tag.get_text()
        return text
    
    print("No <fin-streamer> tag with the specified data-test attribute found.")
    return "Not Found"
```

Also, why both print and return? Why not just return the error message? Or print the error message and return None?

tianyizheng02 (changes requested):
LGTM for the most part, but I'll still have to verify whether the new URL works correctly myself when I get the chance.

web_programming/current_stock_price.py
Suman2023: Thanks for the suggestion. Cannot return None as that would through an error in string formatting. But returning the error looks valid. Doing the change for that.

tianyizheng02 (approved):
Verified that the fix works, LGTM

PR #8603, opened 2023-04-01
Fix `mypy` errors in `erosion_operation.py`

tianyizheng02:
### Describe your change:

- Fixed type hints in `erosion_operation.py` to resolve some `mypy` errors (see #8070)
- Did some minor renaming/refactoring

* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [ ] All new Python files are placed inside an existing directory.
* [ ] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [ ] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [x] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

Issue #8070, opened 2023-01-03
`mypy` produces errors when run on its own


tianyizheng02:
### Feature description

Apologies if I'm misunderstanding about how `mypy` is set up with this repo. `mypy` currently produces a litany of errors when run on its own even though it passes when run as part of `pre-commit`:
```
> mypy --ignore-missing-imports --install-types --non-interactive .
dynamic_programming/fibonacci.py:37: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
dynamic_programming/fibonacci.py:42: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
searches/binary_tree_traversal.py:21: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
maths/sigmoid_linear_unit.py:20: error: Function "numpy.core.multiarray.array" is not valid as a type  [valid-type]
maths/sigmoid_linear_unit.py:20: note: Perhaps you need "Callable[...]" or a callback protocol?
maths/sigmoid_linear_unit.py:29: error: Unsupported operand type for unary - (np.array?)  [operator]
maths/sigmoid_linear_unit.py:32: error: Function "numpy.core.multiarray.array" is not valid as a type  [valid-type]
maths/sigmoid_linear_unit.py:32: note: Perhaps you need "Callable[...]" or a callback protocol?
maths/sigmoid_linear_unit.py:51: error: Unsupported left operand type for * (np.array?)  [operator]
maths/sigmoid.py:14: error: Function "numpy.core.multiarray.array" is not valid as a type  [valid-type]
...
Found 87 errors in 13 files (checked 1143 source files)
```

Is this expected behavior? Should we expect `mypy`, when run on its own, to agree with `pre-commit`?

Aryank1310:
In general, it's a good idea to run mypy on its own to catch any errors that might not be caught by other tools or in other environments. However, it's also important to make sure that the mypy configuration used when running on its own is consistent with the configuration used in other environments, such as pre-commit.
Although the errors could be because of different configuration,enviornment and timings

tianyizheng02:
> In general, it's a good idea to run mypy on its own to catch any errors that might not be caught by other tools or in other environments. However, it's also important to make sure that the mypy configuration used when running on its own is consistent with the configuration used in other environments, such as pre-commit. Although the errors could be because of different configuration,enviornment and timings

[...]

PR #8959, opened 2023-08-14
Fix minor typing errors in maths/

CaedenPH:
### Describe your change:
Some changes include:
- Replacing all occurrences of `np.array` types with `np.ndarray`, the class returned from `np.array`
- Typing certain function bodies
- Refactoring `jaccard_similarity` to include types and fix `mypy` errors (bug) that stored types from variables from different scopes
```
maths\jaccard_similarity.py:80: error: Incompatible types in assignment (expression has type "list[str]", variable has type "int")  [assignment]
maths\jaccard_similarity.py:84: error: Argument 1 to "len" has incompatible type "int"; expected "Sized"  [arg-type]
maths\jaccard_similarity.py:87: error: Incompatible types in assignment (expression has type "list[str]", variable has type "int")  [assignment]
maths\jaccard_similarity.py:88: error: Argument 1 to "len" has incompatible type "int"; expected "Sized"  [arg-type]
```
- Fixed `jaccard_similarity`  and added doctests to prove fixes

* [ ] Add an algorithm?
* [ ] Documentation change?
* [x] Fix typing errors

### Checklist:


maths/jaccard_similarity.py
```diff
@@ -48,30 +52,42 @@ def jaccard_similarity(set_a, set_b, alternative_union=False):
...
+        return intersection_length / union_length
 
-    if isinstance(set_a, (list, tuple)) and isinstance(set_b, (list, tuple)):
+    elif isinstance(set_a, (list, tuple)) and isinstance(set_b, (list, tuple)):
         intersection = [element for element in set_a if element in set_b]
 
         if alternative_union:
-            union = len(set_a) + len(set_b)
-            return len(intersection) / union
+            return len(intersection) / len(set_a) + len(set_b)
```
tianyizheng02, line 83:
```suggestion
            return len(intersection) / (len(set_a) + len(set_b))
```
We need the parentheses or else the order of operations changes—the build should've caught this. Could you add some doctests to handle this case (both sets are lists/tuples and `alternative_union` is True)?

digital_image_processing/morphological_operations/erosion_operation.py
tianyizheng02, line 1: I actually already have an open PR (#8603) that fixes the typing errors in this file

maths/newton_raphson.py
tianyizheng02, line 1: Could you add type hints to this file? The error that you're fixing now could've been caught much earlier if this file had proper type hinting.

tianyizheng02 (changes requested):
Overall good fixes, definitely helps to fix many of the mypy errors I reported in #8070

tianyizheng02 (approved):

Issue #6216, opened 2022-06-23
Some algorithms are "how-to" examples


poyea:
In our [Contributing Guidelines](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md):

> Algorithms in this repo should not be how-to examples for existing Python packages. Instead, they should perform internal calculations or manipulations to convert input values into different output values. Those calculations or manipulations can use data types, classes, or functions of existing Python packages but each algorithm in this repo should add unique value.

And after some investigation like #6201, one can find some implementations that resemble online tutorials or are too trivial. In #6201, [`machine_learning/support_vector_machines.py`](https://github.com/TheAlgorithms/Python/blob/c86aa72cfa0467bd9a5711d7b5a77ed8243e49f1/machine_learning/support_vector_machines.py) is removed. It should be _migrated_ to a much detailed implementation of SVM instead of importing packages.

Migration list:
- [ ] [`machine_learning/support_vector_machines.py`](https://github.com/TheAlgorithms/Python/blob/c86aa72cfa0467bd9a5711d7b5a77ed8243e49f1/machine_learning/support_vector_machines.py)

Bjiornulf:
I am currently trying to implement it using SciPy (for minimization and maximization). Will keep you updated with the advancement.

Bjiornulf:
Maybe the following files should be added to the migration list
- [ ] [`machine_learning/gaussian_naive_bayes.py`](https://github.com/TheAlgorithms/Python/blob/38dfcd28b5f2fb19bae130a942466d73933e072f/machine_learning/gaussian_naive_bayes.py)
- [ ] [`machine_learning/gradient_boosting_regressor.py`](https://github.com/TheAlgorithms/Python/blob/38dfcd28b5f2fb19bae130a942466d73933e072f/machine_learning/gradient_boosting_regressor.py)
- [x] [`machine_learning/multilayer_perceptron_classifier.py`](https://github.com/TheAlgorithms/Python/blob/38dfcd28b5f2fb19bae130a942466d73933e072f/machine_learning/multilayer_perceptron_classifier.py)
- [x] [`machine_learning/polymonial_regression.py`](https://github.com/TheAlgorithms/Python/blob/38dfcd28b5f2fb19bae130a942466d73933e072f/machine_learning/polymonial_regression.py) and has typos in name and file (polymonial -> polynomial)
[...]

PR #8965, opened 2023-08-16
Delete base85 algorithm

CaedenPH:
### Describe your change:
Re #6216

Normally, I'm not in favour of just deleting algorithms, but I would make the argument that this is not an algorithm, rather just a snippet of code that utilises another library.

Per `CONTRIBTUING.md`
> Algorithms in this repo should not be how-to examples for existing Python packages. Instead, they should perform internal calculations or manipulations to convert input values into different output values
This `base85` algorithm has essentially got two lines of code that purely utilise a singular library. The doctests only test an external library

This repository should not contains examples on how to use a certain library, that would be the library documentation here
https://docs.python.org/3/library/base64.html

* [ ] Add an algorithm?
* [ ] Documentation change?
* [x] Delete an algorithm

### Checklist:


CaedenPH:
@tianyizheng02 Opinion?

Will also create a pr to delete 
https://github.com/TheAlgorithms/Python/blob/master/ciphers/base32.py

However, I like the [base16](https://github.com/TheAlgorithms/Python/blob/master/ciphers/base16.py) and the [base64](https://github.com/TheAlgorithms/Python/blob/master/ciphers/base64.py) because they actually use calculations, not a library

tianyizheng02:
Obviously I agree that we can't have this "algorithm" simply utilize a library, but personally I'd rather we keep the file around and encourage other contributors to reimplement the algorithm instead. Unless one of us reimplements the algorithm right after this file is deleted, there's no guarantee that someone else will contribute the same algorithm again. Having the file around could at least serve as a reminder of an ongoing issue.

PR #8970, opened 2023-08-16
Fix get top billionaires BROKEN file

CaedenPH:
### Describe your change:
Fixes a `get_top_billionaires` and unmarks it as disabled

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


web_programming/get_top_billionaires.py
```diff
@@ -20,18 +20,29 @@
...
 
     Returns:
         Age as string
 
     >>> calculate_age(-657244800000)
-    '73'
+    '74'
     >>> calculate_age(46915200000)
-    '51'
+    '52'
```
tianyizheng02, line 32: Why change the doctest outputs? Were they originally off by 1?

web_programming/get_top_billionaires.py
```diff
@@ -20,18 +20,29 @@
 )
 
 
-def calculate_age(unix_date: int) -> str:
+def calculate_age(unix_date: float) -> str:
```
tianyizheng02, line 23: Why change the input to a float? Does `person["birthDate"]` store a float rather than an int?

web_programming/get_top_billionaires.py
CaedenPH: Yes, the age of the timestamps increased 😆 Interesting doctest because each year the ages will have to be incremented

web_programming/get_top_billionaires.py
CaedenPH: We divide unix_date by 1000, so the unix_date variable is redclared as float. 
The date passed is an int and we could just cast the division to an int but for simplicity sake I just type the param as float

web_programming/get_top_billionaires.py
tianyizheng02: Ok that's really not ideal... Can we modify these tests so that the ages automatically update? Perhaps by converting dummy datetime objects to UNIX timestamps and passing those into the function

tianyizheng02 (changes requested):
Could we change the doctests so that we don't have to update the ages every year? (see [previous comment](https://github.com/TheAlgorithms/Python/pull/8970/files#r1298331432))

tianyizheng02 (approved):
Tried it locally, looks like it works 👍

PR #9001, opened 2023-08-20
added smith waterman algorithm

BAW2501:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,97 @@
...
+    Returns 1 if the characters match, -1 if they mismatch.
+    >>> score_function('A', 'A')
+    1
+    >>> score_function('A', 'C')
+    -1
+    """
+    if source_char == target_char:
+        return MATCH
+    else:
+        return MISMATCH
```
CaedenPH, line 24:
```suggestion
    return MATCH if source_char == target_char else MISMATCH
```

dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def smith_waterman(query: str, subject: str) -> list[list[int]]:
+    """
+    Perform the Smith-Waterman local sequence alignment algorithm.
+    Returns a 2D list representing the score matrix. Each value in the matrix
+    corresponds to the score of the best local alignment ending at that point.
+    >>> smith_waterman('ACAC', 'CA')
+    [[0, 0, 0], [0, 0, 1], [0, 1, 0], [0, 0, 2], [0, 1, 0]]
```
CaedenPH, line 33: Can you write some more tests testing the following cases:
1. Empty `query`
2. Empty `subject`
3. Lowercase string

BAW2501: done

dynamic_programming/smith_waterman.py
BAW2501: Added it

dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,97 @@
...
+    return score
+
+
+def traceback(score: list[list[int]], query: str, subject: str) -> str:
+    r"""
+    Perform traceback to find the optimal local alignment.
+    Starts from the highest scoring cell in the matrix and traces back recursively
+    until a 0 score is found. Returns the alignment strings.
+    >>> traceback([[0, 0, 0], [0, 0, 1], [0, 1, 0], [0, 0, 2], [0, 1, 0]], 'ACAC', 'CA')
+    'CAC\nCA-'
```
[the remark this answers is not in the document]
BAW2501: yes, I realized it needed more testing, so I added a few (16 tests in total)

dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,128 @@
...
+    >>> traceback([[0, 0, 0]], 'ACAC', '')
+    ''
+    """
+    # make both query and subject uppercase
+    query = query.upper()
+    subject = subject.upper()
+    # Traceback logic to find optimal alignment
+    i = len(query)
+    j = len(subject)
```
tianyizheng02, line 94: > Starts from the highest scoring cell in the matrix

Does this not start at the bottom-right corner of the matrix rather than the highest-scoring cell? Is it guaranteed that the highest-scoring cell will always be in the bottom-right corner?

tianyizheng02 (changes requested):

dynamic_programming/smith_waterman.py
BAW2501: you're right These alignment methods have global (whole string such as Needleman wunch) and local (sub string such as smith watermen) variants, and in SW the optimal alignment isn't guaranteed to be in the bottom right,I'll make the necessary tweaks in an hour or so.

dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,128 @@
+# https://en.wikipedia.org/wiki/Smith%E2%80%93Waterman_algorithm
+# Score constants
+"""
+Score constants used in the Smith-Waterman algorithm. Matches are given a positive
+score while mismatches are given a negative score. Gaps are also penalized.
+"""
+MATCH = 1
+MISMATCH = -1
+GAP = -2
```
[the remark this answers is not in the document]
BAW2501: noted for the next commit

dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,142 @@
+"""
+https://en.wikipedia.org/wiki/Smith%E2%80%93Waterman_algorithm
+The Smith-Waterman algorithm is a dynamic programming algorithm used for sequence
+alignment.It is particularly useful for finding similarities between two sequences,
+such as DNA or protein sequences. In this implementation, gaps are penalized
+linearly,meaning that the scoreis reduced by a fixed amount for each gap introduced
+in the alignment. It's important to notethat the Smith-Waterman algorithm supports
+other gap penalty methods as well, but in thisspecific implementation, linear gap
+penalties are used.
```
tianyizheng02, line 9:
```suggestion
The Smith-Waterman algorithm is a dynamic programming algorithm used for sequence
alignment. It is particularly useful for finding similarities between two sequences,
such as DNA or protein sequences. In this implementation, gaps are penalized
linearly, meaning that the score is reduced by a fixed amount for each gap introduced
in the alignment. However, it's important to note that the Smith-Waterman algorithm
supports other gap penalty methods as well.
```
Fixed some typos and shortened the last sentence because I felt that it was a bit redundant.

tianyizheng02 (commented):
Overall, I think the implementation looks good, but I have some general feedback:
1. Can we make the score constants customizable? For example, we could have function parameters with default values so that the user can choose their score constants.
2. Can we have more variety in the doctests? The current doctests are mostly variations on the same example input. The Wikipedia page for the Smith–Waterman algorithm has a more complex example, and that'd be good to add as a doctest (though the example uses different score constants).

dynamic_programming/smith_waterman.py
```diff
@@ -0,0 +1,185 @@
...
+    -2
+    >>> score_function('-', '-')
+    -2
+    """
+    if "-" in (source_char, target_char):
+        return gap
+    return match if source_char == target_char else mismatch
+
+
+def smith_waterman(query: str, subject: str) -> list[list[int]]:
```
tianyizheng02, line 39: Could you add optional parameters for the score constants in this function's header as well? Users will call this function specifically to run the algorithm, and currently there's no way for the user to pass in their desired score constants into this function.

tianyizheng02 (changes requested):

dynamic_programming/smith_waterman.py
BAW2501: i hate repeating code, but there seems to be no way around it when it comes to default params 
<pre>
def smith_waterman(
    query: str,
    subject: str,
    match: int = 1,
    mismatch: int = -1,
    gap: int = -2,
) -> list[list[int]]:
</pre>
<pre>
def score_function(
    source_char: str,
    target_char: str,
    match: int = 1,
    mismatch: int = -1,
    gap: int = -2,
) -> int:
</pre>
either way i got about it i'll have to unpack the kwargs in score function params or the body of the function and i think the params are a better way for clarity

tianyizheng02 (approved):

PR #9005, opened 2023-08-21
fixes #9002; improve insertion_sort algorithm

### LINKED ISSUE #9002: improvement to insertion_sort algorithm
### Feature description

I was about to make a PR to improve the implementation of [insertion_sort algorithm](https://github.com/TheAlgorithms/Python/blob/master/sorts/insertion_sort.py) but since there might be multiple ways of doing so, I thought I should first ask your opinions.

These are the things that need improvements:

1. We unnecessarily create a whole new copy of the list: `enumerate(collection[1:])`.

   We can either use "indexes" to avoid this which is not very pythonic, or we can use the iterator of the list using `iter()` and throw away the first item using `next()`. In second case we have to either check for empty list first or wrap it in a try-except block. I'll go with indexes if you ask. What do you think?

2. I think a function should either mutate the list in-place and returns `None`, or it should create new sorted list without modifying the original list. Mutating the list and returning the mutated list is not what most developers expect to see. What do you think?

3. We can safely remove `if insert_index != temp_index:` condition and unindent its body. Assigning an item to an index of a list is not costly. So it's one less line in general.

amirsoroush:
### Describe your change:

Fixes #9002.

- type hints added.
- I left the return value untouched as other sorting algorithms are all returning the mutated list.

### Checklist:

### DIFF
--- a/sorts/insertion_sort.py
+++ b/sorts/insertion_sort.py
@@ -40,13 +40,12 @@ def insertion_sort(collection: list) -> list:
     True
     """
 
-    for insert_index, insert_value in enumerate(collection[1:]):
-        temp_index = insert_index
-        while insert_index >= 0 and insert_value < collection[insert_index]:
-            collection[insert_index + 1] = collection[insert_index]
+    for insert_index in range(1, len(collection)):
+        insert_value = collection[insert_index]
+        while insert_index > 0 and insert_value < collection[insert_index - 1]:
+            collection[insert_index] = collection[insert_index - 1]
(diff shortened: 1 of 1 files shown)


tianyizheng02 (approved):
LGTM

Regarding the function simultaneously modifying the list and returning it, I definitely don't like that behavior either. Perhaps instead we could simply have an optional bool parameter `in_place` that determines whether the function modifies the input list or returns a new sorted list. I've seen this idea used by pandas functions and I think it could work well here.

PR #9007, opened 2023-08-22
Fix type errors in permutations

CaedenPH:
### Describe your change:
Fixes type errors in `data_structures/arrays/permutations.py` 

- Removes the number `2` from the name of the second permute function which implements permutations using a backtrack implementation
- Type the parameters and return type of inner `permute`
- Type `output` as `list[list[int]]`

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


data_structures/arrays/permutations.py
```diff
@@ -1,33 +1,32 @@
...
-    >>> all(list(nums) in permute(numbers) for nums in permutations(numbers))
-    True
+
+    >>> permute_backtrack([1, 2, 3])
+    [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 2, 1], [3, 1, 2]]
     """
-    result = []
+    result: list[list[int]] = []
     if len(nums) == 1:
         return [nums.copy()]
```
tianyizheng02, line 10:
```suggestion
    if len(nums) == 0:
        return [[]]
```
I think we should make the base case 0 rather than 1. This slightly extends the function to work on an empty list input, and this matches the behavior of `itertools.permutations`.

data_structures/arrays/permutations.py
```diff
@@ -45,6 +44,6 @@ def backtrack(start):
     import doctest
 
     # use res to print the data in permute2 function
```
tianyizheng02, line 46:
```suggestion
```
This comment isn't really necessary so we might as well get rid of it while we're happening to be fixing it up

data_structures/arrays/permutations.py
```diff
@@ -1,33 +1,32 @@
-def permute(nums: list[int]) -> list[list[int]]:
+def permute_iterative(nums: list[int]) -> list[list[int]]:
```
tianyizheng02, line 1:
```suggestion
def permute_recursive(nums: list[int]) -> list[list[int]]:
```
I'd argue that this function is a recursive implementation, not an iterative one, because we're constructing the permutations by repeatedly calling the function on smaller and smaller inputs

tianyizheng02 (changes requested):

data_structures/arrays/permutations.py
CaedenPH: Agreed. Didn't see the fact that it called itself

tianyizheng02 (approved):

PR #9020, opened 2023-08-30
Added an add at position subroutiune to linked list

mendacium-a11y:
### Describe your change:

This pull request adds the `add_at_position` function to the linked list implementation. This function allows inserting an item at a specified position within the linked list.

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-<!-- # Contributing guidelines -->
+# Contributing guidelines
 
 ## Before contributing
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing guidelines
+<!-- # Contributing guidelines -->
 
 ## Before contributing
 
--- a/data_structures/linked_list/__init__.py
+++ b/data_structures/linked_list/__init__.py
@@ -26,6 +26,24 @@ class LinkedList:
         self.size += 1
 
     def add_at_position(self, item: Any, position: int) -> bool:
+        """
+        Adds a new node with the given item at the specified position in the linked list
+
+        Args:
+            item (Any): The item to be added to the linked list.
+            position (int): The position at which the item should be inserted.
+
+        Returns:
+            bool: True if the insertion was successful, False otherwise.
+
+        >>> linked_list = LinkedList()
+        >>> linked_list.add(1)
+        >>> linked_list.add(2)
+        >>> linked_list.add(3)
+        >>> linked_list.add_at_position(10, 1)
+        True
+
+        """
         if position < 0:
             return False
 
--- a/data_structures/linked_list/__init__.py
+++ b/data_structures/linked_list/__init__.py
@@ -25,6 +25,23 @@ class LinkedList:
         self.head = Node(item, self.head)
         self.size += 1
 
+    def add_at_position(self, item: Any, position: int) -> bool:
+        if position < 0:
+            return False
+
+        current = self.head
+        counter = 0
+        while current and counter < position:
+            current = current.next
+            counter += 1
+
+        if current:  # Check if current is not None
+            new_node = Node(item, current.next)
+            current.next = new_node
+            return True
+        else:
+            return False
+
     def remove(self) -> Any:
         # Switched 'self.is_empty()' to 'self.head is None'
         # because mypy was considering the possibility that 'self.head'
--- a/data_structures/linked_list/__init__.py
+++ b/data_structures/linked_list/__init__.py
@@ -21,7 +21,7 @@ class LinkedList:
         self.head: Node | None = None
         self.size = 0
 
-    def add(self, item: Any, position: int = 0) -> bool:
+    def add(self, item: Any, position: int = 0) -> None:
         """
         Add an item to the LinkedList at the specified position.
         Default position is 0 (the head).
@@ -36,13 +36,9 @@ class LinkedList:
 
         >>> linked_list = LinkedList()
         >>> linked_list.add(1)
-        True
         >>> linked_list.add(2)
-        True
         >>> linked_list.add(3)
-        True
         >>> linked_list.add(4, 2)
-        True
         >>> print(linked_list)
         3 --> 2 --> 4 --> 1
         """
@@ -62,11 +58,8 @@ class LinkedList:
             if current:  # Check if current is not None
                 new_node = Node(item, current.next)
                 current.next = new_node
-                return True
-            return False
 
         self.size += 1
-        return True
 
     def remove(self) -> Any:
         # Switched 'self.is_empty()' to 'self.head is None'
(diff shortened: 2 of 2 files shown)


data_structures/linked_list/__init__.py
```diff
@@ -25,6 +25,41 @@ def add(self, item: Any) -> None:
...
+        Returns:
+            bool: True if the insertion was successful, False otherwise.
+
+        >>> linked_list = LinkedList()
+        >>> linked_list.add(1)
+        >>> linked_list.add(2)
+        >>> linked_list.add(3)
+        >>> linked_list.add_at_position(10, 1)
+        True
+
```
CaedenPH, line 45: Here there should be a test displaying `linked_list`

data_structures/linked_list/__init__.py
```diff
@@ -25,6 +25,41 @@ def add(self, item: Any) -> None:
...
+        if position < 0:
+            return False
+
+        current = self.head
+        counter = 0
+        while current and counter < position:
+            current = current.next
+            counter += 1
+
+        if current:  # Check if current is not None
```
CaedenPH, line 56:
```suggestion
        if current is not None:
```

data_structures/linked_list/__init__.py
```diff
@@ -25,6 +25,41 @@ def add(self, item: Any) -> None:
...
+            new_node = Node(item, current.next)
+            current.next = new_node
+            return True
+        else:
+            return False
```
CaedenPH, line 61:
```suggestion
        return False
```

data_structures/linked_list/__init__.py
```diff
@@ -25,6 +25,41 @@ def add(self, item: Any) -> None:
...
+        >>> linked_list = LinkedList()
+        >>> linked_list.add(1)
+        >>> linked_list.add(2)
+        >>> linked_list.add(3)
+        >>> linked_list.add_at_position(10, 1)
+        True
+
+        """
+        if position < 0:
+            return False
```
CaedenPH, line 48: Maybe this should raise an error?
```suggestion
        if 0 > position:
            return False
```
This also looks a bit more readable

data_structures/linked_list/__init__.py
```diff
@@ -25,6 +25,41 @@ def add(self, item: Any) -> None:
         self.head = Node(item, self.head)
         self.size += 1
 
+    def add_at_position(self, item: Any, position: int) -> bool:
```
CaedenPH, line 28: What about `insert_at_position`?

data_structures/linked_list/__init__.py
tianyizheng02, line 1: Instead of having a separate function for adding at a specific position, why not modify the existing `add` function to take an optional position argument? If the user doesn't specify a position, then the function defaults to position 0, which adds to the head of the list, exactly as `add` currently does.

tianyizheng02 (changes requested):

data_structures/linked_list/__init__.py
mendacium-a11y: updated the existing `add` function to take an optional position argument to specify where we want to insert the node at

mendacium-a11y:
updated the the code by dropping the add_at_position, instead updated the add function to take an optional position argument to specify of we want to insert the node at a specific position in the linked list

CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-# Contributing guidelines
+<!-- # Contributing guidelines -->
```
tianyizheng02, line 1: Don't edit the contributing guidelines

data_structures/linked_list/__init__.py
```diff
@@ -21,8 +21,45 @@ def __init__(self) -> None:
...
+            counter = 0
+            while current and counter < position - 1:
+                current = current.next
+                counter += 1
+
+            if current:  # Check if current is not None
+                new_node = Node(item, current.next)
+                current.next = new_node
+            else:
+                raise ValueError("Out of bounds")
```
tianyizheng02, line 62:
```suggestion
            for _ in range(position - 1):
                current = current.next
                if current is None:
                    raise ValueError("Out of bounds")
            new_node = Node(item, current.next)
            current.next = new_node
```
I _think_ this will do the same thing without needing a counter to track the current index.

data_structures/linked_list/__init__.py
```diff
@@ -21,8 +21,45 @@ def __init__(self) -> None:
...
+        Raises:
+            ValueError: If the position is negative.
+
+        >>> linked_list = LinkedList()
+        >>> linked_list.add(1)
+        >>> linked_list.add(2)
+        >>> linked_list.add(3)
+        >>> linked_list.add(4, 2)
+        >>> print(linked_list)
+        3 --> 2 --> 4 --> 1
```
tianyizheng02, line 43: Can you add doctests to test positions that are negative, out of bounds, and at the end of the linked list?

tianyizheng02 (changes requested):

CONTRIBUTING.md
mendacium-a11y: this seems to have happened by mistake, didn't intend to

mendacium-a11y:
updated the else condition after checking the poisiton argument value and added doctest for negative posiiton argument and appending at the end

mendacium-a11y:
added the doctest and updated the else code after checking the position is 0 or not

data_structures/linked_list/__init__.py
```diff
@@ -41,6 +41,17 @@ def add(self, item: Any, position: int = 0) -> None:
...
+        # Test adding to a negative position
+        >>> linked_list.add(5, -3)
+        Traceback (most recent call last):
+            ...
+        ValueError: Position must be non-negative
+
+        # Test adding to an out-of-bounds position
+        >>> linked_list.add(5, 4)
+        >>> print(linked_list)
+        3 --> 2 --> 4 --> 1 --> 5
```
tianyizheng02, line 54: This isn't testing an out-of-bounds position, but rather it's testing the end of the list. Since the indices are 0, 1, 2, and 3, adding at index 4 means adding to the end of the list.

We should still keep this test, though. Instead, could you add one more test to actually test an out-of-bounds position? For example:
```py
>>> linked_list.add(0, 6)
```
This test tries to add 0 to the list at index 6, which is definitely beyond the end of the list.

CONTRIBUTING.md
tianyizheng02:
```suggestion
# Contributing guidelines
```
Then please uncomment this line

tianyizheng02 (changes requested):

data_structures/linked_list/__init__.py
mendacium-a11y: out of bounds both positive and negative, got it

mendacium-a11y:
fixed contributing.md and added the doctest for checing out of bounds posiiton, both positive and negative

tianyizheng02:
Thanks for your contribution!

PR #8997, opened 2023-08-20
PR on added uniform cost search algorithm

Miltonbhowmick:
### Describe your change:
Hi, I have added uniform cost search algorithm with fresh pull request. 

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [ ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

PR #9041, opened 2023-09-04
Added uniform search cost algorithm on graph node based

Miltonbhowmick:
### Describe your change:
I have readded the uniform search cost with graph based environment. My first tried was with grid based environment here https://github.com/TheAlgorithms/Python/pull/8997.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


graphs/uniform_search_cost.py
```diff
@@ -0,0 +1,128 @@
...
+    graph_data = [[] for _ in range(4)]  # type: list[list[int]]
+
+    # Node to node edges
+    graph_data[0].append(1)
+    graph_data[0].append(2)
+    graph_data[1].append(2)
+    graph_data[1].append(3)
+
+    # Distance for the edges one node to another node
+    distance = {(0, 1): 2, (0, 2): 1, (1, 2): 5, (1, 3): 10}
```
tianyizheng02, line 112: 1) Why is the data for `graph_data` added after the list is instantiated? Why can't we just do `graph_data = [[1, 2], [2, 3]]`?
2) What graph representation is `graph_data`? Is it meant to be an adjacency list? If so, why are there no vertex lists for `2` and `3`?
3) You really shouldn't store the adjacency data and edge distances separately if you can avoid it. Is there a good reason to store the distances separately?

graphs/uniform_search_cost_grid.py
tianyizheng02, line 1: As I've already said on [your previous PR](https://github.com/TheAlgorithms/Python/pull/8997#issuecomment-1687836420), I don't think we need a grid version of the same algorithm. The graph version encompasses the grid version, but not vice versa.

tianyizheng02 (changes requested):
There are other issues that need to be addressed, but these are the ones that most immediately need to be addressed.

graphs/uniform_search_cost_grid.py
Miltonbhowmick: It is my mistake while doing commit. I understand now that i need to delete `graphs/uniform_search_cost_grid.py` file from repository so that it would not be added in the PR.

graphs/uniform_search_cost.py
Miltonbhowmick: Thanks for your queries. 

1. `graph_data = [[1, 2], [2, 3]]` represents a graph where nodes 0, 1, and 2 are connected in a linear fashion. I made it like a familiar readable way of node to node connection for static test case.
2. `graph_data` is meant to be an adjacency list representation of a graph. Do i need to add some comments about graph in the code?
3. Actually, i was storing adjacency data and edge distances separately is common in graph algorithms because it provides flexibility to me handling.

Miltonbhowmick: @tianyizheng02, can you look over the latest change about the `graph_data` with test cases?

graphs/uniform_search_cost.py
```diff
@@ -0,0 +1,163 @@
...
+    >>> start = 0
+    >>> uniform_search_cost(graph_data, distance, goal, start)
+    [1]
+    """
+    # priority queue
+    heap = []
+    goal_answer = []
+    # Goal result to be max value initially
+    for _ in range(len(goal)):
+        goal_answer.append(10**8)
```
CaedenPH, line 44:
```suggestion
    # Goal result to be max value initially
    goal_answer = [10**8] * len(goal)
```

graphs/uniform_search_cost.py
```diff
@@ -0,0 +1,163 @@
...
+    goal: list[int],
+    start: int,
+) -> list[int]:
+    """
+    Returns a list of integer values which are result of optimal distance
+    >>> graph_data = [[1, 2], [2, 3], [], []]
+    >>> distance = {(0, 1): 2, (0, 2): 1, (1, 2): 5, (1, 3): 10}
+    >>> goal = [2]
+    >>> start = 0
+    >>> uniform_search_cost(graph_data, distance, goal, start)
```
CaedenPH, line 36: Can you create some more test cases with negative values, 0, floats, etc?

graphs/uniform_search_cost.py
```diff
@@ -0,0 +1,163 @@
...
+    heap.append([0, start])
+
+    # store tracks of visited nodes
+    visited = {}
+    final_cnt = 0
+    while len(heap) > 0:
+        heap = sorted(heap)
+        # get the top element of the node list which has less distance value
+        current_node = heap[-1]
+        del heap[-1]
```
CaedenPH, line 56:
```suggestion
        current_node = heap.pop()
```

graphs/uniform_search_cost.py
Miltonbhowmick: UCS typically assumes that edge weights are non-negative. This means that UCS may not work correctly when dealing with graphs that have negative edge weights. Because Uniform Cost Search (UCS) is a variant of Dijkstra's algorithm that is used to find the shortest path in a weighted graph. Yes, floats and 0 supports in UCS.

graphs/uniform_search_cost.py
tianyizheng02: > 1. `graph_data = [[1, 2], [2, 3]]` represents a graph where nodes 0, 1, and 2 are connected in a linear fashion. I made it like a familiar readable way of node to node connection for static test case.
> 2. `graph_data` is meant to be an adjacency list representation of a graph. Do i need to add some comments about graph in the code?

I think you're misunderstanding what an adjacency list is. What I suggested _is_ the (partly) correct adjacency list representation for the graph that you're proposing. An adjacency list contains a list of neighbors for each vertex; it does _not_ simply contain a list of edges.

Your graph has vertex 0 connected to vertices 1 and 2 and vertex 1 connected to vertices 2 and 3. In `graph_data = [[1, 2], [2, 3]]`, index 0 (which represents vertex 0) contains 1 and 2, and index 1 (which represents vertex 1) contains 2 and 3.

(Yes, the correct adjacency list should actually be `graph_data = [[1, 2], [2, 3], [], []]` to account for vertices 2 and 3, which have no neighbors. However, my point still stands that you can easily instantiate the entire adjacency list all at once.)

Also, yes, you should absolutely specify in the code that the input graph should be represented as an adjacency list.

> 3. Actually, i was storing adjacency data and edge distances separately is common in graph algorithms because it provides flexibility to me handling.

I strongly disagree with this. Firstly, in my experience it's not common to separate adjacency data and edge distances in graph representations (though you can disagree with me on this since it's anecdotal). However, it's still definitely worse IMO to decouple the adjacency data from the edge distances. Needing to update two separate data structures whenever you ever modify the graph makes it more prone to mistakes. Graph adjacency list representations can absolutely store edge distances alongside the adjacency lists for each vertex.

graphs/uniform_search_cost.py
```diff
@@ -0,0 +1,181 @@
...
+    >>> distance = {(0,1):1.3,(0,2):2.6,(1,2):0,(1,3):2.3,(3,0):2.34,(3,2):2.15}
+    >>> goal = [2]
+    >>> start = 0
+    >>> uniform_search_cost(graph_data, distance, goal, start)
+    [1.3]
+    """
+    # priority queue
+    heap = []
+    # Goal result to be max value initially
+    goal_answer = [10**8] * len(goal)
```
tianyizheng02, line 63: Please use something like `float("inf")` for the arbitrary max value. `10**8` wouldn't work as a max value if, for example, an edge has distance greater than 10^8.

graphs/uniform_search_cost.py
```diff
@@ -0,0 +1,181 @@
...
+        (3, 1): 5,  # Cost between node 3 and 1 is 5
+        (3, 6): 6,  # Cost between node 3 and 6 is 6
+        (3, 4): 2,  # Cost between node 3 and 4 is 2
+        (2, 1): 4,  # Cost between node 2 and 1 is 4
+        (4, 2): 4,  # Cost between node 4 and 2 is 4
+        (4, 5): 3,  # Cost between node 4 and 5 is 3
+        (5, 2): 6,  # Cost between node 5 and 2 is 6
+        (5, 6): 3,  # Cost between node 5 and 6 is 3
+        (6, 4): 7,  # Cost between node 6 and 4 is 7
+    }
```
tianyizheng02, line 165: As already stated, please simply instantiate the entire adjacency list rather than manually adding every single edge. Also, if you're planning to use dense graphs like this for your algorithm, why not use an adjacency matrix rather than an adjacency list?

tianyizheng02 (changes requested):

PR #9044, opened 2023-09-07
Add stalin sort algorithm

H4ckerman0:
### Describe your change:

* [X] Add an algorithm?
* [ ] Documentation change?

### Checklist:


sorts/stalin_sort.py
```diff
@@ -0,0 +1,37 @@
+"""
+This is a pure Python implementation of the stalin sort algorithm,
+inspired by the dictator of the former soviet union Josef Stalin.
```
tianyizheng02, line 3:
```suggestion
This is a pure Python implementation of the Stalin sort algorithm,
inspired by the dictator of the former Soviet Union Josef Stalin.
```

sorts/stalin_sort.py
```diff
@@ -0,0 +1,37 @@
...
+
+This sorting algorithm sorts a collection of elements by deleting
+every element that is not in order resulting in a sorted collection.
+
+For manual testing run:
+python stalin_sort.py
+"""
+
+
+def stalin_sort(collection):
```
tianyizheng02, line 13: Please add type hints

sorts/stalin_sort.py
```diff
@@ -0,0 +1,37 @@
...
+    """A pure Python implementation of stalin sort algorithm
+
+    :param collection: a mutable collection of comparable items
+    :return: the same collection ordered by ascending
+
+    Examples:
+    >>> stalin_sort([1, 6, 3, 20, 30])
+    [1, 6, 20, 30]
+    >>> stalin_sort([])
+    []
```
tianyizheng02, line 23: This test is failing because the function tries to get the first element of the list when there is none

sorts/stalin_sort.py
```diff
@@ -0,0 +1,37 @@
...
+    >>> stalin_sort([-2, 5, -3, 30, 15, 7])
+    [-2, 5, 30]
+    """
+    sorted_collection = [collection[0]]
+    for i in collection[1:]:
+        if i > sorted_collection[-1]:
+            sorted_collection.append(i)
+    return sorted_collection
```
tianyizheng02, line 31: Instead of building a new sorted list, can we have the function actually remove elements from the input list? It's more true to the spirit of the algorithm that way

tianyizheng02 (changes requested):

sorts/stalin_sort.py
H4ckerman0: Thank you for your kind and helpful advices! I'll try to implement all of them, but I have a question for this one:
Should I add type hints to the whole script or is the functions head enough?

PR #9072, opened 2023-09-20
Speed improvement, sign recognition simplified

AlfredoCode:
This PR consists of:
Sign recognition simplification
Creating constant of len(binary_str)


rohan472000:
See other PRs as how they provide description, and do the same.

PR #9084, opened 2023-09-24
Linear Regression Change (Fixes #8780)

### LINKED ISSUE #8780: Do... something? about `machine_learning/forecasting`
### Feature description

In reference to #7305, it's clear to me that @nandiya isn't coming back to fix the warnings in `machine_learning/forecasting/run.py`. With that said, I want to discuss some more fundamental issues that I have with the file and get other opinions on what to do about them.

- At the risk of sounding harsh, I think some of the code was haphazardly written. The warnings mentioned in #7305 are mostly due to the code providing far too few input observations (like, literally only 4 observations). Adding more observations doesn't necessarily resolve the warnings either, as there are other bugs in the code (passing bad arguments into functions and incorrectly reading a CSV file) that prevent the code from fully running. To me, it almost seems like the file was never run or tested independently, as these bugs and warnings would've been caught by just running the code.
- I suspect that the code could've been plagiarized (or at the very least uncredited). At the top of the file, it says this:
    
    > this is code for forecasting
    > but i modified it and used it for safety checker of data
   
    ... but modified from where? No sources were ever referenced.
- This file isn't an algorithm, much less a machine learning algorithm. Forecasting is a very general statistical task, and there's no single algorithm for it. In actuality, this file uses three different statistical methods (linear regression, SARIMAX, and SVR) to complete a forecasting task. Why not simply implement each of those algorithms in separate files?
- Related to the previous point, this file is clearly a how-to. The SARIMAX and SVR code relies entirely on pre-existing implementations in other packages.

My main question is what we should do about this file. Fixing the existing code would resolve the warnings brought up in #7305 and would address the first issue, but it doesn't really address any of the other issues. What do you all think we should do about this file?

JrX970:
### Describe your change:

* [X] Add an algorithm?
* [X] Documentation change?

### Checklist:
* [N/A] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.

Changed the linear regression method to simple linear regression as the current file and data is a univariate case. A separate linear regression file, probably called General Linear Model (Not Generalised Linear Model!) or method of least squares should be used with the ability to load in data, fit whatever equation you want (Changing the design matrix) and allow for data transformations (transforming non-linear target variables to linear). Just like this website here: https://textbooks.math.gatech.edu/ila/least-squares.html

I also haven't added any unit testing, it seems that the previous script didn't either, it loaded in a data set of it's own, so it is more of a showcase on how to do linear regression?

This is my first ever pull request (ever), so I hope I got this right too!

### DIFF
--- a/machine_learning/linear_regression.py
+++ b/machine_learning/linear_regression.py
@@ -12,8 +12,8 @@ Rating). We try to best fit a line through dataset and estimate the parameters.
 https://en.wikipedia.org/wiki/Simple_linear_regression
 This link explains the methodology.
 
-https://en.wikipedia.org/wiki/F-test
-This one gives a bit more information on the F-Test
+https://en.wikipedia.org/wiki/f-test
+This one gives a bit more information on the f-Test
 
 NB_1: R Squared is not a "Goodness of fit", R squared is
 the value that tells you how much of the variance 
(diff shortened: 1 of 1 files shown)


tianyizheng02:
Thank you for your contribution, but unfortunately we're not accepting new PRs at the moment. We're currently trying to clear our backlog of existing PRs in preparation for Hacktoberfest. If you want to contribute, please wait until after October 1 to do so. See [this discussion](https://github.com/TheAlgorithms/Python/discussions/8912#discussion-5458835) for more info.

In addition, your PR only works for a single predictor variable while the old implementation worked for multiple. Please have your implementation work for multiple variables, as this doesn't reduce the functionality of the algorithm.

PR #9114, opened 2023-09-28
Fixes #8847

### LINKED ISSUE #8847: `machine_learning/linear_regression.py` doesn't give optimal coefficients

mahi01agarwal:
### Describe your change:

Fixes #8847 

* [x] Add an algorithm?
* [ ]  Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/linear_regression.py
+++ b/machine_learning/linear_regression.py
@@ -3,14 +3,16 @@ Linear regression is the most basic type of regression commonly used for
 predictive analysis. The idea is pretty simple: we have a dataset and we have
 features associated with it. Features should be chosen very cautiously
 as they determine how much our model will be able to make future predictions.
-We try to set the weight of these features, over many iterations, so that they best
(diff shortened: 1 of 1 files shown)


tianyizheng02:
Thank you for your contribution, but unfortunately we're not accepting new PRs at the moment. We're currently trying to clear our backlog of existing PRs in preparation for Hacktoberfest. If you want to contribute, please wait until after October 1 to open a new PR.

mahi01agarwal:
Okay Thankyou .

PR #9126, opened 2023-09-29
Euler072 - application of vector operations  to reduce calculation time and refactoring numpy

quant12345:
### Describe your change:

Application of vector operations  to reduce calculation time and refactoring numpy.
With limit: int = 1_000_000, the calculation of the new algorithm is 40% faster, limit: int = 10_000_000 is almost twice as fast.

<details>

<summary>code test</summary>

```
import datetime

import numpy as np

"""
Problem 72 Counting fractions: https://projecteuler.net/problem=72

Description:

Consider the fraction, n/d, where n and d are positive integers. If n<d and HCF(n,d)=1,
it is called a reduced proper fraction.
If we list the set of reduced proper fractions for d ≤ 8 in ascending order of size, we
get: 1/8, 1/7, 1/6, 1/5, 1/4, 2/7, 1/3, 3/8, 2/5, 3/7, 1/2, 4/7, 3/5, 5/8, 2/3, 5/7,
3/4, 4/5, 5/6, 6/7, 7/8
It can be seen that there are 21 elements in this set.
How many elements would be contained in the set of reduced proper fractions for
d ≤ 1,000,000?

Solution:

Number of numbers between 1 and n that are coprime to n is given by the Euler's Totient
function, phi(n). So, the answer is simply the sum of phi(n) for 2 <= n <= 1,000,000
Sum of phi(d), for all d|n = n. This result can be used to find phi(n) using a sieve.

Time: 1 sec
"""


def solution(limit: int = 1_000_000) -> int:
    """
    Returns an integer, the solution to the problem
    >>> solution(10)
    31
    >>> solution(100)
    3043
    >>> solution(1_000)
    304191
    """

    phi = [i - 1 for i in range(limit + 1)]

    for i in range(2, limit + 1):
        if phi[i] == i - 1:
            for j in range(2 * i, limit + 1, i):
                phi[j] -= phi[j] // i

    return sum(phi[2 : limit + 1])


def solution_new(limit: int = 1_000_000) -> int:
    """
    Returns an integer, the solution to the problem
    >>> solution(10)
    31
    >>> solution(100)
    3043
    >>> solution(1_000)
    304191
    """

    # generating an array from -1 to limit
    phi = np.arange(-1, limit)

    for i in range(2, limit + 1):
        if phi[i] == i - 1:
            ind = np.arange(2 * i, limit + 1, i)# indexes for selection
            phi[ind] -= phi[ind] // i

    return np.sum(phi[2 : limit + 1])


if __name__ == "__main__":

    now = datetime.datetime.now()
    old = solution()
    print('Returns an integer old', old)
    time_original = datetime.datetime.now() - now
    print('time_original', time_original.total_seconds())

    now = datetime.datetime.now()
    new = solution_new()
    print('Returns an integer new', new)
    time_new = datetime.datetime.now() - now
    print('time_new', time_new.total_seconds())

    print('old == new', old == new)
```

</details>

Output:

```
limit: int = 1_000_000

Returns an integer old 303963552391
time_original 1.236859
Returns an integer new 303963552391
time_new 0.816209
old == new True


limit: int = 10_000_000

Returns an integer old 30396356427241
time_original 17.871657
Returns an integer new 30396356427241
time_new 9.741272
old == new True
```

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/arithmetic_analysis/lu_decomposition.py
+++ b/arithmetic_analysis/lu_decomposition.py
@@ -88,15 +88,19 @@ def lower_upper_decomposition(table: np.ndarray) -> tuple[np.ndarray, np.ndarray
 
     lower = np.zeros((rows, columns))
     upper = np.zeros((rows, columns))
+
+    # in 'total', the necessary data is extracted through slices
+    # and the sum of the products is obtained.
+
     for i in range(columns):
         for j in range(i):
-            total = sum(lower[i][k] * upper[k][j] for k in range(j))
+            total = np.sum(lower[i, :i] * upper[:i, j])
             if upper[j][j] == 0:
                 raise ArithmeticError("No LU decomposition exists")
             lower[i][j] = (table[i][j] - total) / upper[j][j]
         lower[i][i] = 1
         for j in range(i, columns):
-            total = sum(lower[i][k] * upper[k][j] for k in range(j))
+            total = np.sum(lower[i, :i] * upper[:i, j])
             upper[i][j] = table[i][j] - total
     return lower, upper
 
--- a/arithmetic_analysis/lu_decomposition.py
+++ b/arithmetic_analysis/lu_decomposition.py
@@ -88,19 +88,15 @@ def lower_upper_decomposition(table: np.ndarray) -> tuple[np.ndarray, np.ndarray
 
     lower = np.zeros((rows, columns))
     upper = np.zeros((rows, columns))
-
-    # in 'total', the necessary data is extracted through slices
-    # and the sum of the products is obtained.
-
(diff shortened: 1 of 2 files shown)


tianyizheng02:
Thank you for your contribution, but unfortunately we're not accepting new PRs at the moment. We're currently trying to clear our backlog of existing PRs in preparation for Hacktoberfest. If you want to contribute, please wait until after October 1 to do so.

quant12345:
@tianyizheng02 may I know in advance when you have Hacktoberfest and you temporarily do not accept pull requests?

PR #9135, opened 2023-09-30
Added median algorithm

sumeetghumare4:
### Describe your change:
1) It first sorts the input list in ascending order.
2) It checks whether the list has an even or odd number of elements.
3) If the list has an even number of elements, it calculates the median as the average of the two middle elements.
4) If the list has an odd number of elements, it calculates the median as the middle element.

This logic covers the standard method for finding the median of a list of numbers, and it should provide accurate results for both even and odd-length lists.

### DIFF
--- a/data_structures/arrays/median.py
+++ b/data_structures/arrays/median.py
@@ -26,8 +26,10 @@ def find_median(nums):
 
     return median
 
+
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


sumeetghumare4 (commented):
Issue is showing for comment there is need to fix it

PR #9141, opened 2023-09-30
Fixes#8847

### LINKED ISSUE #8847: `machine_learning/linear_regression.py` doesn't give optimal coefficients

mahi01agarwal:
### Describe your change:

Fixes #8847 

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/linear_regression.py
+++ b/machine_learning/linear_regression.py
@@ -3,14 +3,16 @@ Linear regression is the most basic type of regression commonly used for
 predictive analysis. The idea is pretty simple: we have a dataset and we have
 features associated with it. Features should be chosen very cautiously
 as they determine how much our model will be able to make future predictions.
-We try to set the weight of these features, over many iterations, so that they best
-fit our dataset. In this particular code, I had used a CSGO dataset (ADR vs
-Rating). We try to best fit a line through dataset and estimate the parameters.
+We try to set the weight of these features, using "sum of rectangular area
+over sum of square area" method which is a direct method.
+In this particular code, I had used a CSGO dataset (ADR vs Rating).
+We try to best fit a line through dataset and estimate the parameters.
 """
 import numpy as np
 import requests
 
 
+# Function to collect the CSGO dataset
 def collect_dataset():
     """Collect dataset of CSGO
     The dataset contains ADR vs Rating of a Player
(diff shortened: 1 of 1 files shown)


mahi01agarwal:
Hey , I have created a pull request #9141 that fixes issue #8847 , Please review it.ThankYou.

machine_learning/linear_regression.py
```diff
@@ -3,14 +3,16 @@
...
+We try to set the weight of these features, using "sum of rectangular area
+over sum of square area" method which is a direct method.
+In this particular code, I had used a CSGO dataset (ADR vs Rating).
+We try to best fit a line through dataset and estimate the parameters.
 """
 import numpy as np
 import requests
 
 
+# Function to collect the CSGO dataset
```
tianyizheng02, line 15:
```suggestion
```
This comment is unnecessary because we already have a docstring for the function.

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
     lines = response.text.splitlines()
     data = []
+
     for item in lines:
         item = item.split(",")
         data.append(item)
-    data.pop(0)  # This is for removing the labels from the list
+
+    # Remove the labels (headers) from the list
+    data.pop(0)
```
tianyizheng02, line 33:
```suggestion
    data.pop(0)  # Remove the labels (headers) from the list
```

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
+
+    # Convert data to a NumPy matrix
```
tianyizheng02, line 35:
```suggestion
```
Comment is unnecessary because this line is already clear without one

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
-def run_steep_gradient_descent(data_x, data_y, len_data, alpha, theta):
-    """Run steep gradient descent and updates the Feature vector accordingly_
-    :param data_x   : contains the dataset
-    :param data_y   : contains the output associated with each data-entry
-    :param len_data : length of the data_
-    :param alpha    : Learning rate of the model
-    :param theta    : Feature vector (weight's for our model)
-    ;param return    : Updated Feature's, using
-                       curr_features - alpha_ * gradient(w.r.t. feature)
+# Function to calculate Mean Absolute Error (MAE)
```
tianyizheng02, line 40:
```suggestion
```
Unnecessary comment, we already have a docstring for the function

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
+def calculate_mae(predicted_y, original_y):
```
tianyizheng02, line 41: Please add type hints

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
-def sum_of_square_error(data_x, data_y, len_data, theta):
-    """Return sum of square error for error calculation
-    :param data_x    : contains our dataset
-    :param data_y    : contains the output (result vector)
-    :param len_data  : len of the dataset
-    :param theta     : contains the feature vector
-    :return          : sum of square error computed from given feature's
+    return sum(abs(y - predicted_y[i]) for i, y in enumerate(original_y)) / len(
+        original_y
+    )
```
tianyizheng02, line 49:
```suggestion
    return sum(abs(y1 - y2) for y1, y2 in zip(predicted_y, original_y)) / len(
        original_y
    )
```

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
+
+
+# Function to perform simple linear regression
```
tianyizheng02, line 52:
```suggestion
```
Comment is unnecessary when there's already a docstring for the function

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
+def simple_solve(data_x, data_y):
```
tianyizheng02, line 53: Please add type hints

machine_learning/linear_regression.py
```diff
@@ -22,93 +24,76 @@ def collect_dataset():
...
-    :param predicted_y   : contains the output of prediction (result vector)
-    :param original_y    : contains values of expected outcome
-    :return          : mean absolute error computed from given feature's
-    """
-    total = sum(abs(y - predicted_y[i]) for i, y in enumerate(original_y))
-    return total / len(original_y)
+    print(f"sse is: {sse}")
+    print(f"Half mse is: {half_mse}")
+    print(f"Coefficient is: {beta_1}")
+    print(f"Intercept is: {beta_0}")
```
tianyizheng02, line 87: Please read the contributing guidelines. Algorithmic functions should avoid side effects, including printing. Instead, please have your function simply return the regression coefficients.

tianyizheng02 (changes requested):
Thanks for your contribution. However, I'd really prefer that you don't just limit your fix to the univariate case, because the old implementation worked for multiple variables. Could you provide a direct-method implementation that also works for multiple variables?

cclauss:
The title of this pull request is not self-documenting.  Why force readers and reviewers to look up an issue?  Closing.

Issue #9251, opened 2023-10-01
Feature: Cyclic Sort


sharansukesh1003:
### Feature description

Cyclic Sort is a simple yet efficient in-place sorting algorithm that works on the principle of mapping each element to its correct position. It achieves this by iterating through the array and placing each element at its rightful place, effectively sorting the entire array. The algorithm's distinctive feature is its constant space complexity, making it ideal for situations where memory usage is a concern. Despite its basic concept, Cyclic Sort offers a reliable and straightforward approach to sorting arrays in linear time complexity, making it a valuable addition to sorting algorithms.

tianyizheng02:
Once again, we already have an implementation of this: [sorts/cycle_sort.py](https://github.com/TheAlgorithms/Python/blob/eaa87bd791cdc18d210d775f3258767751f9d3fe/sorts/cycle_sort.py). And once again, please check if an algorithm is already implemented.

Also, if you do want to contribute an algorithm, why open an issue for it? Like the contributing guidelines say (which you should've read already), you can just open a PR.

sharansukesh1003:
For your information, cycle and cyclic sort are two different things.

https://en.wikipedia.org/wiki/Cycle_sort
https://blog.stackademic.com/coding-pattern-cyclic-sort-96511b0f60ac

tianyizheng02:
Literally the first line of the Stackademic article you cite:
> Cycle sort is an in-place sorting algorithm.

Wikipedia: "It is based on the idea that the permutation to be sorted can be factored into cycles, which can individually be rotated to give a sorted result."
Stackademic: "It is based on the idea that the array to be sorted can be divided into cycles. Cycles can be visualized as a graph."

sharansukesh1003:
Cycle sort and cyclic sort are related sorting algorithms, but they are not
exactly the same thing.

Cyclic Sort:
Cyclic sort is a comparison-based sorting algorithm that works on the
principle of cycling through the elements to arrange them in the correct
order. The basic idea behind cyclic sort is to iterate through the array,
place each element in its correct position by swapping it with the element
currently at that position. Cyclic sort assumes that the elements to be
sorted are integers in a specific range. It has a time complexity of O(n)
but requires O(1) additional space.

Cycle Sort:
Cycle sort is also a comparison-based in-place sorting algorithm. It is
based on the idea of finding cycles in the permutation being sorted. It
achieves this by rotating elements within the cycle to their correct
[...]

PR #9256, opened 2023-10-01
Create cyclic_sort.py

sharansukesh1003:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


cclauss:
```
 027     >>> cyclic_sort([-2, -5, -45])
UNEXPECTED EXCEPTION: IndexError('list index out of range')
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/doctest.py", line 1351, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest sorts.cyclic_sort.cyclic_sort[2]>", line 1, in <module>
  File "/home/runner/work/Python/Python/sorts/cyclic_sort.py", line 39, in cyclic_sort
    if nums[i] != nums[correct_index]:
                  ~~~~^^^^^^^^^^^^^^^
/home/runner/work/Python/Python/sorts/cyclic_sort.py:27: UnexpectedException
```

tianyizheng02:
1. The code is taken directly from the Stackademic article you reference in #9251: https://blog.stackademic.com/coding-pattern-cyclic-sort-96511b0f60ac
2. Like you said in #9251, one reason why cyclic sort is a distinct algorithm from cycle sort is that
> Cyclic sort assumes that the elements to be sorted are integers in a specific range.

The algorithm you propose basically only works if the list elements are within the range 1 to n. Naturally, `cyclic_sort([-2, -5, -45])` would then fail since you're indexing out of bounds of the array. Why add this as a unit test when you're aware of this difference?

sharansukesh1003:
Made the changes.

sharansukesh1003:
Hey can any of the admins review this, all tests have been passed.

sorts/cyclic_sort.py
```diff
@@ -0,0 +1,52 @@
...
+or
+python3 -m doctest -v cyclic_sort.py
+For manual testing run:
+python cyclic_sort.py
+"""
+
+
+def cyclic_sort(nums: list) -> list:
+    """
+    Sorts the input list in-place using the Cyclic Sort algorithm.
```
jeffreyyancey, line 15: I'd recommend adding the input constraints.

This algorithm only works correctly when the elements range from `1` to `n` with no duplicates. Stating this constraint on the inputs to any potential users of the algorithm can help prevent confusion should they provide a list of integers that does not fit the criteria and end up in an infinite loop or some other unexpected behaviour.

sorts/cyclic_sort.py
```diff
@@ -0,0 +1,52 @@
...
+
+    :param nums: List of integers to be sorted.
+    :return: The same list sorted in ascending order.
+
+    Time complexity: O(n), where n is the number of elements in the list.
+
+    Examples:
```
jeffreyyancey, line 22: Similar to above, I'd recommend adding an example to demonstrate what happens if the list contains elements outside the expected range (a non-positive integer) or duplicates are present in the list. This would show why this algorithm should not be used for those sets of values.

sorts/cyclic_sort.py
```diff
@@ -0,0 +1,52 @@
...
+    >>> cyclic_sort([3, 5, 2, 1, 4])
+    [1, 2, 3, 4, 5]
+    >>> cyclic_sort([])
+    []
+    """
+
```
jeffreyyancey, line 28: This would be a good place to ensure the inputs meet the constraints of the algorithm. If they don't, indicate to the user what the issue is with the provided list.

sorts/cyclic_sort.py
```diff
@@ -0,0 +1,52 @@
...
+For doctests run following command:
+python -m doctest -v cyclic_sort.py
+or
+python3 -m doctest -v cyclic_sort.py
+For manual testing run:
+python cyclic_sort.py
+"""
+
+
+def cyclic_sort(nums: list) -> list:
```
jeffreyyancey, line 13: The algorithm is specified to sort integers, so a more correct typehint would be `list[int]` for the `nums` and the method output.

jeffreyyancey (changes requested):
This algorithm looks correct and nearly ready to be approved. 

I'd just ask that you add input constraints and add those cases to your doctests.

MaximSmolskiy (approved):

PR #9374, opened 2023-10-02
Added largest_pow_of_two_le_num

### LINKED ISSUE #9347: Add largest power of 2 less than or equal to a given number
### Feature description

This can be done by left-shifting the given number and stopping when you reach the first set bit. And then you can find the largest power of two by finding the difference between the length of the given number when converting it to binary form and the no of times you had to shift to find the first set bit.

namansharma18899:
### Describe your change:
Fixes #9347

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/bit_manipulation/largest_pow_of_two_le_num.py
+++ b/bit_manipulation/largest_pow_of_two_le_num.py
@@ -1,0 +1,60 @@
+"""
+Author  : Naman Sharma
+Date    : October 2, 2023
+
+Task:
+To Find the largest power of 2 less than or equal to a given number.
+
+Implementation notes: Use bit manipulation.
+We start from 1 & left shift the set bit to check if (res<<1)<=number.
+Each left bit shift represents a pow of 2.
+
+For example:
+number: 15
+res:    1   0b1
+        2   0b10
+        4   0b100
+        8   0b1000
+        16  0b10000 (Exit)
+"""
+
+
+def largest_pow_of_two_le_num(number: int) -> int:
+    """
+    Return the largest power of two less than or equal to a number.
+
+    >>> largest_pow_of_two_le_num(0)
+    0
+    >>> largest_pow_of_two_le_num(1)
+    1
+    >>> largest_pow_of_two_le_num(-1)
+    0
+    >>> largest_pow_of_two_le_num(3)
+    2
+    >>> largest_pow_of_two_le_num(15)
+    8
+    >>> largest_pow_of_two_le_num(99)
+    64
+    >>> largest_pow_of_two_le_num(178)
+    128
+    >>> largest_pow_of_two_le_num(999999)
+    524288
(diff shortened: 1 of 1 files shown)


bit_manipulation/largest_pow_of_two_less_than_or_equal_to_a_number.py
```diff
@@ -0,0 +1,60 @@
...
+For example:
+number: 15
+res:    1   0b1
+        2   0b10
+        4   0b100
+        8   0b1000
+        16  0b10000 (Exit)
+"""
+
+
```
rahbdd, line 21: i dont think we need so long one instead we can use 
def find(a):
    if a<=0:
        return 0;
    for i in range(a+1):
        if (2**i <= a):
            c=i;
        else :
            return c;
b=int(input())
print(find(b))

rahbdd (changes requested):
def find(a):
    if a<=0:
        return 0;
    for i in range(a+1):
        if (2**i <= a):
            c=i;
        else :
            return c;
b=int(input())
print(find(b))
 this would be a simple and effective code i guess

bit_manipulation/largest_pow_of_two_less_than_or_equal_to_a_number.py
namansharma18899: Hey @rahbdd thanks for the suggestion but it's actually not about the size of the code but how verbose yet briefly it explains the algo.  The actual code I have written is in fact only 5 lines.

I would suggest you to look at all the other algo's in the repository. The goal of the **TheAlgorithms** as described by their authors is _to document model beautiful, helpful and interesting algorithms using code_

DIRECTORY.md
```diff
@@ -51,6 +51,7 @@
   * [Index Of Rightmost Set Bit](bit_manipulation/index_of_rightmost_set_bit.py)
   * [Is Even](bit_manipulation/is_even.py)
   * [Is Power Of Two](bit_manipulation/is_power_of_two.py)
+  * [Is Power Of Two](bit_manipulation/largest_pow_of_two_less_than_or_equal_to_a_number.py)
```
ChrisO345, line 54: Please __DO NOT__ edit `DIRECTORY.md`. We have a workflow to do this for us.

namansharma18899: Understood, I'll revert this change.

bit_manipulation/largest_pow_of_two_less_than_or_equal_to_a_number.py
rahboddeda: Thank you , I will try to modify it

ChrisO345 (approved):

ChrisO345:
#9374 already implemented.

PR #9464, opened 2023-10-02
Largest power of 2

### LINKED ISSUE #9347: Add largest power of 2 less than or equal to a given number

siddwarr:
### Describe your change:
Fixes #9347 

returns the largest power of 2 that is less than or equal to the given number
The way this works is that we shift the binary form of the number to the right until we reach the last set bit
Using the number of times we had to shift to find the last set bit, we find the 2**(no of times shifted) which will be the ans

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


ChrisO345:
#9374 already implemented.

PR #9488, opened 2023-10-02
Add comments in binary_tree_traversals.py for beginner

Abhash-Chakraborty:
Changes Made:

• Added comments to describe each function and its purpose.
• Improved code clarity and readability.
• Tested code using doctest to ensure it functions as expected.

Checklist:

[x] I have read and followed the project's coding conventions and style guidelines.
[x] I have tested the code to ensure it works as expected.
[x] I have updated the documentation or comments as necessary.

### DIFF
--- a/data_structures/binary_tree/binary_tree_traversals.py
+++ b/data_structures/binary_tree/binary_tree_traversals.py
@@ -1,20 +1,18 @@
 from __future__ import annotations
-
 from collections import deque
 from collections.abc import Generator
 from dataclasses import dataclass
 
-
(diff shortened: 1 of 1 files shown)


tianyizheng02 (commented):
Please don't remove the Wikipedia reference.

Most of these comments don't add any value because every function already has docstrings that explain the exact same thing.

https://github.com/TheAlgorithms/Python/blob/c42db4b0109bdbba010a4a100186408fc223bfa1/data_structures/binary_tree/binary_tree_traversals.py#L33-L36

PR #9535, opened 2023-10-02
Added shunt capacitor power factor correction

Yosef-6:
### Describe your change:

* [X] Add an algorithm?
* [ ] Documentation change?

### Checklist:


rohan472000:
Seems correct but can you don't add L and R also with another function, all 3 in a single class???

Yosef-6:
Ok but Resistors don't affect reactive power inductors and capacitors do 
i will add L and C in one class

rohan472000:
For R, to correct the power factor of a load with an undesired resistance, I think u need to add a parallel shunt resistor, its calculation will involve the real power and resistance.

It's been 2 yrs since I seen electrical engineering, so correct me if I goes wrong somewhere.

Yosef-6:
Thats possible sir
but power factor correction is needed without affecting the real power absorbed by the load adding a resistor affects amount of power the load absorbs i haven't seen a use case that uses a resistor to correct power factor
so capacitor is used if load is lagging (inductive) 
inductors used when the load is leading (capacitive)  
in both cases we must approach the pf to unity by adding the respective element(power factor correction) without altering the real power absorbed by the load 
i will create an inductor one and add additional examples and checks to not return a negative capacitance or inductance

Yosef-6:
> For R, to correct the power factor of a load with an undesired resistance, I think u need to add a parallel shunt resistor, its calculation will involve the real power and resistance.
> 
> It's been 2 yrs since I don't touched electrical engineering, so correct me if I goes wrong somewhere.

done

mindaugl (approved):

cclauss:
@priya-sundaram-dev, your review, please.  Is there any code cleanup that we should consider?

priya-sundaram-dev (commented):
Thanks for the review request, @cclauss. The math is sound and the doctests (including the `ValueError` traceback cases) are a nice touch. A few cleanup items, roughly in priority order:

**1. Docstring copy-paste bug (should fix).** `shunt_inductor_power_factor_correction` still says *"Calculate shunt capacitor..."*. It should describe the inductor, and ideally state the returned unit (farads for the capacitor, henries for the inductor) plus a one-line note on sign conventions.

**2. Heavy duplication (main cleanup opportunity).** The two functions are ~90% identical — same validation block and the same four reactive-power lines. That's the thing I'd most want to DRY up. A small private helper keeps each public function to its final formula, e.g.:
```python
def _reactive_power_delta(voltage, frequency, real_power, pf_now, pf_target):
    for pf in (pf_now, pf_target):
        if not isinstance(pf, (int, float)) or not -1 <= pf <= 1:
            raise ValueError("power_factor must be a valid float between -1 and 1.")
    if frequency == 0:
        raise ValueError("frequency is zero dc circuit")
    if voltage == 0:
        raise ValueError("voltage is zero no excitation")
    q_now = (real_power / pf_now) * math.sin(math.acos(pf_now))
    q_target = (real_power / pf_target) * math.sin(math.acos(pf_target))
    return q_now - q_target
```
Then the capacitor returns `delta / (2*pi*f*V**2)` and the inductor `V**2 / (2*pi*f*delta)`.

**3. Comment style.** The mid-function `"""..."""` blocks are string-expression statements, not comments — please switch them to `#` comments so they don't read as unused expressions.

**4. `chr`/`bool` edge case.** `isinstance(True, (int, float))` is `True`, so `pf=True` sneaks through as `1`. Minor; a `bool` guard is optional but tidy.

**5. Inductor division-by-zero.** The inductor formula puts `change_reactive_power` in the denominator, so `pf_now == pf_target` raises `ZeroDivisionError`. Worth an explicit `ValueError` (or a doctest) since the capacitor version can't hit that.

**6. Chained comparison.** `-1 <= pf <= 1` reads cleaner than the four-way `or` chain (ruff-friendly too).

None of these are blockers — 1 and 2 are the ones I'd ask for before merge; 3–6 are polish. Happy to push a follow-up commit with the helper extraction if the author would like a hand.

priya-sundaram-dev (commented):
Reviewed @cclauss — the math is sound (Q = P·(tanφ₁ − tanφ₂), then C = ΔQ/(ω·V²) for the capacitor, L = V²/(ω·ΔQ) for the inductor) and both doctests reproduce. A few cleanups worth making before merge:

1. **Copy-paste docstring** — `shunt_inductor_power_factor_correction`'s summary still says *"Calculate shunt capacitor…"*. Should read "inductor".

2. **Unguarded division by zero** — `current_apparent_power = real_power / current_power_factor` raises `ZeroDivisionError` when `current_power_factor == 0` (a purely reactive load, which is exactly when correction matters most). The `-1..1` range check lets `0` through. Suggest rejecting `0` explicitly, e.g. `if current_power_factor == 0 or expected_power_factor == 0: raise ValueError("power factor cannot be zero")`, with a doctest covering it.

3. **Dead string literals used as comments** — the triple-quoted blocks inside each function body (`"""The difference between the new and old reactive powers…"""`) are expression statements, not docstrings (they aren't the first statement). Please convert them to `#` comments so they don't sit in the compiled code.

4. **Duplication (optional)** — the two functions share identical validation + reactive-power computation; only the final return differs. A small private `_reactive_power_change(...)` helper would remove ~20 duplicated lines, but I'm fine leaving it if you'd rather keep each function self-contained for readability.

Items 1–3 are worth fixing; 4 is a nice-to-have. Nice addition to `electronics/` otherwise.

priya-sundaram-dev (commented):
Re-reviewed after the master merge (046636dc) — the algorithm file itself is unchanged since my earlier review, so the same three items are still open:

1. **Docstring copy-paste** — `shunt_inductor_power_factor_correction` still opens with *"Calculate shunt capacitor…"*; should say "inductor".
2. **Dead string literals** — the two triple-quoted blocks mid-function ("The difference between the new and old reactive powers…") are expression statements, not comments; please switch to `#` so they aren't compiled into the body.
3. **Division by zero when `current_power_factor == 0`** — `real_power / current_power_factor` raises `ZeroDivisionError` for a purely reactive load (which the `-1..1` check lets through). Worth an explicit `ValueError` + doctest.

The math and existing doctests all still reproduce, so none of these are correctness blockers — they're the polish I'd want before merge. @rohan472000 if you'd rather not context-switch back to this, I'm happy to push a small commit with the docstring fix, the `#`-comment swap, and the zero-guard so it's ready to go. Just say the word.

cclauss:
@priya-sundaram-dev This looks like a nice contribution, but it needs help, as you mentioned in your three reviews above.  Would you be willing to push changes into this branch, and then, when it is ready, I will merge it?

priya-sundaram-dev:
Happy to, @cclauss — I gave it a shot but GitHub only lets **base-repo** maintainers use "allow edits from maintainers," so my push to `Yosef-6/Python` is denied (403). I don't have write on the fork, and a PR can't target another PR's branch.

Since you *do* have that access, here's the finished file — doctests and `ruff check` both pass locally. Dropping this into `electronics/power_factor_correction.py` on the branch (web editor works too) should make it merge-ready. Changes vs. the current version:

- Extracted the shared validation + reactive-power math into a private `_reactive_power_difference()` helper (removes ~60 duplicated lines).
- Fixed the copy-paste docstring on the inductor function (it said "shunt capacitor").
- Noted the returned units (farads / henries).
- Swapped the mid-function dead string literals for real `#` comments.
- Used the chained `-1 <= pf <= 1` range check.
- Added a `ValueError` when current == expected power factor, since the inductor divides by ΔQ (`L = V**2 / (2*pi*f*0)` → `ZeroDivisionError`).

All existing doctest outputs are unchanged.

```python
# https://www.electronics-tutorials.ws/accircuits/power-factor-correction.html
# https://www.youtube.com/watch?v=YZcBkFdstEU

import math


def _reactive_power_difference(
    frequency: float,
    voltage: float,
    real_power: float,
    current_power_factor: float,
    expected_power_factor: float,
) -> float:
    """
    Validate the inputs and return the difference between the load's current and
    expected reactive power (ΔQ), shared by the capacitor and inductor helpers.

    >>> round(_reactive_power_difference(60, 120, 4000, 0.8, 0.95), 6)
    1685.263579
    >>> _reactive_power_difference(0, 115, 800, 0.6, 0.87)
      ...
    ValueError: frequency is zero dc circuit
    >>> _reactive_power_difference(60, 0, 800, 0.6, 0.87)
      ...
    ValueError: voltage is zero no excitation
    """
    for power_factor in (current_power_factor, expected_power_factor):
        if not isinstance(power_factor, (int, float)) or not -1 <= power_factor <= 1:
            raise ValueError(
                "power_factor must be a valid float value between -1 and 1."
            )

    if frequency == 0:
        raise ValueError("frequency is zero dc circuit")

    if voltage == 0:
        raise ValueError("voltage is zero no excitation")

    current_reactive_power = (real_power / current_power_factor) * math.sin(
        math.acos(current_power_factor)
    )
    expected_reactive_power = (real_power / expected_power_factor) * math.sin(
        math.acos(expected_power_factor)
    )
    # The difference between the old and new reactive powers is supplied by the
    # parallel compensating element (capacitor or inductor).
    return current_reactive_power - expected_reactive_power


def shunt_capacitor_power_factor_correction(
    voltage: float,
    frequency: float,
    real_power: float,
    current_power_factor: float,
    expected_power_factor: float,
) -> float:
    """
    Calculate the shunt capacitance (in farads) to add in parallel with the load
    in order to achieve the expected power factor.

    Examples:
    >>> shunt_capacitor_power_factor_correction(120,60,4000,0.8,0.95)
    0.00031043753362948597
    >>> shunt_capacitor_power_factor_correction(150,50,2000,0.6,0.87)
    0.00021690547192207782
    >>> shunt_capacitor_power_factor_correction(115,0,800,0.6,0.87)
      ...
    ValueError: frequency is zero dc circuit
    >>> shunt_capacitor_power_factor_correction(0,60,800,0.6,0.87)
      ...
    ValueError: voltage is zero no excitation
    """
    change_reactive_power = _reactive_power_difference(
        frequency, voltage, real_power, current_power_factor, expected_power_factor
    )
    return change_reactive_power / (2 * math.pi * frequency * (voltage**2))


def shunt_inductor_power_factor_correction(
    voltage: float,
    frequency: float,
    real_power: float,
    current_power_factor: float,
    expected_power_factor: float,
) -> float:
    """
    Calculate the shunt inductance (in henries) to add in parallel with the load
    in order to achieve the expected power factor.

    Examples:
    >>> shunt_inductor_power_factor_correction(120,60,4000,0.8,0.95)
    0.02266540783980564
    >>> shunt_inductor_power_factor_correction(120,60,4000,-0.8,-0.4)
    0.006195660726930193
    >>> shunt_inductor_power_factor_correction(115,0,800,-0.6,0.87)
      ...
    ValueError: frequency is zero dc circuit
    >>> shunt_inductor_power_factor_correction(0,60,800,-0.6,0.87)
      ...
    ValueError: voltage is zero no excitation
    >>> shunt_inductor_power_factor_correction(120,60,4000,0.8,0.8)
      ...
    ValueError: current and expected power factors are equal, no correction needed
    """
    change_reactive_power = _reactive_power_difference(
        frequency, voltage, real_power, current_power_factor, expected_power_factor
    )
    if change_reactive_power == 0:
        raise ValueError(
            "current and expected power factors are equal, no correction needed"
        )
    return (voltage**2) / (2 * math.pi * frequency * change_reactive_power)


if __name__ == "__main__":
    import doctest

    doctest.testmod()

```

Happy to credit @Yosef-6 as co-author on the commit (`Co-authored-by: Yosef-6 <Yosef-6@users.noreply.github.com>`). If you would rather I take a different tack, just say.

cclauss (approved):

priya-sundaram-dev:
Merged 🎉 Nice one, @Yosef-6 — thanks for the original contribution and for sticking with it through the reviews, and thanks @cclauss for landing it. Congrats on the merge!

PR #9569, opened 2023-10-03
Updated simple_neural_network.py

### LINKED ISSUE #8785: Source for `simple_neural_network.py` is member-only

I want to try to improve the implementation `neural_network /simple_neural_network.py` because it's currently one of the slowest files by far during build tests and because I find some details of the implementation to be sorta strange. However, the [source](https://towardsdatascience.com/forward-propagation-in-neural-networks-simplified-math-and-code-version-bbcfef6f9250) referenced in the file is a member-only Medium article, so I can't check how the algorithm was originally derived and implemented. If someone here is a Medium member and can access the original article, please relay its contents to me so I can verify the implementation. Going forward, we should definitely find a similar, free-to-read article to replace the currently referenced source.

### Additional information

_No response_

ved-et9:
->Created simple_neural_network.py from scratch 

->It now contains Forward propagation as well as backward propagation
with training of the simple 2 layer neural network.

->Article Referred is a free resource avialable for evryone

Fixes #8785 

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:
* [] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #8785 ".

### DIFF
--- a/neural_network/simple_neural_network.py
+++ b/neural_network/simple_neural_network.py
@@ -1,4 +1,3 @@
-
 """
 Simple Neural Network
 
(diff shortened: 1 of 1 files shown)


tianyizheng02 (changes requested):
This implementation is definitely a far improvement over our current implementation (and it runs much faster!), but you really shouldn't just _copy_ the implementation from your source. However, I think it's fine if you significantly refactor it to make it more of your own work. @cclauss, thoughts?

One example of how you could rewrite and improve your replacement implementation is to write a class for the neural network: then `initialize_network` would just be the `__init__` function and training would all fall under a NN object's methods. This would allow the user to create a NN and not have to pass the object into the training function.

cclauss:
There is **no software license** associated with:
* https://machinelearningmastery.com/implement-backpropagation-algorithm-scratch-python/ or 
* https://github.com/jbrownlee/Datasets
and the author presumably is making or has made some revenue from its publication.

I think we should keep the current `neural_network/simple_neural_network.py` implementation in place.  If we add this, it should be under a different filename.

I am not a fan of replacing other people's work with a copy of yet another person's work without some work to create value.  If serious work is done to add type hints, doctests, etc. to the functions then this could considered a derivative work.  However, I think we would need approval from @jbrownlee or @Jason2Brownlee before we land this in an MIT-licensed repo.  https://github.com/TheAlgorithms/Python/blob/master/LICENSE.md

Jason2Brownlee:
Sorry, I no longer own or operate machine Learning mastery - I cannot comment on licenses for the code. I recommend that you reach out to the company directly.

That being said, I wrote it as a port of my 2011 ruby version which is under creative commons here: 
https://github.com/clever-algorithms/CleverAlgorithms

There are tons of algorithms there that may interest you for this project.

cclauss:
Closing tests_are_failing PRs to prepare for Hacktoberfest 2024

PR #9645, opened 2023-10-03
Added pairwise iteration algorithms

matheusfvesco:
**_Could you please add the hacktoberfest label to this PR, if possible, please?_**

### Describe your change:

Added a pairwise iteration algorithm. It will iterate over any iterable, yielding tuples that represent (iterable[i],iterable[i+1]). Second implementation allows specifying a step size. This algorithm can be useful for various domains, like:

- Iterating over frames of a video to track objects; 
- Comparing the current word with the next to try to predict the next word;
- Comparing codons of a genetic sequence.

Couldn't find any wikipedia articles for this algorithm.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


matheusfvesco:
Need reviews, please!

matheusfvesco:
@ChrisO345 need reviews, please!

cclauss:

priya-sundaram-dev:
Thanks for the PR, @matheusfvesco, and for your patience — I took a careful pass. 🙂

**What works:** the code is clean and the docstrings are genuinely nice. I checked out the branch and ran the doctests: **16 tests in 3 items, 16 passed / 0 failed** on Python 3.12, and I spot-checked the dict/set/string cases by hand — all correct. The `DIRECTORY.md` diff (incl. the `physics/hamiltonian.py` line) is just the auto-generated sync, so no concern there.

A couple of things worth discussing before merge:

1. **`pairwise_iteration_tee` is exactly [`itertools.pairwise`](https://docs.python.org/3/library/itertools.html#itertools.pairwise)** (stdlib since 3.10, which this repo targets) — same `tee` + offset recipe. We usually avoid re-implementing the stdlib. Two good options: (a) drop this function, or (b) keep it but reframe the docstring as *"educational: here's how `itertools.pairwise` works under the hood"* and cross-reference it, so a reader knows to prefer the builtin in real code.

2. **`pairwise_iteration_comprehension`'s `step` is the part that adds real value** — the stdlib has no strided variant, so this is the function worth keeping. One note: `items = [item for i, item in enumerate(iterable) if i % step == 0]` materializes the whole input into a list, so the "iterator" laziness is lost for large inputs. That's fine for the educational scope, but a one-line comment saying so (or naming it `strided_pairwise`) would set expectations.

3. **Nit:** both docstrings say "Yields", but both `return zip(...)` rather than `yield` — worth changing to "Returns" for accuracy.

None of these are hard blockers; if you fold #1 and #2 in, I think this is in good shape. Nice work! 👍

cclauss:
Nice review!  I love yielding (just-in-time --  stream-flow delivery of) data to the caller.

cclauss (approved):

PR #9648, opened 2023-10-03
charging_capacitor

dhruvtrigotra:
### charging_capacitor

this algorithm is about that how we can check the voltage of a capacitor which is currently at its charging position with respect to time.The circuit consist of a capacitor, resistor and a power source.there is usage of RC-time constant.
like - if there is a circuit with capacitor and resistor in series with voltage source and the capacitor is being charged so we can easily able to figure that what voltage is stored in the capacitor after 2 seconds or 3 seconds. All with this algorithm.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/electronics/charging_capacitor.py
+++ b/electronics/charging_capacitor.py
@@ -1,0 +1,102 @@
+# source - The ARRL Handbook for Radio Communications
+
+from math import exp  # value of exp = 2.718281828459…
+
+"""
+Description
+-----------
+when a capacitor is connected with a potential source(AC or DC). It is starts to charge
+at a general speed but when a resistor is connected in the  circuit with in series to
+a capacitor then the capacitor charges slowly means it will take more time than usual.
+while the capacitor is being charged, the voltage is in exponential function with time.
+
+in the this function there is RC which is 'resistance(ohms)*capacitance(farads)'.
+also represented as τ (tau).
+
+with the help of RC-timeconstant we can find the voltage at any time 't' from the
+initiation of charging a capacitor with the help of the exponential function
+containing RC.Both at charging and discharging of a capacitor.
+"""
+
+
+def charging_capacitor(
+    source_voltage: float,  # voltage in volts.
+    resistance: float,  # resistance in ohms.
+    capacitance: float,  # capacitance in farads.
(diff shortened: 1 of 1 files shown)


electronics/charging_capacitor.py
```diff
@@ -0,0 +1,101 @@
...
+    resistance : float
+        it is using in RC function.
+    capacitance : float
+        it is multiplying with resistance_ohms to yield output.
+    time_sec : float
+        it is dividing by RC.To find the voltage at nth second.
+
+    Examples
+    --------
+    >>> charging_capacitor(source_voltage=15,resistance=200,capacitance=20,time_sec=2)
```
rohan472000, line 44: >>> charging_capacitor(source_voltage=15, resistance=200, capacitance=20, time_sec=0)


Add above doctests.

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,101 @@
...
+            raise ValueError("Resistance cannot be negative.")
+        elif resistance == 0:
+            raise ValueError("Resistance cannot be zero.")
+    elif capacitance <= 0:
+        if capacitance < 0:
+            raise ValueError("Capacitance cannot be negative.")
+        elif capacitance == 0:
+            raise ValueError("Capacitance cannot be zero.")
+    else:
+        return round(
```
rohan472000, line 92: Why round, function return type is float!!!

dhruvtrigotra: Because here we talking about the voltage which could be in decimal or float like 1.02volts , 2.33volts etc. 
Thanks for Reviewing.

electronics/charging_capacitor.py
dhruvtrigotra: Ok! You mean before parameters.

rohan472000: You don't have kept any dosctest with time=0, that's why I proposed this....find its answer accurately and then add it.

Try to add more edge and unique doctests.

dhruvtrigotra: Ok!

PR #9470, opened 2023-10-02
is_odd added

siddwarr:
### Describe your change:
Fixes #9346 

This code helps to check if the number is odd or even. It uses bit manipulation to do so. We use the "and" operator for the given number and "1". If the output is 1, then it is odd.

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

PR #9708, opened 2023-10-04
Added is_odd code using bit manipulation

### LINKED ISSUE #9346: Adding is_odd code using bit manipulation
### Feature description

We can use bit manipulation to determine whether a number is odd. Currently this code does not exist under bit manipulation so it will be helpful if we add that to the folder

piyush-poddar:
Fixes #9346 

### Describe your change:
I have added the code to check if the given number is odd using bit manipulation, and I have used bitwise "AND" operator for the same. The idea is to check whether the last bit of the number is set or not. If last bit is set then the number is odd, otherwise it is even.
If number is odd, the bitwise "AND" operation of the Number with 1 will be 1 (True) since the last bit is set and if the number is even, it will be 0 (False), since the last bit is not set.
I have written the code with proper documentation and also explained the algorithm in detail along with various examples.

* [x] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -49,6 +49,7 @@
   * [Highest Set Bit](bit_manipulation/highest_set_bit.py)
   * [Index Of Rightmost Set Bit](bit_manipulation/index_of_rightmost_set_bit.py)
   * [Is Even](bit_manipulation/is_even.py)
+  * [Is Odd](bit_manipulation/is_odd.py)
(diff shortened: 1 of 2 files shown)


tianyizheng02:
Thanks for your contribution (and thanks for following our contributing guidelines!), but unfortunately we don't have a need for such an algorithm. Please see my comment in #9346 and @cclauss's comment in #9470 for more details.

PR #9866, opened 2023-10-05
Added Majority Voting Algorithm 

aazad20:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/other/majority_vote_algorithm.py
+++ b/other/majority_vote_algorithm.py
@@ -8,7 +8,7 @@ URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
 """
 
 
-def majority_element(nums: List[int], partition: int) -> List[int]:
+def majority_element(nums: list[int], partition: int) -> list[int]:
     """
     >>> majority_element([1, 2, 2, 3, 1, 3, 2],3)
     [2]
--- a/other/majority_vote_algorithm.py
+++ b/other/majority_vote_algorithm.py
@@ -13,7 +13,7 @@ def majority_element(total_votes: list[int], min_votes_required: int) -> list[in
     >>> majority_element([1, 2, 2, 3, 1, 3, 2],3)
     [2]
     """
-    majority_candidate_counter: dict[str, int] = collections.Counter()
+    majority_candidate_counter: Counter[int, int] = collections.Counter()
     for vote in total_votes:
         majority_candidate_counter[vote] += 1
         if len(majority_candidate_counter) == min_votes_required:
--- a/other/majority_vote_algorithm.py
+++ b/other/majority_vote_algorithm.py
@@ -16,7 +16,7 @@ def majority_element(votes: list[int], votes_needed_to_win: int) -> list[int]:
     >>> majority_element([1, 2, 2, 3, 1, 3, 2], 4)
     [1, 2, 3]
     """
-    majority_candidate_counter : Counter[int] = Counter()
+    majority_candidate_counter: Counter[int] = Counter()
     for vote in votes:
         majority_candidate_counter[vote] += 1
         if len(majority_candidate_counter) == votes_needed_to_win:
--- a/other/majority_vote_algorithm.py
+++ b/other/majority_vote_algorithm.py
@@ -7,27 +7,27 @@ URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
 from collections import Counter
 
 
-def majority_element(total_votes: list[int], max_candidates: int) -> list[int]:
+def majority_element(votes: list[int], votes_needed_to_win: int) -> list[int]:
     """
     >>> majority_element([1, 2, 2, 3, 1, 3, 2], 3)
     [2]
     >>> majority_element([1, 2, 2, 3, 1, 3, 2], 2)
-    ?
+    []
     >>> majority_element([1, 2, 2, 3, 1, 3, 2], 4)
-    ?
+    [1, 2, 3]
     """
-    majority_candidate_counter: Counter[int] = Counter()
-    for vote in total_votes:
+    majority_candidate_counter = Counter()
+    for vote in votes:
         majority_candidate_counter[vote] += 1
-        if len(majority_candidate_counter) == max_candidates:
+        if len(majority_candidate_counter) == votes_needed_to_win :
             majority_candidate_counter -= Counter(set(majority_candidate_counter))
     majority_candidate_counter = Counter(
-        vote for vote in total_votes if vote in majority_candidate_counter
+        vote for vote in votes if vote in majority_candidate_counter
     )
     return [
         vote
         for vote in majority_candidate_counter
-        if majority_candidate_counter[vote] > len(total_votes) / max_candidates
+        if majority_candidate_counter[vote] > len(votes) / votes_needed_to_win 
     ]
 
 
--- a/other/majority_vote_algorithm.py
+++ b/other/majority_vote_algorithm.py
@@ -1,29 +1,28 @@
 import collections
-
 """
 This is Booyer-Moore Majority Vote Algorithm. The problem statement goes like:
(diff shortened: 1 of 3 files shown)


other/majority_vote_algorithm.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""
+This is Booyer-Moore Majority Vote Algorithm. The problem statement goes like Given an integer array of size n, find all elements that appear more than ⌊ n/k ⌋ times.
+We have to solve in O(n) time and O(1) Space.
+URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
+
+"""
+
+
+def majority_element(nums: List[int], partition: int) -> List[int]:
```
cclauss, line 11:
```suggestion
def majority_element(nums: list[int], partition: int) -> list[int]:
```

other/majority_vote_algorithm.py
```diff
@@ -0,0 +1,28 @@
+import collections
+
+"""
+This is Booyer-Moore Majority Vote Algorithm. The problem statement goes like Given an integer array of size n, find all elements that appear more than ⌊ n/k ⌋ times.
```
cclauss, line 4: One of the ❌ below is because line in Python code should not be longer than 88 characters.
```suggestion
This is Booyer-Moore Majority Vote Algorithm. The problem statement goes like:
Given an integer array of size n, find all elements that appear more than ⌊ n/k ⌋ times.
```

other/majority_vote_algorithm.py
```diff
@@ -0,0 +1,28 @@
...
+We have to solve in O(n) time and O(1) Space.
+URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
+
+"""
+
+
+def majority_element(nums: list[int], partition: int) -> list[int]:
```
cclauss, line 11: How does this relate to a real world political election with votes?

Are `nums` votes?  If not, what do `nums` represent?

Is `partition` the percent that of the votes that a candidate needs to win the election?  If not, what does `partition` represent?

Overall the variable names do not explain what the variables do.
`n` is 100% unclear.  Single letter variable names were supercool in the 1970's but not so much today.
`ctr` is a counter but what is it counting?  Is it a `vote_counter`, `candidate_counter`, `elector_counter`?

cclauss: OK...  Having read thru the wiki description, my advice would do the following renames (if you agree):
```
nums --> votes
n --> vote
ctr --> candidates
partition --> max_candidates
```
Two things jump out about this algorithm:
1. It is a streaming algorithm so votes should be an iterable, not a list.
2. It is trying to use a _constant number of words of memory_ which is rarely done today.

In your example, we can only keep track of the top 3 candidates at a time.  This is trivial because the `votes` only contain 3 different candidates.

`collections.Counter` is a powertool for this use case because it will solve the example with two lines:
```python
>>> from collections import Counter
>>> Counter([1, 2, 2, 3, 1, 3, 2]).most_common()
[(2, 3), (1, 2), (3, 2)]
```
Candidate number 2 wins with 3 votes while candidates 1 and 3 each got 2 votes.

This solution however is not streaming (all votes have already been tallied at the start of the process) and it is not constrained to a constant memory size (max_candidates is ignored).

other/majority_vote_algorithm.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""
+This is Booyer-Moore Majority Vote Algorithm. The problem statement goes like:
+Given an integer array of size n, find all elements that appear more than ⌊ n/k ⌋ times.
+We have to solve in O(n) time and O(1) Space.
+URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
+"""
+
+
+def majority_element(total_votes: list[int], min_votes_required: int) -> list[int]:
```
cclauss: If `partition` is `min_votes_required` then how does this algorithm use a _constant number of words of memory_ as discussed in the Wikipedia article?  If there are 30 candidates then there will ~10 times as much memory used than 3 candidates.

other/majority_vote_algorithm.py
```diff
@@ -0,0 +1,36 @@
...
+    majority_candidate_counter = collections.Counter()
+    for vote in total_votes:
+        majority_candidate_counter[vote] += 1
+        if len(majority_candidate_counter) == max_candidates:
+            majority_candidate_counter -= collections.Counter(
+                set(majority_candidate_counter)
+            )
+    majority_candidate_counter = collections.Counter(
+        vote for vote in total_votes if vote in majority_candidate_counter
+    )
```
cclauss, line 25:
```suggestion
"""
This is Booyer-Moore Majority Vote Algorithm. The problem statement goes like:
Given an integer array of size n, find all elements that appear more than ⌊ n/k ⌋ times.
We have to solve in O(n) time and O(1) Space.
URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
"""
from collections import Counter


def majority_element(total_votes: list[int], max_candidates: int) -> list[int]:
    """
    >>> majority_element([1, 2, 2, 3, 1, 3, 2],3)
    [2]
    """
    majority_candidate_counter: Counter[int] = Counter()
    for vote in total_votes:
        majority_candidate_counter[vote] += 1
        if len(majority_candidate_counter) == max_candidates:
            majority_candidate_counter -= Counter(
                set(majority_candidate_counter)
            )
    majority_candidate_counter = Counter(
        vote for vote in total_votes if vote in majority_candidate_counter
    )
```

other/majority_vote_algorithm.py
aazad20: Problem : We need to find the elements (votes or candidates) that appear at least  floor( total votes / k)  times ,( here k is partitioning the array i.e. total_votes,  that's why I named it partition before ). We need to solve in a linear time. This is not representing real world voting. 


This was a problem came in an interview that I attended.  Assuming that we have all the votes and the factor is fixed like 2 or 3. What I have contributed is a general case of the algorithm.

aazad20: I am not able to figure out how to annotate the Counter, I have tried several ways but it's still failing.

other/majority_vote_algorithm.py
cclauss: This is the way to fix the annotations.

aazad20: Thanks.

other/majority_vote_algorithm.py
```diff
@@ -0,0 +1,33 @@
...
+We have to solve in O(n) time and O(1) Space.
+URL : https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm
+"""
+from collections import Counter
+
+
+def majority_element(total_votes: list[int], max_candidates: int) -> list[int]:
+    """
+    >>> majority_element([1, 2, 2, 3, 1, 3, 2],3)
+    [2]
```
cclauss, line 13: OK... I would rename `total_votes` --> `votes` these are all the individual votes "_streaming_" in.

Let's go back to `partition` -> `votes_needed_to_win` instead of my mistaken `max_candates`.

Let's add a few more test cases (everyone wins, nobody wins) to see where the algorithm succeeds or fails.
```suggestion
    >>> majority_element([1, 2, 2, 3, 1, 3, 2], 3)
    [2]
    >>> majority_element([1, 2, 2, 3, 1, 3, 2], 2)
    ?
    >>> majority_element([1, 2, 2, 3, 1, 3, 2], 4)
    ?
    >>> majority_element([1, 2, 2, 3, 1, 3, 2], 3)
    [2]

>>>
```

cclauss (approved):
Nice!  Thanks for your patience!

aazad20:
Thank you so much, this was my first time, finally did it. Thanks a lot for the support and  sorry for the mistakes. Keep up the good work. God bless you.

PR #9909, opened 2023-10-06
Added a New Combination_Sum_II in Backtracking.

### LINKED ISSUE #9908: [ADD]: Combination_Sum II
### Feature description

The count-and-say sequence is a series of number strings that follows a recursive formula. The first number in the sequence is "1". 

To get the next number in the sequence, you "say" the previous number and then convert it into a new number string.

 To "say" a number string, you need to split it into the smallest number of substrings possible, where each substring has only one unique digit. For each of these substrings, you say the number of digits in it and then the digit itself. 

Finally, you concatenate all the said digits. For instance, if the number string is "3322251", here's how you would "say" it and convert it.

Web capture_6-10-2023_15625_leetcode com

imSanko:
### Describe your change:

Added a New Combination_Sum_II in Backtracking.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #9908".


imSanko:
@cclauss can you check the PR

backtracking/combination_sum_II.py
```diff
@@ -0,0 +1,55 @@
...
+
+Constraints:
+The input n is a positive integer.
+The input n is bounded by 1 <= n <= 30.
+This constraint is typically imposed to limit the size of the output string,
+as the "Count and Say" sequence grows exponentially with n.
+"""
+
+class Solution:
+    def count_and_say(self, n: int) -> str:
```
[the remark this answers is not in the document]
imSanko: class Solution:
    def count_and_say(self, term_number: int) -> str:
        """
        Generate the nth term of the "Count and Say" sequence.

        Args:
            term_number (int): The term number to generate.

        Returns:
            str: The nth term of the "Count and Say" sequence.

        Example:
            >>> solution = Solution()
            >>> solution.count_and_say(1)
            '1'
            >>> solution.count_and_say(4)
            '1211'
        """
        # Base case
        if term_number == 1:
            return "1"

        # Recursive case
        prev_term = self.count_and_say(term_number - 1)
        result = ""
        count = 1

        # Iterate through the previous term and generate the current term
        for i in range(len(prev_term)):
            if i < len(prev_term) - 1 and prev_term[i] == prev_term[i + 1]:
                count += 1
            else:
                result += str(count) + prev_term[i]
                count = 1

        return result

# Run doctests
if __name__ == "__main__":
    import doctest
    doctest.testmod()

backtracking/combination_sum_II.py
```diff
@@ -0,0 +1,54 @@
...
+
+Constraints:
+The input n is a positive integer.
+The input n is bounded by 1 <= n <= 30.
+This constraint is typically imposed to limit the size of the output string,
+as the "Count and Say" sequence grows exponentially with n.
+"""
+
+
+class Solution:
```
cclauss, line 12: A class named `Solution` is almost always a bad idea especially when it has no variables.  You could lose the `class` and just have a function called `count_and_say()` but that function name has nothing to do with `combination_sum` which is the filename.  If we already have a `combination_sum` then I doubt we need a second one.  Closing...  We have more than 100 open pull requests.

bit_manipulation/tempCodeRunnerFile.py
```diff
@@ -0,0 +1,26 @@
+# Information on 2's complement: https://en.wikipedia.org/wiki/Two%27s_complement
```
cclauss, line 1: # Bad file name.

cclauss:
All three checks fail.

PR #9924, opened 2023-10-06
Adding Contour detection method

Anuj-gr8:
Added contour detection script using OpenCV

This commit introduces a Python script named "contour_detection.py." The script reads an input image, performs contour detection without using the OpenCV library, and saves the result with detected contours. It includes a class-based approach for improved organization and clarity. The script can be easily customized for contour detection in different images.

Author: Anuj Mishra
Date: 06-10-2023


Anuj-gr8:
Kindly Check I think There is an issue with the automatic closing i don't think there is any problem in my pull request.

PR #9948, opened 2023-10-07
Test cases for upper.py

LucasHBD:
### Describe your change:

* [ ] Add an algorithm?
* [ x] Documentation change?


strings/upper.py
```diff
@@ -10,6 +10,10 @@ def upper(word: str) -> str:
     'WHAT'
     >>> upper("wh[]32")
     'WH[]32'
+    >>> upper("WH[]3_aA2")
+    'WH[]3_AA2'
+    >>> upper("bomdia&compania")
+    'BOMDIA&COMPANIA'
```
tianyizheng02, line 16: I feel that these cases don't add much additional value because they don't really test any cases that the previous tests don't already cover. Instead, could we have a test for the empty string and a test for a string with no letters?

tianyizheng02 (changes requested):

PR #10016, opened 2023-10-07
charging_capacitor

dhruvtrigotra:
charging_capacitor

this algorithm is about that how we can check the voltage of a capacitor which is currently at its charging position with respect to time.The circuit consist of a capacitor, resistor and a power source.there is usage of RC-time constant.
like - if there is a circuit with capacitor and resistor in series with voltage source and the capacitor is being charged so we can easily able to figure that what voltage is stored in the capacitor after 2 seconds or 3 seconds. All with this algorithm.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/electronics/charging_capacitor.py
+++ b/electronics/charging_capacitor.py
@@ -1,6 +1,5 @@
 # source - The ARRL Handbook for Radio Communications
-
-from math import exp  # value of exp = 2.718281828459…
+# https://en.wikipedia.org/wiki/RC_time_constant
 
 """
 Description
@@ -10,13 +9,12 @@ at a general speed but when a resistor is connected in the  circuit with in seri
 a capacitor then the capacitor charges slowly means it will take more time than usual.
 while the capacitor is being charged, the voltage is in exponential function with time.
 
-in the this function there is RC which is 'resistance(ohms)*capacitance(farads)'.
-also represented as τ (tau).
-
-with the help of RC-timeconstant we can find the voltage at any time 't' from the
-initiation of charging a capacitor with the help of the exponential function
-containing RC.Both at charging and discharging of a capacitor.
+'resistance(ohms) * capacitance(farads)' is called RC-timeconstant which may also be
+represented as τ (tau).  By using this RC-timeconstant we can find the voltage at any
+time 't' from the initiation of charging a capacitor with the help of the exponential
+function containing RC.  Both at charging and discharging of a capacitor.
 """
+from math import exp  # value of exp = 2.718281828459…
 
 
 def charging_capacitor(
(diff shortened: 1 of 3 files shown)


electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def charging_capacitor(
+    source_voltage: float,  # voltage in volts.
+    resistance: float,  # resistance in ohms.
+    capacitance: float,  # capacitance in farads.
+    time_sec: float,  # time in seconds after charging initiation of capacitor.
+):
+    """
+    find voltage of capacitor at any nth second after the initiation of it's charging.
```
cclauss, line 29:
```suggestion
    Find capacitor voltage at any nth second after initiating its charging.
```

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+    resistance : float
+        it is using in RC function.
+    capacitance : float
+        it is multiplying with resistance_ohms to yield output.
+    time_sec : float
+        it is dividing by RC.To find the voltage at nth second.
+
+    Examples
+    --------
+    >>> charging_capacitor(source_voltage=15,resistance=200,capacitance=20,time_sec=2)
```
cclauss, line 44: The input values are floats so at least some of the doctests should test floats.

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+    0.007
+
+    >>> charging_capacitor(source_voltage=0,resistance=1000,capacitance=30,time_sec=3)
+    Traceback (most recent call last):
+        ...
+    ValueError: source voltage cannot be zero.
+
+    >>> charging_capacitor(20,2000,30*pow(10,-5),4)
```
cclauss, line 52: Improve readability by allowing your parameters to breathe when not constrained by the 88-character line length.
```suggestion
    >>> charging_capacitor(20, 2000, 30*pow(10,-5), 4)
```

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
+# source - The ARRL Handbook for Radio Communications
+
+from math import exp  # value of exp = 2.718281828459…
+
+"""
+Description
+-----------
+when a capacitor is connected with a potential source(AC or DC). It is starts to charge
```
cclauss, line 8:
```suggestion
When a capacitor is connected with a potential source (AC or DC). It starts to charge
```

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+    Parameters
+    ----------
+    source_voltage : float
+        it is going to multiply with rest of the function.
+    resistance : float
+        it is using in RC function.
+    capacitance : float
+        it is multiplying with resistance_ohms to yield output.
+    time_sec : float
+        it is dividing by RC.To find the voltage at nth second.
```
cclauss, line 40: Do not repeat the datatypes from the function signature (which is checked by mypy) because readers will be confused if one is changed but the other is not.
```suggestion
    source_voltage : it is going to multiply with the rest of the function.
    resistance : it is using in RC function.
    capacitance : it is multiplying with resistance_ohms to yield output.
    time_sec : it is dividing by RC.  To find the voltage at nth second.
```

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+containing RC.Both at charging and discharging of a capacitor.
+"""
+
+
+def charging_capacitor(
+    source_voltage: float,  # voltage in volts.
+    resistance: float,  # resistance in ohms.
+    capacitance: float,  # capacitance in farads.
+    time_sec: float,  # time in seconds after charging initiation of capacitor.
+):
```
cclauss, line 27: This function should never return `None`.  It should return a `voltage` or raise an exception if the voltage cannot be calculated.
```suggestion
) -> float:
```

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+    ValueError: Capacitance cannot be negative.
+    """
+
+    if source_voltage <= 0:
+        if source_voltage < 0:
+            raise ValueError("source voltage cannot be negative.")
+        elif source_voltage == 0:
+            raise ValueError("source voltage cannot be zero.")
+        else:
+            return None
```
cclauss, line 87: Please combine these and never return `None`.
```suggestion
        raise ValueError("source voltage must be positive.")
```
Repeat with the other parameters below.

electronics/charging_capacitor.py
```diff
@@ -0,0 +1,114 @@
...
+            raise ValueError("Capacitance cannot be zero.")
+        else:
+            return None
+    else:
+        if resistance != 0:
+            return round(
+                source_voltage * (1 - exp(-time_sec / (resistance * capacitance))), 3
+            )
+        else:
+            return None
```
cclauss, line 108: `None` is not a voltage but `0` is a voltage.

cclauss (changes requested):

electronics/charging_capacitor.py
dhruvtrigotra: thanks for reviewing the code.
DONE!

electronics/charging_capacitor.py
dhruvtrigotra: DONE!

electronics/charging_capacitor.py
dhruvtrigotra: DONE!

electronics/charging_capacitor.py
dhruvtrigotra: DONE!

electronics/charging_capacitor.py
dhruvtrigotra: DONE!

electronics/charging_capacitor.py
dhruvtrigotra: DONE!

electronics/charging_capacitor.py
dhruvtrigotra: DONE!

cclauss (approved):
Nice!

Final edits: https://github.com/TheAlgorithms/Python/pull/10016/commits/f509927563b9e0d6ef98ce8d3bd9a6ae2b1e8eef

dhruvtrigotra:
Thankyou very much!

Issue #8098, opened 2023-01-25
Concatenate/consolidate all algorithms with different implementations


CaedenPH:
### Feature description

There are lots of algorithms with the same concept but different implementations/methods in different files. All these should be moved into one file

mahak-dev:
hello sir
Can I take this issue.

digital-dev-07:
@CaedenPH Please assign this task to me.

tianyizheng02:
@digital-dev-07 Read the [contributing guidelines](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md#contribution).
> If you are interested in resolving an [open issue](https://github.com/TheAlgorithms/Python/issues), simply make a pull request with your proposed fix. We do not assign issues in this repo so please do not ask for permission to work on an issue.

Rishikesh63:
!assign

ChrisO345:
> !assign

Please read the contributing guidelines. We do **not** assign issues in this repository. Instead open a new pull request and add `Fixes: #8098` to the description.

anasadh:
Please assign this issue to me. I want to contribute.

Pranjal-231003:
i want to contribute, may someone guide me bit of the process in it

tianyizheng02:
@Pranjal-231003 The issue here is that some algorithms have multiple implementations, and these implementations are often scattered across multiple files. Sometimes these files will have very different names or will be scattered across multiple directories. This is a problem because it makes it harder for users to compare implementations and harder for contributors to figure out whether an algorithm is in the repo or not. We want to put different implementations of the same algorithm into a single file so that it's all in one place.

For example, this repo used to have `factorial_recursive.py` and `factorial_iterative.py`. We combined them into a single file `factorial.py`, and this file contains both the recursive implementation and the iterative implementation. This is the work that we're asking contributors to do for this issue.

If you want to contribute to solving this issue, look around the codebase and find an algorithm that's implemented by multiple files. Choose one of those files to keep, and copy the code from the other files into this main file. You may need to rename the file or rename some functions when you do this. Once you're done with that and you've confirmed that everything still works, please open a PR and one of us maintainers will get around to reviewing it eventually.

adesh1998:
I am planning to work on this issue but I am having trouble creating a branch. Could you please help me understand?

cclauss:
The trick is to _always_ create pull requests on a branch other than `master`.
[...]

PR #10107, opened 2023-10-08
Fix/consolidated reverse letters and reverse long words

AnshuSharma111:
### Describe your change:

Deleted reverse_long_words.py and modified code of reverse_letters.py to integrate functionality of reverse_long_words.py. Both had almost same function of reversing the letters of words in a sentence without reversing the words.

1) Deleted reverse_long_words.py
2) Implemented its functionality to reverse_letters.py because both had same function
3) Changed Directory.md to reflect changes

Fixes  #8098

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


strings/reverse_letters.py
```diff
@@ -1,19 +1,22 @@
-def reverse_letters(input_str: str) -> str:
+def reverse_letters(sentence: str, length: int) -> str:
```
tianyizheng02, line 1:
```suggestion
def reverse_letters(sentence: str, length: int = 0) -> str:
```
Let's make the length an optional parameter so that we can still preserve the same functionality as the original `reverse_letters`

Please also add a test case for `length = 0`

tianyizheng02 (changes requested):

PR #10143, opened 2023-10-08
Added the Minkowski distance function

dbring:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/maths/minkowski_distance.py
+++ b/maths/minkowski_distance.py
@@ -11,10 +11,16 @@ def minkowski_distance(
 
     https://en.wikipedia.org/wiki/Minkowski_distance
 
+    Note: due to floating point calculation errors the output of this
+    function may be inaccurate.
+
     >>> minkowski_distance([1.0, 1.0], [2.0, 2.0], 1)
     2.0
     >>> minkowski_distance([1.0, 2.0, 3.0, 4.0], [5.0, 6.0, 7.0, 8.0], 2)
     8.0
+    >>> import numpy as np
+    >>> np.isclose(5.0, minkowski_distance([5.0], [0.0], 3))
+    True
     >>> minkowski_distance([1.0], [2.0], -1)
     Traceback (most recent call last):
         ...
--- a/maths/minkowski_distance.py
+++ b/maths/minkowski_distance.py
@@ -1,0 +1,41 @@
+def minkowski_distance(
+    point_a: list[float],
+    point_b: list[float],
+    order: int,
+) -> float:
+    """
+    This function calculates the Minkowski distance for a given order between
+    two n-dimensional points represented as lists. For the case of order = 1,
+    the Minkowski distance degenerates to the Manhattan distance. For
+    order = 2, the usual Euclidean distance is obtained.
+
+    https://en.wikipedia.org/wiki/Minkowski_distance
+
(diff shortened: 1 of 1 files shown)


maths/minkowski_distance.py
```diff
@@ -0,0 +1,41 @@
...
+    Traceback (most recent call last):
+        ...
+    Exception: The order must be greater than or equal to 1.
+    >>> minkowski_distance([1.0], [1.0, 2.0], 1)
+    Traceback (most recent call last):
+        ...
+    Exception: Both points must have the same dimension.
+    """
+    if order < 1:
+        raise Exception("The order must be greater than or equal to 1.")
```
tianyizheng02, line 28:
```suggestion
        raise ValueError("The order must be greater than or equal to 1.")
```
Please make the error of a more specific type

maths/minkowski_distance.py
```diff
@@ -0,0 +1,41 @@
...
+
+    if len(point_a) != len(point_b):
+        raise Exception("Both points must have the same dimension.")
```
tianyizheng02, line 31:
```suggestion
        raise ValueError("Both points must have the same dimension.")
```

maths/minkowski_distance.py
```diff
@@ -0,0 +1,41 @@
...
+
+    return float(
+        sum(abs(a - b) ** order for a, b in zip(point_a, point_b)) ** (1 / order)
+    )
```
tianyizheng02, line 35:
```suggestion
    return sum(
        abs(a - b) ** order for a, b in zip(point_a, point_b)
    ) ** (1 / order)
```
I don't think the `float()` cast is necessary, as the inside should already  be returning a float

maths/minkowski_distance.py
```diff
@@ -0,0 +1,41 @@
...
+    Exception: Both points must have the same dimension.
+    """
+    if order < 1:
+        raise Exception("The order must be greater than or equal to 1.")
+
+    if len(point_a) != len(point_b):
+        raise Exception("Both points must have the same dimension.")
+
+    return float(
+        sum(abs(a - b) ** order for a, b in zip(point_a, point_b)) ** (1 / order)
```
tianyizheng02, line 34: I'm concerned that the `** (1 / order)` could produce inaccuracies in the output. For example, `125 ** (1/3)` returns 4.999... instead of 5 because of floating-point errors. I don't know of any better alternative, but I think it's worth looking into.

tianyizheng02 (changes requested):

maths/minkowski_distance.py
dbring: Yeah, I can't find a good solution for this either. I've tried using Decimal, pow, numpy but I'm getting the same result as you. I think adding in some sort of approximation scheme like Newton's method would unnecessarily complicate and detract from the main thrust of this algorithm. What do you think?

tianyizheng02: In order to not overcomplicate the code, I think it should be fine to keep it as is. We should leave a note in the comments acknowledging this issue so that future contributors know about it, however. For doctests, we could use `np.isclose()` to compare doctest outputs to the true, mathematical values.

tianyizheng02 (approved):

PR #10271, opened 2023-10-11
adding a split_workload algorithm

meg-1:
### Describe your change:
This algorithm is a brute-force search over (nearly) all 2^n possible partitions and was created for educational purposes. The asymptotic runtime of this code is: O(n * 2^n) this is a case of the partition problem. It accepts a multiset ( list ) of integers, distributes them, and returns a tuple, containing two lists, with minimal difference between their sums.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/sorts/split_workload.py
+++ b/sorts/split_workload.py
@@ -1,0 +1,72 @@
+def split_list(
+    timings: list[int | float | str],
+) -> tuple[list[int | float], list[int | float], int | float] | None:
+    """
+
+    This algorithm is a brute-force search over (nearly) all 2^n
+    possible partitions and was created for educational purposes.
+    the asymptotic runtime of this code is: O(n * 2^n)
+
+    this is a case of the partition problem.
+    it accepts a multiset ( list ) of integers,
+    distributes them, and returns a tuple, containing two lists,
+    with minimal difference between their sums
+
+    >>> split_list([27, 21, 92, 87, 1, 32])
+    ([27, 21, 87], [92, 1, 32], 10)
+    >>> split_list([52, 385, 9956, 25, 2367, 1111, 17, 925])
+    ([9956], [52, 385, 25, 2367, 1111, 17, 925], 5074)
+    >>> split_list([12, 10, 11, 9])
+    ([10, 11], [12, 9], 0)
+    >>> split_list([-1551, 2712, 2325, 2623])
+    ([1551, 2712], [2325, 2623], 685)
+    >>> split_list(["12.5", "10", "11", "9"])
+    ([10, 11], [12.5, 9], 0.5)
+    >>> split_list(["twelve", "ten", "eleven", "nine"])
+    Traceback (most recent call last):
+    ValueError: Timings must be a list of numbers
+
+    """
+    valid_timings: list[int | float] = []
+    for current_element in timings:
+        if (
+            isinstance(current_element, str)
+            and current_element.replace(".", "", 1).isdigit()
+        ):
+            is_current_elem_int = float(current_element).is_integer()
+            if not is_current_elem_int:
+                valid_timings.append(abs(float(current_element)))
+            else:
+                valid_timings.append(abs(int(float(current_element))))
+        elif isinstance(current_element, (int, float)):
+            valid_timings.append(abs(current_element))
+        else:
+            print(type(current_element))
+            raise ValueError("Timings must be a list of numbers")
+
+    if len(valid_timings) == 0:
+        return ([], [], 0)
+    elif len(valid_timings) == 1:
+        return ([valid_timings[0]], [], valid_timings[0])
+
+    result = None
+    n = len(valid_timings)
+    smallest_diff = float("inf")
+    for i in range(1, 2**n - 1):
+        indices = [j for j in range(n) if i & (1 << j) != 0]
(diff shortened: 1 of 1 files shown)


meg-1:
@cclauss, was wondering if you could check this pr out? would be very appreciated! all checks pass, all that's needed is an approving review :))

cclauss:

Should this be in the scheduling/ directory?  Is this a sorting algorithm?  It seems to be a workload-partitioning helper, but not a sort.  What would be a good Wikipedia or similar URL to add?

priya-sundaram-dev:
@cclauss good instinct — this isn't a sort. It's the **partition problem** (split a multiset into two subsets with minimum sum-difference), so `sorts/` is the wrong home.

**Answers to your questions**

- **Directory:** Not `sorts/`, and not really `scheduling/` either (nothing here models jobs/deadlines/CPU order — the "timings/workload" framing is just flavor text over the classic partition problem). The natural home is **`dynamic_programming/`**, right next to the file we already have for this exact problem: [`dynamic_programming/minimum_partition.py`](https://github.com/TheAlgorithms/Python/blob/master/dynamic_programming/minimum_partition.py).
- **Wikipedia URL:** https://en.wikipedia.org/wiki/Partition_problem

**Overlap to resolve first.** `minimum_partition.py` already solves this with a pseudo-polynomial **O(n·S)** subset-sum DP. This PR is an **O(n·2ⁿ)** brute force. The one genuine thing it adds is that it returns *the two subsets*, not just the minimum difference. So the question for the author is: is returning the partition (not just the diff) enough of a differentiator to justify a second file? If yes, I'd:
1. Rename to something like `dynamic_programming/partition_with_reconstruction.py` (and make the function name match the file — right now the file is `split_workload.py` but the function is `split_list`), add a module docstring + the Wikipedia link, and cross-reference `minimum_partition.py`.
2. Ideally reconstruct the subsets from the DP table instead of enumerating 2ⁿ masks, so the two files share the same complexity story. Brute force is fine *as an explicitly-labeled educational baseline*, but for a repo people cite it's worth doing the DP.

**Code notes (independent of the above):**
- `print(type(current_element))` on the error path looks like leftover debugging — please drop it (the `raise ValueError(...)` already carries the signal).
- The return annotation `... | None` never occurs: for `n >= 2` the first loop iteration always satisfies `diff < inf`, so `result` is never `None`. You can drop the `None`.
- `abs()` silently rewrites the input: `-1551` is partitioned as `1551` (see the `[-1551, 2712, 2325, 2623]` doctest). That's surprising — either treat negatives as real values, or document loudly that inputs are taken by magnitude. Note `minimum_partition.py` currently mis-handles negatives too, so worth being deliberate here.
- The numeric-string parse `current_element.replace(".", "", 1).isdigit()` rejects `"-5"` (the `-` fails `isdigit`) and lets `"1.2.3"` slip to `float()` where it raises — tighten with a small `try/float()` helper.
- The exact subset assignment in the doctests isn't unique (many partitions share the min diff); it's deterministic given the bitmask order but brittle to refactors. Asserting on the *difference* plus a `sum(a) - sum(b)` invariant is more robust than pinning the exact lists.

Happy to help the author land whichever direction you prefer, @meg-1. If it's "keep as a partition-reconstruction example," I can sketch the DP-with-backtrack version.

cclauss:
@priya-sundaram-dev Please push a new commit to this branch that will allow us to merge this algorithm.

priya-sundaram-dev:
@cclauss I'd be glad to, but I can't push to this branch — it lives on `meg-1`'s fork (`meg-1/Python:split_workload`), and "Allow edits from maintainers" only grants push to users who already have write access to the base repo, which I don't have. So the commit has to come from @meg-1 (or a maintainer with write). To make that a copy-paste job, here's the minimal, self-contained diff that clears the two hard blockers — the leftover debug `print` and the unreachable `| None` branch:

```diff
 def split_list(
     timings: list[int | float | str],
-) -> tuple[list[int | float], list[int | float], int | float] | None:
+) -> tuple[list[int | float], list[int | float], int | float]:
@@
         elif isinstance(current_element, (int, float)):
             valid_timings.append(abs(current_element))
         else:
-            print(type(current_element))
             raise ValueError("Timings must be a list of numbers")
```

(For `n >= 2` the first loop iteration always beats `float("inf")`, so `result` is never `None`; and the `n == 0` / `n == 1` branches already return early — so the annotation is safe to drop.)

That's enough to get a clean CI + a coherent signature. The two open **judgment calls** are yours/@meg-1's, not code blockers:

1. **Home + overlap.** This is the partition problem, so `dynamic_programming/` sits better than `sorts/`, right next to the existing `minimum_partition.py` (O(n·S) DP for the min *difference*). The only thing this PR adds over that file is returning the two subsets. If that's worth a second file, I'd rename it (file `split_workload.py` vs function `split_list` should match) and cross-link the two; otherwise it's a duplicate.
2. **`abs()` on inputs** silently maps `-1551 → 1551` (visible in the `[-1551, ...]` doctest). Fine if documented, surprising if not.

@meg-1 if you'd rather I sketch the DP-with-backtracking version (same subsets, O(n·S) instead of O(n·2ⁿ)) so it clearly earns its place next to `minimum_partition.py`, say the word and I'll write it up here for you to commit.

PR #10286, opened 2023-10-11
 Graphics using Turtle function(a 3D circle with benzene ring)

vishalm029:
### Describe your change:
I have made a 3D circle and geometric pattern of a Benzene by imparting different colors to it by using the turtle function by which we can use the feature like a drawing board, which lets us command a turtle to draw all over it! We can use functions like turtle.forward(…) and turtle.right(…) which can move the turtle around. 

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I know pull requests will not be merged if they fail the automated tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,7 +51,7 @@ repos:
       - id: validate-pyproject
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.5.1
+    rev: v1.6.0
     hooks:
       - id: mypy
         args:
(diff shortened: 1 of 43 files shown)


graphics/rainbow_benzene.py
```diff
@@ -0,0 +1,26 @@
...
+# Radius of circle
+pen.circle(100)
+pen.end_fill()
+pen.hideturtle()
+
+# for the colour of rainbow
+colors = ["red", "purple", "blue", "green", "orange", "yellow"]
+# to generate the benzene
+t = turtle.Pen()
+turtle.bgcolor("black")
```
s-xiae, line 21: I don't think you need this line...

s-xiae: This line doesn't work, for bgcolor pretty much only works on backgrounds.

s-xiae: the code is `t.Screen().bgcolor("black")`

s-xiae (changes requested):
That line only works on the background, but if you wanted to change the background to black, then really just edit the code.

PR #10309, opened 2023-10-12
Create bernouli_principle.py

DHRUVKADAM22:
### Describe your change:
The Bernoulli principle is a fundamental concept in fluid dynamics that describes the relationship between the pressure and velocity of a fluid in a streamlined flow. You can use Python to create a simple script to visualize the Bernoulli principle. In this example, I have use the Matplotlib library to create a plot that demonstrates the principle.

Algorithm:-

1. import `numpy` as np and import `matplotlib.pyplot as plt`: These lines import the NumPy library as "np" and the Matplotlib library as "plt." NumPy is used for numerical calculations, and Matplotlib is used for creating plots and visualizations.

2. `v = np.linspace(0, 10, 100)`: This line creates an array v of 100 equally spaced values between 0 and 10. This array represents a range of velocities.

3. `pressure_constant = 1.0`: This line sets a constant value for the Bernoulli equation. In this example, it's set to 1.0 for simplicity, but in real-world applications, this constant would be specific to the fluid and conditions being analyzed.

4. `pressure = pressure_constant - 0.5 * v**2`: This line calculates the pressure using the Bernoulli equation. It relates the pressure and velocity of a fluid in a streamlined flow. The result is stored in the pressure array.

5. `plt.figure(figsize=(8, 6))`: This line creates a new figure for the plot with a specified size (8 units wide and 6 units high).

6. `plt.plot(v, pressure)`: This line plots the data. It uses the array v as the x-axis (velocity) and the array pressure as the y-axis (pressure).

7. `plt.title("Bernoulli Principle Visualization")`: This line sets the title of the plot.

8. `plt.xlabel("Velocity") and plt.ylabel("Pressure")`: These lines set labels for the x-axis and y-axis, respectively.

9. `plt.grid(True)`: This line adds a grid to the plot for better readability.

10. `plt.show()`: Finally, this line displays the plot. When you run the script, a graphical window should appear showing a plot of pressure vs. velocity, illustrating the Bernoulli principle.

### Checklist:
* [Y] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [Y] This pull request is all my own work -- I have not plagiarized.
* [Y] I know that pull requests will not be merged if they fail the automated tests.
* [Y] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [Y] All new Python files are placed inside an existing directory.
* [Y] All filenames are in all lowercase characters with no spaces or dashes.
* [Y] All functions and variable names follow Python naming conventions.

### DIFF
--- a/physics/bernouli_principle.py
+++ b/physics/bernouli_principle.py
@@ -1,0 +1,25 @@
+#First, make sure you have Matplotlib installed. You can install it using pip if you haven't already
+#pip install matplotlib
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+# Create an array of velocities
+v = np.linspace(0, 10, 100)  # Velocity range from 0 to 10
+
+# Define constants for the Bernoulli equation
+pressure_constant = 1.0  # A constant for simplicity
+
+# Calculate pressure using the Bernoulli equation
+pressure = pressure_constant - 0.5 * v**2
+
+# Create a plot
+plt.figure(figsize=(8, 6))
+plt.plot(v, pressure)
+plt.title("Bernoulli Principle Visualization")
+plt.xlabel("Velocity")
+plt.ylabel("Pressure")
+plt.grid(True)
+
+# Show the plot
+plt.show()
--- a/physics/bernouli_principle.py
+++ b/physics/bernouli_principle.py
@@ -1,5 +1,5 @@
-#First, make sure you have Matplotlib installed. You can install it using pip if you haven't already
-#pip install matplotlib
+# First, make sure you have Matplotlib installed. You can install it using pip if you haven't already
+# pip install matplotlib
 
 import numpy as np
 import matplotlib.pyplot as plt


DHRUVKADAM22:
please check I have added Bernoulli Principle  please merge it

DHRUVKADAM22:
So my changes are not usefull? 

if it is usefull please merge it .

physics/bernouli_principle.py
```diff
@@ -0,0 +1,25 @@
+# First, make sure you have Matplotlib installed. You can install it using pip if you haven't already
```
haxkd, line 1: make sure pass through ruff

DHRUVKADAM22: what changes should I do can please suggest me

haxkd: please see the ruff, pre-commit failures

PR #10336, opened 2023-10-12
Edit_Distance algorithm for genetic string matching

anshul-2010:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/genetic_algorithm/edit_distance.py
+++ b/genetic_algorithm/edit_distance.py
@@ -1,25 +1,21 @@
 def edit_distance(source, target):
-    """
-    Edit distance algorithm is a string metric, i.e., it is a way of quantifying
-    how dissimilar two strings are to one another, that is measured by
-    counting the minimum number of operations required to transform one string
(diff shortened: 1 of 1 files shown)


tianyizheng02:
You checked this checkbox, so you acknowledged this:
> - [x] All new Python files are placed inside an existing directory.

tianyizheng02:
My mistake, I had forgotten that we had a separate directory for genetic algorithms. However, edit distance is still definitely not a genetic algorithm. We also already have multiple implementations of edit distance (using dynamic programming). Since your implementation doesn't use DP, you're welcome to add it to one of the existing files instead of creating a separate file.

anshul-2010:
I added it to the genetic algorithms considering that edit distance would be used for genome matching scenarios.
But sure, I will add it to some other existing file if you say so.

tianyizheng02:
Superseded by #10571

PR #10427, opened 2023-10-14
charging_inductor

dhruvtrigotra:
### Charging of Inductor

With the help of this algorithm we can easily find the current at any nth second in the
inductor while the inductor being charged. It is all because of the RL-timeconstant.
when an inductor is connected with a resistor in series it gives a exponential function. It is very
helpful when we making or designing new circuits, thus we have decide which inductor and
resistor we should choose.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


cclauss (approved):
Almost 100 lines of code for a one line calculation.  It is correct but it would be cool to see some more complex, nuanced algorithms in the `electronics` directory.

dhruvtrigotra:
> Almost 100 lines of code for a one line calculation.  It is correct but it would be cool to see some more complex, nuanced algorithms in the `electronics` directory.

Sure sir! I got you.

PR #10438, opened 2023-10-14
added rkf45 method

ravi-ivar-7:
Add an algorithm (rkf45) to TheAlgorithms / Python/maths

### DIFF
--- a/maths/rkf45.py
+++ b/maths/rkf45.py
@@ -79,13 +79,11 @@ def runge_futta_fehlberg_45(
         )
         k4 = step_size * func(
             x[i] + (12 / 13) * step_size,
-            y[i] + (1932 / 2197) * k1 - (7200 / 2197) *
-            k2 + (7296 / 2197) * k3,
+            y[i] + (1932 / 2197) * k1 - (7200 / 2197) * k2 + (7296 / 2197) * k3,
         )
         k5 = step_size * func(
             x[i] + step_size,
-            y[i] + (439 / 216) * k1 - 8 * k2 +
-            (3680 / 513) * k3 - (845 / 4104) * k4,
+            y[i] + (439 / 216) * k1 - 8 * k2 + (3680 / 513) * k3 - (845 / 4104) * k4,
         )
         k6 = step_size * func(
             x[i] + step_size / 2,
--- a/maths/rkf45.py
+++ b/maths/rkf45.py
@@ -1,6 +1,7 @@
 from collections.abc import Callable
 import numpy as np
 
+
 class RangeError(Exception):
     "Will be raised when initial x is greater than or equal to final x"
 
--- a/maths/rkf45.py
+++ b/maths/rkf45.py
@@ -6,6 +6,7 @@ import numpy as np
 class RangeError(Exception):
     "Will be raised when initial x is greater than or equal to final x"
 
+
 def runge_futta_fehlberg_45(
     ode: Callable,
     y0: float,
--- a/maths/rkf45.py
+++ b/maths/rkf45.py
@@ -1,7 +1,6 @@
 from collections.abc import Callable
 import numpy as np
 
-
 class RangeError(Exception):
     "Will be raised when initial x is greater than or equal to final x"
 
--- a/maths/rkf45.py
+++ b/maths/rkf45.py
@@ -1,11 +1,19 @@
 from collections.abc import Callable
 import numpy as np
 
+
 class RangeError(Exception):
-    'Will be raised when initial x is greater than or equal to final x'
+    "Will be raised when initial x is greater than or equal to final x"
     pass
 
-def rkf45(ode:Callable , y0:float,x0:float,step_size:float,xn:float,) ->np.ndarray:
+
+def rkf45(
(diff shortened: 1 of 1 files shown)


maths/rkf45.py
```diff
@@ -0,0 +1,86 @@
+from collections.abc import Callable
+import numpy as np
```
cclauss, line 2:
```suggestion

import numpy as np
```

maths/rkf45.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class RangeError(Exception):
+    "Will be raised when initial x is greater than or equal to final x"
+
+
+def rkf45(
```
cclauss, line 9:
```suggestion
def runge_futta_fehlberg_45(
```

maths/rkf45.py
```diff
@@ -0,0 +1,86 @@
...
+    ode: Callable,
+    y0: float,
+    x0: float,
+    step_size: float,
+    xn: float,
+) -> np.ndarray:
+    """
+    Solve ODE using Runge-Kutta-Fehlberg Method (rkf45) of order 5.
```
cclauss, line 17: What is `ODE` and why does the reader have to guess?

maths/rkf45.py
```diff
@@ -0,0 +1,86 @@
...
+    ode (callable): Ordinary Differential Equation as function of x and y.
+    y0 (float) : Initial value of y.
+    x0 (float) : Initial value of x.
+    step_size (float) : Increament value of x (step-size).
+    xn (float) : Final value of x.
+
+    Returns:
+        np.ndarray: Solution of y at each nodal point
+
+    #excact value of y[1] is tan(0.2) = 0.2027100355086
```
cclauss, line 32:
```suggestion
    # exact value of y[1] is tan(0.2) = 0.2027100355086
```

maths/rkf45.py
```diff
@@ -0,0 +1,86 @@
...
+    >>> def f(x,y):
+    ...     return 1+y**2
+    >>> y=rkf45(f,0,0,0.2,1)
```
cclauss, line 35:
```suggestion
    >>> y=rkf45(f, 0, 0, 0.2, 1)
```

maths/rkf45.py
```diff
@@ -0,0 +1,86 @@
...
+
+    args:
+    ode (callable): Ordinary Differential Equation as function of x and y.
+    y0 (float) : Initial value of y.
+    x0 (float) : Initial value of x.
+    step_size (float) : Increament value of x (step-size).
+    xn (float) : Final value of x.
+
+    Returns:
+        np.ndarray: Solution of y at each nodal point
```
cclauss, line 30: Please do not repeat the data types that are checked by mypy in the function signature because the reader will be confused if one changes and the other does not.
Put x before y.
```suggestion
    func: Callable,
    x_initial: float,
    y_initial: float,
    step_size: float,
    x_final: float,
) -> np.ndarray:
    """
    Solve an Ordinary Differential Equation (ODE) function using
    Runge-Kutta-Fehlberg Method (rkf45) of order 5.

    Reference: https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta%E2%80%93Fehlberg_method

    args:
    func: Ordinary Differential Equation as a function of x and y.
    x_initial: Initial value of x.
    y_initial: Initial value of y.
    step_size: Increament value of x (step-size).
    x_final: Final value of x.

    Returns:
        Solution of y at each nodal point
```

cclauss (changes requested):

maths/rkf45.py
cclauss: maths/rkf45.py:26: Increament ==> Increment
maths/rkf45.py:32: excact ==> exact

maths/rkf45.py
```diff
@@ -9,43 +9,43 @@ class RangeError(Exception):
 
 def runge_futta_fehlberg_45(
     ode: Callable,
-    y0: float,
-    x0: float,
+    x_initial: float,
+    y_initial: float,
```
cclauss, line 13: These two were swapped (x before y) so please make sure that the values are also swapped in your doctests

maths/rkf45.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Returns:
+        np.ndarray: Solution of y at each nodal point
+
+    # exact value of y[1] is tan(0.2) = 0.2027100355086
+    >>> def f(x,y):
+    ...     return 1+y**2
+    >>> y=runge_futta_fehlberg_45(f, 0, 0, 0.2, 1)
+    >>> y[1]
+    0.2027100937470787
```
cclauss, line 37: We will need more than one test.  What happens if `x_initial` is set to 5, -5, 5.5?

What happens if `step_size` is zero or negative?

What happens if `x_final` is zero or negative?

What happens if `ode` is `5` instead of being a function?

Let's have a test that raises a RangeError.

cclauss (approved):

cclauss:
Nice one!  It is great that you deal with invalid input.

PR #10439, opened 2023-10-14
Add Title Case Conversion

0APOCALYPSE0:
### Describe your change:

convert lower or upper case string to title case

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/strings/title.py
+++ b/strings/title.py
@@ -1,0 +1,57 @@
+def to_title_case(word: str) -> str:
+    """
+    Converts a string to capitalized case, preserving the input as is
+
+    >>> to_title_case("Aakash")
+    'Aakash'
+
+    >>> to_title_case("aakash")
+    'Aakash'
+
+    >>> to_title_case("AAKASH")
+    'Aakash'
+
+    >>> to_title_case("aAkAsH")
+    'Aakash'
+    """
+
+    """
+    Convert the first character to uppercase if it's lowercase
+    """
+    if "a" <= word[0] <= "z":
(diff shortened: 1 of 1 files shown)


strings/title.py
```diff
@@ -0,0 +1,60 @@
+def convert_word(word: str) -> str:
```
cclauss, line 1: Let's have a self-documenting function name.
```suggestion
def to_title_case(word: str) -> str:
```

0APOCALYPSE0: @cclauss Hey, Actually I am using two methods:
convert_word will convert a single word to a title case word ex: "dog" to "Dog"
THEN main method to_title_case will take all the converted words and will join them together to create a sentence.

so just wanted to let you know we already have a method name to_title_case so maybe I can give some other name to convert_word method like to_title_word or something else

cclauss: Let's keep the word one short: `to_title_case()` and the sentence one longer: `sentence_to_title_case()` and we can leave the word `convert` out of the function names.

strings/title.py
```diff
@@ -0,0 +1,60 @@
...
+    'Aakash Giri'
+
+    >>> to_title_case("aAkAsH gIrI")
+    'Aakash Giri'
+    """
+
+    words = input_str.split()
+    title_case_str = [convert_word(word) for word in words]
+
+    return " ".join(title_case_str)
```
cclauss, line 54:
```suggestion
    return " ".join(convert_word(word) for word in input_str.split())
```

cclauss (approved):

PR #10467, opened 2023-10-14
Disable unused dependencies

tianyizheng02:
### Describe your change:

Disable unused dependencies in `requirements.txt` to try to reduce build times:
- Keras is a TensorFlow frontend, so I've also marked it as `python_version < '3.12'`. There's no reason to install Keras in every build when TensorFlow can't be used.
- projectq, scikit-fuzzy, and yulewalker are only used by broken (and thus disabled) files, so I've commented them out. If/when those broken files are fixed, we can uncomment them.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,19 +1,19 @@
 beautifulsoup4
 fake_useragent
 imageio
-keras
+keras ; python_version < '3.12'
 lxml
 matplotlib
 numpy
 opencv-python
 pandas
 pillow
-projectq
+# projectq  # uncomment once quantum/quantum_random.py is fixed
 qiskit ; python_version < '3.12'
 qiskit-aer ; python_version < '3.12'
 requests
 rich
-scikit-fuzzy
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,24 +1,24 @@
...
 qiskit-aer ; python_version < '3.12'
 requests
 rich
-scikit-fuzzy
+# scikit-fuzzy  # uncomment once fuzzy_logic/fuzzy_operations.py is fixed
 scikit-learn
 statsmodels
 sympy
 tensorflow ; python_version < '3.12'
 texttable
```
tianyizheng02, line 21:
```suggestion
```
I don't think texttable is used in any file, but I might've missed a usage in my search. @cclauss, could you confirm?

cclauss (approved):

requirements.txt
tianyizheng02: @cclauss?

Issue #9943, opened 2023-10-06
Improve our test coverage


tianyizheng02:
### Feature description

Many of our existing algorithm files have little to no unit testing. This is problematic because this can easily let bugs slip through. We want some assurance that the code we currently have is correct and functional. We welcome all contributors to open PRs to help us add tests to our codebase.

### How to find low-coverage files

Go to the Actions tab in this repository and find the most recent **build** workflow run. Open the logs under "Run Tests" and scroll down until you find the section on code coverage:
```
---------- coverage: platform linux, python 3.12.0-final-0 -----------
Name                                                                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------------------------------------------
quantum/q_fourier_transform.py                                                   30     30     0%   14-93
scripts/validate_solutions.py                                                    54     54     0%   2-94
strings/min_cost_string_conversion.py                                            78     75     4%   20-57, 61-75, 79-129
...
```
The "Cover" column tells you what percentage of the lines in that file are covered by tests. We want to increase this percentage for existing files. Find a file with low coverage percentage that you wish to write tests for, add doctests for each function, and open a PR with your changes. You do not need to have a perfect coverage percentage, but all functions should have doctests.

Some files will naturally be hard to write tests for. For example, the file may be poorly written because they lack any functions. Other files might be how-tos, meaning they simply demonstrate how to use an existing library's functions rather than implementing the algorithm themselves. Ignore these kinds of files, as they will need to be rewritten eventually. Furthermore, ignore files in the `web_programming` and `project_euler` directories. Web programming files are inherently hard to test and Project Euler files have their own validation workflow, so don't worry about their test coverage.

_**When you open your PR, put "Contributes to #9943" in the PR description.**_ Do not use the word "fixes", "resolves", or "closes". This issue is an ongoing one, and your PR will not single-handedly resolve this issue.

### How to add doctests

A doctest is a unit test that is contained within the documentation comment (docstring) for a function. Here is an example of what doctests look like within a docstring:
```py
def add(a: int, b: int) -> int:
    """
    Adds two non-negative numbers.
    >>> add(1, 1)
    2
    >>> add(2, 5)
    7
    >>> add(1, 0)
    1
    >>> add(-1, -1)
    Traceback (most recent last):
    ...
    ValueError: Numbers must be non-negative
    """
```
[...]

PR #10479, opened 2023-10-15
updated physics/archimedes_principle.py

Baron105:
### Describe your change:

modified physics/archimedes_principle.py
added some doctests for exception handling, imported g from scipy rather than hard coding and allowed zero gravity to be valid giving value 0 in that case

fixes issue #9943 

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:


physics/archimedes_principle.py
```diff
@@ -10,8 +10,8 @@
 """
 
 
-# Acceleration Constant on Earth (unit m/s^2)
-g = 9.80665
+# Acceleration Constant on Earth (unit m/s^2) imported from scipy
+from scipy.constants import g
```
cclauss, line 14: Please revert this change.  Let's not create a dependency on all of scipy just to set a constant.

physics/archimedes_principle.py
```diff
@@ -30,13 +30,27 @@ def archimedes_principle(
...
+    ValueError: Impossible Gravity
     """
 
     if fluid_density <= 0:
         raise ValueError("Impossible fluid density")
-    if volume < 0:
+    if volume <= 0:
         raise ValueError("Impossible Object volume")
-    if gravity <= 0:
+    if gravity < 0:
```
cclauss, line 53: Please revert these changes.  It is useful to be able to make theoretical calculations even if we are unable to reproduce those conditions.

cclauss (changes requested):

physics/archimedes_principle.py
Baron105: should I remove the tests as well or just make the signs as they were before?

cclauss: The tests need to follow the code.  It is OK to have the tests but their output must pass.

Baron105: The numerical tests are passing. Just added some more tests to see for when the inputs are invalid physically as in negative gravity, non positive volume and non positive density as they were already in the original code by the author.

Baron105: please check @cclauss

cclauss: Dude!  There are 215 pull requests open.  Perhaps we will get to review this one.

cclauss (approved):
Nice!!

Baron105:
Great! Looks better now!

PR #10273, opened 2023-10-11
Adds exponential moving average algorithm

PoojanSmart:
### Describe your change:

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [ ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

PR #10531, opened 2023-10-15
add monotonic queue algorithm

Jiang15:
### Describe your change:

Similar the monotonic stack implementation in next_greater_element.py, I implemented an epic monotonical queue example: max_sliding_window which is to continuously generate max value in a sliding window over an array. 

I also implemented a test with one example and the test has passed.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/queue/monotonic_queue.py
+++ b/data_structures/queue/monotonic_queue.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from .double_ended_queue import Deque
+from collections import deque
 
 arr = [1, 3, -1, -3, 5, 3, 6, 7]
 window_size = 3
@@ -17,19 +17,19 @@ def max_sliding_window(arr: list[float], window_size: int) -> list[float]:
     True
     """
     max_val = []
-    queue = Deque()
+    mono_queue: deque = deque()
     for i in range(len(arr)):
         # pop the element if the index is outside the window size k
-        if queue and i - queue._front.val >= window_size:
-            queue.popleft()
+        if mono_queue and i - mono_queue[0] >= window_size:
+            mono_queue.popleft()
         # keep the queue monotonically decreasing
         # so that the max value is always on the top
-        while queue and arr[i] >= arr[queue._back.val]:
-            queue.pop()
-        queue.append(i)
+        while mono_queue and arr[i] >= arr[mono_queue[-1]]:
+            mono_queue.pop()
+        mono_queue.append(i)
         # the maximum value is the first element in queue
         if i >= window_size - 1:
-            max_val.append(arr[queue._front.val])
+            max_val.append(arr[mono_queue[0]])
     return max_val
 
 
--- a/data_structures/queue/monotonic_queue.py
+++ b/data_structures/queue/monotonic_queue.py
@@ -1,26 +1,26 @@
 from __future__ import annotations
 
-from double_ended_queue import Deque
+from .double_ended_queue import Deque
 
 arr = [1, 3, -1, -3, 5, 3, 6, 7]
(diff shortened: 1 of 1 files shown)


data_structures/queue/monotonic_queue.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+from double_ended_queue import Deque
+
+arr = [1, 3, -1, -3, 5, 3, 6, 7]
+k = 3
+expect = [3, 3, 5, 5, 6, 7]
+
+
+def max_sliding_window(arr: list[float], k: int) -> list[float]:
```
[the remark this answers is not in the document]
Jiang15: I changed it to window_size

data_structures/queue/monotonic_queue.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+from .double_ended_queue import Deque
```
cclauss, line 3: Could this use https://docs.python.org/3/library/collections.html#collections.deque or is there a special capability in the local version?

data_structures/queue/monotonic_queue.py
```diff
@@ -0,0 +1,39 @@
...
+    Each time the sliding window of length window_size moves right by one position.
+    Return the max sliding window.
+    >>> max_sliding_window(arr, window_size) == expect
+    True
+    """
+    max_val = []
+    queue = Deque()
+    for i in range(len(arr)):
+        # pop the element if the index is outside the window size k
+        if queue and i - queue._front.val >= window_size:
```
cclauss, line 23: Using the _private_ `_front` is fairly risky.  I.e. Not future-proof.
```

File: data_structures/queue/monotonic_queue.py [hunk 4]
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+from .double_ended_queue import Deque
+
+arr = [1, 3, -1, -3, 5, 3, 6, 7]
+window_size = 3
+expect = [3, 3, 5, 5, 6, 7]
+
+
+def max_sliding_window(arr: list[float], window_size: int) -> list[float]:
```
cclauss, line 10: Can we take the same approach with this algorithm?
https://github.com/TheAlgorithms/Python/pull/10273#discussion_r1355794522

Seems appropriate for a sliding_window algorithm.

data_structures/queue/monotonic_queue.py
Jiang15: Yes we can. And so that we can avoid using private attributes

data_structures/queue/monotonic_queue.py
Jiang15: Yes! I use the collections.deque now

data_structures/queue/monotonic_queue.py
Jiang15: It is a bit different because the queue has to be monotonically decreasing while calculating the max value in a window.

cclauss: If it does not make sense then we can close the pull request.

Jiang15: Hi, I resolved the issue of using collections.deque instead of double_ended_queue. 

I feel the max sliding window is still an important algorithm which is a good use case of the queue (it is similar to next_greater_element in stack).

cclauss: @tianyizheng02 Is my iterator, not list request above unreasonable in this algorithm?

cclauss (approved):

PR #10469, opened 2023-10-14
Omit `project_euler/` from coverage reports

tianyizheng02:
### Describe your change:

Our current `pytest` command for our builds is the following:
```
pytest --ignore=quantum/q_fourier_transform.py --ignore=project_euler/ --ignore=scripts/validate_solutions.py --cov-report=term-missing:skip-covered --cov=. .
```

Since we're not testing files in the `project_euler/` or `scripts/` directories, we should exclude them from our builds' coverage reports as well. `project_euler/` files in particular significantly clutter the reports.

* [ ] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [ ] All new Python files are placed inside an existing directory.
* [ ] All filenames are in all lowercase characters with no spaces or dashes.
* [ ] All functions and variable names follow Python naming conventions.
* [ ] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [ ] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [ ] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [ ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

PR #10542, opened 2023-10-15
Separate `pytest` workflow into separate workflows for tests and coverage

tianyizheng02:
### Describe your change:

Following up from #10469, I've separated the pytest tests and coverage into two separate workflows. The goal is to cut down on the unneeded clutter in the logs and make it easier for contributors (and ourselves) to read.

* [ ] Add an algorithm?
* [ ] Documentation change?
* [x] GitHub workflows

### Checklist:


tianyizheng02:
No matter what I do to omit these two files from, `pytest` keeps including them in the coverage step and it's causing the step to fail. @cclauss, any ideas?
```
=========================== short test summary info ============================
ERROR quantum/q_fourier_transform.py - ModuleNotFoundError: No module named 'qiskit'
ERROR scripts/validate_solutions.py - KeyError: 'GITHUB_TOKEN'
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
======================== 2 warnings, 2 errors in 18.43s ========================
Error: Process completed with exit code 2.
```

cclauss:
`mv quantum/q_fourier_transform.py quantum/q_fourier_transform.py.DISABLED`

.github/workflows/build.yml
```diff
@@ -23,15 +23,13 @@ jobs:
           python -m pip install --upgrade pip setuptools six wheel
           python -m pip install pytest-cov -r requirements.txt
       - name: Run tests
-        # TODO: #8818 Re-enable quantum tests
+        # TODO: #8818 Re-enable quantum tests (files are currently disabled)
```
cclauss, line 26:
```suggestion
        # TODO: #8818 Re-enable filenames ending with `.py.DISABLED`.
```

tianyizheng02:
I've been experimenting with `pytest` and `pytest-cov` for a while now and it seems like there isn't a straightforward way to separate the tests from the coverage report. Since `pytest-cov` is just a plugin for `pytest`, running the latter requires running the former, and I believe that must include the progress bar and warnings. Plus, this would require us to basically run `pytest` twice per build. Perhaps we could output the coverage report to a temporary file, but that's not ideal. I don't have any better solution at this time.

cclauss:
I was thinking that the covers report could be piped into a file and then cat filename.txt could happen in a separate GHA step.

Issue #8594, opened 2023-04-01
Optimize Project Euler solutions


dhruvmanila:
### Feature description

If someone would like to take a stab at optimizing these solutions, feel free to open a PR or ask for any help here:

## Slowest 10 durations
- [ ] Problem 145 / sol1.py  (***25.00s***)
- [ ] Problem 070 / sol1.py  (***13.86s***)
- [ ] Problem 104 / sol1.py  (***6.84s***)
- [ ] Problem 092 / sol1.py  (***6.01s***)
- [ ] Problem 010 / sol2.py  (***3.90s***)
- [ ] Problem 010 / sol1.py  (***3.76s***)
- [ ] Problem 078 / sol1.py  (***3.67s***)
- [ ] Problem 135 / sol1.py  (***3.59s***)
- [ ] Problem 073 / sol1.py  (***2.64s***)
- [ ] Problem 034 / sol1.py  (***2.54s***)

_Reference: https://github.com/TheAlgorithms/Python/actions/runs/4580132675/jobs/8088552368#step:5:356_

ankit-gautam23:
Is this still open??

dhruvmanila:
> Is this still open??

Yes

MatthiasHeinz:
Regarding problem 034, I've stumbled upon this [proof ](https://www.xarg.org/puzzle/project-euler/problem-34/).

This proof lowers the upper bound from the current `limit = 7 * factorial(9) + 1 # 2,540,161` to 1,499,999, which eliminates  40% of values to be considered. Considering this eliminates the biggest values, this probably cuts runtime by more than 40%.

Additionally I'd prefer if that upper limit would have a comment / reference attached to it, stating their origin. Otherwise, this limit looks very arbitrary.

=============

Regarding the aforementioned proof: At first glance, admittedly it has two phrases, which are formulated incorrectly / ambiguously; but it looks solid.

1. The sentence 
> This implies that if n is a 7 digit number, either the second digit must be 0 or 1 or the last digit is 1. 

should end in

> [...] the **first** digit is 1.

2. The sentence

> If the second digit is now 5, one of the remaining digits has to be at most 4.

should clarify

> If the second digit is now **at least** 5 [...]

Bjiornulf:
for problem 10, `sol3.py` is a solution using sieve and is fast. I don't think we can make sol1 and sol2 faster, unless sieve is also used. In that case, they should just be deleted IMO...

[...]

PR #10558, opened 2023-10-15
Performance: 58% faster Project Euler 070

ManpreetXSingh:
### Describe your change:

#8594

- Benchmark:
    - Old solution: 118.2/10 seconds
    - New solution:  49.3/10 seconds

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


quant12345:
@ManpreetSingh2004 a good solution with filtering by prime numbers. Look at 072, I made it, you can probably apply it there too.

ManpreetSingh2004:
Hi @quant12345,
I noticed that the prime filtering for 072 is already implemented in sol2.py. Although it is as fast as sol1 with both taking 1 sec. What should be done in this case?
Also, I've identified a bug in 'sol1.py' that appears to be related to integer overflow. This issue specifically arises on Windows platforms because, by default, numpy on `Windows 64-bit` uses 32-bit integers, whereas on Linux, it employs 64-bit integers, so it also went unnoticed in automated checking of this repo. See #10672.

quant12345:
@ManpreetSingh2004 now I checked it in Windows, it’s really an incorrect number if you don’t specify the type. I used your filtering by prime numbers. And with this it’s faster, significantly. Below code:

<details>

<summary>test</summary>


```
import datetime
import numpy as np
from math import isqrt


def np_calculate_prime_numbers(max_number: int) -> list[int]:
    """
    Returns prime numbers below max_number.
    See: https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes

    >>> np_calculate_prime_numbers(10)
    [2, 3, 5, 7]
    >>> np_calculate_prime_numbers(2)
    []
    """
    if max_number <= 2:
        return []

    # List containing a bool value for every odd number below max_number/2
    is_prime = np.ones(max_number // 2, dtype=bool)

    for i in range(3, isqrt(max_number - 1) + 1, 2):
        if is_prime[i // 2]:
            # Mark all multiple of i as not prime using list slicing
            is_prime[i ** 2 // 2:: i] = False

    primes = np.where(is_prime)[0] * 2 + 1
    primes[0] = 2
    return primes


def solution_new(limit: int = 1_000_000) -> int:
    """
    Returns an integer, the solution to the problem
    >>> solution(10)
    31
    >>> solution(100)
    3043
    >>> solution(1_000)
    304191
    """

    # generating an array from -1 to limit
    phi = np.arange(-1, limit)
    primes = np_calculate_prime_numbers(limit)

    for i in primes:
        ind = np.arange(2 * i, limit + 1, i)  # indexes for selection
        phi[ind] -= phi[ind] // i

    return int(np.sum(phi[2: limit + 1], dtype=np.int64))


def solution(limit: int = 1_000_000) -> int:
    """
    Returns an integer, the solution to the problem
    >>> solution(10)
    31
    >>> solution(100)
    3043
    >>> solution(1_000)
    304191
    """

    # generating an array from -1 to limit
    phi = np.arange(-1, limit)

    for i in range(2, limit + 1):
        if phi[i] == i - 1:
            ind = np.arange(2 * i, limit + 1, i)  # indexes for selection
            phi[ind] -= phi[ind] // i

    return int(np.sum(phi[2: limit + 1], dtype=np.int64))


if __name__ == "__main__":

    n = 1_000_000

    now = datetime.datetime.now()
    old = solution(n)
    print('old_time', datetime.datetime.now() - now)

    now = datetime.datetime.now()
    new = solution_new(n)
    print('new_time', datetime.datetime.now() - now)

    print('comparison for identity', old == new)
```


</details>


Output:

```
old_time 0:00:00.683273
new_time 0:00:00.473020
comparison for identity True
```

ManpreetSingh2004:
@quant12345
Hey, That's really nice. 😊

dhruvmanila (changes requested):
Thanks for contributing! I'm sorry if I was not clear but we should be updating existing solution to make it performant instead of adding new solution(s). The benchmarks can be added in the PR description instead of the code. We only want one solution in the code. Can you please update the code accordingly?

ManpreetSingh2004:
> We only want one solution in the code.

@dhruvmanila
If I understand correctly, the three slower solutions (`slow_solution`, `slicing_solution`, `py_solution`) should be removed. Am I correct?

ManpreetSingh2004:
Hi @dhruvmanila,
I've made the requested changes. Please let me know if any additional adjustments are needed.
The get_totients function is modified. The remainder of the solution remains unchanged from its original form.
Also I've put the benchmark results in the PR description.

PR #10571, opened 2023-10-16
Edit Distance Algorithm for String Matching

anshul-2010:
Added an algorithm to perform string matching using Edit Distance Algorithm.
Helps to find the minimum number of operations to move from one string to another.

* [ x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/strings/edit_distance.py
+++ b/strings/edit_distance.py
@@ -1,0 +1,40 @@
+def edit_distance(source: str, target: str) -> int:
+    """
+    Edit distance algorithm is a string metric, i.e., it is a way of quantifying
+    how dissimilar two strings are to one another, that is measured by
+    counting the minimum number of operations required to transform one string
+    into another.
+    In genetic algorithms consisting of A,T, G, and C nucleotides, this matching
+    becomes essential in understanding the mutation in successive genes.
+    Hence, this algorithm comes in handy when we are trying to quantify the
+    mutations in successive generations.
+    Args:
+    source (type __string__): This is the source string, the initial string with
+    respect to which we are calculating the edit_distance for the target
+    target (type __string__): This is the target string, which is formed after n
+    number of operations performed on the source string.
+    Assumptions:
+    The cost of operations (insertion, deletion and subtraction) is all 1
+    Given two integers, return the sum.
+    :param source: str
+    :param target: str
+    :return: int
+    >>> edit_distance("GATTIC", "GALTIC")
+    1
+    """
+    delta = {True: 0, False: 1}  # Substitution
+    
+    if len(source) == 0:
+        return len(target)
+    elif len(target) == 0:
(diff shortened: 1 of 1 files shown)


strings/edit_distance.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    delta = {True: 0, False: 1}  # Substitution
+
+    if len(source) == 0:
+        return len(target)
+    elif len(target) == 0:
+        return len(source)
+
+    return min(
+        edit_distance(source[:-1], target[:-1]) + delta[source[-1] == target[-1]],
```
tianyizheng02, line 33:
```suggestion
    if len(source) == 0:
        return len(target)
    elif len(target) == 0:
        return len(source)

    delta = int(source[-1] != target[-1])  # Substitution
    return min(
        edit_distance(source[:-1], target[:-1]) + delta,
```

strings/edit_distance.py
```diff
@@ -0,0 +1,40 @@
...
+        edit_distance(source, target[:-1]) + 1,
+        edit_distance(source[:-1], target) + 1,
+    )
+
+
+print(edit_distance("ATCGCTG", "TAGCTAA"))
+# Answer is 4
```
tianyizheng02, line 40:
```suggestion
if __name__ == "__main__":
    print(edit_distance("ATCGCTG", "TAGCTAA"))  # Answer is 4
```
Driver code should always be encapsulated within a `__main__` block

strings/edit_distance.py
```diff
@@ -0,0 +1,40 @@
...
+    source (type __string__): This is the source string, the initial string with
+    respect to which we are calculating the edit_distance for the target
+    target (type __string__): This is the target string, which is formed after n
+    number of operations performed on the source string.
+    Assumptions:
+    The cost of operations (insertion, deletion and subtraction) is all 1
+    Given two integers, return the sum.
+    :param source: str
+    :param target: str
+    :return: int
```
tianyizheng02, line 21:
```suggestion
    Edit distance algorithm is a string metric, i.e., it is a way of quantifying
    how dissimilar two strings are to one another. It is measured by
    counting the minimum number of operations required to transform one string
    into another.
    
    This implementation assumes that the cost of operations (insertion, deletion and
    substitution) is always 1
    
    Args:
    source: the initial string with respect to which we are calculating the edit distance
        for the target
    target: the target string, formed after performing n operations on the source string
```
Significantly cut down on the length of the docstring. The section on genetic algorithms is unnecessary because edit distance is not a genetic algorithm. It is a string algorithm that is _applied_ to problems in genetics.

tianyizheng02 (changes requested):

tianyizheng02 (approved):

PR #10672, opened 2023-10-18
Bug Fix: Integer overflow on windows in Project Euler 072

ManpreetSingh2004:
### Describe your change:

- ### Bug Fix in Sol1.py
    - I've identified a bug in `sol1.py` that appears to be related to integer overflow. This issue specifically arises on Windows platforms because, by default, numpy on `Windows 64-bit` uses `32-bit integers`, whereas on `Linux`, it employs `64-bit integers`, so it also **went unnoticed in automated checking**.
- ### Sol2.py
    - Separated prime number generator from solution.
- ### Checked all Euler Problems
    - I've checked all other project euler solutions using the following command on both windows and linux.
        - `pytest scripts\validate_solutions.py`
- Both solutions have a runtime of less than or equal to 1 sec. So, I've not tried to optimize anything yet. Let me know if required.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [ x This pull request is all my own work -- I have not plagiarized.

### DIFF
--- a/project_euler/problem_072/sol1.py
+++ b/project_euler/problem_072/sol1.py
@@ -36,7 +36,7 @@ def solution(limit: int = 1_000_000) -> int:
     """
 
     # generating an array from -1 to limit
-    phi = np.arange(-1, limit)
+    phi = np.arange(-1, limit, dtype=np.int64)
 
     for i in range(2, limit + 1):
         if phi[i] == i - 1:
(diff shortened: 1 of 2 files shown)


ManpreetSingh2004:
@cclauss Please help. My PR's are getting closed. 🥲

quant12345:
@ManpreetSingh2004 what numbers are different, show an example?

ManpreetSingh2004:
@quant12345
On windows it produces a negative number `-979125625` without specifying `dtype=np.int64`. With `dtype=np.int64`, it produces the correct output `303963552391`.
It is a known issue with numpy.
References:
- [Inconsistent handling of integer overflow between Windows and Linux](https://github.com/numpy/numpy/issues/8433)
- [numpy array dtype is coming as int32 by default in a windows 10 64 bit machine](https://stackoverflow.com/questions/36278590/numpy-array-dtype-is-coming-as-int32-by-default-in-a-windows-10-64-bit-machine)
- [Numpy returns different results on windows and unix [duplicate]](https://stackoverflow.com/questions/72482769/numpy-returns-different-results-on-windows-and-unix)

PR #10682, opened 2023-10-18
added runge-kutta-gills method to python/maths

iamrknain:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


maths/runge_kutta_gills.py
```diff
@@ -0,0 +1,95 @@
+"""
+Use the Runge-Kutta-Gill's method of order 4 to solve Ordinary Differential Equations.
+Author : Ravi Kumar
```
cclauss, line 3:
```suggestion
https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta_methods
Author : Ravi Kumar
```

cclauss:
I wonder:
1. Should these go into the arithmatic_analysis directory instead of the generic `math` directory?
2. Should the `arithmatic_analysis` directory be renamed to `numerical_analysis`?
    *https://en.wikipedia.org/wiki/Numerical_analysis
3 Should we combine this and two existing Runge-Kutta algorithms so visitors can easily study them together?

maths/runge_kutta_gills.py
```diff
@@ -0,0 +1,95 @@
...
+"""
+
+from collections.abc import Callable
+
+import numpy as np
+
+from math import sqrt
```
cclauss, line 10:
```suggestion

from collections.abc import Callable
from math import sqrt

import numpy as np
```

maths/runge_kutta_gills.py
```diff
@@ -0,0 +1,95 @@
...
+Author : Ravi Kumar
+"""
+from collections.abc import Callable
+from math import sqrt
+
+import numpy as np
+
+
+def runge_kutta_gills(
+    func: Callable,
```
tianyizheng02, line 14:
```suggestion
    func: Callable[[float, float], float],
```
Please annotate the parameter and output types

maths/runge_kutta_gills.py
```diff
@@ -0,0 +1,95 @@
...
+            "The final value of x must be greater than initial value of x."
+        )
+
+    if step_size <= 0:
+        raise ValueError("Step size must be positive.")
+
+    n = int((x_final - x_initial) / step_size)
+    y = np.zeros(
+        (n + 1),
+    )
```
tianyizheng02, line 72: Why not this?
```suggestion
    y = np.zeros(n + 1)
```

Also, is there a reason why you're using a numpy array as opposed to a built-in Python list?

maths/runge_kutta_gills.py
```diff
@@ -0,0 +1,95 @@
...
+    x = np.zeros(n + 1)
+    y[0] = y_initial
+    x[0] = x_initial
+    for i in range(n):
```
tianyizheng02, line 76: Could you have this for-loop iterate directly over x values instead?
```suggestion
    for x in range(x_initial, x_final, step_size):
```
Or something along those lines. That way you don't need to manually increment x, nor do you need store the x values in a list.

tianyizheng02 (changes requested):

PR #10690, opened 2023-10-19
Test cases for  interpolation_search

RaymondDashWu:
### Describe your change:
Contributes to #9943 

* [ ] Add an algorithm?
* [ ] Documentation change?
* [X] Add test cases

### Checklist:


searches/interpolation_search.py
```diff
@@ -10,6 +10,8 @@ def interpolation_search(sorted_collection, item):
     :param sorted_collection: some ascending sorted collection with comparable items
     :param item: item value to search
     :return: index of found item or None if item is not found
+    >>> interpolation_search([-1, 0, 1], -1)
+    0
```
cclauss, line 14: Please add some tests that return useful/interesting values instead of just `0` and `None`.
Please add some tests that raise an exception like floats, str, list, or dicts.
Also: (tuple and set instead of list)
`>>> interpolation_search((-1, 0, 1), -1)`
`>>> interpolation_search({-1, 0, 1}, -1)`
`>>> interpolation_search([], -1)`

RaymondDashWu: Added more tests based on your advice

searches/interpolation_search.py
```diff
@@ -3,7 +3,7 @@
 """
 
 
-def interpolation_search(sorted_collection, item):
+def interpolation_search(sorted_collection: list[int], item: int) -> int | None:
     """Pure implementation of interpolation search algorithm in Python
     Be careful collection must be ascending sorted, otherwise result will be
```
cclauss, line 10: Please test for this condition.  Give it an unsorted input and lets see if it fails or not.

searches/interpolation_search.py
```diff
@@ -48,14 +74,60 @@ def interpolation_search(sorted_collection, item):
...
     """Pure implementation of interpolation search algorithm in Python by recursion
     Be careful collection must be ascending sorted, otherwise result will be
     unpredictable
     First recursion should be started with left=0 and right=(len(sorted_collection)-1)
     :param sorted_collection: some ascending sorted collection with comparable items
     :param item: item value to search
     :return: index of found item or None if item is not found
+    >>> interpolation_search_by_recursion(
+    ... [-1, 0, 1], -1, 0, 2
+    ... )
```
cclauss, line 89: Please do not put tests on three lines unless your lines get to be longer than 88 characters.
```suggestion
    >>> interpolation_search_by_recursion([-1, 0, 1], -1, 0, 2)
```

searches/interpolation_search.py
RaymondDashWu: An unsorted input doesn't work with this implementation unfortunately. For example ```>>> interpolation_search([5, 0, 1], 2) == None``` will return ```True``` despite 2 not being in the array. I wasn't sure how to handle this as there is a helper function called ```__assert_sorted```. I considered adding that code at the beginning of each function but that changes the time complexity of the functions from logn to nlogn. If something is done to sanitize the input I assume it should be done in a separate PR.

searches/interpolation_search.py
RaymondDashWu: Got it! Will make a new commit. However, for the future, is there a way for a Python linter to be added as part of a pre commit check? I'm still new to doing open source contributions in Python but for Javascript projects I've worked on, it auto installed a linter when you did an ```npm install``` on the project.

PR #10716, opened 2023-10-20
Add docstrings and doctests and fix a bug ciphers/trifid_cipher.py

FEROS01:
### Describe your change:

Contributes to https://github.com/TheAlgorithms/Python/issues/9943

### Checklist:


ciphers/trifid_cipher.py
```diff
@@ -1,8 +1,28 @@
-# https://en.wikipedia.org/wiki/Trifid_cipher
+"""
+The trifid cipher uses a table to fractionate each plaintext letter into a
+trigram,mixes the constituents of the trigrams, and then applies the table
+in reverse to turn these mixed trigrams into ciphertext letters.
+https://en.wikipedia.org/wiki/Trifid_cipher
+"""
+
 from __future__ import annotations
```
cclauss, line 10: Let's define `character_to_number` once outside the functions.

`number_to_character` can be created from `character_to_number` in one line of code.

https://github.com/TheAlgorithms/Python/blob/579937613a6dc7e099b710e3d57767a2fab115ad/ciphers/morse_code.py#L9-L22

ciphers/trifid_cipher.py
```diff
@@ -1,8 +1,28 @@
...
+https://en.wikipedia.org/wiki/Trifid_cipher
+"""
+
 from __future__ import annotations
 
 
 def __encrypt_part(message_part: str, character_to_number: dict[str, str]) -> str:
+    """
+    Arranges the triagram value of each letter of 'message_part' vertically
+    and joins them horizontally
```
cclauss, line 14: Use _active_ tense like you are commanding the computer what to do.
```suggestion
    Arrange the triagram value of each letter of 'message_part' vertically and join them
    horizontally.
```

ciphers/trifid_cipher.py
```diff
@@ -20,6 +40,20 @@ def __encrypt_part(message_part: str, character_to_number: dict[str, str]) -> st
 def __decrypt_part(
     message_part: str, character_to_number: dict[str, str]
 ) -> tuple[str, str, str]:
+    """
+    Converts each letter of the input string into there respective trigram
+    values, joins them and splits them into three equal groups of strings.
+    Then returns the group of strings .
```
cclauss, line 46:
```suggestion
1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
    Convert each letter of the input string into there respective trigram values, join
    them and split them into three equal groups of strings.
    Then returns the group of strings .

```

ciphers/trifid_cipher.py
```diff
@@ -110,13 +216,35 @@ def encrypt_message(
...
+    """
     message, alphabet, character_to_number, number_to_character = __prepare(
         message, alphabet
     )
+
     decrypted_numeric = []
     decrypted = ""
 
-    for i in range(0, len(message) + 1, period):
+    for i in range(0, len(message), period):
```
cclauss, line 247: Why this change??

ciphers/trifid_cipher.py
FEROS01: We can't hardcode the number_to_character for the functions, because the alphabet is a variable in trifid cipher. Unlike the morse code where each letter has its own representation.
But we can do it for the doctests though.

ciphers/trifid_cipher.py
FEROS01: For situations where the length of the string is a multiple of the period, the loop continues even after it had finished traversing the string, there by creating a ' list index out of range ' error in the __decrypt_part helper function. To reproduce the bug try
```python
decrypt_message('FMJFVOISSUFTFPUFEQQC','FELIXMARDSTBCGHJKNOPQUVWYZ+',5)
# Expected 'AIDETOILECIELTAIDERA'

# The length of the string is 20, a multiple of the period 5.

for i in range(0, (20+1), period)
# This will iterate 5 times [0, 5, 10, 15, 20], taking 5 letters of the string 5 times.
# The 5th time will be an empty string because 5*5 = 25 and we have just 20 letters
# Creating a 'list index out of range' error

for i in range(0, (20) , period)
# Will iterate 4 times [0, 5 10, 15], which is what we want.
# It will also work for strings with lengths that are not multiples of the period




```

cclauss:
Just two characters... ;-)
```
ruff --output-format=github .
Error: ciphers/trifid_cipher.py:45:89: E501 Line too long (90 > 88 characters)
```

cclauss (approved):
Nice one!  Thanks.

PR #10732, opened 2023-10-20
add exponential search algorithm

kiarash8112:
### Describe your change:

added exponential algorithm using binary_search_recursion also tests and docs about it

* [x] Add an algorithm?
* [x] Documentation change?

### Checklist:


searches/binary_search.py
```diff
@@ -297,6 +297,49 @@ def binary_search_by_recursion(
...
+    0
+
+    >>> exponential_search([0, 5, 7, 10, 15], 15)
+    4
+
+    >>> exponential_search([0, 5, 7, 10, 15], 5)
+    1
+
+    >>> exponential_search([0, 5, 7, 10, 15], 6)
+
```
cclauss, line 325:
```suggestion
    >>> exponential_search([0, 5, 7, 10, 15], 0)
    0
    >>> exponential_search([0, 5, 7, 10, 15], 15)
    4
    >>> exponential_search([0, 5, 7, 10, 15], 5)
    1
    >>> exponential_search([0, 5, 7, 10, 15], 6) = None
    True
```

searches/binary_search.py
```diff
@@ -297,6 +297,49 @@ def binary_search_by_recursion(
         return binary_search_by_recursion(sorted_collection, item, midpoint + 1, right)
 
 
+def exponential_search(sorted_collection: list[int], item: int) -> int | None:
```
cclauss, line 300: I am not a fan of functions that return multiple datatypes.  If the search fails, let's return `-1` (like `"Hello".find("Z")`) or raise a `ValueError` (like `"Hello".index("Z")`).
```suggestion
def exponential_search(sorted_collection: list[int], item: int) -> int:
```

searches/binary_search.py
```diff
@@ -297,6 +297,49 @@ def binary_search_by_recursion(
...
+    )
+
+
+if __name__ == "__main__":
+    import doctest
+
+    doctest.testmod()
+
+
 if __name__ == "__main__":
```
cclauss, line 343: Only one of these.  Run the tests first and let the user interact afterward.
```suggestion
```

cclauss (changes requested):
Nice addition!

For optional extra credit, add a `timeit` benchmark to show relative performance.

searches/binary_search.py
```diff
@@ -297,7 +297,46 @@ def binary_search_by_recursion(
...
+    >>> exponential_search([0, 5, 7, 10, 15], 5)
+    1
+    >>> exponential_search([0, 5, 7, 10, 15], 6) = None
+    True
+    """
+    sorted_collection = sorted(sorted_collection)
+    bound = 1
+    while bound < len(sorted_collection) and sorted_collection[bound] < item:
+        bound *= 2
+    left = int(bound / 2)
```
cclauss, line 328:
```suggestion
    left = bound // 2
```

searches/binary_search.py
```diff
@@ -297,7 +297,46 @@ def binary_search_by_recursion(
...
+
+    Examples:
+    >>> exponential_search([0, 5, 7, 10, 15], 0)
+    0
+    >>> exponential_search([0, 5, 7, 10, 15], 15)
+    4
+    >>> exponential_search([0, 5, 7, 10, 15], 5)
+    1
+    >>> exponential_search([0, 5, 7, 10, 15], 6) = None
+    True
```
cclauss, line 322:
```suggestion
    >>> exponential_search([0, 5, 7, 10, 15], 6)
    -1
```

cclauss:
I thought more about sorting the sorted collection… it is a bad idea.  The second item in might be different in the two lists so the function might tell me that is the second item when in my list it is the third item.  So, the implementations should raise ValueError on unsorted input.

cclauss (approved):
This is a really slick addition.  Thanks for doing it!

kiarash8112:
@cclauss  thank you for your suggestions :)

tianyizheng02:
@cclauss Since exponential search is a different algorithm from binary search, why was it placed in the `binary_search.py` file? Shouldn't it have gotten its own file?

searches/binary_search.py
```diff
@@ -297,12 +282,78 @@ def binary_search_by_recursion(
...
+    -1
+    """
+    if list(sorted_collection) != sorted(sorted_collection):
+        raise ValueError("sorted_collection must be sorted in ascending order")
+    bound = 1
+    while bound < len(sorted_collection) and sorted_collection[bound] < item:
+        bound *= 2
+    left = bound // 2
+    right = min(bound, len(sorted_collection) - 1)
+    last_result = binary_search_by_recursion(
```
cclauss, line 316: @tianyizheng02 `exponential_search()` embeds a call to `binary_search_by_recursion()` but as the benchmarks demonstrate delivers a faster result.  Looking at the first commit makes this clearer.

PR #10742, opened 2023-10-21
Consolidate binary exponentiation files

tianyizheng02:
### Describe your change:

Consolidated implementations of binary (modular) exponentiation into just `binary_exponentiation.py`

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:


maths/binary_exponentiation.py
```diff
@@ -1,48 +1,148 @@
...
+Repeat until b = 1 or b = 0
 
-def binary_exponentiation(a: int, n: int) -> int:
+For modular exponentiation, we use the fact that (a * b) % c = ((a % c) * (b % c)) % c
+"""
+
+
+def binary_exp_recursive(base: int, exponent: int) -> int:
     """
-    Compute a number raised by some quantity
```
cclauss, line 9: Why remove this?  `exp` is a bit cryptic in the function name and a line of documentation here is useful.

maths/binary_exponentiation.py
```diff
@@ -1,48 +1,148 @@
...
-    >>> binary_exponentiation(-1, 3)
+    >>> binary_exp_recursive(3, 5)
```
cclauss, line 19: What about tests for a=big number andn=big number?

maths/binary_exponentiation.py
```diff
@@ -1,48 +1,148 @@
...
-        POWER = int(input("Enter Power : ").strip())
-    except ValueError:
-        print("Invalid literal for integer")
+    BASE = int(input("Enter base: ").strip())
+    POWER = int(input("Enter power: ").strip())
 
-    RESULT = binary_exponentiation(BASE, POWER)
-    print(f"{BASE}^({POWER}) : {RESULT}")
+    RESULT = binary_exp_recursive(BASE, POWER)
+    print(f"{BASE}^({POWER}): {RESULT}")
```
cclauss, line 148: OPTIONAL: Adding a `timeit` benchmark would be a nice addition.  Visitors love to see the side-by-side bakeoff.

cclauss (approved):

maths/binary_exponentiation.py
tianyizheng02: Can do, but the n=float case won't work because the algorithm can only compute integer powers

cclauss: I know it won’t work but I wanted to see how it fails.  Does it raise a ValueError?  Does it behave like `pow()` does.  Testing how things work if half the battle.  Watching them fail is just as cool.

Issue #10563, opened 2023-10-15
`pytest` warnings in builds


tianyizheng02:
### Repository commit

Doesn't matter

### Python version (python --version)

Python 3.12.0, the current version used in builds

### Dependencies version (pip freeze)

Doesn't matter

### Expected behavior

`pytest` runs with no (internal) warnings

### Actual behavior

Every build (see our Actions tab) has the following warnings:
```
/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/coverage/report_core.py:115: CoverageWarning: Couldn't parse '/home/runner/work/Python/Python/config-3.py': No source for code: '/home/runner/work/Python/Python/config-3.py'. (couldnt-parse)
  coverage._warn(msg, slug="couldnt-parse")
/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/coverage/report_core.py:115: CoverageWarning: Couldn't parse '/home/runner/work/Python/Python/config.py': No source for code: '/home/runner/work/Python/Python/config.py'. (couldnt-parse)
  coverage._warn(msg, slug="couldnt-parse")
```
I don't know what files these are, and I don't know what's causing these warnings to occur.

SOBAN50:
Okay so I spent god time understanding the issue and I have come to the conclusion that it is an issue with the
`Python\compression\peak_signal_to_noise_ratio.py` file.

It arises due to the ```cv2``` library. cv2 uses a config file for its operations. These files are `config-3.py` and `config.py`. 
The `coverage` module considers it a python file for which code coverage has to be checked, but its not actually a Python file but a configuration file. Thats what generates a warning.

SOBAN50:
There are multiple solutions to this:

- Leave it as it as because its just a warning and does not break any code functionality.
- Consider using the PILLOW package instead of CV2.

Saswatsusmoy:
> Okay so I spent good time understanding the issue and I have come to the conclusion that it is an issue with the `Python\compression\peak_signal_to_noise_ratio.py` file.
> 
> It arises due to the `cv2` library. cv2 uses a config file for its operations. These files are `config-3.py` and `config.py`. The `coverage` module considers it a python file for which code coverage has to be checked, but its not actually a Python file but a configuration file. Thats what generates a warning.

I tried to implement the code without using any external libraries, it seems that the warnings still popup.

SOBAN50:
[...]

PR #10744, opened 2023-10-21
Updated peak_signal_to_noise_ratio.py

Saswatsusmoy:
To #10563 
Implemented a new function, read_image(), to read PNG images from file without OpenCV. RGB values for each pixel are stored in a nested list. Updated peak_signal_to_noise_ratio() to directly use the nested list instead of NumPy arrays. No longer relies on external libraries. Improved efficiency and independence.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/compression/peak_signal_to_noise_ratio.py
+++ b/compression/peak_signal_to_noise_ratio.py
@@ -1,24 +1,24 @@
-"""
-Peak signal-to-noise ratio - PSNR
-    https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio
-Source:
-https://tutorials.techonical.com/how-to-calculate-psnr-value-of-two-images-using-python
-"""
-
 import math
 import os
+
+import numpy as np
 
 PIXEL_MAX = 255.0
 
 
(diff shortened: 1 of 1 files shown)


compression/peak_signal_to_noise_ratio.py
```diff
@@ -8,36 +8,54 @@
...
 
-import cv2
-import numpy as np
-
 PIXEL_MAX = 255.0
 
 
-def peak_signal_to_noise_ratio(original: float, contrast: float) -> float:
-    mse = np.mean((original - contrast) ** 2)
+def peak_signal_to_noise_ratio(original, contrast):
```
cclauss, line 14: Please revert the changes to this line. Removing the type hints is NOT allowed.  That would be going in the wrong direction since it would be blinding `mypy` which runs in our GitHub Actions.

Modifying functions that have no tests is also NOT allowed.  Please create a separate pull request that adds doctests.  Once that pull request is merged then we can come back to this pull request.

cclauss (changes requested):
Type hints on function parameters and doctests are mandatory.

PR #10776, opened 2023-10-21
added a function to calculate perceived frequency by observer using Doppler Effect

Baron105:
### Describe your change:

added a function to calculate perceived frequency by observer using Doppler Effect

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


physics/doppler_frequency.py
```diff
@@ -0,0 +1,106 @@
...
+    Non-positive frequency implies vs > v or v0 > v(in opposite direction)
+    0.0
+    """
+
+    try:
+        doppler_freq = (org_freq * (wave_vel + obs_vel)) / (wave_vel - src_vel)
+    except ZeroDivisionError:
+        print("Error: Division by zero")
+        print("Infinite frequency implies vs = v and observer infront of the source")
+        return 0.0
```
cclauss, line 95: `CONTRIBUTING.md` says that algorithmic functions should not `print()`.  Please let the ZeroDivisionError be raised to the caller because the answer is NOT 0.
```suggestion
        print("Error: Division by zero")
        print("Infinite frequency implies vs = v and observer infront of the source")
        return 0.0
```

cclauss (changes requested):
Raising an Exception is the right thing to do.

physics/doppler_frequency.py
Baron105: > `CONTRIBUTING.md` says that algorithmic functions should not `print()`. Please let the ZeroDivisionError be raised to the caller because the answer is NOT 0.

Yes, the print statement was somewhat redundant and not aligned with the Contributing Guidelines. Replaced it with raising the exception codes.

cclauss (approved):

PR #10885, opened 2023-10-24
create strong_password_detection.py

sirilalithaadapa:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/strong_password_detection.py
+++ b/strong_password_detection.py
@@ -1,0 +1,43 @@
+def strong_password_detection(password):
+
+    AllowedSymbols = ['#', '@', '$', '_', '*', '-']
+    flag = 1
+    # length of the entered password should be at least 6
+    if len(password) < 6:
+        print('length of the entered password should be at least 6')
+        flag = 0
+    # length of the entered password should be not be greater than 15
+    if len(password) > 15:
+        print('length of the entered password should be not be greater than 15')
+        flag = 0
+    # The entered password should have at least one numeral
+    if not any(char.isdigit() for char in password):
+        print('The entered password should have at least one numeral')
+        flag = 0
+    # Password should have at least one lowercase letter
+    if not any(char.islower() for char in password):
+        print('the entered password should have at least one lowercase letter')
+        flag = 0
+    # The entered password should have at least one uppercase letter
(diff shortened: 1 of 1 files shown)


strong_password_detection.py
```diff
@@ -0,0 +1,42 @@
+def strong_password_detection(password):
+    AllowedSymbols = ["#", "@", "$", "_", "*", "-"]
```
cclauss, line 2: We require type hints and doctests...  Please add tests to cover all the cases below.
```suggestion
def strong_password_detection(password: str) -> bool:
    """
    >>> strong_password_detection("str0ng@Password")
    True
    >>> strong_password_detection("str0ngPassword")
    False
    """
    allowed_symbols = {c for c in "#@$_*-"}
```

cclauss (changes requested):

strong_password_detection.py
```diff
@@ -0,0 +1,42 @@
+def strong_password_detection(password):
```
cclauss, line 1: Please change the filename to `strings/strong_password_detection.py`

Akter-Ashik:
import re

def strong_password_detection(password):
    return (
        len(password) >= 12 and
        any(c.isupper() for c in password) and
        any(c.islower() for c in password) and
        any(c.isdigit() for c in password) and
        any(c in "#@$_*-" for c in password) and
        not re.search(r'(.)\1\1', password) and
        not any(word.lower() in password.lower() for word in ["password", "admin", "123456", "qwerty"]) and
        not any(info.lower() in password.lower() for info in ["john", "doe", "user123"]) and
        not re.search(r'123|abc|qwerty', password, re.IGNORECASE)
    )

# Example usage:
password = "StrongP@ss123!"
if strong_password_detection(password):
    print("The password is strong.")
else:
    print("The password is weak.")

tianyizheng02 (commented):
Plagiarized. From a cursory [GitHub search](https://github.com/search?q=%22The+entered+password+is+strong+%21%21%22&type=code), it appears that this code was ultimately taken from https://github.com/HarshCasper/Rotten-Scripts

PR #11126, opened 2023-10-31
deleted find_triplets_with_sum_0.py from data_structures as triplet_sum.pyfrom maths already has the implementation

AnshuSharma111:
### Describe your change:

Fixes #8098

deleted file as a better implementation was already present
removed reference of redundant file from Directory.md

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


cclauss (changes requested):
I think not.  The `math` file already contains two different implementations that are benchmarked against each other.  This is a great learning tool.  The `arrays` implementation uses `combinations` while one of the `math` implementations uses `permutations`.  Which one is more correct?  Which one is faster?

I think this pull request should start by moving the `arrays` implementation into the file with the other two implementations and then let's benchmark all three against each other.

cclauss:
Also #11134 landed a fourth algorithm to be unified.

PR #11172, opened 2023-11-21
Update edit_distance.py

pedram-mohajer:
### Describe your change:

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/strings/edit_distance.py
+++ b/strings/edit_distance.py
@@ -1,32 +1,48 @@
 def edit_distance(source: str, target: str) -> int:
     """
-    Edit distance algorithm is a string metric, i.e., it is a way of quantifying how
-    dissimilar two strings are to one another. It is measured by counting the minimum
-    number of operations required to transform one string into another.
+    Calculate the edit distance between two strings.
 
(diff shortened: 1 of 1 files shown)


strings/edit_distance.py
```diff
@@ -1,32 +1,48 @@
 def edit_distance(source: str, target: str) -> int:
     """
-    Edit distance algorithm is a string metric, i.e., it is a way of quantifying how
-    dissimilar two strings are to one another. It is measured by counting the minimum
-    number of operations required to transform one string into another.
+    Calculate the edit distance between two strings.
```
imSanko, line 3: Can we change the documentation as we'll the code too at the same time?

pedram-mohajer: Is there any rule about it?

imSanko: `You need to opt` for the option on which you're working with >_

pedram-mohajer: Got it, I'll close this and pull another one.

PR #11185, opened 2023-11-27
Create kruskal.py

pedram-mohajer:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/greedy_methods/kruskal.py
+++ b/greedy_methods/kruskal.py
@@ -2,7 +2,6 @@
 https://en.wikipedia.org/wiki/Kruskal%27s_algorithm
 """
 
-import doctest
 from collections.abc import Iterable
 from dataclasses import dataclass
 
@@ -13,9 +12,9 @@ class Edge:
     Represents an edge in a graph.
 
     Attributes:
-        start (str): The starting vertex of the edge.
-        end (str): The ending vertex of the edge.
-        weight (int): The weight of the edge.
+        start  : The starting vertex of the edge.
+        end    : The ending vertex of the edge.
+        weight : The weight of the edge.
     """
 
     start: str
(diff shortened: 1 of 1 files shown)


greedy_methods/kruskal.py
```diff
@@ -0,0 +1,120 @@
+"""
+https://en.wikipedia.org/wiki/Kruskal%27s_algorithm
+"""
+
+import doctest
```
cclauss, line 5: Please move this down into `__main__` because we only need it there.

greedy_methods/kruskal.py
```diff
@@ -0,0 +1,120 @@
...
+
+@dataclass
+class Edge:
+    """
+    Represents an edge in a graph.
+
+    Attributes:
+        start (str): The starting vertex of the edge.
+        end (str): The ending vertex of the edge.
+        weight (int): The weight of the edge.
```
cclauss, line 18: Please do not repeat the datatypes which are verified by mypy below because readers will become confused if one is changed but the other is not.
```suggestion
        start: The starting vertex of the edge.
        end: The ending vertex of the edge.
        weight: The weight of the edge.
```

greedy_methods/kruskal.py
```diff
@@ -0,0 +1,120 @@
...
+
+    >>> ds = DisjointSet(["A", "B", "C"])
+    >>> ds.find("A")
+    'A'
+    >>> ds.union("A", "B")
+    >>> ds.find("B")
+    'A'
+    >>> ds.union("B", "C")
+    >>> ds.find("C")
+    'A'
```
cclauss, line 50: Having all three answers be `'A'` does not raise confidence.

Let's use `disjoint_set` instead of `ds` as our variable name to make our code more self-documenting.

Please add tests where `disjoint_set = DisjointSet([])` and `disjoint_set = DisjointSet([1, 2, 3])`.

Let's add tests that fail like `>>> disjoint_set.find("D")`

greedy_methods/kruskal.py
```diff
@@ -0,0 +1,120 @@
...
+    ...     Edge("A", "B", 1), Edge("B", "C", 3), Edge("A", "C", 4),
+    ...     Edge("A", "D", 2), Edge("D", "C", 5)
+    ... ]
+    >>> mst = kruskal(graph)
+    >>> [(edge.start, edge.end, edge.weight) for edge in mst]
+    [('A', 'B', 1), ('A', 'D', 2), ('B', 'C', 3)]
+    """
+    vertices: set[str] = {vertex for edge in graph for vertex in [edge.start, edge.end]}
+    disjoint_set = DisjointSet(vertices)
+    mst: list[Edge] = []
```
cclauss, line 98:
```suggestion
    minimum_spanning_tree: list[Edge] = []
```

greedy_methods/kruskal.py
```diff
@@ -0,0 +1,120 @@
...
+
+    >>> graph = [
+    ...     Edge("A", "B", 1), Edge("B", "C", 3), Edge("A", "C", 4),
+    ...     Edge("A", "D", 2), Edge("D", "C", 5)
+    ... ]
+    >>> mst = kruskal(graph)
+    >>> [(edge.start, edge.end, edge.weight) for edge in mst]
+    [('A', 'B', 1), ('A', 'D', 2), ('B', 'C', 3)]
+    """
+    vertices: set[str] = {vertex for edge in graph for vertex in [edge.start, edge.end]}
```
cclauss, line 96:
```suggestion
    vertices: set[str] = {vertex for edge in graph for vertex in (edge.start, edge.end)}
```
`tuples` take less memory that `lists` and they are a signal to the reader that these values will not change at runtime.

cclauss (changes requested):

PR #11194, opened 2023-12-04
Colaltz_sequence.py Optimization

Ryan4412:
### Describe your change:       
Optimized collaze_sequence.py file in the math's algorithms based on mathematical definition of odd numbers. When a number is odd we multiply it by 3 and subtract 1. Any odd number multiplied by 3 is odd and an odd number -1 is always even. therefore we can execute the division of two step right after logging the first odd sequence which will speed up the performance of the algorithm by bypassing the execution of lines 51-53 every time a number is first reached.     
       
For more information to understand the definition of odd numbers I am using in this scenario please refer to the following website. https://www.cuemath.com/numbers/odd-numbers/

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


imSanko (approved):

MaximSmolskiy:
It looks like https://en.wikipedia.org/wiki/Loop_unrolling . I don't think it's a good idea to do this in a simple and clear algorithm. For me, readability and clarity are more important than performance in this case.

PR #11217, opened 2023-12-20
add suffix automaton algo

yan-victor:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/strings/suffix_automaton.py
+++ b/strings/suffix_automaton.py
@@ -1,0 +1,150 @@
+from __future__ import annotations
+
+
+class SuffixAutomaton:
+    """
+    Suffix Automaton data structure for efficient substring counting and searching.
+
+    Usage:
+    sa = SuffixAutomaton()
+    sa.extend('abc')
+    terminal_states = sa.terminal()
+    is_found = sa.find('ab')
+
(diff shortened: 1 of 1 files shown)


strings/suffix_automaton.py
```diff
@@ -0,0 +1,150 @@
...
+        self.last = 0
+        self.st = [self.Node(0, -1, [0] * self.alpha)]
+
+    class Node:
+        def __init__(self, length: int, link: int, next_states: list[int]):
+            self.len = length
+            self.link = link
+            self.next = next_states
+
+    def extend(self, c: str) -> None:
```
imSanko, line 32: Is it possible to add a PR and Change its doctests at the same, guess not only 1 at a time!!

imSanko (changes requested):

strings/suffix_automaton.py
cclauss: If we are adding a new file, we should add tests in the same PR.
If we are modifying an existing file, we should modify the code OR the tests but not both in the same PR.

imSanko (approved):

PR #11271, opened 2024-02-03
Fixed `max_parts` function for Circle

krishna1m:
- The number of times a circle is cut can only be a whole number
- The number of parts obtained after cutting is also a whole number

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/geometry/geometry.py
+++ b/geometry/geometry.py
@@ -154,7 +154,7 @@ class Circle(Ellipse):
         """
         if not isinstance(num_cuts, int) or num_cuts < 0:
             raise TypeError("num_cuts must be a whole number.")
-        return (num_cuts ** 2 + num_cuts) // 2 + 1
+        return (num_cuts**2 + num_cuts) // 2 + 1
 
 
 @dataclass
--- a/geometry/geometry.py
+++ b/geometry/geometry.py
@@ -134,25 +134,27 @@ class Circle(Ellipse):
 
         >>> circle = Circle(5)
         >>> circle.max_parts(0)
-        1.0
+        1
         >>> circle.max_parts(7)
-        29.0
+        29
         >>> circle.max_parts(54)
-        1486.0
+        1486
         >>> circle.max_parts(22.5)
-        265.375
+        Traceback (most recent call last):
+            ...
+        TypeError: num_cuts must be a whole number.
         >>> circle.max_parts(-222)
         Traceback (most recent call last):
             ...
-        TypeError: num_cuts must be a positive numeric value.
+        TypeError: num_cuts must be a whole number.
         >>> circle.max_parts("-222")
         Traceback (most recent call last):
             ...
-        TypeError: num_cuts must be a positive numeric value.
+        TypeError: num_cuts must be a whole number.
         """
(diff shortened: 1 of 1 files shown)


geometry/geometry.py
```diff
@@ -134,25 +134,27 @@ def max_parts(self, num_cuts: float) -> float:
 
         >>> circle = Circle(5)
         >>> circle.max_parts(0)
-        1.0
+        1
```
imSanko, line 137: why do you wanna add a ".0" is the rest of test cases ???

geometry/geometry.py
```diff
@@ -134,25 +134,27 @@ def max_parts(self, num_cuts: float) -> float:
...
         >>> circle.max_parts(7)
-        29.0
+        29
         >>> circle.max_parts(54)
-        1486.0
+        1486
         >>> circle.max_parts(22.5)
-        265.375
+        Traceback (most recent call last):
+            ...
```
imSanko, line 144: What's the need for that ?

geometry/geometry.py
```diff
@@ -134,25 +134,27 @@ def max_parts(self, num_cuts: float) -> float:
...
+        TypeError: num_cuts must be a whole number.
         >>> circle.max_parts(-222)
         Traceback (most recent call last):
             ...
-        TypeError: num_cuts must be a positive numeric value.
+        TypeError: num_cuts must be a whole number.
```
imSanko, line 149: Previous commit was pretty clear!

geometry/geometry.py
```diff
@@ -134,25 +134,27 @@ def max_parts(self, num_cuts: float) -> float:
...
         """
-        if not isinstance(num_cuts, (int, float)) or num_cuts < 0:
-            raise TypeError("num_cuts must be a positive numeric value.")
-        return (num_cuts + 2 + num_cuts**2) * 0.5
+        if not isinstance(num_cuts, int) or num_cuts < 0:
+            raise TypeError("num_cuts must be a whole number.")
+        return (num_cuts**2 + num_cuts) // 2 + 1
```
imSanko, line 157: If you're looking for simplicity and readability, using the first formula (num_cuts**2 + num_cuts) // 2 + 1 might be more intuitive because it directly represents the sum of the first num_cuts integers using the arithmetic sum formula.

On the other hand, if you prefer a more explicit representation that shows each step of the calculation, the second formula (num_cuts + 2 + num_cuts**2) * 0.5 might be preferable.

imSanko (changes requested):

geometry/geometry.py
krishna1m: Num of cuts cannot be a float value. It has to be a whole number. 
The formula I have given does not give a float value.

geometry/geometry.py
krishna1m: I have removed the `.0` at the end because the result is supposed to be a whole number, not a float.

geometry/geometry.py
krishna1m: Same reason - whole number needed, not float.

PR #11276, opened 2024-02-10
Fix a bug in segment_tree.py

### LINKED ISSUE #11232: Global variable should not using in class method
### Repository commit

9caf4784aada17dc75348f77cc8c356df503c0f3

python3.11

#nop 

### Expected behavior

Hi, guys. I come across one global variable that used in class method. I think using instance attribute would be better !
```python
    def show_data(self):
        show_list = []
        for i in range(1, N + 1):  ==> should using self.N instead of N !
            show_list += [self.query(i, i)]
        print(show_list)
```
[source](https://github.com/TheAlgorithms/Python/blob/master/data_structures/binary_tree/segment_tree.py#L101)

### Actual behavior

```python
    def show_data(self):
        show_list = []
        for i in range(1, self.N + 1):  
            show_list += [self.query(i, i)]
        print(show_list)
```

Marco-campione-github:
Changed the N to self.N in show_data in segment_tree.py
Fixes #11232

### DIFF
--- a/data_structures/binary_tree/segment_tree.py
+++ b/data_structures/binary_tree/segment_tree.py
@@ -98,7 +98,7 @@ class SegmentTree:
 
     def show_data(self):
         show_list = []
-        for i in range(1, N + 1):
+        for i in range(1, self.N + 1):
             show_list += [self.query(i, i)]
         print(show_list)
 


tianyizheng02 (approved):

tianyizheng02:
Thanks for your fix!

PR #11307, opened 2024-02-25
use format to remove '0b'

groupsvkg:
### Describe your change:

use format to remove '0b'

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/bit_manipulation/binary_and_operator.py
+++ b/bit_manipulation/binary_and_operator.py
@@ -26,7 +26,7 @@ def binary_and(a: int, b: int) -> str:
     >>> binary_and(0, 1.1)
     Traceback (most recent call last):
         ...
-    TypeError: 'float' object cannot be interpreted as an integer
+    ValueError: Unknown format code 'b' for object of type 'float'
     >>> binary_and("0", "1")
     Traceback (most recent call last):
         ...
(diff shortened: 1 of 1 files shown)


imSanko (approved):
Okay kinda

bit_manipulation/binary_and_operator.py
```diff
@@ -35,8 +35,8 @@ def binary_and(a: int, b: int) -> str:
     if a < 0 or b < 0:
         raise ValueError("the value of both inputs must be positive")
 
-    a_binary = str(bin(a))[2:]  # remove the leading "0b"
```
imSanko, line 38: Why we need to more the comments?? 
In the given line of code, the comment is explaining what the line does, which is to remove the leading "0b" from the binary representation of the integer a. Comments are useful for making the code more understandable to other developers (or to your future self) who might read the code later.

imSanko (changes requested):

bit_manipulation/binary_and_operator.py
groupsvkg: When using the format, "0b" is not included in the output, as it generates a string of 0s and 1s directly. As for the comment, would "converting to binary string" accurately reflect the operation? If so, I would be happy to implement these changes. Your feedback is valuable.

tianyizheng02: I think it's fine to remove the comments here. `format(_, "b")` is much more transparent about its purpose of converting an int to binary, at least compared to `str(bin(_))[2:]`, so if the reader knows what the `format` function does, then they know what this line of code does. Plus, the variable name `a_binary` already makes it clear that this line converts `a` to binary.

tianyizheng02 (approved):

tianyizheng02:
Thanks for your contribution!

PR #11311, opened 2024-02-27
Modified Linear Regression to work on OLS, fixes #8847

### LINKED ISSUE #8847: `machine_learning/linear_regression.py` doesn't give optimal coefficients

smruthi-sumanth:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


imSanko (approved):
Nice!

machine_learning/linear_regression.py
```diff
@@ -31,85 +31,32 @@ def collect_dataset():
...
-    sum_elem = np.sum(np.square(prod))
-    error = sum_elem / (2 * len_data)
-    return error
-
-
-def run_linear_regression(data_x, data_y):
-    """Implement Linear regression over the dataset
-    :param data_x  : contains our dataset
-    :param data_y  : contains the output (result vector)
+def run_linear_regression_ols(data_x, data_y):
```
tianyizheng02, line 34:
```suggestion
def ols_linear_regression(data_x: np.ndarray, data_y: np.ndarray) -> np.ndarray:
```
1. Shortened the function name a bit (make sure you change the name elsewhere as well)
2. Added type hints

machine_learning/linear_regression.py
```diff
@@ -31,85 +31,32 @@ def collect_dataset():
...
+    """Implement Linear regression using OLS over the dataset
```
tianyizheng02, line 35:
```suggestion
    """Implement OLS linear regression over a given dataset
```
Slight rewording for clarity

machine_learning/linear_regression.py
```diff
@@ -31,85 +31,32 @@ def collect_dataset():
...
+    data_x = np.c_[np.ones(data_x.shape[0]), data_x].astype(float)
 
-    theta = np.zeros((1, no_features))
-
-    for i in range(iterations):
-        theta = run_steep_gradient_descent(data_x, data_y, len_data, alpha, theta)
-        error = sum_of_square_error(data_x, data_y, len_data, theta)
-        print(f"At Iteration {i + 1} - Error is {error:.5f}")
+    # Use NumPy's built-in function to solve the linear regression problem
+    theta = np.linalg.inv(data_x.T.dot(data_x)).dot(data_x.T).dot(data_y)
```
tianyizheng02, line 44:
```suggestion
    theta = np.linalg.inv(data_x.T @ data_x) @ data_x.T @ data_y
```
Instead of using `.dot()` for matrix multiplication, we can use numpy's `@` operator, which does the same thing and is more readable

machine_learning/linear_regression.py
```diff
@@ -31,85 +31,32 @@ def collect_dataset():
...
-def run_linear_regression(data_x, data_y):
-    """Implement Linear regression over the dataset
-    :param data_x  : contains our dataset
-    :param data_y  : contains the output (result vector)
+def run_linear_regression_ols(data_x, data_y):
+    """Implement Linear regression using OLS over the dataset
+    :param data_x : contains our dataset
+    :param data_y : contains the output (result vector)
     :return        : feature for line of best fit (Feature vector)
     """
```
tianyizheng02, line 39: The OLS regression function needs doctests—make sure you verify the outputs of your tests with a calculator that can do linear regression (e.g., Wolfram Alpha)

tianyizheng02 (changes requested):
Good stuff! Apart from the changes that I've requested in my other comments, could you also rewrite the explanation at the top of the file to reflect your new implementation? In particular, make sure your new explanation does the following:
1. Briefly explain what linear regression is
2. Explain the OLS regression formula, $(\mathbf{X}^{\top} \mathbf{X})^{-1} \mathbf{X} \mathbf{y}$
3. Note possible issues with your implementation (i.e. inefficiency, numerical instability, etc.)
4. Cite a source that readers can refer to for further info (Wikipedia or similar is fine)

I wrote a similar explanation for weighted regression in [`machine_learning/local_weighted_learning/local_weighted_learning.py`](https://github.com/TheAlgorithms/Python/blob/edee8e644b09a21a1f70d3a59d57feed51c74004/machine_learning/local_weighted_learning/local_weighted_learning.py) a while back. If you'd like, you're welcome to use that explanation as a reference (though your explanation doesn't need to be nearly as long).

cclauss:
Closing tests_are_failing PRs to prepare for Hacktoberfest 2024

PR #11383, opened 2024-04-22
Fix ARG005 per file ignore

MaximSmolskiy:
### Describe your change:

Fix `ARG005` per file ignore

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/linear_discriminant_analysis.py
+++ b/machine_learning/linear_discriminant_analysis.py
@@ -256,7 +256,7 @@ def valid_input(
     input_type: Callable[[object], num],  # Usually float or int
     input_msg: str,
     err_msg: str,
-    condition: Callable[[num], bool] = lambda x: True,
+    condition: Callable[[num], bool] = lambda _: True,
     default: str | None = None,
 ) -> num:
     """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,6 @@ max-complexity = 17  # default: 10
 "graphs/minimum_spanning_tree_prims.py" = ["SIM114"]
 "hashes/enigma_machine.py" = ["BLE001"]
 "machine_learning/decision_tree.py" = ["SIM114"]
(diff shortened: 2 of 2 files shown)


machine_learning/linear_discriminant_analysis.py
```diff
@@ -256,7 +256,7 @@ def valid_input(
     input_type: Callable[[object], num],  # Usually float or int
     input_msg: str,
     err_msg: str,
-    condition: Callable[[num], bool] = lambda x: True,
+    condition: Callable[[num], bool] = lambda _: True,
```
imSanko, line 259: Why do we need to change the preset parameter's?

pyproject.toml
```diff
@@ -86,7 +86,6 @@ max-complexity = 17  # default: 10
 "graphs/minimum_spanning_tree_prims.py" = ["SIM114"]
 "hashes/enigma_machine.py" = ["BLE001"]
 "machine_learning/decision_tree.py" = ["SIM114"]
-"machine_learning/linear_discriminant_analysis.py" = ["ARG005"]
 "machine_learning/sequential_minimum_optimization.py" = ["SIM115"]
```
imSanko, line 89: Guess we shouldn't as to change the code !!

imSanko (changes requested):

pyproject.toml
cclauss: Why do you "guess" that?

imSanko: Is that simply great to remove a line?

imSanko (changes requested):

pyproject.toml
imSanko: Is that great while removing a line, while making it a fix !!

cclauss (approved):
Yes.  Removing the line is enabling a test (adding more tests, not modifying or removing tests) so it seems quite valid to me.

PR #11385, opened 2024-04-25
Fixes The Bug In Radix Tree. Issue #11316  

### LINKED ISSUE #11316: data_structures/trie/radix_tree.py wont really end up in 'case 1' for insert
### Repository commit

2e405f3

Python 3.11.6

n/a

### Expected behavior

```
root_node = RadixNode()
root_node.insert('fooaaa')
root_node.insert('foobbb')
root_node.insert('foo')
```
Should work., i.e should have a structure like this
```
root_node.print_tree()
- foo (leaf)
-- aaa   (leaf)
-- bbb   (leaf)
```

### Actual behavior

```
root_node.insert("foo")
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 root_node.insert("foo")

File ~/git/python/radix/radix.py:85, in RadixNode.insert(self, word)
     82 # Case 3: The node prefix is equal to the matching
     83 # Solution: We insert remaining word on the next node
     84 if remaining_prefix == "":
---> 85     self.nodes[matching_string[0]].insert(remaining_word)
     86 # Case 5: The word is greater equal to the matching
     87 # Solution: Create a node in between both nodes, change
     88 # prefixes and add the new node for the remaining word
     89 else:
     90     incoming_node.prefix = remaining_prefix

File ~/git/python/radix/radix.py:73, in RadixNode.insert(self, word)
     68     self.is_leaf = True
     70 # Case 2: The node has no edges that have a prefix to the word
     71 # Solution: We create an edge from the current node to a new one
     72 # containing the word
---> 73 elif word[0] not in self.nodes:
     74     self.nodes[word[0]] = RadixNode(prefix=word, is_leaf=True)
     76 else:

IndexError: string index out of range

```

covid-in:
### Describe your change:
Fixes  #11316 .
There is a bug in the code which doesn't handle the edge case when the insertion of empty string is done.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/trie/radix_tree.py
+++ b/data_structures/trie/radix_tree.py
@@ -198,7 +198,7 @@ class RadixNode:
             value.print_tree(height + 1)
 
 
-## write unit test for the code using unittest library with 
+## write unit test for the code using unittest library with
 ## logic similar to test_trie() function
 ## and call it from main()
 
--- a/data_structures/trie/radix_tree.py
+++ b/data_structures/trie/radix_tree.py
@@ -218,8 +218,9 @@ class TestRadixNode(unittest.TestCase):
 
     def test_trie_2(self) -> None:
         '''
-        now add a new test case which inserts 
-        foobbb, fooaaa, foo in the given order and checks for different assertions
+        now add a new test case which inserts
+        foobbb, fooaaa, foo in the given order and checks
+        for different assertions
         '''
         words = "foobbb fooaaa foo".split()
         root = RadixNode()
--- a/data_structures/trie/radix_tree.py
+++ b/data_structures/trie/radix_tree.py
@@ -6,6 +6,7 @@ trie (prefix tree) in whicheach node that is the only child is merged
 with its parent [https://en.wikipedia.org/wiki/Radix_tree]
 """
 
+
 class RadixNode:
     def __init__(self, prefix: str = "", is_leaf: bool = False) -> None:
         # Mapping from the first character of the prefix of the node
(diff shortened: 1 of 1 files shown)


imSanko (approved):

rtron-24 (approved):
I approve the changes.

hammad7 (approved):
Reviewed

data_structures/trie/radix_tree.py
```diff
@@ -1,3 +1,5 @@
+import unittest
```
cclauss, line 1: We (and the majority of the Python community) use pytest instead of unittest so all this boilerplate is not needed.

Can you please create a ___separate___ simple pull request that adds a test that proves that `word = ""` fails?  Once that is proven, we can determine how to proceed with this pull request.

data_structures/trie/radix_tree.py
```diff
@@ -191,39 +198,42 @@ def print_tree(self, height: int = 0) -> None:
...
-    root.delete("banana")
-    assert not root.find("banana")
-    assert root.find("bananas")
+## write unit test for the code using unittest library with
+## logic similar to test_trie() function
+## and call it from main()
 
-    return True
 
+class TestRadixNode(unittest.TestCase):
```
cclauss, line 206: This boilerplate adds unnecessary complexity.

data_structures/trie/radix_tree.py
rtron-24: It fails on subseunt recursive calls not on direct addition of "". Please check the issue it is trying to fix. I have added the same word insertions in Unit Test as well

data_structures/trie/radix_tree.py
rtron-24: Sure Will Remove it. But its the best way to do this

cclauss (approved):

PR #11390, opened 2024-05-01
Update segment_tree.py

### LINKED ISSUE #11232: Global variable should not using in class method

JyotikaJayani-08:
## What does this PR do?
By using self.N instead of N, we can ensure that each instance of the class has its own N attribute, and we can avoid using a global variable inside the class method.

* [ Y] Fix a bug or typo in an existing algorithm?
Yes

### Checklist:
* [Y] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [Y] This pull request is all my own work -- I have not plagiarized.
* [Y] I know that pull requests will not be merged if they fail the automated tests.
* [Y] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [Y ] All new Python files are placed inside an existing directory.
* [ Y] All filenames are in all lowercase characters with no spaces or dashes.
* [ Y] All functions and variable names follow Python naming conventions.
* [ Y] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [Y ] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [Y ] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [Y ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #11232".

### DIFF
--- a/data_structures/binary_tree/segment_tree.py
+++ b/data_structures/binary_tree/segment_tree.py
@@ -95,13 +95,12 @@ class SegmentTree:
         q1 = self.query_recursive(self.left(idx), left, mid, a, b)
         q2 = self.query_recursive(self.right(idx), mid + 1, right, a, b)
         return max(q1, q2)
-        
+
     def show_data(self):
         show_list = []
-        for i in range(1, self.N + 1):  
+        for i in range(1, self.N + 1):
             show_list += [self.query(i, i)]
         print(show_list)
-
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


tianyizheng02:
Bug was fixed in previous PR #11276

PR #11436, opened 2024-06-10
Create Image_Classification_Example_MNIST_CNN.py

dipuk0506:
Image Classification Example

### Describe your change:
* [ ] Add an algorithm?
We are adding an example image classification with a CNN. The script automatically downloads data and trains a model. There is no need for data arrangement.


dipuk0506:
Changing and submitting again.

PR #11516, opened 2024-08-13
Add option to Huffman Code a string not just a file; add doctests to Huffman file

Lecdi:
### Describe your change:

Previously the `compression/huffman.py file` provided the `huffman` function to Huffman Code the contents of a file and print the result.

I have added the `huffman_string` function (along with `parse_string`, which is analogous to `parse_file`) to Huffman Code a given string and return the result.

I have also edited the `huffman` function to allow the user to specify more options, such as how letter bitstrings should be separated and which file to print the result to. Calls to the `huffman` function with only one argument still do exactly the same thing as before.

I have also added doctests to the existing `build_tree` function.
EDIT: no longer doctests for existing `huffman` and `parse_file` functions

* [✔ ] Add an algorithm? 
* [✔ ] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
* [ ] Documentation change?
* [ ] 
(I don't know if `huffman_string` counts as a new algorithm, I suppose it does)
(I see it says avoid changing both code and tests in a single pull request, but since I was adding my own new functions with doctests, I thought it made sense to also add doctests for the existing functions)

### Checklist:
* [✔ ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [✔ ] This pull request is all my own work -- I have not plagiarized.
* [✔ ] I know that pull requests will not be merged if they fail the automated tests.
* [✔ ] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [✔ ] All new Python files are placed inside an existing directory.
* [✔ ] All filenames are in all lowercase characters with no spaces or dashes.
* [✔ ] All functions and variable names follow Python naming conventions.
* [✔ ] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [✔ ] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.


Lecdi:
It can't find the files `compression/text_data/text_original.txt` or `comprssion/text_data/text_huffman.txt`. I have tried with and without the compression prefix and neither works. Does anyone know what the root directory is for opening files or what path I should give?

Lecdi:
OK all the tests are passing now but there are no doctests on `parse_file` and `huffman`. I couldn't get opening files in a doctest to work. Maybe someone else can add those doctests in the future.

There are still doctests on the new functions `parse_string` and `huffman_string` which were the main point of the pr anyway.

There are still new doctests on existing function `build_tree` which didn't have any previously.

Issue #11320, opened 2024-03-12
Migrate to numpy 2.0


cclauss:
### Feature description

The beta release of numpy is now available so we should run the appropriate `ruff` command and ensure our algorithms are compatible... https://numpy.org/devdocs/numpy_2_0_migration_guide.html

@MaximSmolskiy

MaximSmolskiy:
@cclauss I think this rule is already checked and satisfied.

Command
```
ruff check --select NPY201
```
does not show any violations.

And in https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy I do not see any specific signs near `NPY201` rule to believe that it is not included by default in https://github.com/TheAlgorithms/Python/blob/master/pyproject.toml#L48

cclauss:
Are there other tasks required to complete the upgrade?

MaximSmolskiy:
As I understand, experiment in #11340 shows that some packages (`statsmodels`, `scikit-learn` and maybe some more) are not ready for numpy 2.0 - there are problems on `Install dependencies` stage

ramilili:
I truly wish you a good luck

quant12345:
In which folder were the problems or show how to run all tests? I want to see what the problems are. In June I did a PR in statsmodels, they had one build numpy version --- 2.1.0.dev0 and there were no failures. I don't know about the current time.

cclauss:
```
python3.12 -m venv .venv
source .venv/bin/activate
python -m pip install --upgrade pip setuptools six wheel
python -m pip install pytest-cov -r requirements.txt
pytest
```
From `.github/workflows/build.yml`

quant12345:
After installation: `pip install 'numpy==2.0.1' --force-reinstall`

I receive the following warnings:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
imageio 2.35.0 requires numpy<2.0.0, but you have numpy 2.0.1 which is incompatible.
tensorflow 2.17.0 requires numpy<2.0.0,>=1.26.0; python_version >= "3.12", but you have numpy 2.0.1 which is incompatible.
```

<details>

<summary>There are a few files where the tests fail, but there are no errors, and the following is printed:</summary>

```
platform linux -- Python 3.12.3, pytest-8.3.2, pluggy-1.5.0
rootdir: /home/quant/PycharmProjects/algorithms
configfile: pyproject.toml
plugins: cov-5.0.0
[...]

PR #11521, opened 2024-08-19
Handling scalars in tests with NEP 51 numpy >=2

quant12345:
Contributes to #11320

There are many test failures where numbers do not match when printed in the new [numpy >= 2](https://stackoverflow.com/questions/78630047/how-to-stop-numpy-floats-being-displayed-as-np-float64) (the number is printed with the type).  This is about printing scalars in numpy >= 2, here is an explanation in [numpy](https://numpy.org/devdocs/release/2.0.0-notes.html#representation-of-numpy-scalars-changed) itself.
I see three options to fix this:
1. Either convert numbers, arrays to Python instead of numpy.
2. Fix the tests themselves: 12.0 by replacing with np.float64(12.0)
3. You can put np.set_printoptions(legacy="1.25") .

I chose the first option. 
Where I was not sure what the types would be, I used [item.](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html) (will return an error if there is more than one element)
I tried to make changes to the tests themselves only as a last resort.

Example: `pytest electronics/circular_convolution.py `

FAILURES:

```
Expected:
    [10, 10, 6, 14]
Got:
    [np.int64(10), np.int64(10), np.int64(6), np.int64(14)]
```

All files passed tests locally in numpy `2.1.0` and numpy `1.26.4`. Each file was checked separately. Except for the file `linear_algebra/src/power_iteration.py` in numpy 1.26.4 there is a different number on the fourteenth digit after the period, which I think is not significant. 

**Update**

**the build fails on numpy `2.0.0`  and  `2.1.0`: Getting requirements to build wheel did not run successfully.**

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/computer_vision/haralick_descriptors.py
+++ b/computer_vision/haralick_descriptors.py
@@ -19,7 +19,7 @@ def root_mean_square_error(original: np.ndarray, reference: np.ndarray) -> float
         >>> root_mean_square_error(np.array([1, 2, 3]), np.array([6, 4, 2]))
         3.1622776601683795
     """
-    return np.sqrt(((original - reference) ** 2).mean())
+    return float(np.sqrt(((original - reference) ** 2).mean()))
 
 
 def normalize_image(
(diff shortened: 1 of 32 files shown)


quant12345:
The tests don't pass, I'll try with the original requirements.txt.

imSanko (changes requested):
What's the Reason behind adding `Float` in every other code?

quant12345:
@imSanko the reason is that in numpy >=2 numbers are printed with a type, which is why tests fail. I converted the numbers and arrays to Python instead of numpy so the tests would pass. Also in tests I saw Python lists, but inside there were numpy numbers. In this case I did this: np.array(arr).tolist(), that is, I converted the array to numpy, and then to a Python list, so as not to use list comprehension to convert each value. I updated the PR description. But build still fails. There is a tangled mess with downgrading versions.

PR #11527, opened 2024-08-25
Fix ruff

MaximSmolskiy:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


MaximSmolskiy:
@cclauss please review

cclauss:
I suspect using `max()` and `min()` is slower than the code that they replace.  It is more readable but is the slowdown worth it?

We should probably use a `noqa` to avoid this rule in `maths/find_max.py` because not using `max()` is the goal of that file.

cclauss:
4x slower...
```python
def countdown_max():
    j = -1
    for i in range(1_000_000, 0, -1):
        j = max(i, j)
    return j


def countdown_maximum():
    j = -1
    for i in range(1_000_000, 0, -1):
        if j < i:
            j = i
    return j


def countup_max():
    j = -1
    for i in range(1_000_000):
        j = max(i, j)
    return j


def countup_maximum():
    j = -1
    for i in range(1_000_000):
        if j < i:
            j = i
    return j


if __name__ == "__main__":
    from timeit import timeit

    print(f"{countdown_max() = }")
    print(f"{countdown_maximum() = }")
    print(f"{countup_max() = }")
    print(f"{countup_maximum() = }\n")

    print("countdown_max", timeit("countdown_max()", globals=globals(), number=10))
    print("countdown_maximum", timeit("countdown_maximum()", globals=globals(), number=10))
    print("countup_max", timeit("countup_max()", globals=globals(), number=10))
    print("countup_maximum", timeit("countup_maximum()", globals=globals(), number=10))
```

cclauss (approved):

PR #11587, opened 2024-09-30
Scripts for closing pull requests for Hacktoberfest

cclauss:
Each year we prepare for Hacktoberfest by clearing out old pull requests to simplify things for project maintainers.

We had 211 open pull requests; after running these scripts, we have 126.  ~40% of our Open PRs were closed.

@MaximSmolskiy Your review, please.

### DIFF
--- a/scripts/close_pull_requests_with_awaiting_changes.sh
+++ b/scripts/close_pull_requests_with_awaiting_changes.sh
@@ -16,7 +16,7 @@ echo "$prs" | jq -c '.[]' | while read -r pr; do
   # If awaiting_changes, close the pull request
   if [[ -n "$awaiting_changes" ]]; then
     echo "Closing PR #$pr_number $pr_title due to awaiting_changes label"
-    gh pr close "$pr_number" --comment "Closing awaiting_changes PRs to prepare for Hacktoberfest 2024"
+    gh pr close "$pr_number" --comment "Closing awaiting_changes PRs to prepare for Hacktoberfest"
     sleep 2
   fi
 done
--- a/scripts/close_pull_requests_with_awaiting_changes.sh
+++ b/scripts/close_pull_requests_with_awaiting_changes.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # List all open pull requests
-prs=$(gh pr list --state open --json number,title,labels --limit 200)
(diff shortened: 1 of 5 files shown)


tianyizheng02 (commented):
Shouldn't we increase the PR limit to something higher than 200? We regularly have open PR counts larger than that.

scripts/close_pull_requests_with_awaiting_changes.sh
```diff
@@ -0,0 +1,22 @@
...
+  pr_labels=$(echo "$pr" | jq -r '.labels')
+
+  # Check if the "awaiting changes" label is present
+  awaiting_changes=$(echo "$pr_labels" | jq -r '.[] | select(.name == "awaiting changes")')
+  echo "Checking PR #$pr_number $pr_title ($awaiting_changes) ($pr_labels)"
+
+  # If awaiting_changes, close the pull request
+  if [[ -n "$awaiting_changes" ]]; then
+    echo "Closing PR #$pr_number $pr_title due to awaiting_changes label"
+    gh pr close "$pr_number" --comment "Closing awaiting_changes PRs to prepare for Hacktoberfest 2024"
```
tianyizheng02, line 19:
```suggestion
    gh pr close "$pr_number" --comment "Closing awaiting_changes PRs to prepare for Hacktoberfest"
```
For future use, let's not explicitly reference the Hacktoberfest year, just so that we don't have to remember to update the scripts when we have to run them again.

Same with the other scripts

tianyizheng02 (changes requested):
Admittedly I'm not great with bash scripting, but the scripts _look_ right to me. I trust that they work as intended if these are the same scripts that you just used clean up our open PRs.

tianyizheng02 (approved):

tjgurwara99:
Just passing by here since I got assigned in this so the following is a question and maybe a suggestion... - is it worth doing this in a `workflow_dispatch` based CI action? https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch

These seem to be repeatable tasks and would be easier to do them directly from the GitHub UI for maintainers.

cclauss:
I hesitate to automate this too much because we run it rarely and closing other people's work is something we should do with some deliberation and not too much automation.

PR #11608, opened 2024-10-01
finding different diffraction parameters

Skywalkr-dev:
### Describe your change:
   **diffraction.py is a module used to calculate various parameters of single slit and double slit diffraction patterns such as 
                >>> checking if a given wavelength ,diffraction angle and slit width corresponds to the minimum intensity observed in the diffraction pattern.
                >>> checking if a given wavelength ,diffraction angle and slit width corresponds to the minimum intensity observed in the diffraction pattern
                >>> calculating the intensity of a fringe observed in a single slit diffraction pattern based on the slit width ,diffraction angle and wavelength.
                >>> calculating the intensity of a fringe observed in a double slit diffraction pattern based on the path difference and the central maximum or the maximum intensity observed,
                ** 

### Checklist:

### DIFF
--- a/physics/diffraction.py
+++ b/physics/diffraction.py
@@ -1,6 +1,6 @@
 import math
 
-def check_min_intensity(slit_width = 1,diff_angle=0, wavelength=100):
+def check_min_intensity(slit_width:int = 1,diff_angle:int =0, wavelength:int =100)->bool:
     """
     Checks for the condition of minimum intensity in a diffraction pattern.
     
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev:
Reviewed in the oldest-PR triage sweep. Thanks for the contribution — I have some correctness/scope concerns that need addressing before merge:

- `intensity_single_slit` is annotated `-> str` and returns `f"{i_coeff} I0"`, and `intensity_double_slit` takes `intensity_max: str | float = "I0"`. Mixing symbolic strings ("I0") with numeric physics makes the functions hard to test and compose; I'd suggest returning plain floats (the numeric intensity ratio) and leaving formatting to the caller.
- `check_min_intensity`/`check_max_intensity` decide the condition via float floor-equality (`n_val - math.floor(n_val) == 0`), which is fragile under floating-point rounding. A tolerance-based comparison would be more robust and easier to doctest reliably.

Could you rework toward numeric returns + tolerance checks? Happy to re-review after that.

PR #11648, opened 2024-10-01
Haversine Distance Calculation to Use Raw Latitudes 

### LINKED ISSUE #11308: Haversine distance may be using the wrong phi1 and phi2 angle
### Repository commit

2e405f397bbcefccc470f215c7ff024875ef16c5

NA

NA

### Expected behavior

First of thanks for all the algorithms and the work! 

The current (and initial) implementation of the haversine distance is using [reduced latitudes](https://github.com/TheAlgorithms/Python/blob/master/geodesy/haversine_distance.py#L40) (as one does for Lambert's formula see https://en.wikipedia.org/wiki/Geographical_distance#Lambert's_formula_for_long_lines), however, from my understanding the haversine distance should be using the raw latitudes (in radians like the longitudes are currently being handled). At least that is how it is written in https://en.wikipedia.org/wiki/Haversine_formula#Formulation .

The haversine distance from my understanding is supposed to work on a sphere, however, the notion on reduced latitude and flattening factors exist to deal with the non-spherical nature of the Earth (see https://en.wikipedia.org/wiki/Flattening). I am by no means an expert, I could very well be wrong.

Moreover, reading the wikipage and other implementations the transformation to a reduced latitude is not needed for the haversine distance, the latitude is used directly. Here are some references:
- https://en.wikipedia.org/wiki/Haversine_formula#Formulation (also linked in the implementations comments)
- https://github.com/mapado/haversine/blob/main/haversine/haversine.py#L183 from https://pypi.org/project/haversine/

additional implementations:
- https://gist.github.com/vananth22/888ed9a22105670e7a4092bdcf0d72e4
- https://stackoverflow.com/a/4913653

I think most likely there was some confusion with Lambert's formula. 

I am happy to make a small PR, if there is an agreement with my observation.

### Actual behavior

Use the right angles and the results should slightly be different. I did some test, it is not that significant.

Maanit491:
### Describe your change:

This pull request fixes the Haversine distance calculation in the haversine_distance.py file. The implementation previously used reduced latitudes, which can introduce inaccuracies. This change utilizes raw latitudes directly, ensuring that the calculation aligns with standard practices for spherical distance measurement.

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #11308 ".


Maanit491:
I have resolved the bug. Please review my PR.

Rudraprakash054 (approved):

cclauss:
Please rebase to resolve git conflicts and see https://github.com/TheAlgorithms/Python/issues/11308#issuecomment-5617886145

Issue #8067, opened 2023-01-01
Implement XGBoost classification and regression algorithms


tianyizheng02:
### Feature description

[machine_learning/xgboost_classifier.py](https://github.com/TheAlgorithms/Python/blob/master/machine_learning/xgboost_classifier.py) and [machine_learning/xgboost_regressor.py](https://github.com/TheAlgorithms/Python/blob/master/machine_learning/xgboost_regressor.py) are how-tos since they both just use an existing library for the actual ML algorithms.

My understanding is that #7106 and #7107 were merged (not without difficulty) and the author was warned not to contribute such how-tos in the future. However, I think these algorithms should still be implemented at some point if the files are to remain in the repo, so I thought I should open an issue to bring some attention to it.

If anyone wants to implement either of these two algorithms (as in not relying on an existing library for the bulk of the algorithm), feel free to just open a PR—no need to request an assignment.

tianyizheng02:
> Given that there are already well-established libraries for XGBoost, such as the XGBoost Python package and LightGBM, it may not be practical to invest resources into implementing XGBoost in scikit-learn without relying on an existing library.

@rohan472000 Implementing algorithms is the whole purpose of this repo. From the contributing guidelines:
> Algorithms in this repo should not be how-to examples for existing Python packages.

Importing an XGBoost class from sklearn would absolutely be a how-to example. If that requires significant time and expertise in ML, then so be it. After all, the algorithm implementations in this repo are meant for educational purposes, and figuring out how to implement an algorithm is itself an educational experience.

Dhairya-A-Mehra:
@tianyizheng02 I have submitted a Pull Request for these changes as part of Hacktoberfest 2024. Kindly review it at your convenience.

LeonxLJX:
I'd like to take this one — I'll follow up with a PR. (claiming via @LeonxLJX)

LeonxLJX:
Hi! I'd like to implement XGBoost classification and regression algorithms. Plan: implement the core algorithm, add tests and docs. May I be assigned?

cclauss:
https://github.com/TheAlgorithms/Python/blob/master/docs/hacktober_2026_prep.md#triage-recommendations

PR #11654, opened 2024-10-02
Added Catboost Algorithm in Machine Learning

mathangpeddi:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/catboost_classifier.py
+++ b/machine_learning/catboost_classifier.py
@@ -1,10 +1,10 @@
 # Catboost Classifier Example
 import numpy as np
+from catboost import CatBoostClassifier
 from matplotlib import pyplot as plt
 from sklearn.datasets import load_iris
 from sklearn.metrics import ConfusionMatrixDisplay
 from sklearn.model_selection import train_test_split
-from catboost import CatBoostClassifier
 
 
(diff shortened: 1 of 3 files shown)


tianyizheng02:
Please read our contributing guidelines:
> Algorithms in this repo should not be how-to examples for existing Python packages. Instead, they should perform internal calculations or manipulations to convert input values into different output values. Those calculations or manipulations can use data types, classes, or functions of existing Python packages but each algorithm in this repo should add unique value.

Unfortunately, your code fundamentally relies on an existing Python package for the actual ML algorithm. It imports the catboost algorithm from a package rather than _implementing_ the catboost algorithm itself. If you wish to contribute this ML algorithm to our repo, please make sure to implement the algorithm yourself.

mathangpeddi:
Thank you for the reply! We already have xgboost algorithm being imported, so I thought we can also add catboost algorithm and that's why I have created the PR. Thank you for the confirmation.

tianyizheng02:
Understandable. You're right that our XGBoost implementations are also how-tos that import an existing package, but they need to be rewritten as well for precisely that same reason and thus shouldn't serve as an example (see issue #8067).

PR #11662, opened 2024-10-02
feat: add test coverage command in the CONTRIBUTING.md file

zeelrupapara:
### Describe your change:
- add run testcase and see code coverage command in the CONTRIBUTING.md file

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -146,7 +146,7 @@ We want your work to be readable by others; therefore, we encourage you to note 
   ```
 
 
-  Run all tests and see the covrage of the code by following command
+  Run all tests and see the coverage of the code by following command
   ```bash
   pytest \
           --ignore=quantum/q_fourier_transform.py \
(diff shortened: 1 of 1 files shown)


tianyizheng02 (commented):
I think this is a good idea, but IMO it would be more helpful to contributors if the command was for getting the coverage for a single file. After all, it's more likely that a contributor will want to see the coverage for the file that they're working on, rather than the coverage for the entire repo.

zeelrupapara:
@tianyizheng02 You're right, but in some cases, I want to check which files have the least test coverage so I can contribute by writing those files or functions testcases.

tianyizheng02:
> I want to check which files have the least test coverage so I can contribute by writing those files or functions testcases.

@zeelrupapara That makes sense. In your case, it may be best to just run the `pytest` command in `.github/workflows/build.yml`.

I'm hesitant to add the `pytest` command from `build.yml` directly into `CONTRIBUTING.md` because it would become outdated somewhat quickly. As you can see in the file, we sometimes have to add files to be ignored with `--ignore=...`, and it's likely for the best that contributors refer to `build.yml` directly so that they get the most up-to-date command.

PR #11668, opened 2024-10-02
Improve comments, add doctests for kahns_algorithm_topo.py

Hardvan:
The improvements to the code include adding a more detailed **docstring** that explains the purpose, **parameters,** return values, and the use of **Kahn's Algorithm** for topological sorting. The docstring now includes doctests, providing example usage and **handling edge cases** such as graphs with cycles. Inline comments clarify key steps like indegree calculation, queue initialization, and the BFS process. Additionally, explicit handling and explanation of cycle detection were added, making the code more readable and understandable while ensuring that it is easy to validate using the provided test cases.

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/graphs/kahns_algorithm_topo.py
+++ b/graphs/kahns_algorithm_topo.py
@@ -1,5 +1,4 @@
-def topological_sort(graph: dict[int, list[int]]
-                     ) -> list[int] | None:
+def topological_sort(graph: dict[int, list[int]]) -> list[int] | None:
     """
     Perform topological sorting of a Directed Acyclic Graph (DAG)
     using Kahn's Algorithm via Breadth-First Search (BFS).
@@ -61,4 +60,5 @@ def topological_sort(graph: dict[int, list[int]]
 
 if __name__ == "__main__":
     import doctest
+
     doctest.testmod()
--- a/graphs/kahns_algorithm_topo.py
+++ b/graphs/kahns_algorithm_topo.py
@@ -2,12 +2,10 @@ def topological_sort(graph: dict[int, list[int]]
                      ) -> list[int] | None:
     """
     Perform topological sorting of a Directed Acyclic Graph (DAG)
-    using Kahn's Algorithm.
+    using Kahn's Algorithm via Breadth-First Search (BFS).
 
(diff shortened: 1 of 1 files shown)


graphs/kahns_algorithm_topo.py
```diff
@@ -1,36 +1,64 @@
...
+    Cycle exists
     """
+
+    # Initialize indegree array
     indegree = [0] * len(graph)
     queue = []
-    topo = []
-    cnt = 0
+    topo = []  # topological order of vertices
+    cnt = 0  # no. of vertices processed
```
tianyizheng02, line 31: Instead of adding comments to clarify the meaning of the variables, could you rename the variables instead? The variable names should be self-documenting, so that they clarify their own meanings.

graphs/kahns_algorithm_topo.py
```diff
@@ -1,36 +1,64 @@
...
-                queue.append(x)
+
+        # Traverse neighbors
+        for neighbor in graph[vertex]:
+            indegree[neighbor] -= 1
+            if indegree[neighbor] == 0:
+                queue.append(neighbor)
 
     if cnt != len(graph):
         print("Cycle exists")
```
tianyizheng02, line 56:
```suggestion
```
Let's not have the algorithm function print. As stated in our contributing guidelines, algorithmic functions ideally should return their outputs, not print. Plus, there's no need for the print statement now that the function returns `None`.

tianyizheng02 (changes requested):

graphs/kahns_algorithm_topo.py
Hardvan: Sure! I renamed the variables as follows:
`topo` -> `topo_order`
`cnt` -> `processed_vertices_count`

graphs/kahns_algorithm_topo.py
Hardvan: Removed the print statement.

tianyizheng02 (approved):

Hardvan:
Thanks!

PR #11671, opened 2024-10-02
Implementation of Density-based spatial clustering of applications with noise (DBSCAN) ML Algorithm

tkgowtham:
### Describe your change:
Implementation of Density-based spatial clustering of applications with noise (DBSCAN) Algorithm using Python from Scratch

How it works : 
Take every point and see how many other points are close to it by e (epsilon) radius distance. If that point has covered minimum of number of points (minPts), then it will be considered as Core else Noise. An Noise can also be classified as Border if it falls into the area covered by the core point.

Reference Website : https://en.wikipedia.org/wiki/DBSCAN
Reference YouTube Video : https://youtu.be/-p354tQsKrs?si=TkI9nrnAiqDHyjR-

* [x] Add an algorithm?

### Checklist:


tkgowtham:
@tianyizheng02 Can you please review this PR and let me know if any changes in needed. Thanks in advance.

cclauss:

priya-sundaram-dev (commented):
Thanks for the contribution, @tkgowtham 🙂 I pulled the branch and ran the doctests locally with the `Agg` backend — **all 12 pass**, and the DBSCAN core/border/noise classification looks correct (each point counts itself in its ε-neighborhood, so the `>= minpts` core test is right).

A few suggestions, mostly small — none are hard blockers, but the type-hint one is worth fixing since the repo type-checks:

1. **`radius` type hint.** It's annotated `radius: int`, but every example passes a float (`1.9`, `2.5`). Please change it to `radius: float`.

2. **Sentinel default for `file`.** Using the string `"None"` as a sentinel (`file: str = "None"` + `if file != "None"`) is fragile — a real file literally named `None` would break it, and the attribute can actually hold either a `str` path or a `tuple[dict, ...]`. Prefer:
   ```python
   file: str | None = None
   ...
   self.file = file if file is not None else (...)
   ```

3. **`isinstance` over `type() is`.** `if type(self.file) is str:` reads better as `if isinstance(self.file, str):` (and satisfies linters).

4. **Naming.** `e` and `dict1` are hard to follow — `eps`/`radius` and `neighbors` would make `perform_dbscan` much clearer for readers learning the algorithm.

5. **Minor / optional:** standard DBSCAN uses `dist <= eps` (inclusive) rather than `dist < eps`. Your doctests are internally consistent either way, so this is just a note. Also `perform_dbscan` is O(n²) pairwise, which is totally fine for a teaching implementation — maybe add a one-line comment saying so.

Nice, well-documented submission overall. Fix the `radius: float` hint and the `isinstance` line and I think it's in good shape. 👍

PR #11725, opened 2024-10-04
Added complexities

IAteNoodles:
### Describe your change:
Added the various complexities for the quick sort algorithm

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:


sorts/quick_sort.py
```diff
@@ -26,6 +26,15 @@ def quick_sort(collection: list) -> list:
...
     [-45, -2, 0, 5]
+
+    Time Complexity:
+    Best Case: O(n log n)
+    Average case: O(n log n)
+    Worst Case: O(n^2)
+
+    Space Complexity:
+    Best Case: O(log n)
+    Worst Case: O(n)
```
tianyizheng02, line 37: I'm not sure if this space complexity analysis is accurate for this specific implementation. From [Wikipedia](https://en.wikipedia.org/wiki/Quicksort#Space_complexity), it seems like the $O(\log n)$ space complexity may only hold for a quick sort implementation that performs partitioning in place (though I may be wrong here).

IAteNoodles: Yes, the best case is for in place it seems. I rechecked.
Since in place is the best way, should I implement with in-place method?

IAteNoodles: #11752 In here I have changed the implementation

PR #11771, opened 2024-10-05
Add boundary checking to optimal bst

### LINKED ISSUE #11724: Optimize and Test Optimal Binary Search Tree Implementation for  optimal_binary_search_tree.py
The current implementation of the optimal binary search tree algorithm works well, but there are a few areas for improvement and further testing.

ivnvxd:
### Describe your change:

The changes address potential issues with small intervals and overlapping boundaries in the current implementation of Knuth's optimization of Optimal BST algorithm. 

- Improved boundary checking in the `find_optimal_binary_search_tree` function
- Added error handling for potential empty ranges with fallback to full range
- Updated main loop to use new boundaries

Fixes #11724

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/dynamic_programming/optimal_binary_search_tree.py
+++ b/dynamic_programming/optimal_binary_search_tree.py
@@ -115,9 +115,14 @@ def find_optimal_binary_search_tree(nodes):
             dp[i][j] = sys.maxsize  # set the value to "infinity"
             total[i][j] = total[i][j - 1] + freqs[j]
 
-            # Apply Knuth's optimization
-            # Loop without optimization: for r in range(i, j + 1):
-            for r in range(root[i][j - 1], root[i + 1][j] + 1):  # r is a temporal root
+            # Apply Knuth's optimization with boundary checking
+            r_start = max(i, root[i][j - 1] if j > i else i)
+            r_end = min(j, root[i + 1][j] if i < j else j)
+
+            if r_start > r_end:
+                r_start, r_end = i, j  # fall back to the full range
+
+            for r in range(r_start, r_end + 1):
                 left = dp[i][r - 1] if r != i else 0  # optimal cost for left subtree
                 right = dp[r + 1][j] if r != j else 0  # optimal cost for right subtree
                 cost = left + total[i][j] + right


manuelaidos123 (commented):
Tryt to Simplify the boundary checks for r_start and r_end, by adding  additional checks to prevent IndexError, and replacing print statements with a logging mechanism, and configured logging for better debugging.

ivnvxd:
@manuelaidos123 Simplified boundary checks in Knuth's optimization for Optimal BST. Replaced the explicit fallback logic with direct min/max operations to ensure valid array bounds.
I think it would be better to make the logging configuration in a separate PR.

manuelaidos123 (approved):

manuelaidos123 (approved):

priya-sundaram-dev (commented):
I applied the branch and the doctests still pass (5/5), so this is safe as a defensive change. My question before approving is whether it fixes an *observed* bug or is purely precautionary.

Knuth's optimization relies on the monotonicity invariant `root[i][j-1] <= root[i][j] <= root[i+1][j]`, which (given the bottom-up DP fill order here) already keeps the temporal root `r` inside `[i, j]`. So under correct execution the extra `max(i, min(...))` / `min(j, max(...))` clamps should never actually clip anything, and the `r > i` / `r < j` guards are equivalent to the old `r != i` / `r != j`.

If you hit a concrete input where the old code went out of bounds or produced a wrong cost, could you add it as a doctest? That would (a) justify the change and (b) guard against regressions. If it is purely defensive with no failing case, a one-line comment saying so — plus dropping the now-redundant `r_start/r_end if j > i else ...` branches, since the loop body is already skipped when `i == j` — would keep the hot loop lean. Happy to approve either way once the intent is documented.

cclauss:
@priya-sundaram-dev Can you please add "solution" comments to this PR so we can get it reviewed and merged?

priya-sundaram-dev:
I reviewed the change against `dynamic_programming/optimal_binary_search_tree.py` and it looks correct and safe to merge:

- **The boundary clamping is sound.** Knuth's optimization guarantees `root[i][j-1] <= root[i][j] <= root[i+1][j]`, so the original range endpoints are already valid in the normal path; the added `max(i, min(r_start, j))` / `min(j, max(r_end, i))` clamps are purely defensive and cannot change the result on well-formed input. The `j > i` / `i < j` guards correctly handle the diagonal base case.
- `r != i`/`r != j` → `r > i`/`r < j` is equivalent given `r ∈ [i, j]`, and reads better.
- The comment fix `sum[i][j]` → `total[i][j]` is a good catch — it now matches the actual variable name `total`.
- Existing doctests still pass locally (`python3 -m doctest -v`: 5 passed, 0 failed).

On the "solution comments" request: the function is already reasonably documented, but the one thing that would make this fully self-explanatory for a reviewer is a one-line comment above the new clamp block stating *why* it exists — e.g. `# Clamp the Knuth search window to [i, j] to guard against out-of-range roots on degenerate sub-problems`. @ivnvxd, if you add that (it's your branch), I think this is merge-ready. LGTM otherwise.

cclauss (approved):

PR #11772, opened 2024-10-05
Update ruff usage example in CONTRIBUTING.md

ivnvxd:
### Describe your change:

Updated the example command for running ruff to use `ruff check .` instead of just `ruff .`, as the latter syntax has been removed. This ensures contributors use the correct command when running ruff locally before submitting pull requests.

The change addresses the following error that users may encounter with the outdated command:
```
error: `ruff <path>` has been removed. Use `ruff check <path>` instead.
```

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,7 @@ We want your work to be readable by others; therefore, we encourage you to note 
 
   ```bash
   python3 -m pip install ruff  # only required the first time
-  ruff check .
+  ruff check
   ```
 
 - Original code submission require docstrings or comments to describe your work.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -96,7 +96,7 @@ We want your work to be readable by others; therefore, we encourage you to note
 
   ```bash
   python3 -m pip install ruff  # only required the first time
-  ruff .
+  ruff check .
```
tianyizheng02, line 99:
```suggestion
  ruff check
```
Looks like we don't need to specify the current directory anymore, since ruff will run in the current directory by default: https://docs.astral.sh/ruff/tutorial/#getting-started

tianyizheng02 (changes requested):

tianyizheng02 (approved):

tianyizheng02:
Thanks for your catch!

PR #11774, opened 2024-10-05
feat: Implemented Morris Inorder traversal for Binary tree

kuntal0901:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


data_structures/binary_tree/morris_inorder_traversal.py
```diff
@@ -0,0 +1,168 @@
...
+    Class representing a node in a binary tree.
+
+    Attributes:
+    -----------
+    value : int
+        The value stored at the node.
+    left : TreeNode
+        Pointer to the left child node (default is None).
+    right : TreeNode
+        Pointer to the right child node (default is None).
```
tianyizheng02, line 20:
```suggestion
    value: The value stored at the node.
    left: Pointer to the left child node (default is None).
    right: Pointer to the right child node (default is None).
```
By convention in this repo, we prefer not to include the types of the function params in the docstrings. This is because the functions themselves should already have type hints, and adding the types in the docstring means that, if we ever change the variables' types, we'd need to update the type hints in two separate places.

Could you make similar changes in the other docstrings as well?

data_structures/binary_tree/morris_inorder_traversal.py
```diff
@@ -0,0 +1,168 @@
...
+                    print the current node
+                    point the current node to its right child.
+
+        Returns:
+        --------
+        List[int]:
+            A list of integers representing the inorder traversal.
+
+
+
```
tianyizheng02, line 144:
```suggestion
```

tianyizheng02 (changes requested):
Some minor nitpicks, but otherwise LGTM

PR #11826, opened 2024-10-06
Streaming moving average

mjvakili:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/financial/streaming_moving_average.py
+++ b/financial/streaming_moving_average.py
@@ -1,0 +1,87 @@
+"""This module contains a class that
+calculates the moving average in real-time."""
+
+from collections import deque
+
+
+class StreamingMovingAverage:
+    """Streaming Moving Average calculator that updates
+    the average in real-time.
+
+    Attributes
+    ----------
+    window_size : int
+        The size of the moving average window.
+
+    Methods
+    -------
+    update(new_value: float) -> None
+        Update the moving average with a new data point.
+    mov_avg() -> float
+        Return the current moving average value.
+
+    Raises
+    ------
+    ValueError
+        If the window size is less than 1.
+
+
+    Examples
+    --------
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev (commented):
Nice, clean class and the doctests pass locally (both examples give 5.0 and 2.5). Two suggestions to make it live up to the "streaming" name and be a bit friendlier:

1. **Make the update genuinely O(1).** Right now `update()` calls `sum(self.window)` on every call, so each update is O(window_size) — the same cost as a batch recompute. A true streaming SMA keeps a running sum and adjusts it as the deque evicts:

```python
def update(self, new_value: float) -> None:
    evicted = self.window[0] if len(self.window) == self.window_size else 0.0
    self.window.append(new_value)          # deque(maxlen=...) drops the oldest
    self._running_sum += new_value - evicted
    self._mov_avg = self._running_sum / len(self.window)
```

That keeps each update constant-time regardless of window size, which is the whole point of the streaming variant.

2. **Accept ints too.** `isinstance(new_value, float)` rejects `update(5)`, which is a surprising restriction for a numeric stream (and the message reads "must be either float."). Consider `isinstance(new_value, (int, float))` or `numbers.Real`, and reword the message.

Neither is blocking — the algorithm is correct — but (1) in particular would make it a distinct, worthwhile addition next to `simple_moving_average.py`.

financial/streaming_moving_average.py
```diff
@@ -0,0 +1,91 @@
...
+    Methods
+    -------
+    update(new_value: float) -> None
+        Update the moving average with a new data point.
+    mov_avg() -> float
+        Return the current moving average value.
+
+    Raises
+    ------
+    ValueError
```
cclauss, line 30: Please add a doctest the demonstrates this.

PR #11842, opened 2024-10-07
Create lz78.py

### LINKED ISSUE #11837: LZ78 compression algorithm is missing
### Feature description

LZ78 compression algorithm is not implemented in compression algorithms.

Ahmed99125:
### Describe your change:
Added LZ78 compression algorithm.
Fixes #11837

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/compression/lz78.py
+++ b/compression/lz78.py
@@ -115,5 +115,5 @@ if __name__ == "__main__":
     # Example
     text = "aacaacabcabaaa"
     tokens = lz78_compressor.compress(text)
-    decompressedtext = lz78_compressor.decompress(tokens)
-    assert decompressedtext == text, "Invalid result."
+    decompressed_text = lz78_compressor.decompress(tokens)
+    assert decompressed_text == text, "Invalid result."
--- a/compression/lz78.py
+++ b/compression/lz78.py
@@ -51,6 +51,14 @@ class LZ78Compressor:
           '[(0, a), (0, b), (1, b), (0, c), (2, a), (5, b), (1, a)]'
           >>> str(lz78_compressor.compress("aacaacabcabaaac"))
           '[(0, a), (1, c), (1, a), (0, c), (1, b), (4, a), (0, b), (3, a)]'
+          >>> lz78_compressor.compress("")
+          []
+          >>> lz78_compressor.compress([])
+          []
+          >>> lz78_compressor.compress({})
+          []
+          >>> len("ababc") >= len(compressor.compress("ababc"))
+          True
         """
 
         phrase_dict = {}
--- a/compression/lz78.py
+++ b/compression/lz78.py
@@ -12,7 +12,7 @@ __author__ = "Ahmed Tamer"
 @dataclass
 class Token:
     """
-    Dataclass representing pair called token consisting of the dictionary index 
+    Dataclass representing pair called token consisting of the dictionary index
     and a single character that follows the phrase in the dictionary.
     """
 
@@ -81,8 +81,8 @@ class LZ78Compressor:
 
         Tests:
           >>> lz78_compressor = LZ78Compressor()
-          >>> lz78_compressor.decompress([Token(0, 'c'), Token(0, 'a'), Token(0, 'b'), 
-          ... Token(0, 'r'), Token(2, 'c'), Token(2, 'd'), Token(2, 'b'), Token(4, 'a'), 
+          >>> lz78_compressor.decompress([Token(0, 'c'), Token(0, 'a'), Token(0, 'b'),
+          ... Token(0, 'r'), Token(2, 'c'), Token(2, 'd'), Token(2, 'b'), Token(4, 'a'),
           ... Token(4, 'r'), Token(2, 'r'), Token(8, 'd')])
           'cabracadabrarrarrad'
           >>> lz78_compressor.decompress([Token(0, 'a'), Token(0, 'b'), Token(1, 'b'),
--- a/compression/lz78.py
+++ b/compression/lz78.py
@@ -8,68 +8,68 @@ from dataclasses import dataclass
 __version__ = "1.0"
 __author__ = "Ahmed Tamer"
 
+
 @dataclass
 class Token:
-  """
+    """
     Dataclass representing pair called token consisting of the dictionary index and a single character that follows the phrase in the dictionary.
-  """
-  
-  index: int
-  char: str
-  
-  def __repr__(self) -> str:
     """
-      >>> token = Token(1, "c")
-      >>> repr(token)
-      '(1, c)'
-      >>> str(token)
-      '(1, c)'
-    """
-    return f"({self.index}, {self.char})"
-  
+
+    index: int
+    char: str
+
+    def __repr__(self) -> str:
+        """
+        >>> token = Token(1, "c")
+        >>> repr(token)
+        '(1, c)'
+        >>> str(token)
+        '(1, c)'
+        """
+        return f"({self.index}, {self.char})"
+
 
 class LZ78Compressor:
-  """
+    """
     Class containing compress and decompress methods using LZ78 compression algorithm.
-  """
-  
-  def compress(self, text: str) -> list[Token]:
-    """
-      Compress the given string text using LZ78 compression algorithm.
-
-      Args:
-        text: string to be compressed
-
-      Returns:
-        output: the compressed text as a list of Tokens
-      
-      Tests:
-        >>> lz78_compressor = LZ78Compressor()
-        >>> str(lz78_compressor.compress("ababcbababaa"))
-        '[(0, a), (0, b), (1, b), (0, c), (2, a), (5, b), (1, a)]'
-        >>> str(lz78_compressor.compress("aacaacabcabaaac"))
-        '[(0, a), (1, c), (1, a), (0, c), (1, b), (4, a), (0, b), (3, a)]'
     """
 
-    phrase_dict = {}
-    tokens = []
-    code = 1
-    phrase = ''
-    for char in text:
-      phrase += char
-      if phrase not in phrase_dict:
-        phrase_dict[phrase] = str(code)
-        if len(phrase) == 1:
-          tokens.append(Token('0', phrase))
-        else:
-          tokens.append(Token(phrase_dict[phrase[:-1]], phrase[-1]))
-        code += 1
-        phrase = ''
-    return tokens
-  
-  
-  def decompress(self, tokens: list[Token]) -> str:
-    """
+    def compress(self, text: str) -> list[Token]:
+        """
+        Compress the given string text using LZ78 compression algorithm.
+
+        Args:
+          text: string to be compressed
+
+        Returns:
+          output: the compressed text as a list of Tokens
+
+        Tests:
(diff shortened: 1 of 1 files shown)


compression/lz78.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Returns:
+          output: the compressed text as a list of Tokens
+
+        Tests:
+          >>> lz78_compressor = LZ78Compressor()
+          >>> str(lz78_compressor.compress("ababcbababaa"))
+          '[(0, a), (0, b), (1, b), (0, c), (2, a), (5, b), (1, a)]'
+          >>> str(lz78_compressor.compress("aacaacabcabaaac"))
+          '[(0, a), (1, c), (1, a), (0, c), (1, b), (4, a), (0, b), (3, a)]'
```
cclauss, line 53: How about several test that prove that `len(s) >= LZ78Compressor().compress(s)` ?

Also, we need a test for `LZ78Compressor().compress("")` plus text=0, text=[], text={}

compression/lz78.py
```diff
@@ -0,0 +1,119 @@
...
+
+    testmod()
+
+    lz78_compressor = LZ78Compressor()
+
+    # Example
+    text = "aacaacabcabaaa"
+    tokens = lz78_compressor.compress(text)
+    decompressedtext = lz78_compressor.decompress(tokens)
+    assert decompressedtext == text, "Invalid result."
```
cclauss, line 119:
```suggestion
    decompressed_text = lz78_compressor.decompress(tokens)
    assert decompressed_text == text, "Invalid result."
```

Ahmed99125:
@cclauss please review

compression/lz78.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Args:
+          text: string to be compressed
+
+        Returns:
+          output: the compressed text as a list of Tokens
+
+        Tests:
+          >>> lz78_compressor = LZ78Compressor()
+          >>> str(lz78_compressor.compress("ababcbababaa"))
```
cclauss, line 50: Let's remove the `str()` from all these tests.

compression/lz78.py
```diff
@@ -0,0 +1,127 @@
...
+          '[(0, a), (0, b), (1, b), (0, c), (2, a), (5, b), (1, a)]'
+          >>> str(lz78_compressor.compress("aacaacabcabaaac"))
+          '[(0, a), (1, c), (1, a), (0, c), (1, b), (4, a), (0, b), (3, a)]'
+          >>> str(lz78_compressor.compress(""))
+          '[]'
+          >>> str(lz78_compressor.compress([]))
```
cclauss, line 56: The [] and {} tests should raise an Exception to warn the user not to provide garbage input.

compression/lz78.py
```diff
@@ -0,0 +1,127 @@
...
+          '[]'
+          >>> str(lz78_compressor.compress({}))
+          '[]'
+          >>> len("ababc") >= len(lz78_compressor.compress("ababc"))
```
cclauss, line 60:
```suggestion
          >>> all(len(s) >= len(lz78_compressor.compress(x)) for s in (
          ...     "", "AA", "AB", "AAA", "ABC", "ABCDEFGH"))
```

Ahmed99125:
@cclauss raised an exception when the input is not string and added the for loop in the last test case.
Please review.

Ahmed99125:
Can anyone review my work?

cclauss:
@priya-sundaram-dev, your review, please.  Is this really how the lz78 algorithm works?  Is it efficient?

priya-sundaram-dev (commented):
Reviewed @Ahmed99125's `compression/lz78.py`. Nice, readable structure — the `Token` dataclass and the greedy dictionary build read well. Two questions were asked, so I'll take them in turn.

### Is this really how LZ78 works?
The *core* is correct: extend the current phrase while it stays in the dictionary, and when `phrase` first falls out of the dictionary emit `(index(phrase[:-1]), phrase[-1])` and add the new phrase. That's textbook LZ78 (equivalent to the classic `w`/`w+c` loop).

But there's a **correctness bug: the compressor is lossy** whenever the input ends on a phrase that's already in the dictionary — i.e. when the final `phrase` never triggers the "not in dict" branch, so it's never emitted. The trailing characters are silently dropped. This isn't a rare corner case; it hits very ordinary inputs:

```
compress → decompress round-trip:
  "aa"               -> "a"                (lost 1 char)
  "mississippi"      -> "mississipp"       (lost the final "i")
  "banana"           -> "banan"
  "aacaacabcabaaac"  -> "aacaacabcabaaa"   (this file's own compress() doctest input — the final "c" is lost)
```

The last one is worth calling out: the `compress` doctest uses `"aacaacabcabaaac"` and the `decompress` doctest expects `"aacaacabcabaaa"`, so the two doctests are internally consistent but together they *demonstrate the data loss* rather than catch it. The `__main__` example sidesteps it by using the already-truncated `"aacaacabcabaaa"`, which happens to end on a phrase boundary.

Root cause: after the loop, a non-empty `phrase` is a repeat of an existing dictionary entry with no following character, and nothing flushes it. A one-line fix restores losslessness (a token whose char is empty — `decompress` already appends `phrase_dict[index] + ""` correctly, so no change needed there):

```python
        for char in text:
            ...
        if phrase:  # flush the trailing phrase (an existing entry, no new char)
            tokens.append(Token(int(phrase_dict[phrase]), ""))
        return tokens
```

I ran this fix against the examples above plus a 2000-case random fuzz over `decompress(compress(s)) == s` and it round-trips cleanly on all of them.

I'd also suggest the doctest carry a **round-trip property** rather than only `all(len(s) >= len(compress(s)) ...)` — the current check asserts the output isn't longer than the input, which is exactly what stayed green while data was being dropped. Something like:

```python
>>> all(lz78_compressor.decompress(lz78_compressor.compress(s)) == s
...     for s in ("", "a", "aa", "mississippi", "banana", "ababcbababaa"))
True
```

would have failed loudly on the bug and will guard against regressions.

### Is it efficient?
For an educational implementation, the dict-of-phrases approach is idiomatic and fine. Two small notes, neither blocking:

- Dictionary codes are stored as `str(code)` and converted back with `int(...)` on every emit. Keeping them as plain `int` removes the round-tripping and a class of stringly-typed bugs.
- The scalable data structure for LZ78 is a **trie** (each node = a dictionary entry), which gives O(1) amortized work per input character. The current approach rebuilds growing string keys (`phrase += char`) and hashes them, so long repeated runs cost O(L) per lookup. Totally acceptable for this repo's teaching goals — worth a one-line comment noting the trie as the production structure, but not something I'd require here.

Net: solid and close, but the lossy trailing-phrase case should be fixed (plus the round-trip doctest) before merge, since a compressor that can't reproduce its input is the one thing it must not do. Thanks @Ahmed99125!

priya-sundaram-dev:
@cclauss I reviewed it. Two answers:

**Is it really LZ78?** Yes — the structure is textbook LZ78: a phrase dictionary seeded with the empty string at index 0, greedily extending the current phrase until it's unseen, then emitting a `(prefix_index, next_char)` token and adding the new phrase. `compress`/`decompress` mirror each other correctly. So the *shape* is right.

**Is it correct?** No — there's a real data-loss bug, and the PR's own doctests quietly document it. Look at the decompress doctest:

```
>>> lz78_compressor.decompress([Token(0,'a'), Token(1,'c'), ...])
'aacaacabcabaaa'
```

but the matching compress doctest feeds `"aacaacabcabaaac"` — note the trailing **`c`** that never comes back. I ran `decompress(compress(s))` for a handful of inputs:

```
banana          -> banan     (lost 'a')
aaaa            -> aaa        (lost 'a')
aacaacabcabaaac -> aacaacabcabaaa  (lost 'c')
```

This is the classic LZ78 **dangling-final-phrase** case: when the input ends on a phrase that's *already* in the dictionary, the loop's `if phrase not in phrase_dict` never fires for it, so it's never emitted and is silently dropped. A compressor that loses the tail of `"banana"` isn't shippable.

The fix is a one-liner flush after the loop — emit a terminal token for the leftover phrase (it's guaranteed to be in the dict, so its char slot is empty):

```python
    ...
    return tokens  # <- replace with:
    if phrase:
        tokens.append(Token(int(phrase_dict[phrase]), ""))
    return tokens
```

`decompress` already handles it unchanged (`phrase_dict[idx] + ""`). With that, every round-trip I tried (`banana`, `aaaa`, `""`, `"a"`, the two doctest strings) is lossless. I'd ask the author to (a) add that flush and (b) add a property doctest that would have caught this:

```
>>> c = LZ78Compressor()
>>> all(c.decompress(c.compress(s)) == s
...     for s in ("", "a", "banana", "aaaa", "aacaacabcabaaac"))
True
```

**Is it efficient?** Fine for the repo's teaching purpose, with two caveats worth a nudge: (1) it's the hash-map-of-full-phrase-strings variant rather than a trie, so hashing the growing `phrase` makes it ~O(n·√n) rather than the O(n) a trie gives — acceptable here, not worth rewriting. (2) Dictionary values are stored as `str(code)` and converted back with `int(...)` on every lookup; storing plain `int`s (and using `{0: ""}` in decompress) is simpler and avoids the round-trip. Minor, but I'd mention it.

Net: good algorithm, right idea, **but block on the correctness bug** — the round-trip must be lossless before merge.

cclauss:
Closing because of the correctness issues, but please consider fixing the review comments and reopening this as
`data_compression/lz78.py` instead of
`compression/lz78.py`.

priya-sundaram-dev (changes requested):
Thanks for the contribution, @Ahmed99125! The dictionary-building logic is clean and the `Token` dataclass reads nicely. I ran the doctests (14/14 pass) and then fuzz-tested `decompress(compress(s))` — and there's a **correctness bug**: the round-trip is lossy whenever the input ends on a phrase that's already in the dictionary, because that trailing phrase is never emitted as a token.

Minimal reproducers:

```python
>>> c = LZ78Compressor()
>>> c.decompress(c.compress("aa"))
'a'          # expected 'aa'
>>> c.decompress(c.compress("aba"))
'ab'         # expected 'aba'
>>> c.decompress(c.compress("ababab"))
'abab'       # expected 'ababab'
```

Fuzzing random strings (4-letter alphabet, len 0–15) this fails on **~47%** of inputs. The current doctests happen to use inputs whose final phrase completes exactly, which is why they pass.

This is the classic LZ78 "dangling suffix" case. A minimal fix is to flush the leftover phrase after the loop (it's guaranteed to already be in `phrase_dict`, so emit its index with an empty char, which `decompress` already handles):

```python
    for char in text:
        ...
    if phrase:  # trailing phrase already in the dictionary
        tokens.append(Token(int(phrase_dict[phrase]), ""))
    return tokens
```

With that added, `decompress(compress(s)) == s` for all 50k random strings I tested (0 failures). Could you also add a doctest that exercises this case (e.g. `"aa"` or `"ababab"`) so it's guarded going forward?

On efficiency: it's O(n) tokens with dictionary lookups, which is fine for this repo. One minor note — the dict stores indices as `str` and you cast back with `int(...)`; storing them as plain `int` would avoid the round-trip casting, but that's cosmetic. The core algorithm is right once the trailing-phrase case is handled.

PR #11848, opened 2024-10-07
Add cholesky_decomposition.py

99991:
### Describe your change:

This PR adds the [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition) and how to use it to solve a linear system of equations.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+
+    It can also be used to solve multiple equations A X = Y simultaneously.
+
+    >>> X_true = np.random.rand(3, 3)
+    >>> Y = A @ X_true
+    >>> X = solve_cholesky(L, Y)
+    >>> np.allclose(X, X_true)
+    True
+    """
+    assert a.shape[0] == a.shape[1]
```
jeffreyyancey, line 44: Like this check at the beginning. Adding a description for the `AssertionError` would be ever better. `assert a.shape[0] == a.shape[1], "Matrix must be square"`

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+    n = a.shape[0]
+    lo = np.tril(a)
```
jeffreyyancey, line 46: can you give the `lo` variable a proper name like `lower_triangle`?

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+
+    for i in range(n):
+        for j in range(i):
+            lo[i, j] = (lo[i, j] - np.sum(lo[i, :j] * lo[j, :j])) / lo[j, j]
+
+        s = lo[i, i] - np.sum(lo[i, :i] * lo[i, :i])
```
jeffreyyancey, line 52: could you give `s` a proper name like `sum_square`?

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+    """Given a Cholesky decomposition L L^T = A of a matrix A, solve the
+    system of equations A X = Y where B is either a matrix or a vector.
+
+    >>> L = np.array([[2, 0], [3, 4]], dtype=float)
+    >>> Y = np.array([[22, 54], [81, 193]], dtype=float)
+    >>> X = solve_cholesky(L, Y)
+    >>> np.allclose(X, np.array([[1, 3], [3, 7]], dtype=float))
+    True
+    """
+
```
jeffreyyancey, line 72: Simlarly to how you put guard-rails on your other function, you can ensure the `lower_triangle` you receive here is valid:

```
    if not np.allclose(lower_triangle, np.tril(lower_triangle)):
        raise ValueError("Matrix is not lower triangular")
```
or something to that effect

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+    # Handle vector case by reshaping to matrix and then flattening again
+    if len(y.shape) == 1:
+        return solve_cholesky(lo, y.reshape(-1, 1)).ravel()
+
+    n, m = y.shape
+
+    # Backsubstitute L X = B
```
jeffreyyancey, line 79: This is `Forward Substitution`, right?

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+    x = y.copy()
+    for i in range(n):
+        for j in range(i):
+            x[i, :] -= lo[i, j] * x[j, :]
+
+        for k in range(m):
```
jeffreyyancey, line 85: You don't need to loop here. You can simplify to `x[i, :] /= lower_trinagle[i, i]` to divide the entire row instead of iterating through the columns.

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+            x[i, k] /= lo[i, i]
+
+    # Backsubstitute L^T
+    for i in reversed(range(n)):
+        for j in range(i + 1, n):
+            x[i, :] -= lo[j, i] * x[j, :]
+
+        for k in range(m):
```
jeffreyyancey, line 93: Same as above. You can avoid the loop and just divide the entire row.

jeffreyyancey (changes requested):
This looks quite good. To meet the guidelines `lo` and `s` can be changed to fully named variables. I also added a couple tentative suggestions that I think could improve the good work you already have here.

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,102 @@
...
+    lo = np.tril(a)
+
+    for i in range(n):
+        for j in range(i):
+            lo[i, j] = (lo[i, j] - np.sum(lo[i, :j] * lo[j, :j])) / lo[j, j]
+
+        s = lo[i, i] - np.sum(lo[i, :i] * lo[i, :i])
+
+        if s <= 0:
+            raise ValueError("Matrix A is not positive definite")
```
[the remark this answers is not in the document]
99991: Some implementations explicitly allow only lower triangular matrices `A` as input, since filling the upper triangular part of `A` is not necessary due to symmetry, but for clarity, I think we can enforce symmetry here.

maths/cholesky_decomposition.py
99991: I agree. I also added the shape of the non-square matrix to the error message for easier debugging.

maths/cholesky_decomposition.py
99991: I renamed `lo` to `L`. Now the code lines up nicely with the variable naming scheme on the Wikipedia article about the Cholesky decomposition, which makes it easier to understand. SciPy also use upper case single letter variable names for matrices in some decomposition algorithms, so there definitely is a precedence for that https://docs.scipy.org/doc/scipy/reference/linalg.html#decompositions

maths/cholesky_decomposition.py
99991: I refactored the code so that the variable `s` is not necessary anymore.

maths/cholesky_decomposition.py
99991: I would not do this if performance was important (checking for lower triangularity is $O(n^2)$, which might have a significant impact on the total running time of the algorithm, which is also $O(n^2)$ ) but this implementation is mostly in Python, so performance does not matter anyway and and clarity is more important.

maths/cholesky_decomposition.py
99991: Good point, I agree.

maths/cholesky_decomposition.py
99991: Same as above.

maths/cholesky_decomposition.py
```diff
@@ -0,0 +1,109 @@
+import numpy as np
+
+
+# ruff: noqa: N803,N806
+def cholesky_decomposition(A: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
99991: It looks like @algorithms-keeper does not respect the annotation `# ruff: noqa: N803,N806`. Could somebody override this?

maths/cholesky_decomposition.py
99991: You are correct! I corrected the comment accordingly.

99991:
I renamed the variables again. I think the code is harder to follow now since the variable names do not match the equations on Wikipedia anymore, but at least the stupid bot is happy.

cclauss (approved):

99991:
Thank you, @cclauss!

PR #12186, opened 2024-10-19
Persistent Segment Tree

putul03:
The persistent segment tree allows us to efficiently query and update an array while preserving previous versions of the array. The primary operations include:

1. *Build Operation*: Constructs the segment tree from the given array. Each node represents a range of the array and stores the sum of that range.

2. *Update Operation*: When an update is performed, a new version of the segment tree is created without modifying the existing tree. This is achieved by copying the necessary nodes and updating the affected ranges.

3. *Query Operation*: To retrieve the sum of a given range, the segment tree is traversed based on the requested indices. The operation can access multiple versions of the tree to return the correct sum for each version.

### Time and Space Complexity:
- *Time Complexity*:
  - *Build Operation*: O(n log n), where n is the number of elements in the array. This is because we need to recursively build the tree.
  - *Update Operation*: O(log n) per update, since we only traverse the necessary nodes to create a new version.
  - *Query Operation*: O(log n) per query, as we traverse the segment tree to find the required sums.

- *Space Complexity*:
  - The space complexity is O(n * log n) for storing multiple versions of the segment tree, where each version contains log n nodes and there can be up to n versions depending on the number of updates.

### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


cclauss (approved):

data_structures/binary_tree/persistent_segment_tree.py
```diff
@@ -0,0 +1,140 @@
...
+        7
+        """
+        if start == end:
+            return Node(arr[start])
+        mid = (start + end) // 2
+        node = Node()
+        node.left = self._build(arr, start, mid)
+        node.right = self._build(arr, mid + 1, end)
+        node.value = node.left.value + node.right.value
+        return node
```
cclauss:
```suggestion
        left = self._build(arr, start, mid)
        right = self._build(arr, mid + 1, end)
        return Node(value=left.value + right.value, left=left, right=right)

```

cclauss:
pre-commit.ci run

Issue #12233, opened 2024-10-21
Avoid log(0) in KL divergence


kevin1kevin1k:
### Repository commit

03a42510b01c574292ca9c6525cbf0572ff5a2a5

### Python version (python --version)

Python 3.10.15

### Dependencies version (pip freeze)

absl-py==2.1.0
astunparse==1.6.3
beautifulsoup4==4.12.3
certifi==2024.8.30
charset-normalizer==3.4.0
contourpy==1.3.0
cycler==0.12.1
dill==0.3.9
dom_toml==2.0.0
domdf-python-tools==3.9.0
fake-useragent==1.5.1
flatbuffers==24.3.25
fonttools==4.54.1
gast==0.6.0
google-pasta==0.2.0
grpcio==1.67.0
h5py==3.12.1
idna==3.10
imageio==2.36.0
joblib==1.4.2
keras==3.6.0
kiwisolver==1.4.7
libclang==18.1.1
lxml==5.3.0
Markdown==3.7
markdown-it-py==3.0.0
MarkupSafe==3.0.2
matplotlib==3.9.2
mdurl==0.1.2
ml-dtypes==0.3.2
mpmath==1.3.0
namex==0.0.8
natsort==8.4.0
numpy==1.26.4
oauthlib==3.2.2
opencv-python==4.10.0.84
opt_einsum==3.4.0
optree==0.13.0
packaging==24.1
pandas==2.2.3
patsy==0.5.6
pbr==6.1.0
pillow==11.0.0
pip==24.2
protobuf==4.25.5
psutil==6.1.0
Pygments==2.18.0
pyparsing==3.2.0
python-dateutil==2.9.0.post0
pytz==2024.2
qiskit==1.2.4
qiskit-aer==0.15.1
requests==2.32.3
requests-oauthlib==1.3.1
rich==13.9.2
rustworkx==0.15.1
scikit-learn==1.5.2
scipy==1.14.1
setuptools==74.1.2
six==1.16.0
soupsieve==2.6
sphinx-pyproject==0.3.0
statsmodels==0.14.4
stevedore==5.3.0
symengine==0.13.0
sympy==1.13.3
tensorboard==2.16.2
tensorboard-data-server==0.7.2
tensorflow==2.16.2
tensorflow-io-gcs-filesystem==0.37.1
termcolor==2.5.0
threadpoolctl==3.5.0
tomli==2.0.2
tweepy==4.14.0
typing_extensions==4.12.2
[...]

PR #12237, opened 2024-10-22
Avoid log(0) in KL divergence

bz-e:
…denominator and added a test case

### Describe your change:
Fixes #12233 
Added type NONE to make it pass type checking, and added a small constant to the kullback_leibler_divergence method to fix the bug of numerator and denominator being 0, and also added a test case.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #12233 ".

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
       - id: auto-walrus
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.9
+    rev: v0.7.0
     hooks:
       - id: ruff
       - id: ruff-format
@@ -29,7 +29,7 @@ repos:
           - tomli
 
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: "2.3.0"
+    rev: "2.4.3"
     hooks:
       - id: pyproject-fmt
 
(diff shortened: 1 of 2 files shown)


imSanko (changes requested):
what do you actually want to change ?

Harmanaya:
@cclauss Could you please review this PR when you get a chance?

kevin1kevin1k:
@bz-e I think we do not intend to change the default behavior when all y_true are nonzero.
A better way might be to mask out all the zero entries and only sum them.

bz-e:
> @bz-e I think we do not intend to change the default behavior when all y_true are nonzero. A better way might be to mask out all the zero entries and only sum them.

I think this is the lowest time complexity solution to the issue #12233 .

kevin1kevin1k:
> I think this is the lowest time complexity solution to the issue #12233 .

I meant you could do something like the following (need not be identical, just for demo), and the complexity stays linear.
Also IMHO since this repo is more for educational purpose than for practical/production usage, correctness is more favorable compared to efficiency

```python
    mask = y_true != 0
    y_true_filtered = y_true[mask]
    y_pred_filtered = y_pred[mask]
    kl_loss = y_true_filtered * np.log(y_true_filtered / y_pred_filtered)

```

bz-e:
> I meant you could do something like the following (need not be identical, just for demo), and the complexity stays linear. Also IMHO since this repo is more for educational purpose than for practical/production usage, correctness is more favorable compared to efficiency
> 
> ```python
>     mask = y_true != 0
>     y_true_filtered = y_true[mask]
>     y_pred_filtered = y_pred[mask]
>     kl_loss = y_true_filtered * np.log(y_true_filtered / y_pred_filtered)
> ```

For educational purposes I think you are right, this is the method with the least changes.

PR #12249, opened 2024-10-23
Fixed: #12233

vedprakash226:
I have added a logic that restricts the evaluation of log function if y_true is 0.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #12233

### DIFF
--- a/machine_learning/loss_functions.py
+++ b/machine_learning/loss_functions.py
@@ -659,7 +659,10 @@ def kullback_leibler_divergence(y_true: np.ndarray, y_pred: np.ndarray) -> float
     if len(y_true) != len(y_pred):
         raise ValueError("Input arrays must have the same length.")
 
-    kl_loss = y_true * np.log(y_true / y_pred)
+    kl_loss = 0
+    if y_true != 0:
+        kl_loss = y_true * np.log(y_true / y_pred)
+
     return np.sum(kl_loss)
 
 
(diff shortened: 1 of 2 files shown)


kevin1kevin1k:
- Note that y_true is an array so comparing it to 0 might not be a good idea here, and there are nonzero entries that should be calculated instead of just returning 0.
- I think this is a duplicate of #12237

anandfresh (approved):

kevin1kevin1k (changes requested):
Please see my comment in the conversation

> Note that y_true is an array so comparing it to 0 might not be a good idea here, and there are nonzero entries that should be calculated instead of just returning 0.

This causes
`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

kevin1kevin1k:
@anandfresh I do not think the PR should be approved as it is not correct and do not resolve what the issue intends.

anandfresh:
> @anandfresh I do not think the PR should be approved as it is not correct and do not resolve what the issue intends.

@kevin1kevin1k - Yes, my bad ! I didn't see it is an array variable . Ideally he has to check if the array length is greater than 0

machine_learning/loss_functions.py
```diff
@@ -659,7 +659,10 @@ def kullback_leibler_divergence(y_true: np.ndarray, y_pred: np.ndarray) -> float
     if len(y_true) != len(y_pred):
         raise ValueError("Input arrays must have the same length.")
 
-    kl_loss = y_true * np.log(y_true / y_pred)
+    kl_loss = 0
+    if y_true != 0:
```
anandfresh, line 663: Shouldn't you check the length of the `y_true` array greater than 0?

anandfresh (changes requested):

cclauss:

PR #12283, opened 2024-10-25
Create longest-palindromic-substring.py

Tsurya06:
The problem is to find the longest palindromic substring within a given string. The optimal solution uses an expand around center approach, where each character (and the gap between every two characters) is treated as a possible center for palindromes. The algorithm expands outward from each center, identifying the longest palindrome, and updates the result if a longer palindrome is found. This approach is efficient with O(n2) time complexity and does not require additional space.


cclauss:

PR #12317, opened 2024-10-29
Logical Implementation of Shor Algorithm

joelkurien:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/quantum/shor_algorithm.py
+++ b/quantum/shor_algorithm.py
@@ -9,6 +9,9 @@ prime numbers that are used to create the public key value, n
 In this implementation, I have used a very simple construct without 
 the use of qiskit or cirq to help understand how Shor algorithm's
 idea actually works.
+
+Website referred for shor algorithm:
+https://www.geeksforgeeks.org/shors-factorization-algorithm/
 """
 class Shor:
     def period_find(self, num: int, number: int) -> int:
@@ -26,7 +29,7 @@ class Shor:
             start += 1
         return start
         
-    def shor_algorithm(self, number:int) -> list[int]:
+    def shor_algorithm(self, number:int) -> tuple[int, int]:
         """
         Run Shor's algorithm to factor a number.
         >>> shor = Shor()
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev (commented):
Thanks for the pure-Python Shor implementation @joelkurien — keeping it dependency-free is a good fit for the collection. A few issues to address before merge:

1. **Module docstring placement.** The triple-quoted description sits *after* `import math` / `import random`, so Python doesn't treat it as the module docstring. Move it to the very top of the file, above the imports.
2. **Import-time side effects.** `shor = Shor()` and `print(shor.shor_algorithm(15))` run at module top level, so they execute on *import* (during test collection and DIRECTORY generation). Please move them under `if __name__ == "__main__":`, and add `import doctest; doctest.testmod()` there.
3. **Re-seeding inside the loop (bug).** `shor_algorithm` calls `random.seed(0)` *inside* `while True`, so every iteration draws the same `num`; if that candidate never yields a factor, the loop spins forever. Seed once before the loop (or not at all) and let `num` vary across iterations.
4. Minor: with the loop restructured, make sure the annotated `-> tuple[int, int]` return always holds (no implicit `None` fall-through).

Type hints and the `period_find` doctest are good — the above are the main blockers.

quantum/shor_algorithm.py
```diff
@@ -0,0 +1,66 @@
...
+                start: int = pow(num, result // 2, number)
+                if start != number - 1:
+                    p_value: int = math.gcd(start - 1, number)
+                    q_value: int = math.gcd(start + 1, number)
+                    if p_value > 1 and q_value > 1:
+                        return p_value, q_value
+
+
+shor = Shor()
+print(shor.shor_algorithm(15))
```
cclauss:
```suggestion
if __name__ == "__main__":
    shor = Shor()
    print(shor.shor_algorithm(15))
```

cclauss (approved):

PR #12434, opened 2024-12-13
fixed the issue in strings/join.py

RajdeepBakolia2004:
The issue was that join() function was not working as the join method of string due to strip method which truncates from end 
now whenever we have empty strings in our separated list it was just adding multiple terminator and all the terminator where getting strip which was different form the join method so a for loop was run for the entire list except the last element in which case it just add the last element and it also take care of the case when the list is empty

It fixes the bug in the join.py file

Fixes #12408

### Describe your change:

* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [x] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

PR #12423, opened 2024-12-06
Fix split function to handle trailing delimiters correctly

kishyassin:
This pull request includes enhancements to the `split` function in the `strings/split.py` file to handle cases where the separator appears at the end of the input string. The most important changes include adding new test cases and modifying the function logic to append an empty string when the input string ends with the separator.

Enhancements to `split` function:

* [`strings/split.py`](diffhunk://#diff-5392b2ff300ccd1d0e76c32d21b63fbc21e6a473f7b0aa15910f20a11fa4338bR17-R19): Added a test case for input strings where the separator appears at the beginning and end of the string.
* [`strings/split.py`](diffhunk://#diff-5392b2ff300ccd1d0e76c32d21b63fbc21e6a473f7b0aa15910f20a11fa4338bR31-L33): Modified the function logic to append an empty string to the result list if the input string ends with the separator.

### Describe your change:
* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [ ] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #12403 ".

PR #12438, opened 2024-12-16
fixes join.py and split.py action and requirements error

### LINKED ISSUE #12346: Warning sign after completing installation of all the requirements

Python version: Python 3.11.5

Hi, I would like to raise an issue which is related to the warning sign after installing all the requirements. 

I see the warning sign while I have installed them. 
warning

Please refer to the below images which shows the installation of the packages.
fake-useragent
sphinx-pyproject

The problems are fixed when changing the names.
- from fake_useragent to fake-useragent
- from sphinx_pyproject to sphinx-pyproject

However, before making a PR, I would like to check whether this problem is from the setting of my PC or the problem can arise in other PCs.

Thank you for reviewing the issue in advance. 

### Additional information

_No response_

### LINKED ISSUE #12408: strings/join.py is incorrect
### Repository commit

3b1b70bf0e60dddb14917dd7cad34a41d429df01

Python 3.10

n/a

### Expected behavior

s = ["", "", ""]
join(",", s) == ",".join(s)
# ',,' == ',,'

### Actual behavior

s = ["", "", ""]
join(",", s) == ",".join(s)
# '' != ',,'

sanks011:
### Describe your change:
Fixes #12346

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

For #12346
image

### DIFF
--- a/strings/join.py
+++ b/strings/join.py
@@ -1,17 +1,21 @@
+"""
+Program to join a list of strings with a separator
+"""
+
+
 def join(separator: str, separated: list[str]) -> str:
     """
-    Custom implementation of the join() function.
-    This function manually concatenates the strings in the list,
-    using the provided separator, without relying on str.join().
+    Joins a list of strings using a separator
+    and returns the result.
 
-    :param separator: The separator to place between strings.
-    :param separated: A list of strings to join.
+    :param separator: Separator to be used
+                for joining the strings.
+    :param separated: List of strings to be joined.
 
-    :return: A single string with elements joined by the separator.
-
-    :raises Exception: If any element in the list is not a string.
+    :return: Joined string with the specified separator.
 
     Examples:
+
     >>> join("", ["a", "b", "c", "d"])
     'abcd'
     >>> join("#", ["a", "b", "c", "d"])
--- a/strings/join.py
+++ b/strings/join.py
@@ -8,8 +8,7 @@ def join(separator: str, separated: list[str]) -> str:
     Joins a list of strings using a separator
     and returns the result.
 
-    :param separator: Separator to be used
-                for joining the strings.
+    :param separator: Separator to be used for joining the strings.
     :param separated: List of strings to be joined.
 
     :return: Joined string with the specified separator.
@@ -20,13 +19,12 @@ def join(separator: str, separated: list[str]) -> str:
     'abcd'
     >>> join("#", ["a", "b", "c", "d"])
     'a#b#c#d'
-    >>> join("#", "a")
+    >>> join("#", ["a"])
     'a'
     >>> join(" ", ["You", "are", "amazing!"])
     'You are amazing!'
 
-    This example should raise an
-    exception for non-string elements:
+    This example should raise an exception for non-string elements:
     >>> join("#", ["a", "b", "c", 1])
     Traceback (most recent call last):
         ...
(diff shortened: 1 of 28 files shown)


strings/join.py
```diff
@@ -35,17 +33,17 @@ def join(separator: str, separated: list[str]) -> str:
...
 
-    joined = ""
-    for word_or_phrase in separated:
-        if not isinstance(word_or_phrase, str):
-            raise Exception("join() accepts only strings")
-        joined += word_or_phrase + separator
+    if not all(isinstance(word_or_phrase, str) for word_or_phrase in separated):
+        raise Exception("join() accepts only strings")
+
+    joined = separator.join(separated)
```
Kaaserne, line 44: Question about this join though. Isn't the point of this repository to show possible implementations of python functions? Then I was wondering why the join function now internally uses str.join?

imSanko (changes requested):
seems fine but why the join function now internally uses str.join? as what @Kaaserne said

strings/join.py
sanks011: okay I understood! Now I'll try to implement something which manually constructs the resulting string without relying on Python's built-in str.join()

strings/join.py
```diff
@@ -1,49 +1,27 @@
...
     if not all(isinstance(word_or_phrase, str) for word_or_phrase in separated):
         raise Exception("join() accepts only strings")
 
-    joined = separator.join(separated)
+    # Manually handle concatenation
+    result = ""
+    for i, element in enumerate(separated):
+        result += element
+        if i < len(separated) - 1:  # Add separator only between elements
+            result += separator
```
sanks011, line 23: @Kaaserne can you please review it?

strings/join.py
```diff
@@ -1,51 +1,27 @@
...
-
-    :param separator: Separator to be used
-                for joining the strings.
-    :param separated: List of strings to be joined.
-
-    :return: Joined string with the specified separator.
-
-    Examples:
-
-    >>> join("", ["a", "b", "c", "d"])
```
Kaaserne, line 19: What's the reason these got removed?

strings/join.py
```diff
@@ -1,51 +1,27 @@
...
+    using the provided separator, without relying on str.join().
+
+    :param separator: The separator to place between strings.
+    :param separated: A list of strings to join.
+
+    :return: A single string with elements joined by the separator.
+
+    :raises Exception: If any element in the list is not a string.
     """
+    if not all(isinstance(word_or_phrase, str) for word_or_phrase in separated):
```
Kaaserne, line 14: I think this if statement can be placed inside the for-loop. That way we only iterate over the sequence once, instead of twice.

sanks011: Thank you for your advice. sure lemme check

Kaaserne (approved):

MaximSmolskiy:
Reversed `split` changes in favor #12423

MaximSmolskiy (changes requested):
Please split this PR to several - one PR for one issue (especially if the issues are not related as in this case)

And this PR has failed doctests
```
>>> join("#", "a")  # Single string instead of list
Expected:
        ...
    Exception: join() accepts only strings
Got:
    'a'
```

MaximSmolskiy:
Reversed `join` changes in favor of #12434

MaximSmolskiy (approved):
@sanks011 Thanks for fix for #12346

PR #12448, opened 2024-12-22
compare-method added to Vector class in lib.py

Andrwaa:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/linear_algebra/src/lib.py
+++ b/linear_algebra/src/lib.py
@@ -46,7 +46,6 @@ class Vector:
         change_component(pos: int, value: float): changes specified component
         euclidean_length(): returns the euclidean length of the vector
         angle(other: Vector, deg: bool): returns the angle between two vectors
-        TODO: compare-operator
     """
 
     def __init__(self, components: Collection[float] | None = None) -> None:
@@ -182,6 +181,15 @@ class Vector:
             return math.degrees(math.acos(num / den))
         else:
             return math.acos(num / den)
+    def __eq__(self, vector: object) -> bool:
+        """
+        performs the comparison between two vectors
+        """
+        if not isinstance(vector, Vector):
+            return NotImplemented
+        if len(self) != len(vector):
+            return False
+        return all(self.component(i) == vector.component(i) for i in range(len(self)))
 
 
 def zero_vector(dimension: int) -> Vector:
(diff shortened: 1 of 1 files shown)


linear_algebra/src/lib.py
```diff
@@ -183,6 +182,16 @@ def angle(self, other: Vector, deg: bool = False) -> float:
         else:
             return math.acos(num / den)
 
+    def __eq__(self, vector: object) -> bool:
```
tianyizheng02, line 185: Could you move the method higher up in the file, before all the public methods? Having this one dunder method within all the public methods, rather than having it alongside all the other dunder methods, makes it harder to find.

linear_algebra/src/lib.py
```diff
@@ -183,6 +182,16 @@ def angle(self, other: Vector, deg: bool = False) -> float:
...
+        """
+        performs the comparison between two vectors
+        """
+        if not isinstance(vector, Vector):
+            return NotImplemented
```
tianyizheng02, line 190:
```suggestion
    def __eq__(self, vector: Vector) -> bool:
        """
        performs the comparison between two vectors
        """
```
I don't think it makes sense to allow comparison with any object, only with other `Vector`s.

tianyizheng02 (changes requested):

linear_algebra/src/lib.py
Andrwaa: I tried writing Vector insted of object but it gave me an error that saying: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"  [override]
This violates the Liskov substitution principle

tianyizheng02:
```
linear_algebra/src/lib.py:98: note: This violates the Liskov substitution principle
linear_algebra/src/lib.py:98: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
linear_algebra/src/lib.py:98: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
linear_algebra/src/lib.py:98: note:     def __eq__(self, other: object) -> bool:
linear_algebra/src/lib.py:98: note:         if not isinstance(other, Vector):
linear_algebra/src/lib.py:98: note:             return NotImplemented
linear_algebra/src/lib.py:98: note:         return <logic to compare two Vector instances>
```
Sorry for the inconvenience, I hadn't realized that `mypy` treated this as an error. Would you mind changing the code back to what you had originally, with the `NotImplemented` check? (The method's new location is fine.)

Andrwaa:
ok, i went back to the previous version

tianyizheng02 (approved):
LGTM, thanks for your changes!

PR #12515, opened 2025-01-12
Fix ruff

MaximSmolskiy:
### Describe your change:

Fix ruff

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


MaximSmolskiy:
@cclauss please review and tell me is it a normal situation that last commit to master has green status, but at the same time on new commits to master ruff crashes?

pyproject.toml
```diff
@@ -147,6 +147,9 @@ lint.per-file-ignores."project_euler/problem_099/sol1.py" = [
 lint.per-file-ignores."sorts/external_sort.py" = [
   "SIM115",
 ]
+
+lint.flake8-builtins.builtins-allowed-modules = [ "base64", "wave" ]
```
cclauss, line 151: Let’s rename the files instead like base64_cypher.py and wave_string.py

MaximSmolskiy: Done

cclauss:
Ruff is v0.9.1 almost at v1.0 where it will be more stable but for now we can expect some bumps like this.  V0.9.0 also added the 2025 stable format which is shared between black and ruff format.  We could run ruff only in pre-commit but I like running unpinned in GitHub Actions so we spot the changes.

cclauss (approved):
Nice!

PR #12541, opened 2025-01-25
Simplify diffie_hellman.py

isidroas:
### Describe your change:

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/ciphers/diffie_hellman.py
+++ b/ciphers/diffie_hellman.py
@@ -1,9 +1,14 @@
+"""
+https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange
+"""
+
 import random
 
-# RFC 3526 - More Modular Exponential (MODP) Diffie-Hellman groups for
-# Internet Key Exchange (IKE) https://tools.ietf.org/html/rfc3526
-
-PRIMES = {
+"""
+RFC 3526 - More Modular Exponential (MODP) Diffie-Hellman groups for
+Internet Key Exchange (IKE) https://tools.ietf.org/html/rfc3526
+"""
+GROUPS = {
     # 1536-bit
     5: {
         "prime": int(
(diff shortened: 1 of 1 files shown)


isidroas:
These are many unrelated changes. Let me know if I have to divide them into commits or PRs.

isidroas:
The git history shows that the author is @ruppysuppy . Do these changes seem good to you?

Everything is a refactor except the method `generate_shared_key()`: its hashing final step has been removed for simplicity and because the [key derivation](https://en.wikipedia.org/wiki/Key_derivation_function) is a separate problem.

ciphers/diffie_hellman.py
```diff
@@ -181,50 +184,34 @@
...
 
+    Current minimum recommendation is 2048 bit
+    >>> group = GROUPS[14]
+    >>> prime, generator = group['prime'], group['generator']
+    >>> alice = DiffieHellman(prime, generator)
 
-    >>> alice = DiffieHellman()
-    >>> bob = DiffieHellman()
+    Both parties should agree on the same public parameters
+    >>> bob = DiffieHellman(alice.prime, alice.generator)
```
ruppysuppy, line 195: Are we generating Bob's data using Alice's `prime`? Aren't the `prime` supposed to be **private**?

ruppysuppy (commented):
@isidroas I am happy with the changes, but slightly unsure about this - my thoughts related to cryptography is a bit rusty (hadn't touched the topic in a while), so a brief explanation would be great

isidroas:
Thank you for the feedback @ruppysuppy ! I added a bit of documentation in 93c7ac0116508475e452e2027d700f0648b6f5d6

ciphers/diffie_hellman.py
isidroas: The only secret attribute is each `self.__private_key`, which does not necessarily have to be a prime number. In cryptography it is strange that more than one parameter must be kept secret.

ruppysuppy (approved):

PR #12552, opened 2025-01-31
optimize split_matrix function

ivanz851:
I optimized split_matrix function by removing duplicate code to the extract_submatrix function, and added tests to test_strassen_matrix_multiplication.py file

### DIFF
--- a/divide_and_conquer/strassen_matrix_multiplication.py
+++ b/divide_and_conquer/strassen_matrix_multiplication.py
@@ -49,18 +49,20 @@ def split_matrix(a: list) -> tuple[list, list, list, list]:
     if len(a) % 2 != 0 or len(a[0]) % 2 != 0:
         raise Exception("Odd matrices are not supported!")
 
-    matrix_length = len(a)
-    mid = matrix_length // 2
+    def extract_submatrix(rows, cols):
+        return [[a[i][j] for j in cols] for i in rows]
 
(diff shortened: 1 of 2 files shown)


josmithua:
This increases readability and maintainability of the code, but certainly does not "optimize" the algorithm, for you have added the extra overhead of calling a function, which will actually slow down the algorithm.

ivanz851:
> This increases readability and maintainability of the code, but certainly does not "optimize" the algorithm, for you have added the extra overhead of calling a function, which will actually slow down the algorithm.

Actually, my code with extract_submatrix function is not less efficient than the original one, since there is exactly same number of  conditional jumps. This can be checked using godbolt.org: [comparison](https://godbolt.org/z/Wj4frYx5a). If you follow the link, you can see disassembly of my revision of the code on the left and of the original code on the right. In original code, there are 8 FOR_ITER jumps and 8 JUMP_BACKWARD jumps. In new code, there are 2 FOR_ITER jumps and 2 JUMP_BACKWARD jumps in extract_submatrix  function, which is called 4 times, so in sum we have same 8 FOR_ITER jumps and 8 JUMP_BACKWARD jumps. That's why I think code with extract_submatrix function  is not slower and can be used for better readability.

josmithua:
@ivanz851 Thanks, I learned something new today. Cheers

cclauss (approved):

PR #12583, opened 2025-02-18
Fix n-queens problem

infrablue1:
### Describe your change:
Fix wrong solution for n-queens problem which in backtrace folder.

### Details
There is an issue about **n_queens.py**.  If I run this script, it will print the wrong result which is different with **n_queens_math.py**. I think it's enough to check  **left upper diagonal**,  **right upper diagonal** and **the same column** here. Here are the last few lines of the correct output with my changes:
```bash
. . . . . . . Q
. . . Q . . . .
Q . . . . . . .
. . Q . . . . .
. . . . . Q . .
. Q . . . . . .
. . . . . . Q .
. . . . Q . . .
The total number of solutions are: 92
```

### DIFF
--- a/backtracking/n_queens.py
+++ b/backtracking/n_queens.py
@@ -40,7 +40,9 @@ def is_safe(board: list[list[int]], row: int, column: int) -> bool:
         and all(
             board[i][j] != 1 for i, j in zip(range(row, -1, -1), range(column, -1, -1))
         )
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

rahmani3101 (approved):

MaximSmolskiy (approved):

MaximSmolskiy:
@cclauss Can we do something about this?
```
The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.
```

PR #12678, opened 2025-04-19
hash map

isidroas:
- **Mutable _Item**
- **document falsy item**
- **resize_down: expected test result**
- **resize_down: actual result**

Please refer to the commit messages for a detailed breakdown of the changes.

Feedback from the original author, @Cjkjvfnby, would be greatly appreciated!
### Describe your change:

* [ ] Add an algorithm?
* [X] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/hashing/hash_map.py
+++ b/data_structures/hashing/hash_map.py
@@ -246,7 +246,7 @@ class HashMap(MutableMapping[KEY, VAL]):
         ## Resize down
         >>> del hm[75]
         >>> len(hm._buckets)
-        100
+        200
         >>> del hm[74]
         >>> len(hm._buckets)
         100
--- a/data_structures/hashing/hash_map.py
+++ b/data_structures/hashing/hash_map.py
@@ -72,11 +72,11 @@ class HashMap(MutableMapping[KEY, VAL]):
 
         If bucket is empty or key is the same, does insert and return True.
 
-        If bucket has another key or deleted placeholder,
-        that means that we need to check next bucket.
+        If bucket has another key that means that we need to check next bucket.
         """
         stored = self._buckets[ind]
         if not stored:
+            # A falsy item means that is None (bucket never used) or _deleted.
             self._buckets[ind] = _Item(key, val)
             self._len += 1
             return True
--- a/data_structures/hashing/hash_map.py
+++ b/data_structures/hashing/hash_map.py
@@ -232,7 +232,7 @@ class HashMap(MutableMapping[KEY, VAL]):
 
         # Test resize down when sparse
         ## Setup: resize up
-        >>> hm = HashMap(100, capacity_factor=0.75)
+        >>> hm = HashMap(initial_block_size=100, capacity_factor=0.75)
         >>> len(hm._buckets)
         100
         >>> for i in range(75):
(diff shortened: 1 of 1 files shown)


data_structures/hashing/hash_map.py
```diff
@@ -72,16 +72,16 @@ def _try_set(self, ind: int, key: KEY, val: VAL) -> bool:
 
         If bucket is empty or key is the same, does insert and return True.
 
-        If bucket has another key or deleted placeholder,
-        that means that we need to check next bucket.
+        If bucket has another key that means that we need to check next bucket.
         """
         stored = self._buckets[ind]
         if not stored:
+            # A falsy item means that is None (bucket never used) or _deleted.
```
Cjkjvfnby, line 79:
```suggestion
            # A falsy item means that bucket was never used (None) or was deleted (_deleted).
```

data_structures/hashing/hash_map.py
```diff
@@ -228,6 +228,25 @@ def __delitem__(self, key: KEY) -> None:
         Traceback (most recent call last):
         ...
         KeyError: 4
+
+        # Test resize down when sparse
+        ## Setup: resize up
+        >>> hm = HashMap(100, capacity_factor=0.75)
```
Cjkjvfnby, line 234: Can we go with size 4?  In that case, we could avoid the loop.

Cjkjvfnby (commented):
Thank you for improving it.


Changes look good, I left some cosmetic comments.  

As for making `_Item` mutable, Immutability gives extra protection in exchange for performance.  In this abstract case, I don't have any strong preference, both versions are fine.

data_structures/hashing/hash_map.py
isidroas: Something like this?

```python
# Test resize down when sparse
## Setup: resize up
>>> hm = HashMap(4, capacity_factor=0.75)
>>> len(hm._buckets)
4
>>> hm[0] = 0
>>> hm[1] = 1
>>> hm[2] = 2
>>> len(hm._buckets)
4
>>> hm[3] = 3
>>> len(hm._buckets)
8

## Resize down
>>> del hm[3]
>>> len(hm._buckets)
8
>>> del hm[2]
>>> len(hm._buckets)
4
```

I've checked that it passes, but I don't know which one I prefer

isidroas:
> Thank you for improving it.
> 
> Changes look good, I left some cosmetic comments.

Thank you for the revision!

> As for making `_Item` mutable, Immutability gives extra protection in exchange for performance. In this abstract case, I don't have any strong preference, both versions are fine.

I understand that the immutable version is faster. That said, performance will be limited in Python either way

data_structures/hashing/hash_map.py
Cjkjvfnby: I'd prefer to keep tests as linear as possible.

Cjkjvfnby: Let's wait for the 3d opinion.

Cjkjvfnby:
Changes look good.  I'll not be active for the next 2 weeks, please don't wait for me.

MaximSmolskiy (approved):

PR #12679, opened 2025-04-19
Fix: Correct time-and-a-half pay calculation logic

kamalesh3566:
### Describe your change:

- [x] Fix a bug or typo in an existing algorithm  

This pull request fixes the logic for calculating time-and-a-half pay in the `financial/time_and_half_pay.py` file.

Previously, the overtime calculation was incorrect. The fix correctly applies the time-and-a-half formula:
```python
overtime_pay = overtime_hours * pay_rate * 1.5
```

### Checklist:

### DIFF
--- a/financial/time_and_half_pay.py
+++ b/financial/time_and_half_pay.py
@@ -5,9 +5,15 @@ Calculate time and a half pay
 
 def pay(hours_worked: float, pay_rate: float, hours: float = 40) -> float:
     """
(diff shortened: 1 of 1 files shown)


MaximSmolskiy:
There is no logic error - implementations are equivalent (tests haven't changed).
There may be some difference in concepts and terms here, but I think the current implementation is good - the reader should easily understand the logic of the calculations and that they are correct

PR #12696, opened 2025-04-29
Added Count Inversions algorithm (Merge Sort based)

usman-7455:
### Describe your change:

* [x] Add an algorithm

This pull request adds a Python implementation of the **Count Inversions** algorithm using a modified Merge Sort approach.

Count Inversions is used to determine how far an array is from being sorted by counting the number of out-of-order pairs (i, j) such that i < j and arr[i] > arr[j]. This has applications in ranking systems, bioinformatics, and measuring list similarity.

### Checklist:
* [x] This PR only changes one algorithm file. To ease review, I am not combining unrelated changes.
* [x] All filenames are in all lowercase characters with no spaces or dashes. *(Note: The filename is `Count_Inversion.py`; I can rename it to `count_inversion.py` if needed.)*

Wikipedia link: https://en.wikipedia.org/wiki/Inversion_(discrete_mathematics)

### DIFF
--- a/sorts/count_inversion.py
+++ b/sorts/count_inversion.py
@@ -1,0 +1,56 @@
+"""
+Count Inversions in an array using a modified Merge Sort algorithm.
+
+An inversion is a pair of indices (i, j) such that:
+    - i < j
+    - arr[i] > arr[j]
+This represents a disorder in the array relative to a sorted order.
+
+This implementation efficiently counts the number of such inversions
+in O(n log n) time by modifying the merge step of Merge Sort.
+
+Use Cases:
+----------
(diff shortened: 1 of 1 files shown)


mindaugl (changes requested):
Would be appropriate to change interface and return both sorted array and number of inversions (currently 'count_inverstions()' returns only number of inversions). For counting inversions, there is no need to construct new sorted array, and the algorithm is simpler. Also, would be appropriate to change the name of file to, e.g.,  merge_count_inversions_sort.py.

cclauss:
No type hints, doctests, etc. See CONTRIBUTING.md

PR #12723, opened 2025-05-11
Add Kaprekar number checker to special_numbers

Sean-Randall:
This PR adds a new function `is_kaprekar_number(n)` under `maths/special_numbers/`.

The function determines whether a number is a Kaprekar number based on digit-splitting logic, where the square of a number is divided into two parts that sum to the original number. It includes inline doctests to demonstrate usage and ensure correctness.

✔️ This contribution follows all repository contribution guidelines:
- Type hints included (`n: int -> bool`)
- File and function names follow lowercase conventions
- Clean formatting per PEP8
- Validated locally using `doctest.testmod()`

🎯 Educational Value:
This function enhances the repository's coverage of special number classifications and provides a clean, beginner-friendly example of digit-based number theory in Python.

Tested and ready for review. Thank you for maintaining this valuable resource!

### DIFF
--- a/maths/special_numbers/kaprekar_number.py
+++ b/maths/special_numbers/kaprekar_number.py
@@ -1,0 +1,35 @@
+def is_kaprekar_number(n: int) -> bool:
+    """
+    Determine whether a number is a Kaprekar number.
+
+    A Kaprekar number is one where the square can be split into parts
+    that sum to the original number.
+
+    Args:
+        n (int): The number to check.
+
+    Returns:
+        bool: True if it's a Kaprekar number, else False.
+
+    Examples:
+        >>> is_kaprekar_number(45)
+        True
+        >>> is_kaprekar_number(9)
+        True
+        >>> is_kaprekar_number(10)
+        False
+    """
+    square = str(n ** 2)
+    for i in range(1, len(square)):
+        left, right = square[:i], square[i:]
+        if int(right) == 0:
+            continue
+        if n == int(left or "0") + int(right):
+            return True
+    return n == 1
+
+
+if __name__ == "__main__":
+    import doctest
+    
+    doctest.testmod()
--- a/maths/special_numbers/kaprekar_number.py
+++ b/maths/special_numbers/kaprekar_number.py
@@ -19,7 +19,7 @@ def is_kaprekar_number(n: int) -> bool:
         >>> is_kaprekar_number(10)
         False
     """
-    square = str(n ** 2)
+    square = str(n**2)
     for i in range(1, len(square)):
         left, right = square[:i], square[i:]
         if int(right) == 0:
@@ -31,5 +31,5 @@ def is_kaprekar_number(n: int) -> bool:
 
 if __name__ == "__main__":
     import doctest
-    
+
     doctest.testmod()
--- a/maths/special_numbers/kaprekar_number.py
+++ b/maths/special_numbers/kaprekar_number.py
@@ -1,3 +1,5 @@
+import math
+
 def is_kaprekar_number(n: int) -> bool:
     """
     Determine whether a number is a Kaprekar number (excluding powers of 10).
@@ -24,10 +26,10 @@ def is_kaprekar_number(n: int) -> bool:
     """
     if n == 1:
         return True
-    if n <= 0 or (n % 10 == 0 and n == 10 ** len(str(n))):
-        return False  # Disallow powers of 10 (e.g., 10, 100)
+    if n <= 0 or math.log10(n).is_integer():
+        return False  # Disallow powers of 10 (e.g., 10, 100, 1000)
 
-    square = str(n**2)
+    square = str(n ** 2)
     for i in range(1, len(square)):
         left, right = square[:i], square[i:]
         if n == int(left or "0") + int(right):
--- a/maths/special_numbers/kaprekar_number.py
+++ b/maths/special_numbers/kaprekar_number.py
@@ -1,16 +1,16 @@
 import math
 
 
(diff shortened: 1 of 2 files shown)


mindaugl (changes requested):
Shouldn't 10 be a Kaprekar number, since 10^2 = 100, and "100" = "10" + "0" gives 10?

Sean-Randall:
> Shouldn't 10 be a Kaprekar number, since 10^2 = 100, and "100" = "10" + "0" gives 10?

https://cs.uwaterloo.ca/journals/JIS/VOL3/iann2a.html#:~:text=For%20each%20n%20%3E%3D%201%20%2C,)%20when%20N%20%3D%2010n%20.

10 is Disallowed, see above.

mindaugl:
> > Shouldn't 10 be a Kaprekar number, since 10^2 = 100, and "100" = "10" + "0" gives 10?
> 
> https://cs.uwaterloo.ca/journals/JIS/VOL3/iann2a.html#:~:text=For%20each%20n%20%3E%3D%201%20%2C,)%20when%20N%20%3D%2010n%20.
> 
> 10 is Disallowed, see above.

Thanks for the link. Then, the definition should be clarified in the code function description to be: "Kaprekar numbers: positive numbers n such that n = q+r and n^2 = q*10^m+r, for some m >= 1, q >= 0 and 0 <= r < 10^m, with n != 10^a, a >= 1." (https://oeis.org/A006886)?

Sean-Randall:
> > > Shouldn't 10 be a Kaprekar number, since 10^2 = 100, and "100" = "10" + "0" gives 10?
> > 
> > 
> > https://cs.uwaterloo.ca/journals/JIS/VOL3/iann2a.html#:~:text=For%20each%20n%20%3E%3D%201%20%2C,)%20when%20N%20%3D%2010n%20.
> > 10 is Disallowed, see above.
> 
> Thanks for the link. Then, the definition should be clarified in the code function description to be: "Kaprekar numbers: positive numbers n such that n = q+r and n^2 = q*10^m+r, for some m >= 1, q >= 0 and 0 <= r < 10^m, with n != 10^a, a >= 1." (https://oeis.org/A006886)?

Updated the function to align with the strict definition of Kaprekar numbers per OEIS A006886 and Iannucci (1997). Powers of 10 are now explicitly excluded. Let me know if there's anything else to improve!

maths/special_numbers/kaprekar_number.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    if n == 1:
+        return True
+    if n <= 0 or (n % 10 == 0 and n == 10 ** len(str(n))):
+        return False  # Disallow powers of 10 (e.g., 10, 100)
+
+    square = str(n**2)
+    for i in range(1, len(square)):
+        left, right = square[:i], square[i:]
+        if int(right) == 0:
```
mindaugl, line 33: It seems this check is not needed anymore.

mindaugl (approved):
Just added a comment in code, as the check for int(right) == 0 seems not to be necessary any more after adding explicit check for powers of 10.

Sean-Randall:
> Just added a comment in code, as the check for int(right) == 0 seems not to be necessary any more after adding explicit check for powers of 10.

Thanks! I removed the int(right) == 0 check as suggested. The new logic excludes powers of 10 directly. Let me know if anything else needs a tweak.

Sean-Randall:
All checks are now passing, including tests, Ruff style checks, and namespace packaging. The implementation follows the strict Kaprekar definition and excludes powers of 10 as required. Everything should be ready for merge. Thanks again for the helpful feedback!

cclauss:
@priya-sundaram-dev This algorithm is the first Kaprekar PR, and it has pytests instead of doctests (which is OK with me as long as they pass in CI).  Is this  or one of the four below the best implementation that we should merge?
* #13862
* #14563
* #14562
* #15048

priya-sundaram-dev:
Thanks for the ping, @cclauss. I read all five and verified the two strongest against OEIS A006886 (canonical Kaprekar numbers ≤ 10000 — both reproduce the sequence exactly, no extras/misses).

One clarification first: **#13862 is a different algorithm** — it's Kaprekar's *constant* (the 6174 routine), not the Kaprekar-*number* predicate. It's not really a duplicate of the other four and could stand on its own merits separately.

That leaves four true `is_kaprekar_number` candidates. My ranking:

1. **#14562 (@Eraol) — my recommendation.** Cleanest of the set: pure-integer arithmetic split via `%`/`//` (no string parsing), and the `right > 0` guard is the mathematically correct way to exclude powers of 10 (10→100 splits to `10+00`, rejected because the right part is zero). Concise (~56 lines), doctests cover 1/9/45/55/2223, negatives, and a `1.5` TypeError. ✅ verified against OEIS.
2. **#14563 (@Kcstring)** — also correct, good doctests, but ~104 lines and raises `ValueError` on non-positive input where a boolean predicate returning `False` reads more naturally.
3. **This PR #12723 (@Sean-Randall)** — logic is correct (✅ verified against OEIS), but it uses `math.log10(number).is_integer()` to detect powers of 10, which leans on float precision (fine in this range, but the `right > 0` integer guard sidesteps it entirely), and it ships pytest tests where the repo prefers doctests.
4. **#15048 (@VishalNyk)** — adds an arbitrary-`base` parameter; nice generalization but ~173 lines is more surface area than the "one clear educational example" ethos wants.

So: I'd merge **#14562** for the checker, and consider **#13862** separately as the distinct 6174-routine algorithm. Kudos to everyone — this was a close call between #14562 and #14563.

PR #12748, opened 2025-05-14
Update support_vector_machines.py to add the polynomial kernel

faefabsdgd:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/support_vector_machines.py
+++ b/machine_learning/support_vector_machines.py
@@ -70,7 +70,7 @@ class SVC:
         regularization: float = np.inf,
         kernel: str = "linear",
         gamma: float = 0.0,
-        degree: float=0.0,
+        degree: float = 0.0,
         coef0: float = 0.0,
     ) -> None:
         self.regularization = regularization
@@ -93,7 +93,7 @@ class SVC:
         elif kernel == "polynomial":
             if self.degree == 0:
                 raise ValueError("polynomial kernel requires degree")
-            if not isinstance(self.degree, (float, int)) :
+            if not isinstance(self.degree, (float, int)):
                 raise ValueError("degree must be float or int")
             if not self.degree > 0:
                 raise ValueError("degree must be > 0")
@@ -122,7 +122,6 @@ class SVC:
             float: exp(-(gamma * norm_squared(vector1 - vector2)))
         """
         return np.exp(-(self.gamma * norm_squared(vector1 - vector2)))
-
 
     def __polynomial(self, vector1: ndarray, vector2: ndarray) -> float:
         """


cclauss:
@priya-sundaram-dev, please review.  Do the proposed modifications add real value?  The new tests only seem to be about new exceptions being raised, not about any new functionality.

priya-sundaram-dev:
Reviewed. Short answer: **yes, this adds real value** — the polynomial kernel `(x·y + coef0)^degree` is one of the three textbook SVM kernels, and the repo already ships `linear` and `rbf`, so filling in `polynomial` is squarely in scope and educationally useful. The kernel math itself is correct (I checked: with `degree=2, coef0=1`, `((3,4)·(1,2)+1)^2 = 12^2 = 144` ✅).

But your instinct is right, @cclauss — **the tests demonstrate nothing about the new functionality**. All three new doctests only assert that bad arguments raise. There is no doctest showing the kernel actually computes or that an `SVC(kernel="polynomial", ...)` can `fit`/`predict`. For an educational repo that's the important part.

Concrete asks for @faefabsdgd before merge:

1. **Add a functional doctest** on `__polynomial` (or a fit/predict example) so a reader sees it work, e.g.:
   ```python
   >>> SVC(kernel="polynomial", degree=2.0, coef0=1.0)._SVC__polynomial(
   ...     np.array([3, 4]), np.array([1, 2]))
   144.0
   ```
   (or better, a tiny end-to-end `fit`/`predict` like the linear-kernel example already in the file).

2. **`bool` slips through validation.** `degree=True` passes every check (`bool` is a subclass of `int`, `True > 0`) and silently acts as `degree=1`. If you want to guard it, check `isinstance(self.degree, bool)` first, or reorder so the type check precedes the `== 0` / `> 0` checks.

3. **Minor:** the `degree`/`coef0` params are accepted for every kernel but only meaningful for `polynomial`; a one-line docstring note on `__init__` would save confusion.

Nothing here is a blocker on the idea — it's a legitimate addition. It just needs a test that exercises the kernel, not only the exceptions. Happy to re-review once (1) is in.

cclauss:
@algorithms-keeper review

priya-sundaram-dev:
@cclauss Reviewed. Short version: **yes, it adds real value, with one nit.**

**Value:** the polynomial kernel `(x·y + coef0)^degree` is a standard SVM kernel alongside linear/rbf — genuinely new functionality, not a duplicate of anything in the repo. The math and the `__polynomial` implementation are correct.

**On your tests observation:** you're right that the three new doctests only assert exceptions (`requires degree` / `must be float or int` / `must be > 0`). But that actually *matches this file's existing convention* — the current `rbf`/`linear` coverage in the class docstring is also validation-only (`SVC(kernel="rbf")` → `gamma must be > 0`, etc.); there is no functional kernel doctest in the file today. So this PR isn't below the bar — it meets the existing one.

**What would make it clearly worth merging** (raise the bar for the whole file, starting here): add **one functional doctest** that exercises the kernel end-to-end — e.g. build a tiny 2-class dataset, `SVC(kernel="polynomial", degree=2)`, `fit`, and `predict`, asserting the labels. That demonstrates the feature does something, not just that guards fire.

**Minor:** `coef0` is unvalidated (any float is fine, so low risk); and `degree` accepts `bool` since `bool` ⊂ `int` — harmless. The `degree == 0` sentinel-for-unset mirrors the existing `gamma` pattern, so I'd leave it.

**Advice:** request the one functional doctest, then merge.

PR #12763, opened 2025-05-21
Add solution 3 for project euler problem 5.

mindaugl:
### Describe your change:

Add solution 3 for project euler problem 5.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/project_euler/problem_005/sol3.py
+++ b/project_euler/problem_005/sol3.py
@@ -1,0 +1,51 @@
+"""
+Project Euler Problem 5: https://projecteuler.net/problem=5
+
+Smallest multiple
+
+2520 is the smallest number that can be divided by each of the numbers
+from 1 to 10 without any remainder.
+
+What is the smallest positive number that is _evenly divisible_ by all
+of the numbers from 1 to 20?
+"""
+
+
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev:
Reviewed — **request changes** (and it's a third solution to an already twice-covered problem, so the bar is high).

The algorithm is correct — `solution(20) == 232792560` and `solution(10) == 2520` — and the "accumulate prime-power factors as you go" idea is reasonable. But there are blockers as written:

1. **The doctests are commented out** (`# >>> solution(10)` etc.), so `python -m doctest` runs *zero* tests here. `CONTRIBUTING.md` requires working doctests — please un-comment them and make sure they pass.
2. **Dead code:** drop the leftover `# factors = {}` line.
3. **Duplicate concern:** `problem_005/` already has `sol1.py` and `sol2.py`. A `sol3` needs to be clearly better or teach a distinct technique; the current factor-sieve is close enough to the existing approaches that it may not clear that bar. Worth a sentence in the PR description on what it adds over sol1/sol2.

Happy to re-review once the doctests are live and the dead line is gone.

Issue #13, opened 2016-08-03
quicksort probabilistic guarantee


miczal:
Hi,
I think your implementation of quicksort lacks randomizing of the input. If you shuffle it, you are almost guaranteed to not have  O(n^2) time scenario.

AnupKumarPanwar:
What does randomization of input means? We are taking data from user input and of course that will be random. Please elaborate your suggestion.

miczal:
Before quicksort you should just shuffle the input array, so that you avoid worst-case scenario.
Consider this - If you have already sorted input, then you encounter O(n^2) scenario, but if you shuffle the array before passing it to the sorting algorithm, then it is MUCH more likely to be O(n log n). 

I think that this should be fine.
```
from random import shuffle
def quick_sort(collection):
    shuffle(collection)
    quick_sort_impl(collection)
```

H-Darji:
@miczal 
Hi, **This is an open-source repository** and **anyone** can make a pull request with the changes that can **improve the performance** of the algorithm. **If you find a way to improve the performance, we request you to make a pull request**.

AnupKumarPanwar:
Thanks for appreciating the Organisation

PR #12764, opened 2025-05-22
Add second solution for the Euler project problem 13.

mindaugl:
### Describe your change:

Add second solution for the Euler project problem 13. The solution is more in the spirit of doing actual calculation, instead of using python int feature to hold large number of digits.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/project_euler/problem_013/sol2.py
+++ b/project_euler/problem_013/sol2.py
@@ -1,0 +1,41 @@
+"""
+Problem 13: https://projecteuler.net/problem=13
+
+Problem Statement:
+Work out the first ten digits of the sum of the following one-hundred 50-digit
+numbers.
+"""
+
+import os
+
+
+def solution(n: int = 10) -> str:
+    """
+    Returns the first 'n' digits of the sum of the array elements
+    from the file num.txt. n should be larger than 2.
+
+    >>> solution(3)
+    '553'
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev:
Reviewed — **mergeable, but it's an additional solution to an already-solved problem, so it's a maintainer call on the duplicate policy.**

Correctness is fine: with the existing `problem_013/num.txt` in place, `solution(3) == '553'`, `solution(6) == '553737'`, and `solution() == '5537376230'` (the correct Euler #13 answer), and the doctests pass. (They only fail if run in isolation without `num.txt`, which won't happen in-repo.)

The merit, like `sol3` for problem 16, is that it sums the 100 numbers by explicit column-wise digit addition instead of leaning on Python big ints — a genuinely different technique, so it's not a pure copy of `sol1`. If we're accepting "manual arithmetic" variants as pedagogically distinct `solN` files, this qualifies. One nit: `import os` + reading `num.txt` is fine, but the leading-carry handling (`size_first`) is subtle enough to deserve a one-line comment.

Recommendation: **merge if we're keeping distinct-technique second solutions; otherwise decline on duplicate grounds.** I'm comfortable either way.

cclauss (approved):

PR #12684, opened 2025-04-25
Add solution for the euler project problem 111

mindaugl:
### Describe your change:

* [x] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
* [ ] Documentation change?

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have [doctests](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [x] All new algorithms include at least one URL that points to Wikipedia or another similar explanation.
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

PR #12777, opened 2025-06-02
Add third solution for the Euler project problem 16.

mindaugl:
### Describe your change:

Add solution for explicitly using the carry over of the multiplications to get the sum of digits, and not using the 
property of large integers in Python.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/project_euler/problem_016/sol3.py
+++ b/project_euler/problem_016/sol3.py
@@ -1,0 +1,44 @@
+"""
+Problem 16: https://projecteuler.net/problem=16
+
+2^15 = 32768 and the sum of its digits is 3 + 2 + 7 + 6 + 8 = 26.
+
+What is the sum of the digits of the number 2^1000?
+"""
+
+from numpy import log10, zeros
+
+
+def solution(power: int = 1000) -> int:
+    """
+    Calculates the sum of digits by explicit construction of digits array
+    without the help of Python's large integer property.
+
+    >>> solution(50)
+    76
+    >>> solution(20)
+    31
+    >>> solution(15)
+    26
+    >>> solution(3)
+    8
+    """
+
+    digits = zeros(int(power * log10(2)) + 1)
+    digits[0] = 1
+    dig_count = 1
+    for _ in range(power):
+        carry = 0
+        for d in range(dig_count):
+            digits[d] = digits[d] * 2 + carry
+            carry = digits[d] // 10
+            digits[d] %= 10
+        if carry > 0:
+            digits[dig_count] = carry
+            dig_count += 1
+
+    return int(sum(digits))
+
+
+if __name__ == "__main__":
+    print(solution())


cclauss:
@priya-sundaram-dev, please review this and @mindaugl's other open pull requests.

They are all related to the `project_euler/` directory, which I do not work on.  That directory is loaded with duplicates, which I do not really like.

### Is the proposed solution better, faster, or more correct than the existing solutions?

In reviewing these pull requests, please categorize them as:
* Merge -- Put your comments on the pull request.
* Do not merge
* Request changes -- Put your comments on the pull request so @mindaugl can make the changes.
* Rebase needed -- Put your comments on the pull request so @mindaugl can click the `Update branch` button.

priya-sundaram-dev:
**Verdict: Request changes** (small, then Merge).

The math is right — I ran it: `solution(1000) == 1366` and the doctests (`15→26`, `20→31`, `50→76`, `3→8`) all pass. As an additive third Project Euler solution it's fine in principle.

The one thing I'd fix before merge is the `numpy` usage:

- `zeros(...)` gives a **float64** array, so every digit op (`digits[d]*2 + carry`, `// 10`, `% 10`) is float arithmetic. It's exact here because digits stay in 0–9, but doing exact-integer digit manipulation in floats is fragile and reads oddly.
- It pulls in a numpy dependency for what is deliberately a "don't rely on big ints" exercise — a plain `list` of Python `int`s is dependency-free, avoids the float concern entirely, and is truer to the stated goal.
- `int(power * log10(2)) + 1` also uses float `log10` just to size the buffer; a growing list sidesteps that too.

A pure-stdlib version is a couple of lines different:

```python
def solution(power: int = 1000) -> int:
    digits = [1]
    for _ in range(power):
        carry = 0
        for i in range(len(digits)):
            digits[i] = digits[i] * 2 + carry
            carry, digits[i] = divmod(digits[i], 10)
        while carry:
            carry, d = divmod(carry, 10)
            digits.append(d)
    return sum(digits)
```

Same algorithm and carry idea, no numpy, all-integer. If @<author> makes that swap I'd flip this to **Merge**. Not a rebase issue — branch just needs the cleanup + green CI.

priya-sundaram-dev:
Reviewed all of @mindaugl's other open `project_euler/` PRs as requested. I ran each locally — full `solution()` against the known Project Euler answer, plus the doctests. Verdicts:

### ✅ Merge (correct, doctests pass, fast)
| PR | Problem | Verified | Notes |
|----|---------|----------|-------|
| #12694 | 137 Fibonacci Golden Nuggets | `solution(15)=1120149658760`; doctests `104`, `74049690` pass | Clean, O(n), nice derivation in the docstring. Nothing to change. |
| #12701 | 138 Special Isosceles Triangles | `solution(12)=1118049290473932`; doctests `322`, `1866293` pass | Correct Pell-equation recurrence. Tiny style nit only: `4*m_i + 5*m_i` → `9*m_i` (same for `l_i`). |
| #12706 | 142 Perfect Square Collection | `solution()=1006193` in <1s; doctests `1`, `9` pass | The `min_sum` break prunes well. Additive new problem. |
| #12764 | 13 (sol2) | `solution(10)=5537376230`; doctests `553`, `553737` pass | Genuinely different approach from sol1 (manual column addition, no big-int) so not a pointless dup. Minor: `numbers.append(line)` loop → `numbers = file.readlines()`, and the "n should be larger than 2" contract is a bit fragile. |

### 🔧 Request changes (correct result, small fix — then merge)
| PR | Problem | Issue | Fix |
|----|---------|-------|-----|
| #12763 | 5 (sol3) | The doctests are **commented out** (`# >>> solution(10)`), so the function ships with zero executed tests. I uncommented them locally and they pass (`2520`, `232792560`). | Just un-comment the three doctests. |
| #12690 | 124 Ordered Radicals | Correct (`solution()=21417`, doctests pass) but pulls in `from numpy import sqrt` only for an integer square root. | Use `math.isqrt` — drops the numpy dependency and is exact for ints. |
| #12684 | 111 Primes with Runs | Correct (`solution(10000)=273700`) but same `from numpy import sqrt` for integer sqrt. | Same fix: `math.isqrt`. |

(#12777 / problem 16 sol3 I reviewed separately above — same theme: correct, but swap the numpy float arrays for plain-int digit math.)

Net: **4 ready to merge**, **4 need a one-line change** (uncomment doctests / drop numpy for `math.isqrt` / int digits). None are wrong; the recurring nit is unnecessary `numpy` for integer math. Happy to push the small fixes as suggestions if that's easier for @mindaugl.

priya-sundaram-dev:
Reviewed. Recommendation: **Request changes** — the approach is genuinely distinct and worth having, but drop the numpy dependency in favor of plain-Python integer digits.

What's good: existing `sol1`/`sol2` lean on Python's arbitrary-precision ints (`2**power` then sum the string's digits). This PR instead does the doubling with an explicit digit array — that's a pedagogically different idea (how you'd sum the digits of 2^1000 *without* big-int support), so it earns its place as a `sol3` rather than being a pure duplicate. Output is correct (`solution(1000) == 1366`, and all four doctests check out).

What to change: it uses `numpy` (`log10`, `zeros`) only to size and hold the digit array, which is the wrong tool here:
- `zeros(...)` gives a **float64** array, so the whole "avoid big ints, do exact digit arithmetic" premise is being carried out in floating point. It happens to stay exact because every value is ≤ 19, but that's luck, not design.
- Sizing with `int(power * log10(2)) + 1` is a float computation for what is really `floor(power * log10(2)) + 1`.
- Pulling in numpy for a project-euler digit-sum is a heavy dependency the repo generally avoids when pure Python does the job.

A dependency-free version is just as short and fully exact:

```python
def solution(power: int = 1000) -> int:
    digits = [1]
    for _ in range(power):
        carry = 0
        for i in range(len(digits)):
            value = digits[i] * 2 + carry
            digits[i] = value % 10
            carry = value // 10
        while carry:
            digits.append(carry % 10)
            carry //= 10
    return sum(digits)
```

Same doctests pass. If @mindaugl reworks it this way (no numpy, `list[int]`), I'd be happy to see it merged as `sol3`.

PR #12812, opened 2025-06-28
merge_sort.py

30November:
Here after looking the actual code, I found out that the storing the sorted element in the result one by one on comparing two lists, that is left and right, it was using the pop() function for the seek of reducing the line of code. I wanted to alert about the pop function, as pop function in python extract the element on particular valid index from the list, mentioning in argument (by default the last element of the list). But pop off the first element takes O(n) complexity (where n is the length of the list) which leads to extra time complexity, also kind of making worst case for already sorted list.

Also, there is no need of extracting element, instead we just access the element through pointer ( left_index , right_index )

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/sorts/merge_sort.py
+++ b/sorts/merge_sort.py
@@ -37,10 +37,20 @@ def merge_sort(collection: list) -> list:
         :return: Merged result
         """
         result = []
-        while left and right:
-            result.append(left.pop(0) if left[0] <= right[0] else right.pop(0))
-        result.extend(left)
-        result.extend(right)
+        left_length = len(left)
+        right_length = len(right)
+        left_index = right_index = 0
+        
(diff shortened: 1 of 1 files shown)


30November:
I have worked on <strong> merge_sort.py </strong>, not on the others file
<i>Its getting halt on the test cases of the other file.<br></i>
Message shown : <p. FAILED <mark> web_programming/current_stock_price.py::web_programming.current_stock_price.stock_price </mark>
============ 1 failed, 1969 passed, 5 warnings in 79.33s (0:01:19) =============<br>
Error: Process completed with exit code 1. </p>

mindaugl (approved):

Issue #12732, opened 2025-05-12
Replace dependency `requests` with `httpx`


cclauss:
### Feature description

https://pypi.org/project/requests is becoming less popular.
https://pypi.org/project/httpx is a next-generation HTTP client for Python.

Edit the file `requirements.txt` to remove `requests` and add `httpx`.  Keep the file sorted alphabetically.
https://github.com/TheAlgorithms/Python/blob/485f688d0683ca026959e1b68f12e5d221724860/requirements.txt#L11

Edit the file `.pre-commit-config.yaml` to remove `additional_dependencies: [types-requests]`
https://github.com/TheAlgorithms/Python/blob/485f688d0683ca026959e1b68f12e5d221724860/.pre-commit-config.yaml#L58

Modify the 36 Python files that `import requests` to use httpx instead.
* https://github.com/search?q=repo%3ATheAlgorithms%2FPython+%22import+requests%22&type=code

If an algorithm does multiple `httpx` calls, consider if the `AsyncClient` should be used.

Add PEP723 headers to allow the scripts to be run standalone.
```diff
+ # /// script
+ # requires-python = ">=3.13"
+ # dependencies = [
+ #     "httpx",
+ # ]
+ # ///
```
* #12731

lukazlim:
Hi, I would like to take this issue, by when would you like the change to be completed?

cclauss:
No rush.  Take your time to do it right.  We do not assign issues, so someone else might deliver a PR before yours.

lukazlim:
Submitted a PR, the changes include a .DISABLED file which triggers algorithms-keeper to automatically close the PR

barnavo05:
Hi, I’d like to work on this issue. Could you please assign it to me?

cclauss:
You want to work on an issue that is already closed?!?

As discussed in CONTRIBUTING.md, we do not assign issues.

barnavo05:
Apologies for the confusion. I missed that the issue was already closed and misunderstood the contribution process. Thank you for pointing it out — I’ll make sure to check the issue status and CONTRIBUTING.md more carefully before commenting next time.

PR #12821, opened 2025-07-05
Various ruff fixes

cclauss:
### Describe your change:

Make changes required to pass current ruff tests.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/ciphers/hill_cipher.py
+++ b/ciphers/hill_cipher.py
@@ -79,7 +79,7 @@ class HillCipher:
         >>> hill_cipher.replace_digits(26)
         '0'
         """
-        return self.key_string[round(num)]
+        return self.key_string[(num)]
 
     def check_determinant(self) -> None:
         """
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -443,6 +443,7 @@
   * [Present Value](financial/present_value.py)
   * [Price Plus Tax](financial/price_plus_tax.py)
   * [Simple Moving Average](financial/simple_moving_average.py)
+  * [Straight Line Depreciation](financial/straight_line_depreciation.py)
   * [Time And Half Pay](financial/time_and_half_pay.py)
 
 ## Fractals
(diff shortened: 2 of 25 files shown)


ciphers/hill_cipher.py
```diff
@@ -79,7 +79,7 @@ def replace_digits(self, num: int) -> str:
         >>> hill_cipher.replace_digits(26)
         '0'
         """
-        return self.key_string[round(num)]
+        return self.key_string[(num)]
```
cclauss, line 82: @dhruvmanila I do not understand this change requested by ruff.  Type hints are suggestions, not mandatory in Python.  Just because the type hint says `num: int`, that does not guarantee that the caller will not pass in `17.23`.  This call to `round()` prevents breakage.

cclauss:
@ lighting9999 Let's use this PR to make the fixes to web_programming/current_stock_price.py

We use httpx instead of requests.
* #12732

mindaugl (approved):

MaximSmolskiy (approved):

ciphers/hill_cipher.py
dhruvmanila: I think this is correct and is https://docs.astral.sh/ruff/rules/unnecessary-round/. The `num` is an `int` so it doesn't require a `round` call. If you expect `num` to be a `float` as well then the type hint should be updated to `int | float` which is basically `float`.

cclauss: This misses the point.  The type hint is a suggestion in Python, not an obligation.  The caller may choose to send a float even though I have suggested they send an int.  Without the `round()` call, the function will crash.  With the `round()` call, the function will succeed.  This rule is asking me not to drive defensively.

The workaround for `ruff rule RUF057` is to use `int(num)` instead of `round(num)`:
* #12833

dhruvmanila: I don't think that's correct. From a static analysis perspective, type hints are part of the function contract that says that the `num` parameter here should be an `int` and if the caller passes a `float` then that's an error. So, if the function is expecting to receive a `float` or an `int`, then the contract should be updated instead.

cclauss: The `num` parameter here should be an `int`.  Python will not prevent the caller from breaking that rule.

Using `int(num)` instead of `round(num)` ensures the function remains well-behaved in the face of garbage input.

dhruvmanila: I see. I think that depends on whether this specific function is part of the public API for this module or not. If it is, then one way to perform error handling would be to use something like:
```py
if not isinstance(num, int):
    raise TypeError("expected an integer, but got ...")
```
And, if it's not part of the public API, then we shouldn't do this and instead use a type checker.

PR #12842, opened 2025-07-12
Optimize the Merge Sort 

30November:
### Describe your change:
After reviewing the code, I noticed that while storing the sorted elements in the result one by one by comparing two lists, left and right, the pop() function was used to reduce the lines of code. I wanted to highlight a concern about the pop() function. In Python, the pop() function removes an element at a specified valid index from the list (defaulting to the last element if no index is provided). However, popping the first element has a time complexity of O(n) (where n is the length of the list), leading to additional time complexity and potentially creating a worst-case scenario for an already sorted list.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/sorts/merge_sort.py
+++ b/sorts/merge_sort.py
@@ -37,10 +37,19 @@ def merge_sort(collection: list) -> list:
         :return: Merged result
         """
         result = []
-        while left and right:
-            result.append(left.pop(0) if left[0] <= right[0] else right.pop(0))
-        result.extend(left)
-        result.extend(right)
+        left_index = right_index = 0
+        length_left , length_right = len(left) , len(right)
+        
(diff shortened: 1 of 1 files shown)


30November:
> Docstring Edge Case Note You could note in the main docstring that this implementation is not in-place and uses additional memory due to list slicing and merging.

Yeah true 
Actually, I am focusing on the time complexity (TC), rather than the space complexity (SC).
Anyhow, SC will remain O(n) weather it is using slicing or not.
Great that you have observed and analysis it in detail, earlier I also knew to this situation about using extra minimal space which can be avoid. But it doesn't harm the complexity mention inside the file by comment.

Thus, it is just about to spread knowledge that correctly matches with the practical (code) one.

cclauss:

PR #12896, opened 2025-08-13
Add graph generation algorithm for Knapsack problems

Matti02co:
### Description:
This pull request adds a function to generate a directed acyclic graph (DAG) representation of a classic 0/1 Knapsack problem.
The graph representation enables solving the problem using shortest/longest path algorithms in DAGs, providing an alternative approach to traditional dynamic programming solutions.
Each path from the source node s to the sink node t corresponds to a valid subset of items.

#### Input:

- capacity — maximum knapsack capacity (integer)

- weights — list of item weights (list of integers)

- values — list of item values (list of integers)

#### Output:

- List of edges, each represented as a dictionary containing "from", "to", "cost", and "label".

Implements Python type hints and doctests.

Code style follows PEP 8 and is Black-formatted.

Passes Ruff tests.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/knapsack/knapsack_graph_generation.py
+++ b/knapsack/knapsack_graph_generation.py
@@ -1,0 +1,112 @@
+"""
+This function builds a Directed Acyclic Graph representation of the Knapsack problem.
+This allows us to solve knapsack-style problems with Shortest Path algorithms.
+See https://github.com/Matti02co/graph-based-scheduling for more details.
+
+The graph consists of n+2 layers:
+- Layer 0 contains the source node 's'.
+- Layer n+1 contains the sink node 't'.
+- Intermediate layers (1..n) correspond to the n items.
+
+Each intermediate layer j has (capacity + 1) nodes: j0, j1, ..., jC,
+where jk represents the state of having considered the first j items
+with a total weight of k so far.
+
+From each node jk there are at most two outgoing edges:
+- Skip item j+1 (weight and cost remain the same).
+- Take item j+1 (only if total weight + item's weight ≤ capacity), with a cost
+  equal to the negative of the item's value (we solve via shortest path).
+
+All nodes in the last layer are connected to 't' with zero-cost edges.
+
+In this representation, every path from 's' to 't' corresponds to a feasible
+Knapsack solution, and the shortest path (negative costs) corresponds to
+the maximum total value selection.
+"""
+
+from typing import Any
+
+
+def generate_knapsack_graph(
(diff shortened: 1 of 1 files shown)


cclauss:
@priya-sundaram-dev, please review.  I have not done much with the knapsack, so perhaps I am missing something, but is this really adding educational value?  When would someone reach for this algorithm?

There are tons of single-letter variable names, which make the logic tricky to follow... w vs. weight vs. weights ... Is n even needed?

Functions return `dict[str, Any]` which provides the caller no help.  Would a dataclass or a typed dict make more sense?

priya-sundaram-dev:
Reviewed, @cclauss — I share your skepticism. The *idea* is genuinely nice: reduce 0/1 knapsack to a shortest-path problem on a layered DAG (each layer = one item, each node = a (item, weight-used) state, "take" edges carry `-value`). Connecting knapsack to SPP is a real teaching moment. But as submitted I don't think it's mergeable, for a few concrete reasons:

1. **It never actually solves anything.** `generate_knapsack_graph()` only emits a list of edges — there's no shortest-path call, so a reader can't see knapsack *solved*. An "algorithm" file that stops at building the input to another algorithm is incomplete. To add educational value it should run (e.g.) Bellman-Ford over the DAG and return the optimal value/selection, ideally cross-checked against the classic DP in a doctest.

2. **Types, as you noted.** `list[dict[str, Any]]` with stringly-typed keys `"from"/"to"/"cost"/"label"` gives the caller no help and defeats type checking. A `NamedTuple`/`dataclass Edge(frm, to, cost, label)` would be self-documenting and mypy-friendly.

3. **Inconsistent node identities.** Interior nodes are tuples `(i, w)` but source/sink are strings `"s"/"t"` — mixing types in the same graph is a smell.

4. **Reference is a personal repo** (`Matti02co/graph-based-scheduling`). TheAlgorithms prefers a self-contained implementation with an authoritative reference (Wikipedia "Knapsack problem" / "Shortest path").

5. Minor: single-letter `w`/`n`, and skip-edges are emitted from every `(i, w)` including unreachable states (harmless but noisy).

**Verdict:** the DAG-to-SPP framing is worth encouraging, but this needs (1) an actual solve step with a doctest verifying it matches the DP result, and (2) a typed edge structure. Given the pre-Oct-1 push, I'd **request changes**; if @author can't turn it around, close with a pointer to reopen once it solves the problem end-to-end. Happy to re-review if they add the solver.

PR #12974, opened 2025-09-17
Add Sieve of Atkin algorithm for efficient prime generation

AyhamJo7:
This PR adds an implementation of the **Sieve of Atkin** algorithm to the maths directory.


priya-sundaram-dev (commented):
Reviewed as part of the maths/ awaiting-reviews triage (#15081).

**Verdict: correct algorithm, genuinely new — recommend merge after removing one stray file.**

**Not a duplicate.** We have several Eratosthenes-family sieves (`sieve_of_eratosthenes.py`, `prime_sieve_eratosthenes.py`, `segmented_sieve.py`, `odd_sieve.py`), but the Sieve of Atkin is a genuinely different algorithm (quadratic-form based, not trial-marking of multiples), so it passes the no-duplicate bar.

**Correctness — verified.** I diffed the output against an independent Eratosthenes reference for **every limit 0–400 plus 1000, 5000, 10007, 50000, 100000**: **identical prime sets, 0 discrepancies.** Edge cases (0, 1, 2, negative → `ValueError`) all match the doctests.

**One change required before merge:**
1. **Remove `pr_description.txt`** — it's a PR-writeup file that shouldn't be committed to the repo. Only `maths/sieve_of_atkin.py` should ship. (This is likely also why a reviewer would hesitate; drop it and the PR is a single clean module.)

**Optional polish (non-blocking):**
2. The docstring's `O(n / log log n)` refers to the *theoretical optimized* Atkin; this straightforward `list`-of-bools implementation is closer to `O(n)` work with a smaller constant than a segmented/wheel version. A one-line "this is the basic, non-segmented form" note would set reader expectations accurately.

Clean, well-documented, good doctest coverage on the edge cases. Just drop the stray `.txt` and this is ready.

cclauss (approved):

PR #12976, opened 2025-09-17
Add K-means++

leolamien:
### Describe your change:
Adding K-means++ algorithm based on the existing naive K-means algorithm (k_means_clust.py)

- Implements k-means++ initialization for better centroid selection
- Improves convergence speed and avoids poor cluster initialization

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


leolamien:
Hi!
This PR adds a new K-means++ algorithm based on the naive K-means implementation.
The results are the same, but it converges faster than naive K-means, and in some cases the clustering quality is clearly better.
Please let me know if any improvements are needed. Thanks!

cclauss:

priya-sundaram-dev (commented):
Thanks for the contribution, @leolamien — I ran the code locally and the core K-means++ seeding works well: on a synthetic 3-cluster dataset it recovers all three centroids cleanly, and the D²-weighted seeding is implemented correctly. A few suggestions to bring it in line with the repo's conventions and make it more robust:

**1. Add doctests to the algorithmic functions.** Right now only `report_generator` has one. TheAlgorithms convention is that each function carries a small, deterministic doctest — and `get_initial_centroids_kmeans_plus_plus`, `assign_clusters`, `revise_centroids`, and `compute_heterogeneity` are all easy to test with a tiny fixed array + `seed`. For example:

```python
>>> import numpy as np
>>> data = np.array([[0.0, 0.0], [0.0, 1.0], [10.0, 10.0], [10.0, 11.0]])
>>> centroids = get_initial_centroids_kmeans_plus_plus(data, 2, seed=42)
>>> centroids.shape
(2, 2)
```

**2. Empty-cluster robustness.** If a cluster loses all its members during iteration, `revise_centroids` computes `member_data_points.mean(axis=0)` over an empty slice, which produces `nan` centroids (I reproduced this). That failure is currently hidden by the module-level `warnings.filterwarnings("ignore")`. I'd suggest removing the global warning filter (it silences *all* warnings for anyone importing this module) and instead guarding the empty case explicitly, e.g. keep the previous centroid when a cluster is empty.

**3. Dead code.** In `get_initial_centroids_kmeans_plus_plus`, `probabilities = []` is assigned and then immediately overwritten a few lines later — it can be dropped.

**4. Type hints.** The repo requires type hints on function signatures; most functions here are missing them (`data: np.ndarray, k: int, seed: int | None = None -> np.ndarray`, etc.).

**5. Scope.** The `if False:` mock-test block and the large `report_generator` pandas utility aren't really part of the K-means++ algorithm and add a lot of surface area to the file. Consider replacing the `if False:` block with proper doctests, and either dropping `report_generator` or moving it out — a focused, self-contained algorithm file is easier to review and maintain. (Minor: the header lists `Python: 3.5`, which is well past EOL.)

None of this takes away from the algorithm itself, which is solid. Happy to re-review once the doctests/type hints are in.

*(Disclosure: I'm an AI agent helping triage and review contributions; a human maintainer makes the final merge call.)*

priya-sundaram-dev (changes requested):
Thanks for the contribution, @Leonce-Wilson — K-means++ seeding is a genuinely nice addition. I reviewed it carefully; a couple of things should be addressed before merge.

### 1. This file largely duplicates `machine_learning/k_means_clust.py`
Nine of the ten functions here (`get_initial_centroids`, `centroid_pairwise_dist`, `assign_clusters`, `revise_centroids`, `compute_heterogeneity`, `kmeans`, the `plot_*` helpers, and the ~180-line `report_generator`) are copied almost verbatim from the existing `k_means_clust.py`. The only genuinely new logic is `get_initial_centroids_kmeans_plus_plus`.

Carrying two near-identical k-means implementations is a maintenance hazard (fixes/bugs have to be kept in sync). I'd suggest one of:
- add just `get_initial_centroids_kmeans_plus_plus()` to the existing `k_means_clust.py` and expose it as an optional `init="k-means++"` path, **or**
- keep a small standalone module that contains *only* the k-means++ seeding function and imports the shared helpers from `k_means_clust`.

Either way, please drop the copied `report_generator` — it's unrelated to k-means++.

### 2. Correctness bug: division by zero when D² sums to zero
When every remaining point already coincides with a chosen centroid (e.g. `k` greater than the number of distinct points, or duplicate/identical rows), `np.sum(squared_distances)` is `0`, so `probabilities` becomes all-`NaN` and `rng.choice` raises. Reproducer:

```python
import numpy as np
dup = np.array([[0.0, 0.0]] * 5 + [[1.0, 1.0]] * 5)  # only 2 distinct points
get_initial_centroids_kmeans_plus_plus(dup, 4, seed=1)
# ValueError: probabilities contain NaN
```

Minimal guard (falls back to a uniform draw when all D² are zero):

```python
total = np.sum(squared_distances)
probabilities = (
    np.full(n, 1.0 / n) if total == 0 else squared_distances / total
)
```
I confirmed this makes both the `k > #distinct` and all-identical cases return cleanly.

### 3. Minor
- `probabilities = []` at the top of the loop is dead — it's immediately overwritten. Remove it.
- Module-level `warnings.filterwarnings("ignore")` hides warnings globally for anyone importing this file; better to not suppress, or scope it narrowly.
- The new seeding logic has no doctest, so CI never exercises it. A tiny deterministic doctest (fixed `seed=`, well-separated points, asserting shape / that the two far clusters are picked) would lock in behavior — the repo's [contributing guide](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md#type-hints) also asks for doctests + type hints on new functions.

Happy to re-review once the duplication is trimmed and the zero-distance guard is in. Nice work getting the D² weighting right.

priya-sundaram-dev:
@cclauss @leolamien Reviewed. The core k-means++ seeding is correct — the D² weighting in `get_initial_centroids_kmeans_plus_plus` is exactly right (min distance to already-chosen centroids, squared, normalized to a probability, then `rng.choice`). Nice use of `np.random.default_rng` for reproducibility. A few things I'd fix before merge:

1. **Large duplication with the existing `machine_learning/k_means_clust.py`.** `get_initial_centroids`, `centroid_pairwise_dist`, `assign_clusters`, `revise_centroids`, `compute_heterogeneity`, `kmeans`, and the entire `report_generator` (plus its doctest) are copied almost verbatim from that file. That's ~350 of the 438 lines. The genuinely new contribution is just the `++` initialization. Could this be a small addition to `k_means_clust.py` (or import the shared helpers from it) instead of a near-duplicate module? It keeps the two implementations from drifting out of sync.

2. **The new function has no test.** `report_generator` carries a doctest (inherited from the original), but the actual k-means++ seeding — the point of the PR — isn't exercised. Please add a small deterministic doctest, e.g. build a tiny 2-cluster array, call `get_initial_centroids_kmeans_plus_plus(data, 2, seed=...)`, and assert the returned shape / that both seeds land in distinct clusters. CI runs the doctests, so this is what proves the algorithm works.

3. **`warnings.filterwarnings("ignore")` at import time** suppresses warnings globally for anyone who imports the module — best removed, or narrowed to a specific category inside the function that needs it.

4. **Missing type hints.** New files here are expected to type-annotate parameters and returns (the `ty` check enforces this); the public functions (`get_initial_centroids_kmeans_plus_plus`, `kmeans`, etc.) are currently untyped.

5. Minor: `probabilities = []` at the top of the loop is dead — it's immediately overwritten by the array assignment. And `centroids = initial_centroids[:]` reads more clearly as `initial_centroids.copy()` for a NumPy array.

The `if False:` demo block is fine as an opt-in manual visualization. Happy to re-review once the duplication and the missing test are addressed — the ++ seeding itself is solid.

cclauss:
Cool changes but not aligned with CONTRIBUTING.md.

PR #12992, opened 2025-09-24
feat: optimizing the prune function at the apriori_algorithm.py archive

joaoneto9:
### Describe your change:

Added an optimized version of the `prune` function using `Counter` to improve performance
when checking candidate itemsets for frequent items.

I used as a test base a gradually increasing size of the `itemset` list to demonstrate
the inefficiency of the original algorithm, which had a complexity of O(n * c * i), 
where n is the size of `itemset`, c is the number of candidates, and i is the number of 
items in each candidate. 

The new solution reduces the complexity to O(n + c * i). Previously, the algorithm would
iterate over `itemset` (O(n)) and count occurrences for each item (O(n)) every time it
needed to check a candidate, resulting in repeated costly operations. 

To optimize this, I used an auxiliary dictionary (via `Counter`) where each key is an 
item and its value is the number of occurrences in `itemset`. This allows both the check 
and count operations to be performed in constant time O(1). 

As a result, the performance improvement is significant, at the cost of a small additional 
memory usage, which is a worthwhile trade-off. This improvement can be observed by 
comparing the execution of both algorithms (as shown in the attached image).

Here is the graph comparing both functions: 
[pruneOptimized_prune_algoritm_results.pdf](https://github.com/user-attachments/files/22521191/pruneOptimized_prune_algoritm_results.pdf)

Unit tests were also conducted on my local machine to ensure the consistency of results between the two methods, but they are not included in this PR.

* [ ] Add an algorithm?
* [x] Documentation change?

### Checklist:

### DIFF
--- a/machine_learning/apriori_algorithm.py
+++ b/machine_learning/apriori_algorithm.py
@@ -45,7 +45,7 @@ def prune(itemset: list, candidates: list, length: int) -> list:
     >>> prune(itemset, candidates, 3)
     []
     """
-    itemset_counter = Counter(tuple(x) for x in itemset)
+    itemset_counter = Counter(tuple(item) for item in itemset)
     pruned = []
     for candidate in candidates:
         is_subsequence = True


joaoneto9:
I hadn't realized that the itemset could be a list of lists. As a result, hashing these data structures was not possible, so I switched to using tuples, which are immutable, as keys for the Counter. After this change, I noticed a slight overhead, since each item now needs to be converted into a tuple to be checked within the Counter structure. Nonetheless, there is a significant efficiency gain in the worst-case scenario, and I believe it will also improve performance in average cases. I have not yet tested these other scenarios or generated their corresponding graphs. Below is the graph reflecting the new modification.

[pruneOptimized_prune_algoritm_results.pdf](https://github.com/user-attachments/files/22522246/pruneOptimized_prune_algoritm_results.pdf)

MaximSmolskiy (approved):

PR #13173, opened 2025-10-04
Add optimized and standard implementations for Product of Array Except Self algorithm

kokatesaurabh:
# Product of Array Except Self

## 🧠 Problem Statement
Given an integer array `nums`, return an array `answer` such that  
`answer[i]` equals the **product of all elements in `nums` except `nums[i]`**.

The solution must run in **O(n)** time and use **O(1)** extra space (excluding the output array).  
Division is **not allowed** in the main approach.

---

### 1️⃣ Prefix–Suffix Approach (No Division)
This approach calculates prefix and suffix products in two passes.

**Steps:**
1. Compute prefix products from left to right.
2. Compute suffix products from right to left.
3. Multiply corresponding prefix and suffix products for each index.

**Complexity:**  
- Time: O(n)  
- Space: O(1) (excluding output)

---

### 2️⃣ Optimized Approach (Handles Zeros Efficiently)
This version handles:
- Multiple zeros → all outputs are `0`
- Single zero → only that index gets the product of non-zero elements
- No zeros → uses division safely

---

## 🧩 Example
```python
>>> Solution().productExceptSelf([1, 2, 3, 4])
[24, 12, 8, 6]
>>> Solution().productExceptSelf([-1, 1, 0, -3, 3])
[0, 0, 9, 0, 0]

### DIFF
--- a/data_structures/arrays/ProductOfArrayExceptSelf.py
+++ b/data_structures/arrays/ProductOfArrayExceptSelf.py
@@ -50,6 +50,24 @@ class Solution:
 
         return answer
 
+# #optimized Answer
+# class Solution(object):
+#     def productExceptSelf(self, nums):
+#         m = 1
+#         z = 0
+#         for i in nums:
+#             if i !=0:
+#                 m*= i
+#             else:
+#                 z+=1
+
+#         if z==1:
+#             return [m if i==0 else 0 for i in nums]
+#         if z>1:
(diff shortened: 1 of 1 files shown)


kokatesaurabh:
review the code and let me know i would like to contribute more in python

kokatesaurabh:
ok

PR #13176, opened 2025-10-04
Improve softmax: type checks, stability, axis support, tests, docs

### LINKED ISSUE #13171: Softmax function missing type checking and other enhancements
### Feature description

The softmax function does not have specification for the input type required to compute softmax. At the least, the function argument should have strict type checking, in place, to prevent crashing on non-supported inputs such as string or dict.

Other recommendations are:

1. Numerical stability: subtract the max on the reduction axis before exponentiating to avoid overflow.
2. Make the function N‑D and axis-aware: add an axis parameter (default -1) so it works for 1D/2D/3D tensors and is explicit about where probabilities sum to 1.
3. Input normalization and dtype: convert inputs with np.asarray(..., dtype=float) so lists/tuples/ints behave correctly.
4. Use keepdims=True on reductions so broadcasting keeps shapes correct.
5. Add simple type hints and a clearer docstring describing inputs/outputs and axis semantics.
6. Add unit tests (happy path + edge cases: scalar, single element, identical logits, large values, empty) and a small __main__ demo.
7. Defensive checks: raise meaningful error for empty input and document behavior for NaNs/Infs.

A sample improved version is as follows:

```
import numpy as np
from typing import Optional

def softmax(vector, axis: Optional[int] = -1) -> np.ndarray:
    """
    Compute the softmax of `vector` along `axis` in a numerically-stable way.

    Parameters
    ----------
    vector : array_like
        Input data (vector, matrix, tensor). Will be converted to float ndarray.
    axis : int or None, optional
        Axis along which to compute softmax. If None, compute softmax over
        the flattened array (single distribution). Default is -1 (last axis).

    Returns
    -------
    np.ndarray
        Same shape as `vector`, with softmax applied along `axis`. Probabilities sum to 1
        along `axis` (or to 1 overall if axis is None).

    Raises
    ------
    ValueError
        If input is empty.
    """
    try:
        x = np.asarray(x, dtype=float)
    except Exception as e:
        raise ValueError(f"Could not convert input to float ndarray: {e}")
    
    if vector.size == 0:
        raise ValueError("softmax input must be non-empty")

    if axis is None:
        # flatten to single distribution
        vector_max = np.max(vector)
        e_vector = np.exp(vector - vector_max)
        return e_vector / e_vector.sum()

    # subtract max along axis with keepdims for numerical stability/broadcasting
    vector_max = np.max(vector, axis=axis, keepdims=True)
    e_vector = np.exp(vector - vector_max)
    denom = e_vector.sum(axis=axis, keepdims=True)
    return e_vector / denom

if __name__ == "__main__":
    print(softmax((0,)))

```
credit to Github Copilot

shreyasNaik0101:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #13171".

### DIFF
--- a/maths/softmax.py
+++ b/maths/softmax.py
@@ -1,10 +1,10 @@
 import numpy as np
-from typing import Optional, Union
+from typing import Optional
 
 
 def softmax(
-    vector: Union[np.ndarray, list, tuple],
-    axis: Optional[int] = -1
+    vector: np.ndarray | list | tuple,
+    axis: int | None = -1
 ) -> np.ndarray:
     """
     Compute the softmax of `vector` along `axis` in a numerically-stable way.
@@ -20,8 +20,8 @@ def softmax(
     Returns
     -------
     np.ndarray
-        Same shape as `vector`, with softmax applied along `axis`. Probabilities sum to 1
-        along `axis` (or to 1 overall if axis is None).
+        Same shape as `vector`, with softmax applied along `axis`. Probabilities
+        sum to 1 along `axis` (or to 1 overall if axis is None).
 
     Raises
     ------
@@ -30,8 +30,9 @@ def softmax(
     """
     try:
         vector = np.asarray(vector, dtype=float)
-    except Exception as e:
-        raise ValueError(f"Could not convert input to float ndarray: {e}")
+    except TypeError as e:
+        msg = f"Could not convert input to float ndarray: {e}"
+        raise ValueError(msg)
 
     if vector.size == 0:
         raise ValueError("softmax input must be non-empty")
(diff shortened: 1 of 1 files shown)


maths/softmax.py
```diff
@@ -1,56 +1,96 @@
...
-    Returns:
-        softmax_vec (np.array): The input numpy array  after applying
-        softmax.
+    # Typical 1D input
+    result = softmax([1, 2, 3])
+    npt.assert_almost_equal(result.sum(), 1)
 
-    The softmax vector adds up to one. We need to ceil to mitigate for
-    precision
-    >>> float(np.ceil(np.sum(softmax([1,2,3,4]))))
```
cclauss, line 31: Removing doctests goes against the recommendations in CONTRIBUTING.md.

PR #13231, opened 2025-10-05
Add pytest for heap_sort

NavanithaSri:
- [x] I have read the contribution guidelines
- [x] My code follows PEP8 style
- [x] Added pytest or doctest
- [x] My code is tested
- [ ] Other points not relevant left unchecked


cclauss:
@priya-sundaram-dev I merged this one because it would be good to start following the Python convention of having a `tests/` directory.  We can use if for pytests that cover similar algorithms like a parameterized tests.  For all algorithms in the `sorts` directory, sort([10, -10, -1, 1, 0]), sort([1.1, -1.1, -1, 1, 0]), sort("Python!").

cclauss (approved):

priya-sundaram-dev:
Love it — a `tests/` directory with parameterized tests is the right call. I ran with your three examples in #15185: a single `test_sort_matches_builtin` runs **20 general-purpose sorts × 9 inputs** (your negatives / floats / `"Python!"` string, plus empty, single, duplicate-heavy, sorted, and reverse-sorted), all green, with the original `test_heap_sort` kept intact.

Two real bugs fell out while I was at it (kept out of the green suite so CI stays honest — can fix in separate PRs on your word):
- `tim_sort([])` → `IndexError`
- `tree_sort([3,3,1,2,2,1])` → `[1,2,3]` (drops duplicates)

The module docstring documents why the domain-restricted sorts (`counting_sort`, `radix_sort`, `bead_sort`, `dutch_national_flag_sort`, `bitonic_sort`, `topological_sort`, …) are excluded rather than silently skipped.

<sub>🤖 #ABotWroteThis</sub>

PR #13452, opened 2025-10-12
Implement modified vigenere cipher for all printable ASCII

Abheelash-Mishra:
### Describe your change:
Added a modifed Vigenere cipher script that encrypts and decrypts messages using only printable ASCII characters (32–126). This maintains the original Vigenere cipher behavior with cyclic key repetition, ensures all output remains readable.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -109,6 +109,7 @@
   * [Morse Code](ciphers/morse_code.py)
   * [Onepad Cipher](ciphers/onepad_cipher.py)
   * [Permutation Cipher](ciphers/permutation_cipher.py)
+  * [Printable ASCII Vigenere Cipher](ciphers/printable_ascii_vigenere_cipher.py)
   * [Playfair Cipher](ciphers/playfair_cipher.py)
   * [Polybius](ciphers/polybius.py)
   * [Porta Cipher](ciphers/porta_cipher.py)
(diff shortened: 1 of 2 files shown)


mindaugl (approved):

cclauss:
How is this better, faster, or more correct than our existing ciphers/vigenere_cipher.py?

priya-sundaram-dev:
Thanks for the ping, @cclauss. I read through it — my honest take:

**On "better/faster/more correct" vs `ciphers/vigenere_cipher.py`:** it isn't any of those for the classic cipher. It's the same O(n) translate loop, so no speed win, and for letters-only input the existing file is already correct. What this PR actually is, is a *different variant*: it operates over the full printable-ASCII range (32–126) instead of A–Z, so it also enciphers digits, punctuation and spaces, and the key is any printable string rather than an alphabetic one. That's a legitimately larger keyspace/message-space, but it's a **new cipher, not an improvement to the old one** — and the repo generally discourages near-duplicates unless the added value is clear. I'd ask @Abheelash-Mishra to make that case explicitly in the PR description (why the printable-ASCII variant is worth carrying alongside the classic one) so it's a deliberate keep rather than an accidental fork.

**If it does stay, two concrete fixes:**

1. Empty key crashes instead of erroring cleanly:
   ```python
   >>> encrypt_message('', 'hi')
   ```
   Guard it up front (`if not key: raise ValueError("Key must be non-empty")`) and add a doctest for the raise.

2. `mode` is validated *inside* the per-character loop, so the `ValueError` only fires if the message has a printable char, and re-checks every iteration. Validate `mode` once before the loop.

Minor: the module-level `PRINTABLE_*` constants are nice; consider deriving `key_bytes` outside `translate_message` is already good. Doctests pass locally (6/6). Overall the code is clean and readable — the real question is scope/duplication, which is your call as maintainer.

PR #13609, opened 2025-10-20
feat(searches): Add Floyd's Cycle-Finding Algorithm

AdityaK60:
### Describe your change:

This pull request introduces an implementation of **Floyd's Cycle-Finding Algorithm**, also known as the "Tortoise and Hare" algorithm. This well-known algorithm is used to efficiently detect a cycle in a sequence of iterated values and also finds the cycle's starting point and length.

The file `floyds_cycle_finding.py` has been added to the `searches/` directory.

* [x] Add an algorithm?
* [ ] Add or change doctests?
* [ ] Documentation change?

### Checklist:


cclauss:
@priya-sundaram-dev Are these tests running in an infinite loop in CI?

priya-sundaram-dev:
Yes — confirmed locally. Two of the three doctests spin forever, which is why the `build` job never finishes:

- `floyds_cycle_finding(lambda x: (2*x+3)%17, 0)` → hangs
- `floyds_cycle_finding(lambda x: [1,2,3,4,5,3][x], 0)` → hangs
- the no-cycle example returns fast, but only because it hits the `hare is None` early-return before it can loop.

Root cause is the asymmetric setup in Phase 1: `hare` is initialized one step ahead (`hare = successor_function(start_value)`) and equality is tested at the *top* of the loop before moving. That means the value they meet on isn't guaranteed to sit at an index that's a multiple of the cycle length, which is the invariant Phase 2 relies on. So the Phase-2 `while tortoise != hare` loop never converges and just walks around the cycle forever.

The classic tortoise-and-hare form fixes it — start both at the same point and check equality *after* advancing:

```python
tortoise = successor_function(start_value)
hare = successor_function(successor_function(start_value))
while tortoise != hare:
    if tortoise is None or hare is None:
        return None
    tortoise = successor_function(tortoise)
    hare = successor_function(successor_function(hare))
```

With that Phase 1, both examples terminate. One heads-up though: the list case then returns `(3, 3)`, not the `(2, 3)` in the doctest — and `(3, 3)` is the correct answer, since the sequence is `0,1,2,3,4,5,3,4,5,…` and the cycle `[3,4,5]` first appears at index 3. So the expected value in that doctest is off by one on μ as well and should be updated once the loop is fixed.

Happy to open a small PR with the corrected Phase 1 + fixed doctests if that's useful.

priya-sundaram-dev:
@cclauss Yes — confirmed, two of the three doctests loop forever, which is why `build` never finishes. I traced it locally:

- `floyds_cycle_finding(lambda x: (2 * x + 3) % 17, 0)` → hangs
- `floyds_cycle_finding(lambda x: [1, 2, 3, 4, 5, 3][x], 0)` → hangs
- the "no cycle" example returns `None` fine (it terminates via the `hare is None` guard)

**Root cause** is the phase-1 initialization. The tortoise is seeded with the raw `start_value` (`tortoise = start_value`) while the hare is seeded one step in (`hare = successor_function(start_value)`). The classic tortoise-and-hare invariant needs *both* pointers to start one step in (tortoise at `f(x0)`, hare at `f(f(x0))`). With the current offset, the phase-1 meeting point no longer satisfies the property that phases 2 and 3 rely on, so for any input that is a pure cycle with no `None` sentinel, phase 2's `while tortoise != hare` never converges → infinite loop. The "no cycle" case only escapes because it hits `None` first.

Minimal fix — seed the tortoise one step in as well:

```python
tortoise = successor_function(start_value)
if tortoise is None:
    return None
hare = successor_function(tortoise)
```

With that change all three examples terminate immediately. One heads-up: the second doctest's expected value is also off — after the fix it returns `(3, 3)`, not `(2, 3)`. The sequence of values is `0, 1, 2, 3, 4, 5, 3, 4, 5, …`, so the cycle's first element (value `3`) sits at index `3`, giving `mu = 3`. Worth updating that expected output when you fix the loop.

Because an infinite loop in a doctest wedges the whole CI job, it'd be good to gate this on a bounded check before re-running. Happy to re-review once the init is fixed.

PR #13645, opened 2025-10-21
Add adjacency representation generator for closed Hamiltonian tours

Naman-Vasudev:
This pull request adds a new algorithm that converts a path (permutation form) of a closed Hamiltonian tour into an adjacency representation vector.
Each position in the resulting vector corresponds to a city (in fixed alphabetical order), and the value at that position denotes the next city in the tour.

### DIFF
--- a/genetic_algorithm/adjacency_representation.py
+++ b/genetic_algorithm/adjacency_representation.py
@@ -1,0 +1,72 @@
+# Wikipedia URL- https://en.wikipedia.org/wiki/Adjacency_list
+"""
+Closed Tour Adjacency Representation
+------------------------------------
+
+This converts a path representation (like A→L→G→C→...) into an
+adjacency vector form, where each position corresponds to a city in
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev (commented):
Thanks @Naman-Vasudev — clear docstring, type hints and doctests, and the closed-tour wrap-around logic reads well.

One robustness suggestion: `next_city_map.get(city, "-")` silently yields `"-"` when a node in `nodes` never appears in `path`. Since a valid tour should visit every node, I'd validate `set(path) == set(nodes)` up front and raise a `ValueError` on mismatch, so malformed input fails loudly instead of producing a `"-"` placeholder.

This pairs with your #13647 (ordinal representation) — both are TSP tour encodings used in GA crossover. It might read more cohesively if the two docstrings cross-reference each other (or live together), so a reader sees them as a family of representations.

PR #13690, opened 2025-10-22
Fix ellipse perimeter calculation

avyukth-fll:
Fixed perimeter calculation for the Ellipse class to use Ramanujan's first approximation (based on https://en.wikipedia.org/wiki/Perimeter_of_an_ellipse#First_approximation)

The previous formula was less accurate

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/geometry/geometry.py
+++ b/geometry/geometry.py
@@ -103,9 +103,12 @@ class Ellipse:
     def perimeter(self) -> float:
         """
         >>> Ellipse(5, 10).perimeter
-        47.12388980384689
+        48.44210548835644
         """
-        return math.pi * (self.major_radius + self.minor_radius)
+        a, b = self.major_radius, self.minor_radius
+        # Use Ramanujan's first approximation for ellipse perimeter
+        # https://en.wikipedia.org/wiki/Perimeter_of_an_ellipse#First_approximation
+        return math.pi * (3 * (a + b) - ((3 * a + b) * (a + 3 * b)) ** 0.5)
 
 
 class Circle(Ellipse):


cclauss:
@priya-sundaram-dev, please review. Is the accuracy claim valid?  Is the new property useful?

priya-sundaram-dev:
Thanks for the ping @cclauss — I checked the numbers and the change is correct and a real improvement.

Comparing against the exact perimeter (elliptic integral, `4a·E(e)`) for `Ellipse(5, 10)` (semi-axes 10 and 5):

| formula | value | rel. error |
|---|---|---|
| exact `4a·E(e)` | 48.442241… | — |
| old `π(a+b)` | 47.123890… | **−2.72%** |
| this PR (Ramanujan I) | 48.442105… | **−0.00028%** |

So yes — the accuracy claim holds. The old `π(a+b)` is just the circumference of a circle with radius `(a+b)/2`; it's only exact when `a == b`, and drifts off by a couple of percent for a 2:1 ellipse (worse as eccentricity grows). Ramanujan's first approximation stays within ~3e-4 % here and degrades very gracefully, so it's a genuinely useful property to expose.

Two small suggestions, neither blocking:
- Worth noting in the docstring that this is an *approximation* (Ramanujan's first), since a reader might assume a closed form exists — a one-line note plus the Wikipedia link you already added would do it.
- Optional: a doctest at higher eccentricity (e.g. `Ellipse(1, 10)`) would document that it stays accurate away from the near-circular case.

LGTM on the math. 👍

cclauss (approved):

PR #13707, opened 2025-10-23
Add reverse_string algorithm with type checking and examples

muannn:
### Describe your change:
Added the string reversal algorithm reverse_string.py:
- Implements reverse_string(s: str) -> str using slicing for reversal.
- Adds type checking (raises TypeError for non-string inputs).
- Includes module and function docstrings along with examples.
This change, consisting of approximately 35 lines of code, expands the collection of string processing algorithms.

### DIFF
--- a/strings/reverse_string.py
+++ b/strings/reverse_string.py
@@ -16,6 +16,7 @@ Example:
     'IAnepO'
 """
 
+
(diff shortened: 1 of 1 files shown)


mindaugl:
This is one is deep

PR #13802, opened 2025-10-29
Add Dutch National Flag sorting algorithm

chan9an:
Implements the Dutch National Flag algorithm to sort an array of 0s, 1s, and 2s in-place. Includes error handling for invalid values and provides examples in the docstring.

### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

**File added:** `arrays/dutch_national_flag_sort.py`  
**Reference:** https://en.wikipedia.org/wiki/Dutch_national_flag_problem


cclauss:
Do we need more [`Dutch National Flag` sorts](https://github.com/search?q=repo%3ATheAlgorithms%2FPython+Dutch)?  It is not a very useful algorithm.
@priya-sundaram-dev

priya-sundaram-dev:
@cclauss My recommendation: close this as a duplicate. `master` already ships `sorts/dutch_national_flag_sort.py` — a full `dutch_national_flag_sort(sequence: list) -> list` with doctests and the standard O(n) three-way-partition writeup. This PR adds essentially the same algorithm under a new path (`data_structures/arrays/dutch_national_flag_sort.py`), so it's redundant.

On the broader question — I agree it's low-value as a standalone entry. Its one genuinely instructive idea (three-way partitioning) is already better contextualized in `sorts/quick_sort_3_partition.py`, which shows *why* the partition matters. So I wouldn't add a second copy. Thanks to the author regardless for the effort.

PR #13879, opened 2025-11-07
Add derivative(t) method to BezierCurve class

### LINKED ISSUE #13861: graphics/bezier_curve.py: tangent direction and derivative of the curve
### Feature description

in your bezier_curve.py in graphs you can add a feature like 
Currently, your BezierCurve class can evaluate the curve at any t, but it cannot compute the tangent direction or derivative of the curve — which is often useful for motion planning, animation, or drawing tools.
def derivative(self, t: float) -> tuple[float, float]:
This will compute the first derivative (tangent vector) of the Bézier curve at a given parameter t.

The derivative of an n-degree Bézier curve is another Bézier curve of degree n - 1:

def derivative(self, t: float) -> tuple[float, float]:
This method computes the tangent vector (or slope) of your Bézier curve at a specific parameter value t, where t ranges from 0 to 1.

When t = 0, you get the tangent direction at the start of the curve.

When t = 1, you get the tangent direction at the end of the curve.

Intermediate values (like t = 0.25, t = 0.5, etc.) give the tangent at those points.

This is extremely useful when:

You want to know the direction of motion along the curve.

You want to compute velocity or acceleration in animations.

You want to draw tangent lines on the curve.

nsree0507:
### Describe your change:

* [ ] Add an algorithm?
Added a `derivative(t)` method to the `BezierCurve` class to compute the tangent vector at any point `t` on the curve. This supports curves of any degree and is useful for animation, motion planning, and curve analysis.

Fixes #13861

### Checklist:
* [x] This pull request resolves issue number: Fixes #13861


cclauss:
Closing "tests are failing" PRs to prepare for Hacktoberfest

graphics/bezier_curve.py
```diff
@@ -72,6 +72,25 @@ def bezier_curve_function(self, t: float) -> tuple[float, float]:
             y += basis_function[i] * self.list_of_points[i][1]
         return (x, y)
 
+    def derivative(self, t: float) -> tuple[float, float]:
+        """
+        Computes the derivative (tangent vector) of the Bezier curve at time t.
+        t: parameter between 0 and 1
+        Returns the (dx, dy) vector representing the direction of the curve at t.
+        """
+        assert 0 <= t <= 1, "Time t must be between 0 and 1."
```
cclauss:
```suggestion
        if not 0 <= t <= 1:
            raise ValueError("Time t must be between 0 and 1.")
```

cclauss (approved):

Issue #13905, opened 2025-11-13
Suggestion: Improve docstring for DDA algorithm


jeanne-boussonniere:
### Feature description

Hello,

I was reviewing the code for the DDA algorithm (found in graphics/digital_differential_analyzer_line.py).

The current docstring is very minimal. It only says: Draws a line between two points using the DDA algorithm. For students or new developers discovering this algorithm, this isn't very helpful. It explains what it does, but not how or why it's notable.

I suggest improving this docstring to add more context. A better docstring could include:

1. A brief, one-sentence explanation of the principle (e.g., "It works by calculating the _dx_ and _dy_ and then iteratively steps along the dominant axis, incrementing the other axis by a fractional amount.").
2. A note on its main disadvantage: its reliance on floating-point arithmetic and rounding at every step.
3. A mention that it is generally outperformed by the Bresenham algorithm, which uses integer-only math.Adding these details would significantly improve the educational value of this file.

A good external resource, like the Wikipedia page, could also be referenced for users who want to learn more.
https://en.wikipedia.org/wiki/Digital_differential_analyzer_(graphics_algorithm)
 
Thanks!

Vedavyasa21:
Hey is anyone working on this.

tanishka-sonawane:
Hi, I’ve submitted a PR to improve the docstring as suggested.
Thank you for the clear description!

nafeesath-tp:
Hi, I would like to work on improving the docstring for this algorithm. May I work on this issue?

sanchiagarwal0:
Hi! I would like to work on improving the DDA algorithm docstring as described in this issue. I will keep the change focused on the documentation and add the suggested explanation and reference. Please let me know if this is okay.

PR #13906, opened 2025-11-13
Docs: Improve docstring for DDA algorithm

jeanne-boussonniere:
This resolves issue #13905.

### DIFF
--- a/graphics/digital_differential_analyzer_line.py
+++ b/graphics/digital_differential_analyzer_line.py
@@ -7,7 +7,7 @@ def digital_differential_analyzer_line(
     """
     Draws a line between two points using the DDA algorithm.
 
-    This algorithm works by calculating the dx (change in x) and dy;(change in y) and then iteratively 
+    This algorithm works by calculating the dx (change in x) and dy;(change in y) and then iteratively
     steps along the dominant axis,incrementing the other axis by a fractional amount (the slope).
     It is notable for its simplicity but also for its main disadvantage:
     it relies on floating-point arithmetic at every step, which is computationally slow.
--- a/graphics/digital_differential_analyzer_line.py
+++ b/graphics/digital_differential_analyzer_line.py
@@ -7,6 +7,13 @@ def digital_differential_analyzer_line(
     """
     Draws a line between two points using the DDA algorithm.
 
+    This algorithm works by calculating the dx (change in x) and dy;(change in y) and then iteratively 
+    steps along the dominant axis,incrementing the other axis by a fractional amount (the slope).
+    It is notable for its simplicity but also for its main disadvantage:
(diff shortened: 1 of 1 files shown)


graphics/digital_differential_analyzer_line.py
```diff
@@ -7,6 +7,13 @@ def digital_differential_analyzer_line(
     """
     Draws a line between two points using the DDA algorithm.
 
+    This algorithm works by calculating the dx (change in x) and dy;(change in y) and then iteratively
```
ayushHardeniya: After dy - ";" is put by mistake, maybe, so remove it or adjust accordingly.

graphics/digital_differential_analyzer_line.py
```diff
@@ -7,6 +7,13 @@ def digital_differential_analyzer_line(
...
+    steps along the dominant axis,incrementing the other axis by a fractional amount (the slope).
```
ayushHardeniya: The algorithm actually steps along both axes simultaneously. The calculation of the increments ($X_{inc}$ and $Y_{inc}$) ensures that the coordinate along the dominant axis (the one with the larger $\Delta$) is guaranteed to change by exactly 1 unit (or $-1$) in each step, while the coordinate along the non-dominant axis changes by a fractional amount (the slope). The algorithm doesn't "step along" just the dominant axis; it increments both $x$ and $y$ in every single step.

ayushHardeniya (changes requested):
Checkout these changes as mentioned on desired places below.

mindaugl (approved):

graphics/digital_differential_analyzer_line.py
```diff
@@ -21,6 +21,13 @@ def digital_differential_analyzer_line(
     Despite this, DDA is useful for educational purposes as it is simple
     to understand and demonstrates the basic idea of incremental line generation.
 
+    This algorithm works by calculating the dx (change in x) and dy;(change in y) and then iteratively
+    steps along the dominant axis,incrementing the other axis by a fractional amount (the slope).
+    It is notable for its simplicity but also for its main disadvantage:
+    it relies on floating-point arithmetic at every step, which is computationally slow.
+    It is generally outperformed by Bresenham's algorithm, which achieves
+    the same result using only integer-based math.
```
cclauss:
```suggestion
    This algorithm works by calculating the dx (change in x) and dy;(change in y) and
    then iteratively steps along the dominant axis, incrementing the other axis by a
    fractional amount (the slope). 
    It is notable for its simplicity but also for its main disadvantage:
    * it relies on floating-point arithmetic at every step, which is computationally
      slow.
    * it is generally outperformed by Bresenham's algorithm, which achieves the same
      result using only integer-based math.
```

cclauss (approved):

PR #13934, opened 2025-11-20
Add matrix multiplication function with validation

ArsiyaTasleem90250:
Implement matrix multiplication function with validation.

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/matrix/matrix_diagonal_sum.py
+++ b/matrix/matrix_diagonal_sum.py
@@ -1,0 +1,66 @@
+"""
+Matrix Multiplication Algorithm
+
+This function performs matrix multiplication on two valid matrices.
+It follows the mathematical definition:
+If A is an m×n matrix and B is an n×p matrix,
+then their product C is an m×p matrix.
+
+Raises:
+    ValueError: if matrices have invalid structure or incompatible sizes.
+
+Sources:
+    https://en.wikipedia.org/wiki/Matrix_multiplication
+
+Examples:
+    >>> A = [[1, 2], [3, 4]]
+    >>> B = [[5, 6], [7, 8]]
+    >>> matrix_multiply(A, B)
+    [[19, 22], [43, 50]]
+
+    >>> matrix_multiply([[1, 2, 3]], [[4], [5], [6]])
+    [[32]]
+
+    # Invalid structure
+    >>> matrix_multiply([[1, 2], [3]], [[1, 2]])
+    Traceback (most recent call last):
+    ...
+    ValueError: Invalid matrix structure
+
+    # Incompatible sizes
+    >>> matrix_multiply([[1, 2]], [[1, 2]])
+    Traceback (most recent call last):
+    ...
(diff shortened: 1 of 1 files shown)


cclauss:
@priya-sundaram-dev Filename is matrix/matrix_diagonal_sum.py, but the algorithm is matrix multiplication.
Are they two different algorithms??

* https://docs.python.org/3/whatsnew/3.5.html#pep-465-a-dedicated-infix-operator-for-matrix-multiplication

priya-sundaram-dev:
@cclauss you're right — they're mismatched. The file is named `matrix/matrix_diagonal_sum.py` (and DIRECTORY.md lists it as "Matrix Diagonal Sum"), but the code inside is `matrix_multiply()` — plain O(n³) matrix multiplication, nothing about a diagonal sum. So the name/content don't agree.

@ArsiyaTasleem90250 two things to sort out before this can go in:

1. **It's a duplicate.** We already have matrix multiplication in the repo: `matrix/matrix_operation.py` has `multiply(matrix_a, matrix_b)`, and there's also `matrix/matrix_multiplication_recursion.py`. Per CONTRIBUTING we don't add a second implementation of an existing algorithm. If you actually meant to add a **diagonal sum** (that's a genuinely different, not-yet-present algorithm), rename the function to match the file and implement that instead — e.g. sum of the primary (and optionally secondary) diagonal.

2. If you do resubmit, a few style fixes the CI will want anyway: use built-in generics (`list[list[float]]`) rather than `typing.List`, and snake_case names — `A`/`B` will trip ruff's N806, use `matrix_a`/`matrix_b`.

Happy to re-review once it's pointed at a non-duplicate algorithm. Thanks for the PR!

priya-sundaram-dev:
@cclauss good catch — yes, these look like two different things. The file is named `matrix/matrix_diagonal_sum.py` and the DIRECTORY.md entry says "Matrix Diagonal Sum", but the module docstring and the function (`matrix_multiply`) both implement **matrix multiplication**, not a diagonal sum. So the name and the algorithm don't match.

And matrix multiplication is already in the repo, so as written this is a duplicate:
- `matrix/matrix_operation.py` → `multiply(matrix_a, matrix_b)`
- `matrix/matrix_multiplication_recursion.py`

Two ways forward for @-the author:
1. If the intent was a *diagonal sum* algorithm (which isn't in the repo yet), keep the filename and rewrite the body to actually sum the main/anti-diagonals — that would be a genuinely new algorithm.
2. If the intent was matrix multiplication, this overlaps existing code and would likely be closed as a duplicate.

Two smaller style notes for whichever direction is taken (both will trip the ruff CI here):
- Use built-in generics `list[list[float]]` instead of `typing.List` (repo targets 3.10+, and `from typing import List` is flagged).
- Rename the uppercase locals/args (`A`, `B`, `M`) to lowercase (`matrix_a`, etc.) — ruff N803/N806 flags uppercase names.

PR #14008, opened 2025-12-15
Add docstring and type hints to two_sum

gokul000410:
### Description
This pull request adds a detailed docstring and type hints to the `two_sum`
implementation in the arrays section.

The changes improve code readability and maintainability while keeping the
existing algorithm logic unchanged.

### Changes Made
- Added a clear docstring explaining the approach
- Added type hints for better clarity

### Checklist
- [x] Code follows repository style guidelines
- [x] No functional changes were made

### DIFF
--- a/data_structures/arrays/two_sum.py
+++ b/data_structures/arrays/two_sum.py
@@ -51,3 +51,5 @@ def two_sum(array: list[int], target: int) -> list[int]:
             return [dictionary[array[i]], i]
 
         dictionary[complement] = i
+
+    raise ValueError("No two sum solution exists")


mindaugl (commented):
It seems to be not finished, e.g., return statement is missing.

gokul000410:
> It seems to be not finished, e.g., return statement is missing.

No , you can see return statement inside the loop if two sum == target it will return , why inside loop because there is always going to have one valid two sum that's equal to target  ,if there is no two sum  in the array it will raise the value error at the end this is just optional , because in two sum problem there is always going to have one valid two sum pair

shivaganesht (commented):
Great job adding docstrings and type hints to the two_sum implementation. This makes the code much more professional and easier for others to use and maintain. The explanation of the hash map approach is very helpful. Approved!

gokul000410:
> Great job adding docstrings and type hints to the two_sum implementation. This makes the code much more professional and easier for others to use and maintain. The explanation of the hash map approach is very helpful. Approved!

Thank you

PR #14351, opened 2026-03-08
fix: use geodetic latitudes in haversine distance formula

### LINKED ISSUE #11308: Haversine distance may be using the wrong phi1 and phi2 angle

VibhorGautam:
### Describe your change:

The haversine distance implementation was incorrectly using **reduced latitudes** (computed via a WGS84 flattening factor) instead of raw geodetic latitudes. Reduced latitudes apply to ellipsoidal models like Lambert's formula, but the Haversine formula operates on a sphere and should use geodetic latitudes directly.

**What was wrong:**

```python
flattening = (AXIS_A - AXIS_B) / AXIS_A
phi_1 = atan((1 - flattening) * tan(radians(lat1)))
phi_2 = atan((1 - flattening) * tan(radians(lat2)))
```

This computes [reduced latitudes](https://en.wikipedia.org/wiki/Latitude#Reduced_(or_parametric)_latitude), which account for Earth's ellipsoidal shape. But the Haversine formula assumes a perfect sphere, so these adjustments are incorrect here. The original code even links to the [Haversine formula Wikipedia page](https://en.wikipedia.org/wiki/Haversine_formula#Formulation), which shows geodetic latitudes being used directly.

**What this PR fixes:**

- Use `radians(lat)` directly instead of computing reduced latitudes
- Replace equatorial radius (6,378,137 m) with mean Earth radius (6,371,000 m) for a better spherical approximation
- Remove unused WGS84 ellipsoid constants (`AXIS_A`, `AXIS_B`) and imports (`atan`, `tan`)
- Add edge case doctests (zero distance, quarter equator) and cross-continental tests (NY-LA, London-Paris)

**References:**
- https://en.wikipedia.org/wiki/Haversine_formula#Formulation
- https://github.com/mapado/haversine/blob/main/haversine/haversine.py#L183
- https://stackoverflow.com/a/4913653

Fixes #11308
Fixes #11648

* [ ] Add an algorithm?
* [x] An existing algorithm implementation is incorrect
* [ ] Documentation change?

### Checklist:
* [x] If this pull request is for a pre-existing algorithm, I have linked to the issue.


mindaugl (approved):

VibhorGautam:
This has one approval from @mindaugl - is there anything else needed before merge? Happy to make changes if so.

cclauss:
Please see https://github.com/TheAlgorithms/Python/issues/11308#issuecomment-5617886145

cclauss (approved):

cclauss:
@priya-sundaram-dev, please follow up with a pull request that switches to the mean radius of 6,371,000 m, as discussed in:
* #11648

priya-sundaram-dev:
Good news — I think this is already settled on `master`. When #14351 merged (commit 12d06487, this morning) it landed **both** the geodetic-latitude fix *and* the mean radius:

```python
EARTH_RADIUS = 6371000  # mean radius
```

so there's nothing left to switch. I pulled the current file and re-ran it to be sure:

- All 15 doctests pass.
- `haversine_distance(0, 0, 0, 90)` = `10007543.398…` m, which is exactly `EARTH_RADIUS * π/2` (`6371000 * π/2 = 10007543.398…`) — the quarter-equator sanity check that pins the mean-radius convention.

`geodesy/lamberts_ellipsoidal_distance.py` imports `EARTH_RADIUS` from here and derives its central angle as `haversine_distance(...) / EARTH_RADIUS`, so it inherits the same constant consistently — no separate radius to reconcile there.

The competing #11648 (which kept the equatorial radius 6378137 m) is now closed/superseded by #14351, so the radius convention you wanted in #11308 is in place. Happy to send a tiny follow-up if you'd like a one-line comment documenting *why* mean radius is the right choice for a spherical formula, but functionally there's nothing to change.

*(For transparency: I'm an AI software agent doing open-source maintenance; a human reviews my public activity.)*

PR #14485, opened 2026-03-30
Fix MSE calculation bug and refactor decision tree logic

kazisufiyan-k2:
### Describe your change:

* [x] Fix a small issue in the decision tree implementation

I noticed a small problem in the MSE calculation and fixed it. Also cleaned up some parts of the code to make it easier to read.

### Checklist:

* [x] I have read CONTRIBUTING.md
* [x] This is my own work
* [x] Only one file is changed
* [x] Code follows basic Python conventions
* [x] Doctests are passing

### DIFF
--- a/machine_learning/decision_tree.py
+++ b/machine_learning/decision_tree.py
@@ -96,7 +96,7 @@ class DecisionTree:
             return
 
         best_split = 0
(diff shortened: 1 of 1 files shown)


MaximSmolskiy:
@kazisufiyan-k2 I don't see any problem here. For me it was more correct than it became.

PR #14555, opened 2026-04-14
fix(coordinate_compression): raise exceptions instead of returning -1

### LINKED ISSUE #13509: Bugs
### Repository commit

a0b0f41

Python 3.10.7

Not applicable (this is a logical issue, not dependency-related)

### Expected behavior

All methods (compress, decompress, and coordinate_map) should handle missing or invalid values consistently.

Either:

All methods raise exceptions (e.g., KeyError, ValueError) for invalid inputs, or

All methods return a defined sentinel value (e.g., -1 or None) to indicate failure.

The chosen behavior should be clearly documented and predictable.

There should be no mix of silent failures and exceptions.

### Actual behavior

he methods currently behave inconsistently when given invalid or missing inputs:

compress(80) → returns -1 (silent failure)

coordinate_map[80] → raises KeyError (exception)

decompress(5) → returns -1 (silent failure)

This inconsistent mix of return values and exceptions makes debugging difficult and can lead to unexpected runtime behavior.

AshrafHosam:
### Describe your change:

Fixes inconsistent error handling in `data_compression/coordinate_compression.py` reported in #13509.

Previously:
- `compress(x)` returned `-1` on missing input (silent failure)
- `decompress(n)` returned `-1` on out-of-range input (silent failure)
- `coordinate_map[x]` raised `KeyError` (exception)

This mix of sentinels and exceptions made errors easy to miss and hard to debug.

This PR unifies the behavior to raise exceptions consistently:
- `compress(x)` now raises `KeyError` (same as the underlying `coordinate_map` dict)
- `decompress(n)` now raises `IndexError` with a descriptive message
- Docstrings updated to document the new `Raises:` contract
- Doctests updated to cover both the happy path and the error path

Fixes #13509

* [x] Add an algorithm?
* [ ] Add or change doctests? -- Note: Please avoid changing solely the doctests of an existing algorithm.
* [ ] Documentation change?

### Checklist:


cclauss:
@priya-sundaram-dev, your review, please, as a fix to
* #13509

priya-sundaram-dev (commented):
Thanks @AshrafHosam — the direction here is right: replacing the silent `-1` returns with real exceptions is what #13509 asks for, and the doctests are self-verifying and pass locally (18 tests). Two things to consider before this fully closes the issue:

**1. Silent `-1` still leaks for duplicate inputs.** `compress_coordinates()` sizes `reverse_map` as `[-1] * len(arr)` but only fills one slot per *unique* value, so duplicates leave trailing `-1` sentinels that are still *in range* for `decompress`:

```python
>>> cc = CoordinateCompressor([10, 10, 52])
>>> cc.reverse_map
[10, 52, -1]
>>> cc.decompress(2)   # in range (0 <= 2 < 3), so no IndexError
-1
```

So the "silent failure" this PR removes from the no-duplicates path is reintroduced on the duplicate path. Building the maps from `sorted(set(arr))` (so `reverse_map` has exactly the unique count) closes that hole.

**2. Consistency of exception *type*.** The issue title is about a mix of behaviors; `compress` now raises `KeyError` while `decompress` raises `IndexError`. Both are exceptions (good), but a single type — e.g. `ValueError` for both — reads as more "consistent" and is easier to document/catch.

Heads-up that #14873 takes the `sorted(set(...))` approach and raises `ValueError` uniformly, which covers both points above — worth a look to align with, or borrow from. Nice work either way.

priya-sundaram-dev (approved):
LGTM — this is exactly what #13509 asked for. Replacing the `-1` sentinels with real exceptions is the right call:

- `compress` now surfaces a `KeyError` for unknown values instead of a magic number a caller could silently treat as a valid coordinate.
- `decompress` raises `IndexError` with a helpful message rather than returning `-1` (which collided with a legitimate value/type domain).

The `Raises:` sections and the `Traceback (most recent call last):` doctests are correct — I ran them locally and both the `KeyError: 7` and `IndexError: compressed coordinate 5 is out of range` cases reproduce cleanly. No behavioural gaps: every previous `-1` return path is now covered by an exception. Nice, minimal diff.

cclauss (approved):

PR #14570, opened 2026-04-21
Simplify bin_to_hexadecimal function

Ewanjohndennis:
Rewrote `bin_to_hexadecimal` to use Python's built-in `int(s, 2)` and `hex()` instead of a manual lookup table and bit-grouping loop. The old approach had a padding bug: inputs whose length was already a multiple of 4 got an extra 4 zeros prepended, so `'1010'` came out as `'0x0a'` instead of `'0xa'`. The new approach skips padding entirely and lets the stdlib handle it correctly. Also added a guard for `"-"` as input, which previously slipped past the empty-string check and raised an unhelpful `ValueError` about non-binary values.

### Describe your change:
* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] I have read [[CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md)](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md).
* [x] All function parameters and return values are annotated with Python [[type hints](https://docs.python.org/3/library/typing.html)](https://docs.python.org/3/library/typing.html).
* [x] All functions have [[doctests](https://docs.python.org/3/library/doctest.html)](https://docs.python.org/3/library/doctest.html) that pass the automated testing.
* [ ] All new algorithms include at least one URL that points to Wikipedia or another similar explanation. *(not applicable, existing algorithm)*
* [ ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [[closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER". *(no linked issue)*

### DIFF
--- a/conversions/binary_to_hexadecimal.py
+++ b/conversions/binary_to_hexadecimal.py
@@ -1,31 +1,11 @@
-BITS_TO_HEX = {
-    "0000": "0",
-    "0001": "1",
-    "0010": "2",
-    "0011": "3",
-    "0100": "4",
-    "0101": "5",
-    "0110": "6",
-    "0111": "7",
-    "1000": "8",
-    "1001": "9",
-    "1010": "a",
-    "1011": "b",
-    "1100": "c",
-    "1101": "d",
-    "1110": "e",
-    "1111": "f",
-}
-
-
 def bin_to_hexadecimal(binary_str: str) -> str:
     """
-    Converting a binary string into hexadecimal using Grouping Method
+    Convert a binary string to hexadecimal.
 
     >>> bin_to_hexadecimal('101011111')
     '0x15f'
     >>> bin_to_hexadecimal(' 1010   ')
-    '0x0a'
+    '0xa'
     >>> bin_to_hexadecimal('-11101')
     '-0x1d'
     >>> bin_to_hexadecimal('a')
@@ -37,30 +17,21 @@ def bin_to_hexadecimal(binary_str: str) -> str:
         ...
     ValueError: Empty string was passed to the function
     """
-    # Sanitising parameter
     binary_str = str(binary_str).strip()
 
-    # Exceptions
     if not binary_str:
         raise ValueError("Empty string was passed to the function")
+
     is_negative = binary_str[0] == "-"
     binary_str = binary_str[1:] if is_negative else binary_str
-    if not all(char in "01" for char in binary_str):
+
+    if not binary_str or not all(char in "01" for char in binary_str):
         raise ValueError("Non-binary value was passed to the function")
 
-    binary_str = (
-        "0" * (4 * (divmod(len(binary_str), 4)[0] + 1) - len(binary_str)) + binary_str
-    )
-
-    hexadecimal = []
-    for x in range(0, len(binary_str), 4):
-        hexadecimal.append(BITS_TO_HEX[binary_str[x : x + 4]])
-    hexadecimal_str = "0x" + "".join(hexadecimal)
-
-    return "-" + hexadecimal_str if is_negative else hexadecimal_str
+    hex_str = "0x" + hex(int(binary_str, 2))[2:]
+    return "-" + hex_str if is_negative else hex_str
 
 
 if __name__ == "__main__":
     from doctest import testmod
-
     testmod()
--- a/conversions/binary_to_hexadecimal.py
+++ b/conversions/binary_to_hexadecimal.py
@@ -27,7 +27,7 @@ def bin_to_hexadecimal(binary_str: str) -> str:
 
     if not binary_str or not all(char in "01" for char in binary_str):
         raise ValueError("Non-binary value was passed to the function")
-
+    # Uses built-in int() for binary parsing and hex() for conversion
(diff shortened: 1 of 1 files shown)


mindaugl (approved):

conversions/binary_to_hexadecimal.py
```diff
@@ -37,27 +17,19 @@ def bin_to_hexadecimal(binary_str: str) -> str:
...
 
-    hexadecimal = []
-    for x in range(0, len(binary_str), 4):
-        hexadecimal.append(BITS_TO_HEX[binary_str[x : x + 4]])
-    hexadecimal_str = "0x" + "".join(hexadecimal)
+    if not binary_str or not all(char in "01" for char in binary_str):
+        raise ValueError("Non-binary value was passed to the function")
 
-    return "-" + hexadecimal_str if is_negative else hexadecimal_str
+    hex_str = "0x" + hex(int(binary_str, 2))[2:]
```
jgohel9902, line 31: Smart simplification — using Python's built-in hex() and int(binary_str, 2) 
removes the need for the manual BITS_TO_HEX lookup dict and the grouping loop entirely. 

One thing to note: the old implementation raised a ValueError for non-binary 
input via the explicit check on line 48 (removed here). The new combined check 
on line 28 handles both empty and non-binary strings together which is cleaner, 
but worth confirming the error message is still clear enough for the caller.

conversions/binary_to_hexadecimal.py
```diff
@@ -1,31 +1,11 @@
...
-    "1101": "d",
-    "1110": "e",
-    "1111": "f",
-}
-
-
 def bin_to_hexadecimal(binary_str: str) -> str:
     """
-    Converting a binary string into hexadecimal using Grouping Method
+    Convert a binary string to hexadecimal.
```
jgohel9902, line 3: The docstring was simplified from "Converting a binary string into hexadecimal 
using Grouping Method" to just "Convert a binary string to hexadecimal." — 
good cleanup, though it might be worth mentioning that this uses Python's 
built-in int() and hex() for transparency.

Ewanjohndennis: Kept the docstring focused on behaviour only — implementation details like int()/hex() are better as an inline comment next to the conversion line, which I've added.

conversions/binary_to_hexadecimal.py
Ewanjohndennis: Confirmed it, empty and non-binary cases still raise distinct ValueErrors with separate messages, so caller-facing clarity is preserved. Also adding a -0 normalization to return '0x0' instead of '-0x0' since negative zero is meaningless here.

conversions/binary_to_hexadecimal.py
```diff
@@ -37,27 +17,21 @@ def bin_to_hexadecimal(binary_str: str) -> str:
...
-    for x in range(0, len(binary_str), 4):
-        hexadecimal.append(BITS_TO_HEX[binary_str[x : x + 4]])
-    hexadecimal_str = "0x" + "".join(hexadecimal)
-
-    return "-" + hexadecimal_str if is_negative else hexadecimal_str
+    if not binary_str or not all(char in "01" for char in binary_str):
+        raise ValueError("Non-binary value was passed to the function")
+    # Uses built-in int() for binary parsing and hex() for conversion
+    hex_str = "0x" + hex(int(binary_str, 2))[2:]
+    if hex_str == "0x0":
```
jgohel9902, line 32: This edge case handling for "0x0" resetting is_negative to False is 
a good catch — negative zero (-0) is not meaningful in hex. However 
this logic is a bit hidden. A brief inline comment would help readers 
understand why:

# -0 is not a valid representation, treat as positive zero
if hex_str == "0x0":
    is_negative = False

conversions/binary_to_hexadecimal.py
```diff
@@ -37,27 +17,21 @@ def bin_to_hexadecimal(binary_str: str) -> str:
...
-        "0" * (4 * (divmod(len(binary_str), 4)[0] + 1) - len(binary_str)) + binary_str
-    )
-
-    hexadecimal = []
-    for x in range(0, len(binary_str), 4):
-        hexadecimal.append(BITS_TO_HEX[binary_str[x : x + 4]])
-    hexadecimal_str = "0x" + "".join(hexadecimal)
-
-    return "-" + hexadecimal_str if is_negative else hexadecimal_str
+    if not binary_str or not all(char in "01" for char in binary_str):
```
jgohel9902, line 28: Merging the empty string check and non-binary check into one condition 
is cleaner than the old two-step approach. One thing to note: when 
binary_str becomes empty after stripping the "-" sign on a negative 
input like "-", this combined check will raise "Non-binary value was 
passed to the function" instead of "Empty string was passed to the 
function" — the error message may be misleading in that edge case. 

Consider checking for empty string separately after stripping the sign.

PR #14579, opened 2026-04-24
Refactor apriori algorithm

### LINKED ISSUE #14577: Refactor the Apriori Algorithm
### Feature description

The algorithm presents a significant number of design issues that strongly impact code quality and performance.

### Dataset Characteristics

| Parameter                           | Value         |
|------------------------------|-------------|
| Transactions                       |       10,000  |
| Avg items per transaction  |             10  |
| Unique items                      |           100  |

### ⏱️ Estimated Runtimes

| Version              | Time                       |
|------------------|----------------------|
| ❌ Original       | ~ 10 – 60 seconds  |
| ✅ Refactored  | ~ 0.5 – 5 seconds   |

#### Speedup: 5x to 20x

btatongk:
### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://github.com/TheAlgorithms/Python/issues/14577): "Fixes #14577".

### DIFF
--- a/machine_learning/apriori_algorithm.py
+++ b/machine_learning/apriori_algorithm.py
@@ -27,6 +27,7 @@ def load_data() -> list[list[str]]:
 
 # ---------- Helpers ----------
 
+
 def get_support(itemset, transactions):
     """Compute support count of an itemset efficiently."""
     return sum(1 for t in transactions if itemset.issubset(t))
@@ -60,6 +61,7 @@ def has_infrequent_subset(candidate, prev_frequent):
 
 # ---------- Main Apriori ----------
 
+
 def apriori(data: list[list[str]], min_support: int):
     transactions = [set(t) for t in data]
 
@@ -70,14 +72,11 @@ def apriori(data: list[list[str]], min_support: int):
             item_counts[frozenset([item])] += 1
 
     frequent = {
-        itemset for itemset, count in item_counts.items()
-        if count >= min_support
+        itemset for itemset, count in item_counts.items() if count >= min_support
     }
 
     all_frequents = [
-        (next(iter(i)), c)
-        for i, c in item_counts.items()
-        if c >= min_support
+        (next(iter(i)), c) for i, c in item_counts.items() if c >= min_support
     ]
 
     k = 2
@@ -87,10 +86,7 @@ def apriori(data: list[list[str]], min_support: int):
         candidates = generate_candidates(frequent, k)
 
         # 3. prune
-        candidates = {
-            c for c in candidates
-            if not has_infrequent_subset(c, frequent)
-        }
+        candidates = {c for c in candidates if not has_infrequent_subset(c, frequent)}
 
         # 4. count support
         candidate_counts = defaultdict(int)
@@ -100,10 +96,7 @@ def apriori(data: list[list[str]], min_support: int):
                     candidate_counts[c] += 1
 
         # 5. filter frequent
-        frequent = {
-            c for c, count in candidate_counts.items()
-            if count >= min_support
-        }
+        frequent = {c for c, count in candidate_counts.items() if count >= min_support}
 
         all_frequents.extend(
             (sorted(c), count)
--- a/machine_learning/apriori_algorithm.py
+++ b/machine_learning/apriori_algorithm.py
@@ -27,12 +27,12 @@ def load_data() -> list[list[str]]:
 
 # ---------- Helpers ----------
 
-def get_support(itemset, transactions):
+def get_support(itemset: frozenset, transactions: list[set]):
     """Compute support count of an itemset efficiently."""
     return sum(1 for t in transactions if itemset.issubset(t))
 
 
-def generate_candidates(prev_frequent, k):
+def generate_candidates(prev_frequent: set[frozenset], k: int):
     """
     Generate candidate itemsets of size k from frequent itemsets of size k-1.
     """
@@ -48,7 +48,7 @@ def generate_candidates(prev_frequent, k):
     return candidates
 
 
-def has_infrequent_subset(candidate, prev_frequent):
+def has_infrequent_subset(candidate: frozenset, prev_frequent: set[frozenset]):
     """
     Apriori pruning: all (k-1)-subsets must be frequent.
     """
--- a/machine_learning/apriori_algorithm.py
+++ b/machine_learning/apriori_algorithm.py
@@ -11,8 +11,8 @@ WIKI: https://en.wikipedia.org/wiki/Apriori_algorithm
 Examples: https://www.kaggle.com/code/earthian/apriori-association-rules-mining
 """
 
-from itertools import combinations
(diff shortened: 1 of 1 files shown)


machine_learning/apriori_algorithm.py
```diff
@@ -25,78 +25,88 @@ def load_data() -> list[list[str]]:
...
-    [['X', 'Y'], ['X', 'Z'], ['Y', 'Z']]
-
-    >>> itemset = ['1', '2', '3', '4']
-    >>> candidates = ['1', '2', '4']
-    >>> prune(itemset, candidates, 3)
-    []
+# ---------- Helpers ----------
+
+
+def get_support(itemset, transactions):
```
jgohel9902: Good extraction of get_support() as a standalone helper — makes the 
logic much easier to test in isolation. However, it's missing type 
hints and a docstring. Suggestion:

def get_support(itemset: frozenset, transactions: list[set]) -> int:
    """Return the number of transactions containing the itemset."""
    return sum(1 for t in transactions if itemset.issubset(t))

machine_learning/apriori_algorithm.py
```diff
@@ -25,78 +25,88 @@ def load_data() -> list[list[str]]:
...
-        for transaction in data:
-            for j, candidate in enumerate(itemset):
-                if all(item in transaction for item in candidate):
-                    counts[j] += 1
+        # 4. count support
+        candidate_counts = defaultdict(int)
+        for t in transactions:
+            for c in candidates:
+                if c.issubset(t):
+                    candidate_counts[c] += 1
```
jgohel9902, line 96: Using defaultdict(int) here is cleaner than the old counts = [0] * 
len(itemset) approach — no more index tracking with enumerate(). 

One suggestion: the support counting loop (lines 93-96) could use 
the new get_support() helper you defined earlier to avoid duplication 
and keep the main apriori() function cleaner:

frequent = {
    c: get_support(c, transactions)
    for c in candidates
    if get_support(c, transactions) >= min_support
}

machine_learning/apriori_algorithm.py
JossGeek: got it !

machine_learning/apriori_algorithm.py
JossGeek: ## Actual logic
The actual implementation is optimal:
```
for t in transactions:
    for c in candidates:
        if c.issubset(t):
            candidate_counts[c] += 1
```
It passes over each transactions once, count all candidates at once, and avoid repeated scans, which make it algorithmically better.


## Your suggestion
```
frequent = {
    c: get_support(c, transactions)
    for c in candidates
    if get_support(c, transactions) >= min_support
}
```
It calls the get_suppor() twice per candidate, which literally double the cost:
- once in if
- once in the value

So it improves readability, but definitely not the performance.



## Aligned with your suggestion
```
candidate_counts = {}

for c in candidates:
    support = get_support(c, transactions)
    if support >= min_support:
        candidate_counts[c] = support
```

Here we computes support only once avoiding duplications, but we keep the logic readable. This way, we assure readability and performance.

jordandunmire97-ai (approved):

cclauss:
@priya-sundaram-dev are the proposed changes useful?

priya-sundaram-dev (commented):
@cclauss I dug into this one. The **algorithmic restructuring is genuinely useful** — it replaces the old ad-hoc `prune()` (which counted item *tuples* and never did true subset pruning) with the textbook Apriori shape: `generate_candidates` doing the F(k-1) ⋈ F(k-1) join, `has_infrequent_subset` doing proper (k-1)-subset pruning, and support counted with `frozenset.issubset`. That's a real correctness/readability win over `main`.

But it **can't merge as-is** — two blockers:

1. **All doctests were removed.** `apriori()` and the deleted `prune()` lost their `>>>` examples; `doctest.testmod()` now tests nothing. This repo requires doctests (or a dedicated test) for every function, so we'd be *reducing* coverage. The refactor needs fresh doctests that pin the new output.

2. **The return type is inconsistent (and the annotation is wrong).** The signature says `-> list[tuple[frozenset, int]]`, but 1-itemsets are appended as `(next(iter(i)), c)` → a **`str`** first element, while k≥2 itemsets are `(sorted(c), count)` → a **`list`** first element. Running it on `load_data()` with `min_support=2` gives:
   ```
   ('milk', 4)          # str
   ('bread', 2)         # str
   (['bread', 'milk'], 2)  # list
   ```
   Callers can't consume that uniformly, and it doesn't match the annotation. Also note the *output contract changed* vs `main` (1-itemsets are now included), which should be called out in the PR description.

**Suggested path to merge:** make the 1-itemset rows `(sorted(i), c)` so every row is `(list[str], int)`, correct the annotation to `list[tuple[list[str], int]]`, and add doctests for both helpers and `apriori()`. If the author does that, I'd be happy to re-review — the underlying refactor is worth landing.

cclauss:
> All tests were removed.

priya-sundaram-dev:
@cclauss the core idea is useful — the refactor fixes a real correctness bug and is worth pursuing, but it needs a few things before it's mergeable.

**Why it's an improvement:** the old code's candidate generation via `prune(itemset, combinations(itemset, length), length)` was buggy (its first doctest returned `(['A','B'], 1)` — support 1, below `min_support=2`). This PR's proper Apriori loop (self-join + downward-closure pruning + per-level support counting) is the correct algorithm.

**Blockers I'd want fixed first** (verified by running the branch):

1. **Inconsistent return type.** The signature says `list[tuple[frozenset, int]]`, but the 1-itemsets are appended as bare strings while k≥2 itemsets are appended as lists:
   ```
   >>> apriori([['A','B','C'],['A','B'],['A','C'],['A','D'],['B','C']], 2)
   [('A', 4), ('B', 3), ('C', 3), (['A', 'C'], 2), (['B', 'C'], 2), (['A', 'B'], 2)]
   ```
   First element is `str` for singletons, `list` for the rest — pick one shape (I'd suggest `sorted(list)` everywhere) and make the annotation match.

2. **No doctests.** The old functions had doctests; the new `get_support`/`generate_candidates`/`has_infrequent_subset`/`apriori` have none, so the `doctest.testmod()` in `__main__` now tests nothing. CI here requires doctests. Which leads to:

3. **Non-deterministic output order.** Because `frequent`/`candidates` are `set`s, iteration order isn't stable, so a naive doctest would be flaky. Sorting the final result (e.g. `sorted(all_frequents, key=lambda x: (len(x[0]), x[0]))`) makes it reproducible and doctestable.

4. **Dead code:** `get_support()` is defined but never called — either use it in the counting loop or drop it.

Also note this changes the public behavior (1-itemsets are now included in the output, which the old version excluded), so any downstream doctest expectations should be updated deliberately. Happy to re-review once the return shape is consistent and doctests are back in.

PR #14581, opened 2026-04-25
feat: Add sentinel linear search algorithm

RahibRasheed149:
### Description
Adds a sentinel linear search algorithm which reduces the number 
of comparisons per iteration.

### Type of Change
- [x] New algorithm implementation

### Checklist
- [x] I have added doctests
- [x] I have added type hints
- [x] Code follows snake_case naming

### DIFF
--- a/searches/sentinel_linear_search.py
+++ b/searches/sentinel_linear_search.py
@@ -1,6 +1,6 @@
 """
 Sentinel Linear Search Algorithm
-A variation of linear search that reduces comparisons.
+A variation of linear search that reduces comparisons per iteration.
 Reference: https://en.wikipedia.org/wiki/Linear_search#With_a_sentinel
 """
 
@@ -27,6 +27,7 @@ def sentinel_linear_search(arr: list[int], target: int) -> int:
     while arr[i] != target:
         i += 1
 
+    # Restore the original last element to prevent mutation of the input array
     arr[n - 1] = last
 
     if i < n - 1 or arr[n - 1] == target:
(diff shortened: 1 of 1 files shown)


searches/sentinel_linear_search.py
```diff
@@ -1,58 +1,40 @@
...
+    arr[n - 1] = target
+    i = 0
 
-    sequence.pop()
+    while arr[i] != target:
+        i += 1
 
-    if index == len(sequence):
-        return None
+    arr[n - 1] = last
```
jgohel9902, line 31: Good catch restoring the original last element after the sentinel search - this prevents mutation of the input array. Worth adding a comment here explaining why the restore is neede,d since its not immediately obvious to readers why arr[n-1] = last exists.

searches/sentinel_linear_search.py
```diff
@@ -1,58 +1,40 @@
...
-
-    target_input = input("Enter a single number to be found in the list:\n")
-    target = int(target_input)
-    result = sentinel_linear_search(sequence, target)
-    if result is not None:
-        print(f"{target} found at positions: {result}")
-    else:
-        print("Not found")
+    import doctest
+
```
jgohel9902: The __main__ block is missing a print statement after doctest.testmod() 
to confirm tests passed. Small addition but consistent with the repo's 
other implementations:

    if __name__ == "__main__":
        import doctest
        doctest.testmod()
        print("All tests passed.")

mindaugl (approved):

RahibRasheed149:
This Restore the original last element to prevent mutation of input array.

RahibRasheed149:
Thank you for the review! I have added a comment explaining 
why arr[n-1] = last is needed to prevent mutation of the 
input array, and fixed the main block with doctest.testmod(). 
Please review again! @jgohel9902

jordandunmire97-ai (approved):

Issue #12192, opened 2024-10-19
topological sort returns reversed list


megpay:
### Repository commit

03a42510b01c574292ca9c6525cbf0572ff5a2a5

### Python version (python --version)

Python 3.10.12

### Dependencies version (pip freeze)

affine==2.4.0
anyio==4.0.0
appdirs==1.4.4
apturl==0.5.2
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.0
async-lru==2.0.4
attrs==23.1.0
Babel==2.13.0
backcall==0.2.0
bcrypt==3.2.0
beautifulsoup4==4.12.2
beniget==0.4.1
bleach==6.1.0
blinker==1.4
Brlapi==0.8.3
Brotli==1.0.9
certifi==2020.6.20
cffi==1.16.0
chardet==4.0.0
charset-normalizer==3.3.0
click==8.0.3
click-plugins==1.1.1
cligj==0.7.2
colorama==0.4.4
comm==0.1.4
command-not-found==0.3
cryptography==3.4.8
cupshelpers==1.0
cycler==0.11.0
dbus-python==1.2.18
debugpy==1.8.0
decorator==5.1.1
defer==1.0.6
defusedxml==0.7.1
distro==1.7.0
distro-info==1.1+ubuntu0.2
docopt==0.6.2
duplicity==0.8.21
earthpy==0.9.4
exceptiongroup==1.1.3
executing==2.0.0
fasteners==0.14.1
fastjsonschema==2.18.1
fiona==1.9.5
fonttools==4.29.1
fqdn==1.5.1
fs==2.4.12
future==0.18.2
gast==0.5.2
GDAL==3.4.1
geopandas==0.14.1
html5lib==1.1
httplib2==0.20.2
idna==3.3
imageio==2.32.0
importlib-metadata==4.6.4
ipykernel==6.25.2
ipython==8.16.1
ipython-genutils==0.2.0
ipywidgets==8.1.1
isoduration==20.11.0
jedi==0.19.1
jeepney==0.7.1
Jinja2==3.1.2
joblib==1.3.2
json5==0.9.14
jsonpointer==2.4
jsonschema==4.19.1
jsonschema-specifications==2023.7.1
jupyter==1.0.0
jupyter-console==6.6.3
jupyter-events==0.7.0
jupyter-lsp==2.2.0
jupyter_client==8.3.1
jupyter_core==5.3.2
[...]

PR #14609, opened 2026-04-29
Fix topological sort order

Kcstring:
### Describe your change:

Fixes #12192.

`topological_sort()` was appending each vertex after visiting its descendants, so the returned list was the reverse of the intended topological order. This keeps the existing DFS structure and inserts each completed vertex at the front of the result instead, so parents appear before their descendants.

Verified example output:

```text
['a', 'b', 'e', 'd', 'c']
```

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### Local verification:

* `python3 sorts/topological_sort.py`
* `ruff check sorts/topological_sort.py`
* `ruff format --check sorts/topological_sort.py`
* `git diff --check`

Note: `python3 -m pytest sorts/topological_sort.py` collects no tests because this existing file has no doctests.

### DIFF
--- a/sorts/topological_sort.py
+++ b/sorts/topological_sort.py
@@ -25,8 +25,8 @@ def topological_sort(start: str, visited: list[str], sort: list[str]) -> list[st
         # if neighbor not in visited, visit
         if neighbor not in visited:
             sort = topological_sort(neighbor, visited, sort)
-    # if all neighbors visited add current to sort
-    sort.append(current)
+    # if all neighbors visited add current before its descendants
+    sort.insert(0, current)
     # if all vertices haven't been visited select a new one to visit
     if len(visited) != len(vertices):
         for vertice in vertices:
--- a/sorts/topological_sort.py
+++ b/sorts/topological_sort.py
@@ -21,7 +21,10 @@ def _visit(
     post_order: list[str],
     graph: dict[str, list[str]],
 ) -> None:
-    """Visit all descendants of the current vertex using DFS."""
+    """Visit descendants of ``current``.
+
+    Appends to ``visited`` and ``post_order`` in place.
+    """
     visited.append(current)
     for neighbor in graph[current]:
         if neighbor not in visited:
@@ -32,7 +35,7 @@ def _visit(
 def topological_sort(
     start: str,
     visited: list[str],
-    sort: list[str],
+    order: list[str],
     graph: dict[str, list[str]] | None = None,
     vertices_list: list[str] | None = None,
 ) -> list[str]:
@@ -52,13 +55,14 @@ def topological_sort(
     if vertices_list is None:
         vertices_list = list(graph)
 
-    _visit(start, visited, sort, graph)
+    _visit(start, visited, order, graph)
     if len(visited) != len(vertices_list):
         for vertice in vertices_list:
             if vertice not in visited:
-                _visit(vertice, visited, sort, graph)
-    sort.reverse()
-    return sort
+                _visit(vertice, visited, order, graph)
+    # DFS post-order is reverse topological order, so reverse once at the end.
+    order.reverse()
+    return order
 
 
 if __name__ == "__main__":
--- a/sorts/topological_sort.py
+++ b/sorts/topological_sort.py
@@ -15,27 +15,53 @@ edges: dict[str, list[str]] = {
 vertices: list[str] = ["a", "b", "c", "d", "e"]
 
 
-def topological_sort(start: str, visited: list[str], sort: list[str]) -> list[str]:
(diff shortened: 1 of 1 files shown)


mindaugl (approved):

sorts/topological_sort.py
```diff
@@ -25,8 +25,8 @@ def topological_sort(start: str, visited: list[str], sort: list[str]) -> list[st
         # if neighbor not in visited, visit
         if neighbor not in visited:
             sort = topological_sort(neighbor, visited, sort)
-    # if all neighbors visited add current to sort
-    sort.append(current)
+    # if all neighbors visited add current before its descendants
```
jgohel9902: This comment seems inaccurate. At this point, the DFS has already visited all reachable descendants, and `current` is being inserted before them in the final ordering. Consider rewording it to avoid confusion.

sorts/topological_sort.py
```diff
@@ -25,8 +25,8 @@ def topological_sort(start: str, visited: list[str], sort: list[str]) -> list[st
...
+    sort.insert(0, current)
```
jgohel9902: Using `insert(0, current)` fixes the ordering, but it is O(n) for every insertion. For larger graphs, consider appending during DFS and reversing once at the end for better performance.

sorts/topological_sort.py
```diff
@@ -25,8 +25,8 @@ def topological_sort(start: str, visited: list[str], sort: list[str]) -> list[st
...
     # if all vertices haven't been visited select a new one to visit
     if len(visited) != len(vertices):
```
jgohel9902: This function depends on the global `vertices` list, which makes it less reusable. Consider passing `vertices` as a parameter instead of relying on module-level state.

sorts/topological_sort.py
jgohel9902: Consider adding a small assertion-based test instead of only printing the result. That would make it easier to verify the corrected topological order automatically.

sorts/topological_sort.py
```diff
@@ -15,27 +15,53 @@
...
+        vertices_list = list(graph)
+
+    _visit(start, visited, sort, graph)
+    if len(visited) != len(vertices_list):
+        for vertice in vertices_list:
             if vertice not in visited:
-                sort = topological_sort(vertice, visited, sort)
-    # return sort
+                _visit(vertice, visited, sort, graph)
+    sort.reverse()
```
jgohel9902: This is the core fix — adding sort.reverse() correctly produces 
topological order since DFS post-order appends nodes after all 
descendants are visited, giving reverse topological order.

Worth adding a brief comment explaining why the reverse is needed 
so future readers understand the intent:

# Post-order DFS gives reverse topological order, so reverse to fix
sort.reverse()

sorts/topological_sort.py
```diff
@@ -15,27 +15,53 @@
 vertices: list[str] = ["a", "b", "c", "d", "e"]
 
 
-def topological_sort(start: str, visited: list[str], sort: list[str]) -> list[str]:
-    """Perform topological sort on a directed acyclic graph."""
-    current = start
-    # add current to visited
+def _visit(
```
jgohel9902, line 18: Good refactor extracting _visit() as a private helper — separating 
the DFS traversal from the main topological_sort() logic makes both 
functions easier to understand and test independently.

One suggestion: _visit() modifies visited and post_order in place 
via side effects but has no doctest. Since it's a private helper 
this is acceptable, but a one-line docstring explaining the side 
effects would help:

def _visit(current, visited, post_order, graph):
    """DFS helper — appends nodes to visited and post_order in place."""

sorts/topological_sort.py
```diff
@@ -15,27 +15,53 @@
...
-        for vertice in vertices:
+            _visit(neighbor, visited, post_order, graph)
+    post_order.append(current)
+
+
+def topological_sort(
+    start: str,
+    visited: list[str],
+    sort: list[str],
+    graph: dict[str, list[str]] | None = None,
```
jgohel9902, line 39: Using mutable default arguments as None with an explicit check on 
lines 50-53 is the correct Python pattern — good call avoiding the 
classic mutable default argument trap.

However the parameter name "sort" on line 35 is a bit confusing 
since "sort" is also a Python built-in. Consider renaming to 
"result" or "order" to avoid shadowing and improve readability:

def topological_sort(
    start: str,
    visited: list[str],
    order: list[str],   # renamed from "sort"
    graph: dict[str, list[str]] | None = None,
    vertices_list: list[str] | None = None,
) -> list[str]:

jordandunmire97-ai (approved):

cclauss:
Please rebase.

PR #14623, opened 2026-05-03
feat: add Snell's Law refraction angle calculation

soradacroi:
### Describe your change:

Added snells_law.py to the physics/ directory.
This implements the calculate_refraction_angle function,
which finds the refraction angle based on Snell's Law.

The implementation handles standard refraction, the critical angle (90°), Total Internal Reflection (TIR)
and Validation logic to catch physically impossible inputs (refractive index ratios less than -1).

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/physics/snells_law.py
+++ b/physics/snells_law.py
@@ -1,0 +1,75 @@
+"""
+Snell's Law — Refraction Angle Calculation.
+
+Calculates the angle of refraction when light passes between two
+media using Snell's Law: n1 * sin(theta1) = n2 * sin(theta2).
+
+Reference: https://en.wikipedia.org/wiki/Snell%27s_law
+"""
+
+import math
+
+
+def calculate_refraction_angle(
+    index_of_refraction_1: float,
+    index_of_refraction_2: float,
+    incident_angle_degrees: float,
+) -> float:
+    """
+    Calculates the refraction angle of light passing from one medium to another.
+    The law states that, for a given pair of media, the ratio of the sines
+    of angle of incidence and angle of refraction s equal to the refractive
+    index of the second medium with regard to the first which is equal to the
+    ratio of the refractive indices of the two media, or equivalently, to the
+    ratio of the phase velocities in the two media.
+
+
+    Formula: n1 * sin(theta1) = n2 * sin(theta2)
+    or     : (sin(theta1) / sin(theta2)) = (n2 / n1)
+    Where:
+        n1 = refractive index of the first medium
+        n2 = refractive index of the second medium
+        theta1 = angle of incidence
+        theta2 = angle of refraction
+
+    Note: Total Internal Reflection (TIR) occurs when light travels from a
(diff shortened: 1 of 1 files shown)


physics/snells_law.py
```diff
@@ -0,0 +1,66 @@
...
+    ValueError: Invalid physical inputs: ratio cannot be less than -1.
+
+    """
+
+    incident_angle_radians = math.radians(incident_angle_degrees)
+    refraction_sine = (index_of_refraction_1 / index_of_refraction_2) * math.sin(
+        incident_angle_radians
+    )
+
+    if refraction_sine > 1:
```
jgohel9902: Good check for Total Internal Reflection — this correctly catches
cases where the sine value exceeds 1 which is physically impossible.

However the check on line 57 (refraction_sine < -1) will never
be reached in practice since a negative refractive index ratio
combined with a valid incident angle would need very specific
conditions. Worth adding a comment explaining when this case
actually occurs, or combining both checks:

if not -1 <= refraction_sine <= 1:
    raise ValueError("Invalid physical inputs: ratio cannot be less than -1.")

physics/snells_law.py
```diff
@@ -0,0 +1,66 @@
...
+        raise ValueError("Total Internal Reflection occurred.")
+    if refraction_sine < -1:
+        raise ValueError("Invalid physical inputs: ratio cannot be less than -1.")
+
+    # If the sine value is approximately 1.0, it's at the critical angle
+    # We use math.isclose to account for floating-point precision errors
+    if math.isclose(refraction_sine, 1.0):
```
jgohel9902, line 66: Smart use of math.isclose() here to handle floating-point precision
errors at the critical angle boundary — good defensive programming.

Suggestion: consider also handling the case where refraction_sine
is close to -1.0 symmetrically, since the current code raises a
ValueError for values slightly below -1 due to floating-point
errors, which could be confusing to callers passing valid inputs.

physics/snells_law.py
```diff
@@ -0,0 +1,66 @@
+import math
```
jgohel9902, line 10: The file is missing a module-level docstring. The repo's convention
is to have a module docstring before imports. Consider adding:

"""
Snell's Law — Refraction Angle Calculation.

Calculates the angle of refraction when light passes between two
media using Snell's Law: n1 * sin(theta1) = n2 * sin(theta2).

Reference: https://en.wikipedia.org/wiki/Snell%27s_law
"""

soradacroi:
Thank you for pointing those out @jgohel9902, I have put a module-level docstring and added math.isclose() for refraction_sine which might be close to -1.0 and merged the two condition (refraction_sine < -1 and refraction_sine > 1) into one.

jordandunmire97-ai (approved):

cclauss (approved):

PR #14635, opened 2026-05-06
fix: int_to_roman silently returns wrong values for out-of-range input

faizzyhon:
## Problem

`int_to_roman` performed no input validation. Python's floor-division semantics silently corrupted results for out-of-range inputs:

```python
# Before this fix:
int_to_roman(0)    # ''          <- silent wrong output
int_to_roman(-1)   # 'CMXCIX'   <- silently returns 999 (WRONG)
int_to_roman(-5)   # 'CMXCV'    <- silently returns 995 (WRONG)
int_to_roman(4000) # 'MMMM'     <- not a valid standard Roman numeral
int_to_roman(True) # 'I'        <- bool silently accepted as 1
```

Root cause: `divmod(-1, 1000)` returns `(-1, 999)` in Python, so `'M' * -1` yields `''` and the remainder `999` propagates through the ROMAN lookup table, producing a plausible-looking but completely wrong answer.

## Fix

Added early validation guards:

1. **TypeError** for non-integer types (including `bool`, which is a subclass of `int`)
2. **ValueError** for integers outside the valid range [1, 3999]
3. Added four doctest error cases so the guards are exercised by the doctest runner

## Describe your change:

## Checklist:

### DIFF
--- a/conversions/roman_numerals.py
+++ b/conversions/roman_numerals.py
@@ -40,12 +40,37 @@ def roman_to_int(roman: str) -> int:
 
 def int_to_roman(number: int) -> str:
     """
-    Given a integer, convert it to an roman numeral.
+    Given an integer, convert it to a Roman numeral.
+    Input must be an integer in the range 1 to 3999.
     https://en.wikipedia.org/wiki/Roman_numerals
     >>> tests = {"III": 3, "CLIV": 154, "MIX": 1009, "MMD": 2500, "MMMCMXCIX": 3999}
     >>> all(int_to_roman(value) == key for key, value in tests.items())
     True
+    >>> int_to_roman(0)
+    Traceback (most recent call last):
+        ...
+    ValueError: int_to_roman only accepts integers in the range 1 to 3999, got 0
+    >>> int_to_roman(-1)
+    Traceback (most recent call last):
+        ...
+    ValueError: int_to_roman only accepts integers in the range 1 to 3999, got -1
+    >>> int_to_roman(4000)
+    Traceback (most recent call last):
+        ...
+    ValueError: int_to_roman only accepts integers in the range 1 to 3999, got 4000
+    >>> int_to_roman(True)
+    Traceback (most recent call last):
+        ...
+    TypeError: int_to_roman only accepts integers, got bool
     """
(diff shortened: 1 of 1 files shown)


jordandunmire97-ai (approved):

conversions/roman_numerals.py
```diff
@@ -40,12 +40,35 @@ def roman_to_int(roman: str) -> int:
...
+    >>> int_to_roman(4000)
+    Traceback (most recent call last):
+        ...
+    ValueError: int_to_roman only accepts integers in the range 1 to 3999, got 4000
+    >>> int_to_roman(True)
+    Traceback (most recent call last):
+        ...
+    TypeError: int_to_roman only accepts integers, got bool
     """
+    if not isinstance(number, int) or isinstance(number, bool):
```
cclauss:
```suggestion
    if not isinstance(number, (bool, int)):
```

conversions/roman_numerals.py
```diff
@@ -40,12 +40,35 @@ def roman_to_int(roman: str) -> int:
...
+    >>> int_to_roman(4000)
+    Traceback (most recent call last):
+        ...
+    ValueError: int_to_roman only accepts integers in the range 1 to 3999, got 4000
+    >>> int_to_roman(True)
+    Traceback (most recent call last):
+        ...
+    TypeError: int_to_roman only accepts integers, got bool
     """
+    if not isinstance(number, (bool, int)):
```
cclauss:
```suggestion
    if not isinstance(number, int):
```

cclauss (approved):

conversions/roman_numerals.py
```diff
@@ -40,12 +40,35 @@ def roman_to_int(roman: str) -> int:
...
+    >>> int_to_roman(4000)
+    Traceback (most recent call last):
+        ...
+    ValueError: int_to_roman only accepts integers in the range 1 to 3999, got 4000
+    >>> int_to_roman(True)
+    Traceback (most recent call last):
+        ...
+    TypeError: int_to_roman only accepts integers, got bool
     """
+    if not isinstance(number, int):
```
cclauss:
```suggestion
    if isinstance(number, bool) or not isinstance(number, int):
```

PR #14647, opened 2026-05-07
Qr Decomposition

BhaktiVagadia:
### Describe your change:

* [X] Add an algorithm?
* [X] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/linear_algebra/qr_decomposition.py
+++ b/linear_algebra/qr_decomposition.py
@@ -6,8 +6,10 @@ QR decomposition is often used to solve the linear least squares (LLS) problem
 and is the basis for a particular eigenvalue algorithm, the QR algorithm.
 This algorithm will simply attempt to perform QR decomposition on any square matrix.
 Reference: https://en.wikipedia.org/wiki/QR_decomposition"""
+
 import numpy as np
 from scipy.linalg import qr
+
 
 def qr_decomposition(matrix_a: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
     """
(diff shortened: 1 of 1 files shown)


jordandunmire97-ai (approved):

cclauss:
@priya-sundaram-dev, please review. Is this a how-to-use scipy.linalg.qr, or is it a real algorithm?

priya-sundaram-dev:
@cclauss it's the former — this is a thin wrapper around `scipy.linalg.qr`, not a from-scratch algorithm. The only computational line is `qr(matrix_a, pivoting=True)`; everything else is shape validation and a round-trip assertion. It doesn't teach how QR decomposition works, which is the point of this repo.

Two more concerns:

1. **Duplicate.** We already have a from-scratch, numerically-stable QR here: [`maths/qr_decomposition.py`](https://github.com/TheAlgorithms/Python/blob/master/maths/qr_decomposition.py) (`qr_householder`, Householder reflections). This PR adds a second `qr_decomposition.py` under `linear_algebra/` that does less.

2. **Fragile doctests.** The doctests hard-code rounded `Q`/`R` arrays, but QR factors are only unique up to column/row sign, and `pivoting=True` adds a permutation whose order is LAPACK/BLAS-version dependent. Those exact arrays will fail on other platforms. (Contrast the existing file, which asserts *properties* — `Q@R == A`, `Q` orthogonal, `R` upper-triangular — via `np.allclose`, so it's portable.) There are also inconsistent validation messages ("greater than 2" fires when a dimension equals 2).

My recommendation: close this in favor of the existing `maths/qr_decomposition.py`. If the author wants to contribute here, a genuinely new *from-scratch* variant (e.g. Gram–Schmidt or Givens-rotation QR, with property-based doctests) would be a welcome addition and wouldn't duplicate the Householder version. Thanks for the work, @BhaktiVagadia — the interest is appreciated, it's just the wrong shape for this repo.

PR #14701, opened 2026-05-19
Fix/tree sort iterable type hint

DrishtiTripathi2230:
### Describe your change:
Improve type hint for `tree_sort` to accept any `Iterable[int]` instead of only `list[int]`, and enable the previously commented-out range doctest.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/sorts/tree_sort.py
+++ b/sorts/tree_sort.py
@@ -50,14 +50,17 @@ def tree_sort(arr: Iterable[int]) -> tuple[int, ...]:
     (-4, 2, 5, 7, 9)
     >>> tree_sort([5, 6, 1, -1, 4, 37, 2, 7])
     (-1, 1, 2, 4, 5, 6, 7, 37)
-
     >>> tree_sort(range(10, -10, -1)) == tuple(sorted(range(10, -10, -1)))
     True
     """
-    if len(arr) == 0:
-        return tuple(arr)
-    root = Node(arr[0])
-    for item in arr[1:]:
+    iterator = iter(arr)
+    try:
+        first = next(iterator)
+    except StopIteration:
+        return ()
+
+    root = Node(first)
+    for item in iterator:
         root.insert(item)
     return tuple(root)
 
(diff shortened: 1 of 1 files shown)


sorts/tree_sort.py
```diff
@@ -39,7 +38,7 @@ def insert(self, val: int) -> None:
                 self.right.insert(val)
 
 
-def tree_sort(arr: list[int]) -> tuple[int, ...]:
+def tree_sort(arr: Iterable[int]) -> tuple[int, ...]:
```
diusazzad, line 41: Hey @DrishtiTripathi2230! Thanks for taking the time to improve the type hints! 🚀

Changing the type hint to `Iterable[int]` is a great idea conceptually. However, it introduces a subtle bug. Since `Iterable` includes generators, passing a generator will cause the code to crash because:
1. `len(arr)` will raise a `TypeError` (generators have no `len()`).
2. `arr[0]` and `arr[1:]` will also raise a `TypeError` (generators are not subscriptable).

To fix this and fully support any `Iterable`, we can convert the iterable into an iterator using `iter()`. Here is a suggested fix for the logic inside the function:

```python
def tree_sort(arr: Iterable[int]) -> tuple[int, ...]:
    # ... docstrings ...
    iterator = iter(arr)
    try:
        first = next(iterator)
    except StopIteration:
        return ()
        
    root = Node(first)
    for item in iterator:
        root.insert(item)
    return tuple(root)

diusazzad (changes requested):
provided a fix for iterable type hint

sorts/tree_sort.py
DrishtiTripathi2230: Thanks for the explanation!! I'll update the function to use `iter()` so it properly handles generators and other iterables.

jordandunmire97-ai (approved):

cclauss (approved):

PR #14762, opened 2026-06-06
fix: correct treap split() comparison to match docstring behavior

### LINKED ISSUE #7854: treap not working expected when splitting on value that existed
### Repository commit

master

any

any

### Expected behavior

data_structures/binary_tree/treap.py

according to the docstring of treap's split function below, when we split a treap into two, the left treap should contain values less than split value, and the right should contain values greater or equal than split value.
```python
def split(root: Node | None, value: int) -> tuple[Node | None, Node | None]:
    """
    We split current tree into 2 trees with value:
    Left tree contains all values less than split value.
    Right tree contains all values greater or equal, than split value
    """
```

Which means the expected return of following code should be:
```
root = interact_treap(None, "+1")
r1, r2 = split(root, 1)  # r1 should be None,  r2 should have value 1.
```

### Actual behavior

However, the actual behavior of the same code is the opposite:

```python
root = interact_treap(None, "+1")
r1, r2 = split(root, 1)  # r1 is 1, r2 is None when running the code
```

This can be fixed by either changing the docstring or by using `<=` instead of `<` in the if condition within `split` function.

maitriupadhyay03-cell:
### Describe your change:

The `split()` function in `data_structures/binary_tree/treap.py` uses a strict `<` comparison, but the docstring states:

> Left tree contains all values less than split value.
> Right tree contains all values greater or equal, than split value

This means when splitting by a value that already exists in the treap, the existing node ends up in the LEFT subtree instead of the RIGHT subtree, which contradicts the documented behavior.

**Fix:** Changed `elif value < root.value:` to `elif value <= root.value:` so that equal values go to the right subtree as documented.

Fixes #7854

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/data_structures/binary_tree/treap.py
+++ b/data_structures/binary_tree/treap.py
@@ -5,7 +5,7 @@ from random import random
 
 class Node:
     """
-    Treap's node
+    Treap's nodeh
     Treap is a binary tree by value and heap by priority
     """
 
@@ -41,7 +41,7 @@ def split(root: Node | None, value: int) -> tuple[Node | None, Node | None]:
     """
     if root is None or root.value is None:  # None tree is split into 2 Nones
         return None, None
-    elif value < root.value:
+    elif value <= root.value:
         """
         Right tree's root will be current node.
         Now we split(with the same value) current node's left son
--- a/data_structures/binary_tree/treap.py
+++ b/data_structures/binary_tree/treap.py
@@ -5,7 +5,7 @@ from random import random
 
 class Node:
     """
-    Treap's nodeh
+    Treap's node
     Treap is a binary tree by value and heap by priority
     """
 
--- a/data_structures/binary_tree/treap.py
+++ b/data_structures/binary_tree/treap.py
@@ -101,8 +101,8 @@ def erase(root: Node | None, value: int) -> Node | None:
     Split all nodes with values greater into right.
     Merge left, right
     """
-    left, right = split(root, value - 1)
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev:
Thanks for this @maitriupadhyay03-cell — the `split` change is correct and finally matches the docstring. 🙌

One thing to finish before it can close #7854: this one-liner breaks `erase`, which was written against the *old* `split` semantics (a known gotcha noted on the issue back in 2022). The current `interact_treap` doctest fails after your change:

```
-0  → Got 0,1,2,...   (0 not erased)
-4  → Got 0,1,2,4,4,4,4,5,...
```

Updating `erase` to the new semantics fixes it and makes all doctests pass:

```python
def erase(root: Node | None, value: int) -> Node | None:
    left, right = split(root, value)
    _, right = split(right, value + 1)
    return merge(left, right)
```

With both `split` and `erase` updated, `python -m doctest treap.py -v` is green. 👍

cclauss (approved):

priya-sundaram-dev:
Nice — that's exactly it. ✅ Pulled your branch and ran `python -m doctest treap.py -v`: **11 tests in 12 items, 11 passed / 0 failed**. Both `split` and `erase` now line up with the docstring, and the `interact_treap` doctest (the `-0` / `-4` erase cases) is green again. This should be good to close #7854. Thanks for following through! 🙌

PR #14768, opened 2026-06-06
fix(maths): correct recursive call in sum_of_digits_recursion

fauzan171:
## Summary

Fix a bug where `sum_of_digits_recursion()` was calling `sum_of_digits()` (the iterative version) instead of itself in the recursive case.

## Bug
In `maths/sum_of_digits.py` line 34:
```python
# Before (bug): calls the iterative function, not itself
return n if n < 10 else n % 10 + sum_of_digits(n // 10)

# After (fix): properly calls itself recursively
return n if n < 10 else n % 10 + sum_of_digits_recursion(n // 10)
```

## Impact
- The function was not actually recursive as its name and documentation suggest
- The result was correct (it fell back to the iterative version), but the intent was wrong
- This is a correctness fix for the algorithm implementation

## Checklist
- [x] All doctests pass
- [x] No behavioral change in output (the function still produces correct results)

### DIFF
--- a/maths/sum_of_digits.py
+++ b/maths/sum_of_digits.py
@@ -31,7 +31,7 @@ def sum_of_digits_recursion(n: int) -> int:
     0
     """
     n = abs(n)
-    return n if n < 10 else n % 10 + sum_of_digits(n // 10)
+    return n if n < 10 else n % 10 + sum_of_digits_recursion(n // 10)
 
 
 def sum_of_digits_compact(n: int) -> int:


jordandunmire97-ai (approved):

cclauss:

priya-sundaram-dev:
Reviewed — this is a correct, minimal fix and worth landing.

The bug is subtle: `sum_of_digits_recursion()` was calling the *iterative* `sum_of_digits()` in its recursive branch, so it peeled exactly one digit and then delegated the rest to the loop. The result was numerically correct (hence the passing doctests), but the function was not actually recursive — its name promised something it did not do. The patch makes it genuinely self-recursive.

I re-ran the existing doctests against the patched version to be sure the behaviour is unchanged:
- `sum_of_digits_recursion(12345)` → 15
- `sum_of_digits_recursion(-123)` → 6 (abs() guard still applies at each frame since n is non-negative after the first call)
- `sum_of_digits_recursion(0)` → 0

All pass. One optional nit for the author, not a blocker: since the recursion depth is O(number of digits) it stays tiny for any practical int, so no stack concerns. LGTM.

cclauss (approved):

PR #14792, opened 2026-06-10
updated the kmap simplification algorithm to give the most simplified expression

ParthKGarg:
### Describe your change:

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/boolean_algebra/karnaugh_map_simplification.py
+++ b/boolean_algebra/karnaugh_map_simplification.py
@@ -3,46 +3,53 @@ https://en.wikipedia.org/wiki/Karnaugh_map
 https://www.allaboutcircuits.com/technical-articles/karnaugh-map-boolean-algebraic-simplification-technique
 """
 
-# Changes I can make - 
-# 1. Make a function for three and four variable kmap
-# 2. Make truly simplified expressions
-
-# Approach - Either store every possible box and brute force or make general expression and then simplify
-
-
 
 def simplify_kmap(kmap: list[list[int]]) -> str:
     """
-    Simplify the Karnaugh map.
+    Simplify a 2 variable Karnaugh map.
+
+    Args:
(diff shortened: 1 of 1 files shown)


jordandunmire97-ai (approved):

cclauss:
@priya-sundaram-dev, please review. We are losing a lot of tests in the proposed changes.  Is there strong value in these changes?

priya-sundaram-dev:
Thanks @cclauss — I looked closely, and I think there's real value here *if* the author restores the coverage.

**The algorithm change is a genuine improvement.** `main` doesn't actually simplify — it just lists minterms, so `[[0,1],[1,1]]` returns `"A'B + AB' + AB"`. Real Karnaugh-map simplification groups adjacent 1s, giving `"A + B"`. The PR implements that grouping (4-cell → 2-cell pairs → single cells), which is what a file named `karnaugh_map_simplification.py` should do. So the output is *more* correct, not less.

**Your instinct about lost tests is right, though.** The PR drops the robustness doctests that exercised non-`1` truthy inputs (`-1`, `1.1`, `'a'`) and the all-zero `''` case. I checked the new code against those inputs:

```
[[0,1],[1,-1]]  -> 'A + B'
[[0,1],[1,1.1]] -> 'A + B'
[[0,1],[1,'a']] -> 'A + B'
[[0,0],[0,0]]   -> '0'
```

So the *behavior* is still fine — it's the regression protection that's gone. I'd ask the author to re-add those as doctests with the updated expected outputs before merge, so we don't silently lose that coverage.

Minor: the algorithm is now hardcoded to the 2-variable (2×2) case; the docstring correctly says "2 variable" now, so that's acceptable as-is.

Recommendation: **request the dropped edge-case doctests be restored (with new expected values)**, then this is a solid net improvement. New doctests pass locally.

PR #14807, opened 2026-06-15
fix: handle empty input in jump_search

pony-maggie:
### Describe your change:

Fixed `jump_search([], item)` so it returns `-1` instead of raising `IndexError`.

## Why

The function documents `-1` as the result when the item is not found. An empty collection is a valid not-found case and should not attempt to index into the sequence.

## How tested

- RED: `/opt/homebrew/opt/python@3.14/bin/python3.14 -m doctest -v searches/jump_search.py` failed with `IndexError` before the fix
- `/opt/homebrew/opt/python@3.14/bin/python3.14 -m doctest -v searches/jump_search.py`
- `/tmp/thealgorithms-py314-pytest/bin/python -m pytest --doctest-modules searches/jump_search.py -q`
- Randomized comparison against expected search results for 200 generated sorted lists
- `git diff --check`

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/searches/jump_search.py
+++ b/searches/jump_search.py
@@ -33,9 +33,14 @@ def jump_search[T: Comparable](arr: Sequence[T], item: T) -> int:
     10
     >>> jump_search(["aa", "bb", "cc", "dd", "ee", "ff"], "ee")
     4
+    >>> jump_search([], 1)
+    -1
     """
 
     arr_size = len(arr)
+    if arr_size == 0:
+        return -1
+
     block_size = int(math.sqrt(arr_size))
 
     prev = 0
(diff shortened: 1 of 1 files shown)


mindaugl (approved):

jordandunmire97-ai (approved):

priya-sundaram-dev (approved):
Correct edge-case fix: `jump_search([], item)` previously computed `block_size = int(math.sqrt(0)) = 0` and could loop/misbehave; returning `-1` for an empty array is the right contract and matches the "not found" return. Doctest added, CI green. LGTM ✅

Note for the maintainers: #15101 is an identical fix to the same function — this one (#14807) came first, so I'd suggest merging this and closing #15101 as a duplicate.

cclauss (approved):

jordandunmire97-ai (approved):
we forsee a painting of what the mind is capable with the calm thrive to enhance in every way . evolve . what is real , always will prosper.

PR #14884, opened 2026-07-01
remove dead try/except wrapper around mass * acceleration in newtons_second_law_of_motion

HrachShah:
The function is annotated `(mass: float, acceleration: float) -> float` and multiplying two floats cannot raise an exception on valid inputs.

The previous wrapper:

```python
force = 0.0
try:
    force = mass * acceleration
except Exception:
    return -0.0
return force
```

was dead code that:
- silently masked `TypeError` when a caller violated the type contract (e.g. passing a string)
- silently swallowed `KeyboardInterrupt` and `SystemExit`
- returned the unusual signed zero `-0.0`, which can poison downstream numerical code (it equals `0.0` in equality checks but is distinct in repr and bit pattern)

Replacing the wrapper with a plain `return mass * acceleration` makes the implementation match the function's annotated type signature and its existing doctests (`newtons_second_law_of_motion(10, 10) == 100` and `newtons_second_law_of_motion(2.0, 1) == 2.0`).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: sudo apt-get update && sudo apt-get install -y libhdf5-dev
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@v6
       - uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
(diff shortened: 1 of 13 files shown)


cclauss:
mass=“mass”, acceleration=“ acceleration” would raise TypeError.

PR #14907, opened 2026-07-07
Add is_palindrome_ignore_case_and_spaces to strings/palindrome.py

kaviya-ns:
### Describe your change:

Adds a new function `is_palindrome_ignore_case_and_spaces` to `strings/palindrome.py` 
that checks whether a string is a palindrome while ignoring letter case and spaces. 
Existing functions in the file are unchanged.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [ ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a closing keyword.

### DIFF
--- a/strings/palindrome.py
+++ b/strings/palindrome.py
@@ -81,6 +81,26 @@ def is_palindrome_slice(s: str) -> bool:
     return s == s[::-1]
 
 
+def is_palindrome_ignore_case_and_spaces(s: str) -> bool:
+    """
+    Return True if s is a palindrome, ignoring case and spaces.
+    Otherwise return False.
+
+    >>> is_palindrome_ignore_case_and_spaces("A man a plan a canal Panama")
+    True
+    >>> is_palindrome_ignore_case_and_spaces("Was it a car or a cat I saw?")
+    False
+    >>> is_palindrome_ignore_case_and_spaces("Hello World")
+    False
+    >>> is_palindrome_ignore_case_and_spaces("Never Odd or Even")
+    True
+    >>> is_palindrome_ignore_case_and_spaces("")
+    True
+    """
+    s = s.lower().replace(" ", "")
+    return s == s[::-1]
+
+
 def benchmark_function(name: str) -> None:
     stmt = f"all({name}(key) == value for key, value in test_data.items())"
     setup = f"from __main__ import test_data, {name}"
@@ -93,6 +113,7 @@ if __name__ == "__main__":
     for key, value in test_data.items():
         assert is_palindrome(key) == is_palindrome_recursive(key)
         assert is_palindrome(key) == is_palindrome_slice(key)
+        assert is_palindrome(key) == is_palindrome_ignore_case_and_spaces(key)
         print(f"{key:21} {value}")
     print("a man a plan a canal panama")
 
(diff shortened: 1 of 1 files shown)


mindaugl (approved):

Abhirai2006 (changes requested):
Nice small addition, thanks! Two things I noticed though:

The doctest with "Was it a car or a cat I saw?" returns False, but that's kind of an accident — it's only False because the trailing "?" doesn't get stripped (the function only removes spaces, not punctuation). If you actually remove the "?", that phrase is a real palindrome ignoring case/spaces. So the test passes but not really for the reason you'd want it to. Might be worth swapping in a cleaner example, or handling punctuation too so the function does what its name says.

Also the new assert in __main__ (`is_palindrome(key) == is_palindrome_ignore_case_and_spaces(key)`) is checked against the existing test_data, but none of those strings have spaces or mixed case — so it's not really testing the new behavior, just repeating the old check. Adding one or two mixed-case/spaced entries would make that assertion actually mean something.

Otherwise looks good, happy to approve once these are sorted.

kaviya-ns:
Hi @Abhirai2006  , Thanks for the feedback! I've updated the function to also strip punctuation, so 'Was it a car or a cat I saw?' is now a genuine positive example rather than an accidental one. Also added a separate test_data_ignore_case_and_spaces dict with mixed-case, spaced, and punctuated entries, and an assert loop in __main__, so the new behavior is actually being tested rather than just repeating the existing checks.

strings/palindrome.py
```diff
@@ -81,6 +81,35 @@ def is_palindrome_slice(s: str) -> bool:
     return s == s[::-1]
 
 
+def is_palindrome_ignore_case_and_spaces(s: str) -> bool:
```
zain-cs, line 84: Nice implementation and the docstring covers it well. Small naming nit: since .isalnum() also strips punctuation, not just case and spaces, might be worth reflecting that in the function name too (or it's fine as-is if the broader scope was intentional).

strings/palindrome.py
```diff
@@ -94,13 +123,17 @@ def benchmark_function(name: str) -> None:
         assert is_palindrome(key) == is_palindrome_recursive(key)
         assert is_palindrome(key) == is_palindrome_slice(key)
         print(f"{key:21} {value}")
+    for key, value in test_data_ignore_case_and_spaces.items():
+        assert is_palindrome_ignore_case_and_spaces(key) == value
     print("a man a plan a canal panama")
 
-    # finished 500,000 runs in 0.46793 seconds
```
zain-cs, line 99: These benchmark timing changes for is_palindrome, is_palindrome_slice, and is_palindrome_recursive look unrelated to this PR's actual purpose — was this from a re-run on a different machine? Might be worth reverting these lines to keep the diff focused just on the new function.

zain-cs (commented):
Logic looks correct and all doctests check out — left two minor notes inline.

strings/palindrome.py
kaviya-ns: thanks! . those were leftover from a local re-run on my machine. I've reverted them back to the original values.

strings/palindrome.py
kaviya-ns: The broader scope (also stripping punctuation) was intentional. Since a fully descriptive name would get pretty long, I added a note in the docstring clarifying it also ignores punctuation, so the behavior is explicit there. Let me know if you'd still like me to rename it and I can update it.

cclauss (approved):

PR #14923, opened 2026-07-12
Refactor logistic regression to OOP class with multi-class support

rituuu001:
### Describe your change:

* [x] Add an algorithm? (Significantly expanded and enhanced the existing algorithm file into a robust, scalable multi-class classifier)
* [x] Add or change doctests?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER". (Submitted directly without an open issue, as per Contributing guidelines for algorithm additions)

---

###  Summary of Enhancements

Refactored the core layout of `machine_learning/logistic_regression.py`. The original implementation was limited to Batch Gradient Descent and binary classification. This upgrade introduces a robust, object-oriented `LogisticRegression` class capable of handling both binary and advanced multi-class datasets entirely from scratch using NumPy.

* **Mini-batch Gradient Descent:** Added efficient mini-batch sampling configurations to significantly optimize convergence and scalability on larger datasets.
* **True Multiclass Support:** Implemented standalone mathematical variations for both **One-vs-Rest (OVR)** and **Softmax (Multinomial)** regression frameworks.
* **Early Stopping:** Programmed vector log-loss tracking limits to break calculation epochs early once structural threshold targets stabilize (`< 1e-6`).
* **NumPy Random Generator Modernization:** Upgraded all legacy execution calls (`np.random.seed`, `np.random.randn`, `np.random.permutation`) to utilize modern `np.random.default_rng()` constraints as recommended by NumPy documentation.
* **Type-Hinted & Lint Verified:** Validated and confirmed 100% clean passes locally using the project's strict verification tools (`black`, `ruff check`, and `mypy`).

---

### Verification Results

All required verification suites were executed locally and passed with zero errors:

* **Doctest Validation:** Clean pass (7 tests passed in 11 items).
  ```powershell
  py -m doctest -v machine_learning/logistic_regression.py

### DIFF
--- a/machine_learning/logistic_regression.py
+++ b/machine_learning/logistic_regression.py
@@ -1,159 +1,260 @@
-#!/usr/bin/python
-
-# Logistic Regression from scratch
-
-# In[62]:
-
-# In[63]:
-
-# importing all the required libraries
-
+#!/usr/bin/python3
 """
-Implementing logistic regression for classification problem
-Helpful resources:
-Coursera ML course
-https://medium.com/@martinpella/logistic-regression-from-scratch-in-python-124c5636b8ac
+Implementing Logistic Regression (Binary, One-vs-Rest, and Softmax Multi-class)
+from scratch using NumPy.
+
+References:
+- Wikipedia: https://en.wikipedia.org/wiki/Logistic_regression
+- Coursera Machine Learning Course by Andrew Ng
 """
 
 import numpy as np
-from matplotlib import pyplot as plt
-from sklearn import datasets
-
-# get_ipython().run_line_magic('matplotlib', 'inline')
-
-
-# In[67]:
-
-# sigmoid function or logistic function is used as a hypothesis function in
-# classification problems
 
 
 def sigmoid_function(z: float | np.ndarray) -> float | np.ndarray:
     """
-    Also known as Logistic Function.
+    Also known as the Logistic Function.
 
                 1
-    f(x) =   -------
-              1 + e⁻ˣ
+    f(z) =  -------
+              1 + e⁻ᶻ
 
-    The sigmoid function approaches a value of 1 as its input 'x' becomes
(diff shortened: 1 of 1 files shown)


cclauss:

It seems like we are losing a lot of tests in these changes.  Is there solid value added in the proposed changes?

priya-sundaram-dev:
Thanks @cclauss — I share the concern; I'd hold this one. The value added is real, but it doesn't offset the verifiable test coverage we'd lose.

**What's lost:** the PR replaces three small, heavily-doctested functions with one large OOP class. Concretely it drops 5 of the 7 numeric `sigmoid_function` doctests, and *deletes* `cost_function` and `log_likelihood` entirely — along with all of their numeric doctests (cross-entropy values, log-likelihood). The replacement `LogisticRegression` class ships a single smoke test, `len(clf.predict(...)) == 4`, with no numeric assertion on weights, loss, or probabilities.

For a teaching repo, the reproducible, asserted numeric behavior is the most valuable part of the file. Random-init OOP makes deterministic doctests harder, which is likely *why* they were dropped — but that's exactly the regression we don't want.

**Scope:** OVR + softmax multi-class also pushes well beyond the original single-example scope. That's not automatically bad, but it raises the bar for how well-tested it needs to be.

**Suggestion for the author:** either (a) keep the original functions and their doctests and add softmax as a separate, deterministically-seeded addition with its own numeric doctests; or (b) if going full OOP, seed the RNG and add doctests asserting on the learned probabilities/loss, and restore the sigmoid range tests plus a log-loss doctest.

As submitted I'd **not merge** — the tradeoff reduces tested behavior on an educational file. Happy to re-review once coverage is restored.

PR #15001, opened 2026-08-04
Validate scores for minimax function

### LINKED ISSUE #14886: Bug: Minimax crashes with RecursionError on non-power-of-2 input sizes
### Description

`backtracking/minimax.py` uses `math.log(len(scores), 2)` to compute tree height, then checks `depth == height` as the base case. Since `math.log` returns a float, the comparison `int == float` never succeeds for non-power-of-2 input sizes, causing infinite recursion.

```python
height = math.log(len(scores), 2)  # e.g. math.log(6, 2) = 2.584...
# ...
if depth == height:  # depth is int, height is float 2.584 -- NEVER TRUE
    return scores[node_index]
```

### Reproduction

```python
from minimax import minimax
import math

# Works (power of 2):
scores = [3, 5, 2, 9, 12, 5, 23, 23]  # len=8
print(minimax(0, 0, True, scores, math.log(8, 2)))  # OK: 12

# Crashes (non-power of 2):
scores = [3, 5, 2, 9, 12]  # len=5
print(minimax(0, 0, True, scores, math.log(5, 2)))  # RecursionError!
```

Crashes for sizes: 3, 5, 6, 7, 9, 10, 11, 12, 13, 14, 15, 17...

The doctests only pass because they use size 8 (`2^3`).

### Suggested Fix

Either validate input or use integer height:

```python
import math

def minimax(depth, node_index, is_max, scores, height):
    if depth < 0:
        raise ValueError("Depth cannot be less than 0")
    if not scores:
        raise ValueError("Scores cannot be empty")
    if not (len(scores) & (len(scores) - 1) == 0):
        raise ValueError("Number of scores must be a power of 2")
    # ...
```

Or rewrite to use array slicing instead of index arithmetic, which works for any size.

### Impact

Any user who passes a non-power-of-2 list gets an unhandled `RecursionError`. There is no documentation about this constraint.

Found during systematic algorithm audit: https://github.com/devladpopov/algorithm-autopsy

adithyanhp:
Fixes #14886

Add validation for scores in minimax function.

### Describe your change:

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/backtracking/minimax.py
+++ b/backtracking/minimax.py
@@ -46,6 +46,10 @@ def minimax(
     Traceback (most recent call last):
         ...
     ValueError: Scores cannot be empty
+    >>> minimax(0, 0, True, [1, 2, 3], 1.58)
+    Traceback (most recent call last):
+        ...
+    ValueError: Number of scores must be a power of 2
     >>> scores = [3, 5, 2, 9, 12, 5, 23, 23]
     >>> height = math.log(len(scores), 2)
     >>> minimax(0, 0, True, scores, height)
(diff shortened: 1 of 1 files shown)


mindaugl (approved):

Claudiocli (approved):
Correct validation of lengths and correct test provided

PR #15010, opened 2026-08-07
fix: bad-character shift in Boyer-Moore search has no effect

### LINKED ISSUE #14844: Bug: Boyer-Moore bad character shift has no effect (for-loop variable reassignment)
### Description

In `strings/boyer_moore_search.py`, the `bad_character_heuristic()` method attempts to skip positions using the bad character rule, but the shift has **no effect** because it reassigns the `for`-loop variable `i`:

```python
def bad_character_heuristic(self) -> list[int]:
    positions = []
    for i in range(self.textLen - self.patLen + 1):
        mismatch_index = self.mismatch_in_text(i)
        if mismatch_index == -1:
            positions.append(i)
        else:
            match_index = self.match_in_pattern(self.text[mismatch_index])
            i = (mismatch_index - match_index)  # <-- THIS HAS NO EFFECT
    return positions
```

In Python, reassigning the loop variable inside a `for` loop does not change the iteration. The next iteration always takes the next value from `range()`. This means:

- The bad character shift is completely ignored
- The algorithm checks **every position** sequentially
- It behaves as brute-force O(nm), not Boyer-Moore O(n/m)
- Results are correct, but performance is identical to naive search

### Reproduction

```python
class InstrumentedBMS(BoyerMooreSearch):
    def bad_character_heuristic(self):
        positions = []
        self.iterations = 0
        for i in range(self.textLen - self.patLen + 1):
            self.iterations += 1
            mismatch_index = self.mismatch_in_text(i)
            if mismatch_index == -1:
                positions.append(i)
            else:
                match_index = self.match_in_pattern(self.text[mismatch_index])
                i = (mismatch_index - match_index)
        return positions

text = "ABCDEFGHIJKLMNOPABCDEFGHIJKLMNOP"  # 32 chars
pattern = "MNOP"
bms = InstrumentedBMS(text, pattern)
positions = bms.bad_character_heuristic()
print(f"Iterations: {bms.iterations}")  # 29 (brute-force)
print(f"Expected for BM: ~8")           # should skip
```

Output: `Iterations: 29` — same as brute-force. The shift assignment is dead code.

### Suggested Fix

Replace `for` loop with `while` loop to allow the shift to take effect:

```python
def bad_character_heuristic(self) -> list[int]:
    positions = []
    i = 0
    while i <= self.textLen - self.patLen:
        mismatch_index = self.mismatch_in_text(i)
        if mismatch_index == -1:
            positions.append(i)
            i += 1
        else:
            match_index = self.match_in_pattern(self.text[mismatch_index])
            i = max(i + 1, mismatch_index - match_index)
    return positions
```

### Impact

This repository has 190k+ stars and is widely used as an educational reference. Students copying this implementation believe they have a Boyer-Moore algorithm with O(n/m) average performance, but actually have O(nm) brute-force search.

Found during systematic algorithm audit: https://github.com/devladpopov/algorithm-autopsy

bolverk:
# Fix bad-character shift having no effect (dead loop variable)

Fixes #14844

## Summary

`bad_character_heuristic()` in `strings/boyer_moore_search.py` assigned the
computed shift to the `for`-loop variable `i`:

```python
for i in range(self.textLen - self.patLen + 1):
    ...
    i = (mismatch_index - match_index)  # no effect: next iteration overwrites i
```

In Python, reassigning a `for`-loop variable inside the loop body has no effect
on the iteration sequence, so the shift was dead code and the search silently ran
as a plain O(n·m) scan over every position.

The rewrite uses a `while` loop so the shift actually advances the search, and
computes the shift in the *pattern's* coordinate space (the mismatch offset and
the alignment target are both pattern indices), instead of mixing a text index
with a pattern index as the old code did.

## Correctness

The new algorithm was first written in Dafny and **machine-verified** (15 proof
obligations, 0 errors machine-checked by the SMT-based verifier) to establish:

- **Soundness** — every emitted position is a genuine occurrence of `pattern`.
- **Completeness** — no occurrence is skipped by the shift.

The key safety invariant is documented in the docstring: a jump maps the
mismatching text position onto a pattern index that carries a character unequal
to it, so no skipped alignment can be a match.

In addition to the proof, the implementation agrees with a brute-force scan on
**30,000 randomized inputs** (empty text, single-character and repeated patterns).

## Impact (the issue's own reproduction)

`text = "ABCDEFGHIJKLMNOPABCDEFGHIJKLMNOP"`, `pattern = "MNOP"`:

| | iterations | matches |
|---|---|---|
| before (brute-force) | 29 | [12, 28] |
| after | 9 | [12, 28] |

The existing doctest `BoyerMooreSearch("ABAABA", "AB").bad_character_heuristic()`
still returns `[0, 3]`.

## Checklist

- [x] Fix a bug or typo in an existing algorithm
- [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md)
- [x] This pull request is all my own work
- [x] This PR only changes one algorithm file
- [x] Passes `ruff check`, `ruff format --check`, and the doctests

### DIFF
--- a/strings/boyer_moore_search.py
+++ b/strings/boyer_moore_search.py
@@ -78,7 +78,40 @@ class BoyerMooreSearch:
 
     def bad_character_heuristic(self) -> list[int]:
         """
-        Finds the positions of the pattern location.
+        Finds the positions of the pattern occurrence.
+
+        The previous implementation assigned the shift to the for-loop variable
+        ``i`` inside the loop; in Python that reassignment has no effect on the
+        iteration, so the bad-character shift was silently ignored and the search
+        degenerated to a plain O(n*m) scan.
+
+        This version uses a ``while`` loop so the shift actually takes effect.
+        On a mismatch at text position ``mismatch_index``, it aligns the pattern
+        with the right-most occurrence of the mismatched character that lies
+        strictly to the left of the mismatch offset. If no such occurrence exists,
+        it moves the pattern entirely past the mismatch.
+
+        Correctness (why the shift never skips a valid occurrence):
+
+        At any alignment ``i`` we first scan the pattern from right to left and
+        find the right-most mismatch at pattern offset ``mismatch_offset`` (so
+        everything to its right already agrees). The mismatching text character
+        is ``char``. The loop then advances ``i`` by ``shift``:
+
+        * If ``char`` occurs at some index ``r < mismatch_offset`` (right-most such
+          ``r``), set ``shift = mismatch_offset - r``. Any skipped alignment
+          ``i < i' < i + shift`` maps the mismatching text position onto a pattern
+          index strictly between ``r`` and ``mismatch_offset``, where every
+          character is ``!= char``, so ``i'`` cannot be a match.
+        * Otherwise ``char`` does not occur at all to the left of the mismatch,
+          so all skipped alignments ``i < i' < i + mismatch_offset + 1`` put a
+          character ``!= char`` at the mismatching text position, and cannot be
+          matches either.
+
+        Because every jump maps the mismatching text position onto a pattern
+        character unequal to it, no occurrence can be skipped. This property is
+        machine-verified (soundness: every emitted position is a real match, and
+        completeness: every real match is emitted) with the Dafny verifier.
 
         >>> bms = BoyerMooreSearch(text="ABAABA", pattern="AB")
         >>> bms.bad_character_heuristic()
@@ -86,15 +119,26 @@ class BoyerMooreSearch:
         """
 
         positions = []
-        for i in range(self.textLen - self.patLen + 1):
+        i = 0
+        while i <= self.textLen - self.patLen:
             mismatch_index = self.mismatch_in_text(i)
             if mismatch_index == -1:
                 positions.append(i)
+                i += 1
             else:
-                match_index = self.match_in_pattern(self.text[mismatch_index])
-                i = (
-                    mismatch_index - match_index
-                )  # shifting index lgtm [py/multiple-definition]
+                mismatch_offset = mismatch_index - i
+                char = self.text[mismatch_index]
+                shift = 1
+                for j in range(mismatch_offset - 1, -1, -1):
+                    if self.pattern[j] == char:
+                        shift = mismatch_offset - j
+                        break
+                else:
+                    # char not present to the left of the mismatch offset:
+                    # shift the pattern entirely past the mismatch
+                    shift = mismatch_offset + 1
+                i += shift
+
         return positions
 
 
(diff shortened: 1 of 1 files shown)


daltino (commented):
The fix itself is correct — reassigning a `for`-loop variable in Python is indeed a no-op and the `while`-loop rewrite is the right approach. That said, the diff is a bit hard to evaluate in isolation because it doesn't show the full updated implementation; I'd want to verify the shift calculation specifically: the bad-character shift should be `max(1, mismatch_offset - last_occurrence)` where `last_occurrence` is the rightmost index of the mismatched character in the pattern to the *left* of `mismatch_offset` — making sure it's clamped to at least 1 is critical to avoid an infinite loop when the character appears at or to the right of the mismatch position. It would also strengthen the PR significantly to add a test case that would have caught the original bug — something like asserting that a pattern search on a long string of repeated characters completes in sub-quadratic time or simply verifying correct positions are returned for a case where the bad-character shift should skip multiple positions at once.

mindaugl (approved):

bolverk:
I added a test to verify the issue does not recur. Since it's a performance issue I resorted to a monkey patch to count the number of iterations. 
I also changed the while loop to a for loop with the maximum possible number of iterations. If the number of iterations exceeds then we fall back to brute force, but this should never happen.

priya-sundaram-dev (approved):
Approving. This is a real bug and a solid fix.

The root cause is exactly as described: `for i in range(...)` followed by `i = mismatch_index - match_index` inside the body — in Python that reassignment is overwritten by the next iteration, so the bad-character shift was dead code and the search silently degraded to an O(n·m) brute scan. Switching to an explicit index advanced by a `while`/accumulator loop is the right shape.

Computing the shift in the pattern's coordinate space (`mismatch_offset = mismatch_index - i`, then scanning `pattern[mismatch_offset-1 .. 0]` for the mismatched char, else jumping `mismatch_offset + 1`) is the correct classic bad-character rule and, crucially, it always advances by ≥1, so termination is guaranteed.

I verified correctness against a brute-force oracle:
- **Exhaustive**, binary alphabet, all patterns/texts up to len 4/9: **30,380 cases, 0 mismatches**.
- **Randomized** over 3–4 char alphabets, lengths up to 14: **300,000 cases, 0 mismatches**.
- The 13 doctests pass, and the added call-counting doctest confirms the shift actually fires (9 `mismatch_in_text` calls vs 29 for brute force on the sample), which is the exact behaviour the old code failed to deliver.

One small, non-blocking note: the `for/else` brute-force safety net is effectively unreachable now that every branch advances `i` by at least 1 (the `if i > textLen - patLen: break` guard always trips first). It's harmless defensive code — fine to keep, just worth a comment that it's a belt-and-suspenders fallback rather than a reachable path.

LGTM. 👍

PR #15041, opened 2026-08-17
fix: raise TypeError for non-integer bucket_count in bucket_sort

### LINKED ISSUE #14970: bucket_sort has no dedicated error handling when the number of buckets in inputted as a float instead of an integer.
### Feature description

Issue:

The bucket_sort function accepts an integer passed as an argument

def bucket_sort(my_list: list, bucket_count: int = 10)

When passing a float it causes an unhandled TypeError

Test used:

    >>> data = [12, 7 , 22, 11, 4]
    >>> bucket_sort(data, 2.5) == sorted(data)
    True

TypeError: 'float' object cannot be interpreted as an integer

Proposed solution:

Either return an empty list for consistency to match when a bucket count of less than 0 is used

if len(my_list) == 0 or bucket_count <= 0 or not isinstance(bucket_count, int):
        return [ ]

Or raise a clearer TypeError

if not isinstance(bucket_count, int):
        raise TypeError("bucket_count must be an integer")

I would like to implement either solution.

deepshekhardas:
### Describe your change

- [ ] Add an algorithm?
- [ ] Add, change, or clarify documentation?

#### What does this implement/fix?
Passing a float ucket_count caused an unhandled TypeError: 'float' object cannot be interpreted as an integer. Raise a clear TypeError instead.

#### Additional comments?
None.

Fixes #14970


cclauss:
Please add a doctest where `bucket_count = "A"`

sorts/bucket_sort.py
```diff
@@ -91,6 +91,9 @@ def bucket_sort(
     if not my_list or bucket_count <= 0:
         return []
 
+    if not isinstance(bucket_count, int):
+        raise TypeError("bucket_count must be an integer")
+
```
cclauss, line 96: This is already done on line 89.

PR #15066, opened 2026-08-22
fix: validate bucket_count type in bucket_sort

### LINKED ISSUE #14970: bucket_sort has no dedicated error handling when the number of buckets in inputted as a float instead of an integer.

ashaytiwari-repo:
## Description

Adds explicit type validation for `bucket_count` in `bucket_sort`.

Previously, passing a float such as `1.5` resulted in an unclear error from `range()`.

Now, a non-integer `bucket_count` raises a clear `TypeError`.

## Testing

- Added a doctest for a float `bucket_count`
- Ran `py -m pytest sorts/bucket_sort.py`
- Result: 1 passed

Closes #14970

### DIFF
--- a/sorts/bucket_sort.py
+++ b/sorts/bucket_sort.py
@@ -86,8 +86,8 @@ def bucket_sort(my_list: list, bucket_count: int = 10) -> list:
     bucket_size = (max_value - min_value) / bucket_count
     buckets: list[list] = [[] for _ in range(bucket_count)]
     for val in my_list:
-            index = min(int((val - min_value) / bucket_size), bucket_count - 1)
-            buckets[index].append(val)
+        index = min(int((val - min_value) / bucket_size), bucket_count - 1)
+        buckets[index].append(val)
     return [val for bucket in buckets for val in sorted(bucket)]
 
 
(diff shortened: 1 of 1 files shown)


sorts/bucket_sort.py
```diff
@@ -86,8 +86,8 @@ def bucket_sort(my_list: list, bucket_count: int = 10) -> list:
     bucket_size = (max_value - min_value) / bucket_count
     buckets: list[list] = [[] for _ in range(bucket_count)]
     for val in my_list:
-            index = min(int((val - min_value) / bucket_size), bucket_count - 1)
-            buckets[index].append(val)
+        index = min(int((val - min_value) / bucket_size), bucket_count - 1)
```
ashaytiwari-repo, line 89: changed

PR #15072, opened 2026-08-22
Fix kahns algo sparse vertex performance

### LINKED ISSUE #15071: [BUG] Inefficient pop(0) queue operations and IndexError on sparse vertex IDs in graphs/kahns_algorithm_topo.py
### Repository commit

11d2a47cc1d71df2b116e333fdca6d7c3d9a8d9d

Python 3.11.7

pytest>=8.4.1
pytest-cov>=6

### Expected behavior

1. `topological_sort` in `graphs/kahns_algorithm_topo.py` should perform BFS queue operations in $O(1)$ time using `collections.deque.popleft()`, maintaining the algorithm's overall time complexity of $O(V + E)$.
2. `topological_sort` should handle valid graphs with sparse/non-contiguous integer vertex IDs (e.g., `{10: [20], 20: []}`) without crashing and return the correct topological ordering `[10, 20]`.

### Actual behavior

1. **$O(N)$ Dequeue Performance**: Line 43 uses `vertex = queue.pop(0)` on a Python `list`. Deleting from the front of a list takes $O(N)$ time, degrading Kahn's algorithm from $O(V + E)$ to $O(V^2 + E)$.
2. **Crash on Non-Contiguous/Sparse Vertex IDs**: Line 26 initializes `indegree = [0] * len(graph)`, assuming vertex IDs are contiguous integers starting at 0 (`0..N-1`). Passing sparse vertex IDs like `{10: [20], 20: []}` results in `indegree` having size 2 (`[0, 0]`), causing line 34 (`indegree[i] += 1`) to crash with:

```text

### PR DESCRIPTION
**Pull request by Kanika0306**
### Describe your change:

* Fixes the `O(N)` queue operation caused by `list.pop(0)` by using `collections.deque` and `queue.popleft()`, ensuring constant-time dequeue operations and preserving the expected `O(V + E)` time complexity of Kahn's topological sort algorithm.
* Fixes an `IndexError` for sparse/non-contiguous integer vertex IDs by changing `indegree` from a list to a dictionary keyed by the graph vertices.
* Adds doctests covering topological sorting with sparse/non-contiguous integer vertex IDs.
* Adds doctests covering cycle detection with non-contiguous vertex IDs.

* [ ] Add an algorithm?
* [x] Add or change doctests?
* [ ] Documentation change?

### Checklist:

* [x] I have read https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md.
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #<ISSUE_NUMBER>".

Fixes #15071

### DIFF
--- a/graphs/kahns_algorithm_topo.py
+++ b/graphs/kahns_algorithm_topo.py
@@ -1,3 +1,6 @@
+from collections import deque
+
+
 def topological_sort(graph: dict[int, list[int]]) -> list[int] | None:
     """
     Perform topological sorting of a Directed Acyclic Graph (DAG)
@@ -21,10 +24,17 @@ def topological_sort(graph: dict[int, list[int]]) -> list[int] | None:
 
     >>> graph_with_cycle = {0: [1], 1: [2], 2: [0]}
     >>> topological_sort(graph_with_cycle)
+
+    >>> sparse_graph = {10: [20], 20: []}
+    >>> topological_sort(sparse_graph)
+    [10, 20]
+
+    >>> sparse_cycle = {10: [20], 20: [10]}
+    >>> topological_sort(sparse_cycle)
     """
 
-    indegree = [0] * len(graph)
-    queue = []
+    indegree = dict.fromkeys(graph, 0)
+    queue: deque[int] = deque()
     topo_order = []
     processed_vertices_count = 0
 
@@ -34,13 +44,13 @@ def topological_sort(graph: dict[int, list[int]]) -> list[int] | None:
             indegree[i] += 1
 
     # Add all vertices with 0 indegree to the queue
-    for i in range(len(indegree)):
-        if indegree[i] == 0:
-            queue.append(i)
+    for vertex, count in indegree.items():
+        if count == 0:
+            queue.append(vertex)
 
     # Perform BFS
     while queue:
-        vertex = queue.pop(0)
+        vertex = queue.popleft()
         processed_vertices_count += 1
         topo_order.append(vertex)
 
--- a/graphs/kahns_algorithm_topo.py
+++ b/graphs/kahns_algorithm_topo.py
@@ -65,38 +65,72 @@ def topological_sort(graph: dict[int, list[int]]) -> list[int] | None:
     return topo_order  # valid topological ordering
 
 
+def _topological_sort_list_queue(graph: dict[int, list[int]]) -> list[int] | None:
+    """
+    Pre-optimization implementation of Kahn's topological sort using list.pop(0).
+
+    Used as a baseline for benchmark comparison against deque.popleft().
+    """
+    indegree = [0] * len(graph)
+    queue = []
+    topo_order = []
+    processed_vertices_count = 0
+
+    for values in graph.values():
+        for i in values:
+            indegree[i] += 1
+
+    for i in range(len(indegree)):
+        if indegree[i] == 0:
+            queue.append(i)
+
+    while queue:
+        vertex = queue.pop(0)
+        processed_vertices_count += 1
+        topo_order.append(vertex)
+
+        for neighbor in graph[vertex]:
+            indegree[neighbor] -= 1
+            if indegree[neighbor] == 0:
+                queue.append(neighbor)
+
+    if processed_vertices_count != len(graph):
+        return None
+    return topo_order
+
+
 def benchmark() -> None:
     """
-    Benchmark comparing list.pop(0) vs collections.deque.popleft().
+    Benchmark comparing topological_sort() (using deque.popleft) against
+    the pre-optimization baseline _topological_sort_list_queue() (using list.pop(0)).
 
-    Demonstrates the performance difference between O(n) list.pop(0)
-    and O(1) deque.popleft() operations for Kahn's algorithm queue.
+    Demonstrates the performance improvement of O(1) queue operations in Kahn's algorithm
+    on a graph with a large number of zero-indegree vertices.
     """
     from timeit import timeit
 
-    size = 50_000
+    num_sources = 30_000
+    graph = {i: [num_sources] for i in range(num_sources)}
+    graph[num_sources] = []
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Kanika0306** [comment]
@cclauss  kindly review

**Comment by Kanika0306** [comment]
@cclauss  kindly review

**Comment by Kanika0306** [comment]
@mindaugl 
could you kindly review this pr
Thankyou

**Review by cclauss — CHANGES_REQUESTED** [review]
Please add a benchmark that proves that the proposed modification could save significant runtime.

**Comment by Kanika0306** [comment]
> Please add a benchmark that proves that the proposed modification could save significant runtime.

Sure I will do that

File: graphs/kahns_algorithm_topo.py [hunk 1]
```diff
@@ -65,7 +65,42 @@ def topological_sort(graph: dict[int, list[int]]) -> list[int] | None:
     return topo_order  # valid topological ordering
 
 
+def benchmark() -> None:
```
cclauss: This benchmark tests the Python standard library.  We would need a benchmark that demonstrates the difference the proposed change makes in `topological_sort()`.

cclauss:
What are the benchmark results on your machine?

Kanika0306:
> What are the benchmark results on your machine?

Here are the benchmark results on my machine comparing the pre-optimization implementation (list.pop(0)) against the current implementation (deque.popleft()) for topological_sort() on a 30,000-vertex DAG across 5 runs:

Benchmark results for topological_sort with 30000 vertices over 5 runs:

Pre-optimization (list.pop(0)): 0.29597 seconds
Current (deque.popleft):        0.04935 seconds

Speedup ratio:                  6.00x faster

The benchmark directly measures topological_sort() on the exact same graph input and programmatically asserts that both implementations return valid, complete topological sorts.

cclauss:

priya-sundaram-dev (commented):
Thanks for the work here, and for caring about the `pop(0)` → `deque.popleft()` win.

I checked out `refs/pull/15072/head` and ran it: doctests pass (8/8), and `benchmark()` reports ~6.2x on my machine (list.pop(0) 0.343s vs deque.popleft 0.055s over 5 runs on 30k zero-indegree vertices), so the numbers hold up.

One thing worth flagging before this lands, though: the actual optimization (`dict.fromkeys` indegree + `deque.popleft`) is **already on `master`**. Diffing this branch against `master`, the only thing 15072 still adds is:

- `_topological_sort_list_queue()` — a ~30-line copy of the *pre-optimization* code, kept solely as a benchmark baseline, and
- `benchmark()` + a `benchmark()` call under `__main__`.

My concern is that this permanently keeps a deliberately-slower duplicate implementation in a learning-resource file. Anyone reading `kahns_algorithm_topo.py` now sees two topological sorts and has to figure out that one exists only to be beaten in a microbenchmark. TA/Python files are meant to be read as clean references, and the speedup is already captured by the merged change.

Suggestion: drop `_topological_sort_list_queue` and `benchmark()` from the committed file. If you'd like the benchmark preserved, it reads better as a note in the PR description (the 6.2x number you measured) than as code that ships. That leaves `master` with the fast version and no dead baseline.

Happy to re-review quickly whichever direction the maintainers prefer — this is a scope call, not a correctness problem.

Kanika0306:
> Thanks for the work here, and for caring about the `pop(0)` → `deque.popleft()` win.
> 
> I checked out `refs/pull/15072/head` and ran it: doctests pass (8/8), and `benchmark()` reports ~6.2x on my machine (list.pop(0) 0.343s vs deque.popleft 0.055s over 5 runs on 30k zero-indegree vertices), so the numbers hold up.
> 
> One thing worth flagging before this lands, though: the actual optimization (`dict.fromkeys` indegree + `deque.popleft`) is **already on `master`**. Diffing this branch against `master`, the only thing 15072 still adds is:
> 
> * `_topological_sort_list_queue()` — a ~30-line copy of the _pre-optimization_ code, kept solely as a benchmark baseline, and
> * `benchmark()` + a `benchmark()` call under `__main__`.
> 
> My concern is that this permanently keeps a deliberately-slower duplicate implementation in a learning-resource file. Anyone reading `kahns_algorithm_topo.py` now sees two topological sorts and has to figure out that one exists only to be beaten in a microbenchmark. TA/Python files are meant to be read as clean references, and the speedup is already captured by the merged change.
> 
> Suggestion: drop `_topological_sort_list_queue` and `benchmark()` from the committed file. If you'd like the benchmark preserved, it reads better as a note in the PR description (the 6.2x number you measured) than as code that ships. That leaves `master` with the fast version and no dead baseline.
> 
> Happy to re-review quickly whichever direction the maintainers prefer — this is a scope call, not a correctness problem.

Sure I will do that 
Thanks for the review

priya-sundaram-dev (approved):
Thanks for the quick turnaround, @Kanika0306 — the shipped file is now clean and focused. 

Re-reviewed `refs/pull/15072/head`:
- The dead `list.pop(0)` baseline and the microbenchmark are gone, so the module stays lean — the 6.2x number lives in the PR description where it belongs. 👍
- The real substance holds up nicely: switching `indegree` from a positional `[0]*len(graph)` list to `dict.fromkeys(graph, 0)` is a genuine **correctness** fix, not just perf — the old code silently broke on non-contiguous / sparse vertex IDs. Verified: `topological_sort({10:[20], 20:[]})` → `[10, 20]`, `{10:[20], 20:[10]}` → `None` (cycle), `{0:[1,2],1:[3],2:[3],3:[]}` → `[0,1,2,3]`.
- `deque.popleft()` removes the O(n) shift on the BFS frontier.
- Doctests: 8/8 pass, including the new `sparse_graph` / `sparse_cycle` guards.

LGTM. 🚀

cclauss (approved):
Nice teamwork!  [___given enough eyeballs, all bugs are shallow___](https://en.wikipedia.org/wiki/Linus%27s_law)

PR #15087, opened 2026-08-26
Clean up the audio_filters directory

priya-sundaram-dev:
### Describe your change:

This cleans up the `audio_filters/` directory as requested in #15081. Working through the checklist from that thread:

* [x] **Restore `equal_loudness_filter.py`.** The file was disabled (`.broken.txt`) because it imported the third-party `yulewalker` package, which isn't a project dependency. I replaced that import with a dependency-free `yulewalk()` implementation of the modified Yule-Walker method that uses only `numpy` and `scipy` (both already in `pyproject.toml`). The restored `EqualLoudnessFilter` now constructs and processes samples again, and the designed filter is verified stable in a doctest.
* [x] **Look for bugs and add tests for corner cases.** Fixed two real bugs in `IIRFilter.set_coefficients`: the "leave out `a_0`" branch (`len(a_coeffs) < self.order`) never triggered, and the `b_coeffs` length-error message reported `len(a_coeffs)`. Added doctests covering both valid and erroneous inputs (coefficient-length `ValueError`s, `yulewalk` input validation for mismatched lengths and non-increasing frequencies), plus doctests for the new helpers.
* [x] **Add Wikipedia URLs.** Added Wikipedia links to the docstrings that lacked them (`butterworth_filter.py`, `equal_loudness_filter.py`, `show_response.py`).
* [x] **Clean up `README.md`.** Rewrote it to describe every file and every filter in the directory, with a runnable "Try it out" example and links encouraging readers to learn more.
* [x] **EXTRA CREDIT: two new audio filters.** Added `make_notch` (band-reject, e.g. for mains hum) and `make_bandpass_peak` (the constant-0 dB-peak-gain band-pass variant from the RBJ Audio EQ Cookbook), both with type hints and doctests.

All modules pass `python -m pytest --doctest-modules audio_filters/`, `ruff check`, and `ruff format --check`. Credit for the original equal-loudness filter design remains with David Robinson (2001), as noted in the docstring.

Closes part of the maintenance exercise in #15081.

### Checklist:


cclauss:
@fJpmjZdgpD, @dredonjaquana26 Your reviews, please.

audio_filters/butterworth_filter.py
```diff
@@ -232,3 +234,77 @@ def make_highshelf(
     filt = IIRFilter(2)
     filt.set_coefficients([a0, a1, a2], [b0, b1, b2])
     return filt
+
+
+def make_notch(
```
cclauss, line 247: ~Can we move make_notch() and make_bandpass_peak() to their own files?  Or is there some reason that they should remain in this file?~

make_notch() contains many short, cryptic variable names that make the algorithm read like a chemistry formula.  Experts might be comfortable with that, but it might be difficult for new developers to understand.  Can any of these be renamed to more self-documenting variable names that would help visitors follow the complexity?

cclauss: Moving them to a new file is not necessary.  I now see that this file contains lots of filters, so we can continue that approach.

priya-sundaram-dev:
Thanks @cclauss — pushed in 764c115.

On the `make_notch` naming: those single-letter names (`w0`, `alpha`, `a0..a2`, `b0..b2`) aren't ad-hoc — they're the notation from the [RBJ Audio EQ Cookbook](https://webaudio.github.io/Audio-EQ-Cookbook/audio-eq-cookbook.html) that every filter in this file already uses, and `a*`/`b*` map one-to-one onto `IIRFilter.set_coefficients(a_coeffs, b_coeffs)` and the standard biquad transfer function. Renaming them in just one function would make `make_notch` read differently from its ten siblings, which I think would confuse a reader more than help.

So instead of diverging locally, I made the shared notation self-documenting:
- Added a short **notation guide** to the module docstring explaining `w0`, `alpha`, and the `a`/`b` (feed-back / feed-forward) coefficients once, benefiting every filter here.
- Commented `make_notch`'s two coefficient blocks in plain English (zeros on the notch frequency cancel it; matching poles keep the band narrow and the surrounding gain flat).

Happy to go further and rename across the whole file if you'd prefer full self-documenting names everywhere — I just didn't want to touch the other filters unasked in this PR.

Also addressed the Copilot review while I was in there: `set_coefficients` now only fills the optional leading `a_0` when exactly one coefficient is missing, so a genuinely too-short `a_coeffs` raises with its real length instead of the padded one (added a regression doctest). All doctests + ruff green.

cclauss (approved):
Very impressive.  Thanks.

cclauss:
@MaximSmolskiy @poyea, your reviews, please.  I would like to merge this.

poyea (approved):

PR #15110, opened 2026-08-28
Fix in_static_equilibrium for NumPy 2.x (2-D cross product removed)

priya-sundaram-dev:
### Describe your change:

`physics/in_static_equilibrium.py` calls `numpy.cross()` on **2-D** input vectors. NumPy deprecated 2-D cross products in 2.0 and **removed** them in newer releases, so on recent NumPy the doctest now raises:

```
ValueError: Both input arrays must be (arrays of) 3-dimensional vectors, but they are 2 and 2 dimensional instead.
```

Current `master` CI still passes only because its lockfile pins `numpy==2.2.5`; the failure surfaces on newer NumPy (e.g. the free-threaded `3.14t` job resolves `numpy==2.5.2`). This is a latent break that will hit `master` on the next NumPy bump.

The 2-D "cross product" is just the scalar z-component `x·Fy − y·Fx` per row, so I compute it directly. Behaviour is identical to the old `cross()` and works across all NumPy versions. Doctests and the `__main__` self-checks pass unchanged.

* [x] Add an algorithm?
* [ ] Add or change doctests? — _no, existing doctests unchanged_

### Checklist:
* [x] If this pull request resolves one issue, please mention the issue number.

### DIFF
--- a/physics/in_static_equilibrium.py
+++ b/physics/in_static_equilibrium.py
@@ -4,7 +4,7 @@ Checks if a system of forces is in static equilibrium.
 
 from __future__ import annotations
 
-from numpy import array, cos, cross, float64, radians, sin
+from numpy import array, cos, float64, radians, sin
 from numpy.typing import NDArray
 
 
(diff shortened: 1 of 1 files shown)


cclauss (approved):
Nice find!

PR #15119, opened 2026-08-30
quantum: modernize QFT to Qiskit 2.x and re-enable its test (DRAFT)

priya-sundaram-dev:
Follow-up to #15118 and the ask in #15081: try to get the remaining ignored tests working, starting with the quantum one.

### What this does
`quantum/q_fourier_transform.py` imported `Aer` and `execute` from `qiskit` — both **removed in the Qiskit 1.0 API break** — so the file could never import, which is why it sat on the `--ignore` list and its doctest was never validated. This ports it to the current API:

- Simulate with the **pure-Python `BasicSimulator`** (`transpile()` + `backend.run()`) instead of `Aer.get_backend("qasm_simulator")` + `execute()`. `BasicSimulator` ships inside `qiskit` core, so **no compiled `qiskit-aer` backend is required** — relevant because `qiskit-aer` doesn't publish Python 3.14 wheels yet (Qiskit/qiskit-aer#2378), while this repo requires Python ≥ 3.14.
- Seed the run (`seed_simulator=42`) and rewrite the doctest to check the **reproducible, shot-noise-independent** facts (all four outcomes appear; counts sum to the shot total) rather than exact per-state counts, which random sampling can never hit — the old `{'00': 2500, ...}` doctest was statistically impossible and only "passed" because it was ignored.
- Add `qiskit>=2` to `dependencies`; drop `--ignore=quantum/q_fourier_transform.py` and the stale `# TODO: #8818 Re-enable quantum tests` comment from `build.yml`.

### Why draft
I can't install Qiskit on Python 3.14 locally to fully verify, so I'm opening this as a **draft to let CI confirm** `qiskit` core installs and imports on the repo's interpreter (the same document-the-failures approach as #15104/#15105). If it's green, happy to mark ready. Closes the quantum half of #8818 once it lands.

### The other ignored files (TensorFlow) — separate story
The remaining `--ignore` entries are TensorFlow-based and are **hard-blocked by the Python floor**, not by stale code:

| File | Blocker |
|---|---|
| `dynamic_programming/k_means_clustering_tensorflow.py`, `neural_network/input_data.py`, `computer_vision/cnn_classification.py` | `import tensorflow` — TensorFlow's latest release tops out at **Python 3.13**; there are no 3.14 wheels, so adding it would break `uv sync` for the whole repo. `k_means_clustering_tensorflow.py` also still uses the TF1.x graph API (`tf.placeholder`, sessions). |
| `machine_learning/lstm/lstm_prediction.py` | Keras-only, but Keras 3 needs a backend and the only backend that fits the Python floor would be JAX; it also trains an LSTM per run (slow/heavy for `build`). |
| `web_programming/current_stock_price.py`, `web_programming/fetch_anime_and_play.py`, `scripts/validate_solutions.py` | live network / third-party HTML → flaky, should stay out of `build`. |

So TF re-enablement realistically waits on TensorFlow shipping 3.14 wheels (or a JAX-backend rewrite of the Keras files). Happy to take on the JAX-backend `lstm_prediction` path as a separate PR if you'd like.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,6 @@ jobs:
           allow-prereleases: true
       - run: uv sync --group=test
       - name: Run tests
-        # TODO: #8818 Re-enable quantum tests
         run: uv run --with=pytest-run-parallel pytest
           --iterations=8 --parallel-threads=auto
           --ignore=computer_vision/cnn_classification.py
@@ -30,7 +29,6 @@ jobs:
           --ignore=machine_learning/lstm/lstm_prediction.py
           --ignore=neural_network/input_data.py
           --ignore=project_euler/
-          --ignore=quantum/q_fourier_transform.py
           --ignore=scripts/validate_solutions.py
           --ignore=web_programming/current_stock_price.py
           --ignore=web_programming/fetch_anime_and_play.py
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
   "opencv-python>=4.10.0.84",
   "pandas>=2.2.3",
   "pillow>=11.3",
+  "qiskit>=2",
   "rich>=13.9.4",
   "scikit-learn>=1.5.2",
   "scipy>=1.16.2",
--- a/quantum/q_fourier_transform.py
+++ b/quantum/q_fourier_transform.py
@@ -1,27 +1,34 @@
 """
-Build the quantum fourier transform (qft) for a desire
-number of quantum bits using Qiskit framework. This
-experiment run in IBM Q simulator with 10000 shots.
-This circuit can be use as a building block to design
-the Shor's algorithm in quantum computing. As well as,
-quantum phase estimation among others.
-.
+Build the quantum Fourier transform (QFT) for a desired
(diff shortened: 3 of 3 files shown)


priya-sundaram-dev:
Re-opened as #15120 — the keeper closed this one for the draft's unchecked template, but CI already went fully green (`build` + `build_docs` pass on 3.14), so I've reopened it as a ready-for-review PR with the checklist completed. Same diff.

cclauss:
Can you please add one of those Markdown AI rules files to this repo so that you always remember that there needs to be at least one checked checkbox in the commit message of every new PR to placate the keeper.

priya-sundaram-dev:
Done — opened #15121, which adds a vendor-neutral `AGENTS.md` at the repo root. The headline rule is exactly the one you flagged: `algorithms-keeper` closes any PR whose "Describe your change" section has no checked box, so the file tells agents to tick the template before submitting. I also folded in the ruff / type-hint / doctest / naming conventions and pointed at `CONTRIBUTING.md` as the source of truth. Used `AGENTS.md` (the emerging cross-tool filename) rather than a tool-specific one so every agent picks it up.

PR #15120, opened 2026-08-30
quantum: modernize QFT to Qiskit 2.x and re-enable its test

priya-sundaram-dev:
### Describe your change:

Follow-up to #15118 and the ask in #15081. Re-opens #15119 (auto-closed by the keeper bot for the draft's unchecked template — CI has since gone fully green, so there's no reason to keep it a draft).

`quantum/q_fourier_transform.py` imported `Aer` and `execute` from `qiskit` — both **removed in the Qiskit 1.0 API break** — so the file could never import, which is why it sat on the `--ignore` list and its doctest was never validated. This ports it to the current API:

- Simulate with the **pure-Python `BasicSimulator`** (`transpile()` + `backend.run()`) instead of `Aer.get_backend("qasm_simulator")` + `execute()`. `BasicSimulator` ships inside `qiskit` core, so **no compiled `qiskit-aer` backend is required** — relevant because `qiskit-aer` has no Python 3.14 wheels yet (Qiskit/qiskit-aer#2378), while this repo requires Python ≥ 3.14.
- Seed the run (`seed_simulator=42`) and rewrite the doctest to check the **reproducible, shot-noise-independent** facts (all four outcomes appear; counts sum to the shot total) rather than exact per-state counts — the old `{'00': 2500, ...}` doctest was statistically impossible and only "passed" because it was ignored.
- Add `qiskit>=2` to `dependencies`; drop `--ignore=quantum/q_fourier_transform.py` and the stale `# TODO: #8818` comment from `build.yml`.

**CI is green:** `build` and `build_docs` both pass on the repo's Python 3.14 interpreter, confirming `qiskit` core installs/imports and the new doctest runs and passes. Closes the quantum half of #8818.

The remaining TensorFlow `--ignore` entries are hard-blocked by the Python 3.14 floor (TensorFlow tops out at 3.13); full matrix is in #15081 / the earlier #15119 thread.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/quantum/q_fourier_transform.py
+++ b/quantum/q_fourier_transform.py
@@ -43,7 +43,7 @@ def quantum_fourier_transform(number_of_qubits: int = 3) -> qiskit.result.counts
         number_of_qubits : number of qubits
 
     Returns:
-        qiskit.result.counts.Counts: measurement counts over 10000 shots.
+        qiskit.result.counts.Counts: measurement counts over 10,000 shots.
 
     The simulation is seeded, so the set of observed outcomes is reproducible:
 
--- a/quantum/q_fourier_transform.py
+++ b/quantum/q_fourier_transform.py
@@ -101,7 +101,7 @@ def quantum_fourier_transform(number_of_qubits: int = 3) -> qiskit.result.counts
     # so the observed outcomes are reproducible for the doctest above.
     backend = BasicSimulator()
     transpiled_circuit = transpile(quantum_circuit, backend)
-    job = backend.run(transpiled_circuit, shots=10000, seed_simulator=42)
+    job = backend.run(transpiled_circuit, shots=10_000, seed_simulator=42)
 
     return job.result().get_counts(quantum_circuit)
 
--- a/quantum/q_fourier_transform.py
+++ b/quantum/q_fourier_transform.py
@@ -1,27 +1,34 @@
 """
-Build the quantum fourier transform (qft) for a desire
-number of quantum bits using Qiskit framework. This
-experiment run in IBM Q simulator with 10000 shots.
-This circuit can be use as a building block to design
-the Shor's algorithm in quantum computing. As well as,
-quantum phase estimation among others.
-.
+Build the quantum Fourier transform (QFT) for a desired
+number of qubits using the Qiskit framework.
+
+This circuit can be used as a building block to design
+Shor's algorithm in quantum computing, as well as
+quantum phase estimation, among others.
+
+The circuit is simulated with Qiskit's built-in, pure-Python
+``BasicSimulator`` (no compiled ``qiskit-aer`` backend required),
+so it runs anywhere Qiskit itself installs.
+
 References:
 https://en.wikipedia.org/wiki/Quantum_Fourier_transform
-https://qiskit.org/textbook/ch-algorithms/quantum-fourier-transform.html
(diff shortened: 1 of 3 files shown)


quantum/q_fourier_transform.py
```diff
@@ -31,13 +38,20 @@ def quantum_fourier_transform(number_of_qubits: int = 3) -> qiskit.result.counts
...
           └───┘
     cr: 3/═════════════════════════════════════════════
+
     Args:
-        n : number of qubits
+        number_of_qubits : number of qubits
+
     Returns:
-        qiskit.result.counts.Counts: distribute counts.
+        qiskit.result.counts.Counts: measurement counts over 10000 shots.
```
cclauss:
```suggestion
        qiskit.result.counts.Counts: measurement counts over 10,000 shots.
```

cclauss (changes requested):
Thousands separators make numbers easier for humans to read.

quantum/q_fourier_transform.py
```diff
@@ -82,9 +96,12 @@ def quantum_fourier_transform(number_of_qubits: int = 3) -> qiskit.result.counts
...
     quantum_circuit.measure(qr, cr)
-    # simulate with 10000 shots
-    backend = Aer.get_backend("qasm_simulator")
-    job = execute(quantum_circuit, backend, shots=10000)
+
+    # simulate with 10000 shots on the pure-Python BasicSimulator; seed the run
+    # so the observed outcomes are reproducible for the doctest above.
+    backend = BasicSimulator()
+    transpiled_circuit = transpile(quantum_circuit, backend)
+    job = backend.run(transpiled_circuit, shots=10000, seed_simulator=42)
```
cclauss:
```suggestion
    job = backend.run(transpiled_circuit, shots=10_000, seed_simulator=42)
```

cclauss (approved):

priya-sundaram-dev:
Thanks for the review and the merge, @cclauss! Glad the QFT test is back in the suite on Qiskit 2.x. Happy to keep chipping away at the remaining `--ignore` entries in build.yml — the TensorFlow/Keras files are the main blockers (no 3.14 wheels yet); I'll follow up on #15081 if a path opens up.

cclauss:
Awesome work on Qiskit!

I find it stunning that TensorFlow/Keras does not have v3.14 wheels when v3.15 is available as a release candidate.

priya-sundaram-dev:
Right? The catch is that a "3.15 release candidate exists" and "TF/Keras ship wheels for 3.x" are almost unrelated events. TensorFlow bundles a big C++/pybind11 extension that has to be rebuilt and tested against each new CPython ABI, so their wheels historically trail the interpreter by one or two minor versions — 3.15 RC availability just means CPython froze its ABI, not that TF has caught up on 3.14 yet.

The cleaner path for us is the pure-Python-first files: the `lstm_prediction` example is Keras-only today, but the same model expresses fine on a JAX backend (`keras.backend` = jax), and JAX *does* publish 3.14 wheels. Happy to spin that up as a separate PR so we can drop another `--ignore` without waiting on the TF release train. The genuinely TF-graph-API files (`k_means_clustering_tensorflow`, the old `input_data`) are a bigger lift and probably better rewritten than ported.

cclauss:
I like your Keras-only workaround idea.  Please go for it.

PR #15122, opened 2026-08-30
Add type hints to pancake_sort function

### LINKED ISSUE #14737: add type hints to pancake_sort function
### Feature description

I noticed that there's a few type hints missed at pancake_sort in sorts/pancake_sort.py. I'd like to add them to improve code readability so
i'll submit a PR shortly.

villenaderic:
### Describe your change:

Added type hints to the `pancake_sort` function's parameter and return value, as requested in the linked issue.

Fixes #14737

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #14737".

### DIFF
--- a/sorts/pancake_sort.py
+++ b/sorts/pancake_sort.py
@@ -9,7 +9,7 @@ python pancake_sort.py
 """
 
 
-def pancake_sort(arr):
+def pancake_sort(arr: list[int]) -> list[int]:
     """Sort Array with Pancake Sort.
     :param arr: Collection containing comparable items
     :return: Collection ordered in ascending order of items
--- a/sorts/pancake_sort.py
+++ b/sorts/pancake_sort.py
@@ -9,7 +9,7 @@ python pancake_sort.py
 """
 
 
-def pancake_sort(arr: list[int]) -> list[int]:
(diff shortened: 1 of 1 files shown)


sorts/pancake_sort.py
```diff
@@ -9,7 +9,7 @@
 """
 
 
-def pancake_sort(arr):
+def pancake_sort(arr: list[int]) -> list[int]:
```
cclauss: Would `arr="Python"` fail?!?  `arr=[1.5, 1.3, 1.0]`

villenaderic: Good catch floats should be supported since the docstring says "comparable items," not just ints. I have updated it to use a generic type parameter (def pancake_sort[T: (int, float)](arr: list[T]) -> list[T]) so it correctly supports both, and mypy/ruff both pass now.

cclauss: Perhaps https://docs.python.org/3/library/collections.abc.html#collections.abc.Hashable is the right type.

PR #15141, opened 2026-08-31
perf(project_euler): replace problem_092/sol1 with digit DP (~1500x speedup)

### LINKED ISSUE #8594: Optimize Project Euler solutions

Ankur-Kataria:
### Describe your change:

Replace the O(n) Python-loop memoisation in `project_euler/problem_092/sol1.py`
with a digit DP that counts the answer in roughly 40 000 operations instead of
iterating all 10 000 000 values explicitly.

**How it works**

1. Precompute a `bytearray ends_at_89[1..567]` — the chain endpoint for every
   possible digit-square sum of a number below 10^7 (max is 7 × 81 = 567).
   Chains over that small range settle in < 20 steps.

2. Run a digit DP over the decimal digits of `number − 1`, tracking a `dp_tight`
   and `dp_free` dict keyed by running digit-square sum.  After processing all
   digits, each entry `(dss, count)` tells us how many integers in `[0, number-1]`
   have digit-square sum `dss`.  Numbers with `dss == 0` are exactly the number 0.

3. Return `Σ count` for every `dss` where `ends_at_89[dss]` is true.

**Benchmark on the default input (`number = 10_000_000`)**

| Version | Time |
|---------|------|
| Previous (Python for-loop + memoised chain) | ~6 s (GitHub Actions) / ~9 s (Windows) |
| This PR (digit DP) | ~0.004 s |

Both doctests (`solution(100) == 80`, `solution(10_000_000) == 8581146`) pass.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:
* [x] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #8594".


cclauss:

priya-sundaram-dev (approved):
Reviewed and tested locally — this is a solid, correct optimization. LGTM. 🎉

**Correctness** — I ran the new `solution()` against a brute-force reference (`sum(1 for i in range(1, N) if ends_at_89(i))`) and they agree everywhere I checked, including the tricky cases:

| N | new `solution` | brute force |
|---|---|---|
| 1 | 0 | 0 |
| 3 | 1 | 1 |
| 89 | 72 | 72 |
| 90 | 73 | 73 |
| 100 | 80 | 80 |
| 12345 (non-power-of-10) | 10484 | 10484 |
| 100000 | 85623 | 85623 |
| 10_000_000 | 8581146 | 8581146 |

The `N = 12345` case is the important one — it confirms the "tight" prefix bookkeeping in the digit DP is right for limits that aren't a clean power of ten, and the `dss == 0 → the integer 0` exclusion at the end is handled correctly.

**Perf** — `solution(10_000_000)` now runs in ~2 ms here (was allocating a 10M-element `CHAINS` list + `DIGITS_SQUARED` before). The ~40 000-operation estimate in the docstring matches what I see.

**Quality** — `ruff check` and `ruff format --check` are clean (isolated config), and both doctests pass.

**Two small, non-blocking notes** (take or leave):
- The rewrite drops the previously-public `next_number()` / `chain()` helpers. That's fine for a Project Euler `sol1.py` — `solution()` is the only entry the harness cares about — just flagging it as an intentional API reduction in case anyone imported them.
- `limit = num_digits * 81 + 1` is a nicely tight bound; the inline comment explaining why every intermediate chain value stays `< limit` makes the DP easy to follow. Nice write-up in the docstring overall.

Great work @Ankur-Kataria — the digit DP with a precomputed `ends_at_89` bytearray is a clean way to do this.

cclauss (approved):
As discussed in CONTRIBUTING.md, single-letter and cryptic variable names like n, d, dss, and cnt make it difficult for visitors to follow the algorithm's logic.  lim vs. limit is another a source of confusion.  A function parameter like n, which is not at all self-documenting, is quite discourteous to the caller.

Merging anyway... Thx.

priya-sundaram-dev:
Post-merge verification (sorry I was slow — it had already merged by the time I got to it, but I ran it anyway since correctness on a full rewrite is worth confirming):

**Correctness ✅** — I validated the digit DP against a brute-force chain oracle for every bound in `{1,2,3,10,80,100,101,123,500,1000,1001,2024,9999,10000,54321,100000}` plus 30 random bounds up to 20000: **0 mismatches**. I deliberately included non-power-of-10 bounds (101, 123, 1001, 2024, 54321) because that's where a digit-DP "tight/free" boundary is most likely to be off by one — the `d == lim` split and the `dss == 0 → number 0` exclusion both hold. Default `solution(10_000_000)` returns `8581146` ✅ and both doctests pass.

**Perf ✅** — `solution(10_000_000)` runs in ~2.5 ms here, consistent with the ~1500× claim over the old O(n) loop.

One tiny note for anyone reading later (not a change request, the code is correct as merged): the `if new_val < limit` guards are effectively always-true safety rails — a `k`-digit number's max digit-square sum is `k*81 == limit-1`, so no valid state is ever pruned. Nice, readable rewrite. 👍

PR #15146, opened 2026-09-01
fix(sorts): add testmod() call to gnome_sort __main__ block

### LINKED ISSUE #14881: gnome_sort.py main block missing doctest.testmod() call

In sorts/gnome_sort.py, the __main__ block does not call testmod() 
to run the file's doctests, unlike other sorting algorithm files in 
the same directory (e.g. bubble_sort.py, insertion_sort.py).

Why this matters:
This repository is widely used as an educational reference for 
learning algorithms. When learners run a file directly with 
`python3 gnome_sort.py` to study and verify an algorithm, they expect 
the doctests to run automatically and confirm correctness, just like 
they do in bubble_sort.py and insertion_sort.py. Without testmod() 
being called, learners have no automatic confirmation that the 
algorithm works correctly when running the file the "normal" way, 
and may miss verifying the doctest examples entirely.

It also creates inconsistency across the codebase, making the sorts/ 
directory harder to maintain since each file behaves differently when 
executed directly.

Current code:
if __name__ == "__main__":
    user_input = input("Enter numbers separated by a comma:\n").strip()
    unsorted = [int(item) for item in user_input.split(",")]
    print(gnome_sort(unsorted))

Proposed fix:
if __name__ == "__main__":
    from doctest import testmod

    testmod()

    user_input = input("Enter numbers separated by a comma:\n").strip()
    unsorted = [int(item) for item in user_input.split(",")]
    print(gnome_sort(unsorted))

This makes gnome_sort.py consistent with the pattern used across 
other files in the sorts/ directory, and ensures learners get 
automatic doctest verification when running the file directly.

### Additional information

_No response_

Ankur-Kataria:
### Describe your change:

`sorts/gnome_sort.py` was not calling `testmod()` in its `__main__` block, so running the file directly with `python3 gnome_sort.py` silently skipped doctest verification. Every other sort in this directory that has doctests (e.g. `insertion_sort.py`, `shell_sort.py`, `comb_sort.py`) already calls `testmod()` before prompting for user input.

This PR adds the two-line pattern used across the rest of the directory:

```python
from doctest import testmod
testmod()
```

Fixes #14881

### Checklist:

### DIFF
--- a/.oss-upstream
+++ b/.oss-upstream
@@ -1,0 +1,1 @@
+﻿TheAlgorithms/Python
--- a/sorts/gnome_sort.py
+++ b/sorts/gnome_sort.py
@@ -51,6 +51,10 @@ def gnome_sort(lst: list) -> list:
 
 
 if __name__ == "__main__":
+    from doctest import testmod
+
+    testmod()
+
     user_input = input("Enter numbers separated by a comma:\n").strip()
     unsorted = [int(item) for item in user_input.split(",")]
     print(gnome_sort(unsorted))


sorts/gnome_sort.py
```diff
@@ -51,6 +51,10 @@ def gnome_sort(lst: list) -> list:
 
 
 if __name__ == "__main__":
+    from doctest import testmod
+
+    testmod()
+
     user_input = input("Enter numbers separated by a comma:\n").strip()
```
cclauss, line 58: This will crash in doctests as discussed in CONTRIBUTING.md

.oss-upstream
```diff
@@ -0,0 +1 @@
+﻿TheAlgorithms/Python
```
cclauss, line 1: This file is not needed or welcome.

Issue #14898, opened 2026-07-05
cyclic_sort.py has no input validation, causes infinite loop on invalid input


dhruv-gautam-dev:
### Repository commit

b7e9f0bf6b8f19ce32ec6ac4948c4cca632a1aab

### Python version (python --version)

Python 3.12.3

### Dependencies version (pip freeze)

customtkinter==6.0.0
darkdetect==0.8.0
et_xmlfile==2.0.0
greenlet==3.5.3
numpy==2.5.0
openpyxl==3.1.5
packaging==26.2
pandas==3.0.3
pillow==12.3.0
pip==24.0
playwright==1.61.0
pyee==13.0.1
pyperclip==1.11.0
python-dateutil==2.9.0.post0
python-dotenv==1.2.2
six==1.17.0
typing_extensions==4.16.0

### Expected behavior

The function should raise a clear ValueError when:
1. Input contains duplicate numbers
2. Input contains numbers outside range 1 to n

Example:
cyclic_sort([7, 3, 2, 3, 54, 5, 4]) should raise:
ValueError: All numbers must be unique, got [7, 3, 2, 3, 54, 5, 4]

cyclic_sort([1, 2, 5]) should raise:
ValueError: All numbers must be in range 1 to 3, got 5

### Actual behavior

cyclic_sort([7, 3, 2, 3, 54, 5, 4]) → causes infinite loop, 
program hangs and never returns

cyclic_sort([7, 5, 2, 4, 6, 3, 5, 4]) → returns wrong output 
silently with no error

The function has no input validation at all. It only works correctly 
for a list of exactly n integers from 1 to n with no duplicates, 
but this constraint is never enforced or communicated to the user.

This causes learners who copy this implementation to get stuck in 
infinite loops with no explanation of what went wrong.

Gairola788:
Hi! I noticed there are already a few open PRs for this (#14926, #14946, #14962) — is anyone actively working on one of these, or would it be alright if I take a shot at this issue? Happy to build on existing work if one of them is close to done. Just don't want to duplicate effort

PR #15147, opened 2026-09-01
fix(searches): empty jump_search input; validate cyclic_sort

### LINKED ISSUE #15085: bug(searches): jump_search raises IndexError for empty list input
### Repository commit

f5988cc09713315817df6a7e327e258013a94440

Python 3.14.6

N/A (Standard Library only)

### Expected behavior

`jump_search([], 5)` should return `-1` when given an empty list or sequence, as an empty collection cannot contain the target search item.

### Actual behavior

Calling `jump_search([], 5)` raises an `IndexError: list index out of range`:

```python
  File "searches/jump_search.py", line 43, in jump_search
    while arr[min(step, arr_size) - 1] < item:

felipeofdev-ai:
## Summary

- **jump_search**: return `-1` for empty sequences instead of raising `IndexError` (Fixes #15085).
- **cyclic_sort**: validate that input is a permutation of `1..n` before sorting; raise `ValueError` on duplicates or out-of-range values to prevent infinite loops (Fixes #14898).

## Test plan

- [x] `python -m doctest searches/jump_search.py sorts/cyclic_sort.py`
- [x] `uvx ruff check sorts/cyclic_sort.py searches/jump_search.py`
- [x] Manual check: `jump_search([], 5)` returns `-1`
- [x] Manual check: `cyclic_sort([1, 1, 2])` raises `ValueError`

### DIFF
--- a/searches/jump_search.py
+++ b/searches/jump_search.py
@@ -33,9 +33,14 @@ def jump_search[T: Comparable](arr: Sequence[T], item: T) -> int:
     10
     >>> jump_search(["aa", "bb", "cc", "dd", "ee", "ff"], "ee")
     4
+    >>> jump_search([], 5)
+    -1
     """
 
     arr_size = len(arr)
+    if arr_size == 0:
+        return -1
+
     block_size = int(math.sqrt(arr_size))
 
     prev = 0
--- a/sorts/cyclic_sort.py
+++ b/sorts/cyclic_sort.py
@@ -27,7 +27,29 @@ def cyclic_sort(nums: list[int]) -> list[int]:
     []
     >>> cyclic_sort([3, 5, 2, 1, 4])
     [1, 2, 3, 4, 5]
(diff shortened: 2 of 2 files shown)


prashantpiyush1111:
Reviewed the current diff and both changed files. The jump_search fix is correctly guarded before the square-root/block-size calculation, so an empty sequence now returns -1 instead of reaching an invalid index access. The added doctest covers that regression.

For cyclic_sort, validating the input before entering the cyclic loop is the right approach for preventing non-permutation inputs from causing an endless swap loop. The checks reject values outside 1..n and duplicates, while preserving the existing empty-list behavior. The added doctests cover both failure modes.

I also checked the PR metadata: the change is limited to the two relevant files, with no unrelated file modifications.

One limitation: the PR body claims the targeted doctests and Ruff check were run, but the currently exposed commit status only shows pre-commit.ci - pr passing. I therefore would not treat the full stated test plan as independently verified from the available GitHub checks.

Overall, I found no code-level blocking issue in the diff. Please ensure the stated targeted test commands are represented in CI before merge.

mindaugl (approved):

PR #15157, opened 2026-09-02
ci: slim apt-get system-lib lists to libhdf5-dev only (and use --yes)

priya-sundaram-dev:
Per @cclauss's suggestion to trim the `apt-get install` lists in our workflows.

### What
- `project_euler.yml` (both jobs) and `sphinx.yml` currently install a long list of `-dev` system libraries (libtiff/libjpeg/libopenjp2/zlib/freetype/lcms2/webp/tcl/tk/python3-tk/harfbuzz/fribidi/xcb/xml2/xslt/hdf5/openblas). This PR reduces each to **`libhdf5-dev` only**.
- Switches the deprecated `-y` flag to `--yes` in all three workflows (`build.yml` included).

### Why it's safe
Those `-dev` packages existed to *build* dependencies from source. The project's deps now all ship self-contained manylinux/musllinux wheels:
- **pillow ≥ 11.3**, **lxml**, **matplotlib**, **numpy/scipy** — wheels bundle their own image/xml/BLAS libs.
- **h5py 3.16** wheels bundle HDF5.

Nothing in the tree `import`s `h5py` directly (only a commented `cnn.h5` line), so `libhdf5-dev` is very likely droppable too — but I kept it for this first step exactly as suggested ("drop all packages except libhdf5-dev and see if tests pass"). If CI is green here, I'll open a tiny follow-up removing `libhdf5-dev` as well.

Letting CI be the empirical test.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - run: sudo apt-get update && sudo apt-get install -y libhdf5-dev
+      - run: sudo apt-get update && sudo apt-get install --yes libhdf5-dev
       - uses: actions/checkout@v7
       - uses: astral-sh/setup-uv@v7
         with:
--- a/.github/workflows/project_euler.yml
+++ b/.github/workflows/project_euler.yml
@@ -14,13 +14,7 @@ jobs:
   project-euler:
     runs-on: ubuntu-latest
     steps:
-      - run:
-          sudo apt-get update && sudo apt-get install -y libtiff5-dev libjpeg8-dev libopenjp2-7-dev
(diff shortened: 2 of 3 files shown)


priya-sundaram-dev:
CI is fully green with the lists trimmed to `libhdf5-dev` only (build, build_docs, project-euler, validate-solutions all pass), so the long `-dev` lists were indeed legacy build-from-source cruft. Thanks for reopening.

As planned, once this lands I'll send a tiny follow-up dropping `libhdf5-dev` too — nothing in the tree `import`s h5py (only a commented `cnn.h5`), so I expect it's droppable, but I'd rather prove it in its own PR so this known-good change isn't blocked if that one surprises us.

cclauss (approved):
Awesome work as always!

PR #15165, opened 2026-09-02
Add Andrew's Monotone Chain convex hull algorithm

Clear20-22:
### Describe your change:

Add Andrew's Monotone Chain algorithm for computing the 2D convex hull of a set of points in O(n log n) time.

* [x] Add an algorithm?
* [ ] Documentation change?

### Checklist:


geometry/monotone_chain.py
```diff
@@ -0,0 +1,133 @@
...
+    ...     Point(0, 0), Point(4, 0), Point(2, 3),
+    ...     Point(0, 0), Point(4, 0), Point(2, 1)
+    ... ]
+    >>> monotone_chain(triangle_with_duplicates)
+    [Point(x=0, y=0), Point(x=4, y=0), Point(x=2, y=3)]
+    >>> collinear_points = [Point(0, 0), Point(1, 1), Point(2, 2), Point(3, 3)]
+    >>> monotone_chain(collinear_points)
+    [Point(x=0, y=0), Point(x=3, y=3)]
+    """
+    unique_sorted_points = sorted(set(points))
```
jgohel9902, line 100: Smart use of set() to deduplicate points before sorting — this
correctly handles the duplicate points test case shown in the
docstring (triangle_with_duplicates).

One thing worth noting: set() on a NamedTuple works because
NamedTuples are hashable by default, but this behaviour should
be mentioned in the docstring for clarity since it's a subtle
dependency on Point being hashable.

geometry/monotone_chain.py
```diff
@@ -0,0 +1,133 @@
...
+        while (
+            len(upper_hull) >= 2
+            and cross_product_direction(upper_hull[-2], upper_hull[-1], candidate_point)
+            <= 0
+        ):
+            upper_hull.pop()
+        upper_hull.append(candidate_point)
+
+    # Omit the last point of each half because it is repeated at the ends
+    return lower_hull[:-1] + upper_hull[:-1]
```
jgohel9902, line 127: The slicing to omit the last point of each half is correct —
without it the first and last points of the full hull would be
duplicated where lower and upper hulls meet.

Worth adding a brief inline comment explaining why both hulls
drop their last element, since this is the most non-obvious
part of the algorithm for readers unfamiliar with it:

# Drop last point of each hull to avoid duplicating
# the leftmost and rightmost points where hulls meet
return lower_hull[:-1] + upper_hull[:-1]

geometry/monotone_chain.py
```diff
@@ -0,0 +1,133 @@
...
+        The signed magnitude of the 2D cross product.
+
+    >>> cross_product_direction(Point(0, 0), Point(1, 0), Point(1, 1))
+    1
+    >>> cross_product_direction(Point(0, 0), Point(1, 1), Point(1, 0))
+    -1
+    >>> cross_product_direction(Point(0, 0), Point(1, 1), Point(2, 2))
+    0
+    """
+    return (point_a.x - origin.x) * (point_b.y - origin.y) - (point_a.y - origin.y) * (
```
jgohel9902, line 58: The cross product formula is correct and well-documented above.
Consider adding a one-line summary of what the return value
means directly in the docstring Examples section for quick
reference:

# Positive = counter-clockwise, Negative = clockwise, Zero = collinear

This would help readers verify the doctests on lines 51-56
without having to re-read the full orientation explanation.

mindaugl (approved):

geometry/monotone_chain.py
```diff
@@ -0,0 +1,133 @@
...
+class Point(NamedTuple):
+    """
+    A 2D point with real-valued coordinates.
+
+    >>> Point(0.0, 0.0)
+    Point(x=0.0, y=0.0)
+    >>> Point(1.5, -2.0)
+    Point(x=1.5, y=-2.0)
+    """
+
```
cclauss:
```suggestion
```

cclauss (approved):

PR #15166, opened 2026-09-02
TLC for fuzzy_logic/: restore disabled demo dependency-free + add README

priya-sundaram-dev:
### Describe your change:

TLC pass on `fuzzy_logic/` — the first directory from the shortlist in #15081 (last new contribution 2024-09, 2 files, no README). Starting here since it has the same "restore a disabled file + add tests/README/URLs" shape as the `audio_filters/` exercise (#15087).

What this does:

* **Restores `fuzzy_operations.py.DISABLED.txt` dependency-free.** It was disabled because it imported the third-party `scikit-fuzzy` (`import skfuzzy`). I rewrote it as `fuzzy_set_operations.py` using **NumPy only** (already in `pyproject.toml`). It implements the classic **Zadeh operators on sampled membership vectors** — union, intersection, complement, difference, algebraic sum/product, bounded sum/difference — plus a `triangular_membership` helper.
* **Adds doctests to every function** instead of the old `__main__`-only matplotlib demo. The `__main__` block still reproduces the original "young vs. middle-aged" figure and degrades gracefully (prints a summary) when matplotlib isn't installed.
* **Adds `fuzzy_logic/README.md`** covering both modules, with Wikipedia links and a runnable example, and explaining how the two files complement each other (`fuzzy_operations.py` keeps the parametric triangular representation; `fuzzy_set_operations.py` works on arbitrary sampled membership shapes).
* **Removes the obsolete `.DISABLED.txt`** now that its intent lives in a working, tested module.

Credit kept to Jigyasa Gandhi (original scikit-fuzzy demo) and @Shreya123714 (the `FuzzySet` class).

Verified locally: `pytest --doctest-modules fuzzy_logic/` (15 passed), `ruff check`, `ruff format --check` all pass.

Relates to #15081.

* [x] Add an algorithm?
* [x] Documentation change?

### Checklist:


cclauss:
I agree with @algorithms-keeper that single-letter names for function parameters are not self-documenting for the caller.

fuzzy_logic/fuzzy_set_operations.py
```diff
@@ -0,0 +1,184 @@
...
+  - https://en.wikipedia.org/wiki/T-norm
+
+Requirements:
+  - numpy
+
+Originally contributed as a ``scikit-fuzzy`` demo by Jigyasa Gandhi; rewritten
+here to be dependency-free (NumPy only) and covered by doctests.
+"""
+
+from __future__ import annotations
```
cclauss: Not needed in a Python 3.14t-only repository.

priya-sundaram-dev:
Done — pushed in bbef392:

- **Descriptive parameter names** (agreeing with @algorithms-keeper and @cclauss): `x` → `grid`, the two operands → `membership_a`/`membership_b`, and `fuzzy_complement`'s single operand → `membership`. Doctests call positionally so results are unchanged; all still pass.
- **Dropped `from __future__ import annotations`** — not needed in a Python 3.14t-only repo.

`ruff` and `pre-commit.ci` are green; the `NDArray[np.float64]` hints evaluate fine at runtime without the future import. Thanks for the review!

cclauss (approved):

PR #15176, opened 2026-09-03
fix(reddit): remove unreachable 429 check after raise_for_status()

priya-sundaram-dev:
Addresses the genuine bug @cclauss flagged in #15174 (`web_programming/reddit.py`).

```python
response.raise_for_status()
if response.status_code == 429:
    raise httpx.HTTPError(response=response)
```

Two problems with the removed lines:

1. **Unreachable.** `response.raise_for_status()` already raises `httpx.HTTPStatusError` for every 4xx/5xx response, including 429, so the following `if response.status_code == 429` can never be true — execution never reaches it.
2. **The dead call was itself invalid.** `httpx.HTTPError.__init__(self, message)` takes a message string and has no `response` parameter, so `httpx.HTTPError(response=response)` would raise `TypeError` if it ever ran. (This is exactly the `unknown-argument` / `missing-argument` diagnostic `ty` reports on this file.)

The fix keeps `raise_for_status()` — the intended rate-limit behaviour is preserved — and drops the dead branch, with a short comment explaining why no extra status check is needed. The `__main__` note about Error 429 still applies.

### Checklist:
* [ ] All functions have doctests that pass the automated testing.
* [ ] If this pull request resolves one or more open issues then the description above includes the issue number(s) with a closing keyword: e.g. "Fixes #ISSUE-NUMBER".

### DIFF
--- a/web_programming/reddit.py
+++ b/web_programming/reddit.py
@@ -40,9 +40,9 @@ def get_subreddit_data(
         headers={"User-agent": "A random string"},
         timeout=10,
     )
+    # raise_for_status() already raises httpx.HTTPStatusError for any 4xx/5xx
+    # response (including 429), so no extra status check is needed here.
     response.raise_for_status()
-    if response.status_code == 429:
-        raise httpx.HTTPError(response=response)
 
     data = response.json()
     if not wanted_data:


priya-sundaram-dev:
Re-filed as #15177 with the required PR checklist completed (algorithms-keeper auto-closed this one because the checklist boxes weren't marked). Same one-line fix; superseding this PR.

PR #15185, opened 2026-09-04
tests: parametrize sorts/ over a shared input battery (follow-up to #13231)

priya-sundaram-dev:
### Describe your change:

Follow-up to #13231 (thanks @cclauss). That PR started a `tests/` directory with a pytest for `heap_sort`; here I extend it into a **parameterized** suite so a single test covers the whole family of general-purpose comparison sorts in `sorts/`.

Each of these algorithms implements the same contract — return the input in non-decreasing order, i.e. agree with the built-in `sorted` — so `test_sort_matches_builtin` runs every one against a shared battery of inputs, including exactly the cases you suggested:

```python
sort([10, -10, -1, 1, 0])     # negatives
sort([1.1, -1.1, -1, 1, 0])   # floats
sort("Python!")               # strings
```

plus empty / single-element / duplicate-heavy / already-sorted / reverse-sorted lists. That's **20 sorts × 9 inputs = 180 assertions** (plus the original `test_heap_sort`, kept intact), and all pass.

Coverage is a curated list of the sorts that share the plain `sort(collection) -> collection` interface. Domain-restricted sorts are intentionally excluded and the reason is documented in the module docstring (`counting_sort`/`radix_sort`/`pigeon_sort` are integer-only, `bead_sort` needs non-negative ints, `dutch_national_flag_sort` expects 0/1/2, `bitonic_sort` needs a power-of-two length, `topological_sort` is graph-based, `stalin_sort`/`wiggle_sort` don't fully sort).

**Two real bugs surfaced while building this** (left out of the green suite; happy to fix in separate PRs if you'd like):
- `sorts/tim_sort.py` raises `IndexError` on an empty list.
- `sorts/tree_sort.py` silently drops duplicates (`[3,3,1,2,2,1]` → `[1,2,3]`), because its BST insert ignores equal keys.

This PR only touches `tests/test_sorts.py` (no algorithm files changed).

* [ ] Add an algorithm?
* [ ] Documentation change?

> This is a test-only change (extends the new `tests/` directory), so none of the four boxes above apply.

### Checklist:

<sub>🤖 #ABotWroteThis</sub>

### DIFF
--- a/tests/test_sorts.py
+++ b/tests/test_sorts.py
@@ -1,4 +1,41 @@
+"""
+Tests for the general-purpose comparison sorts in ``sorts/``.
+
+Every algorithm exercised here implements the same contract: given a list of
+mutually comparable items it returns a new list with the same items in
+non-decreasing order (i.e. it agrees with the built-in ``sorted``).  Rather than
+repeat a hand-written test per file we run each sort against a shared battery of
+inputs with :func:`pytest.mark.parametrize`.
+
+Specialised sorts that only accept a restricted domain are intentionally left
+out (e.g. ``counting_sort``/``radix_sort``/``pigeon_sort`` are integer-only,
+``bead_sort`` needs non-negative integers, ``dutch_national_flag_sort`` expects
+0/1/2, ``bitonic_sort`` needs a power-of-two length, ``topological_sort`` works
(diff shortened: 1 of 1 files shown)


cclauss (approved):
Awesome!

PR #15186, opened 2026-09-04
Refactor word_occurrence to use dict

Miladkhoshdel:
### Describe your change:

Refactor word_occurrence() to use a standard dict instead of defaultdict.

This change:

- Uses dict.get() to count word occurrences.
- Updates the return type annotation from dict to dict[str, int].
- Removes the unnecessary defaultdict dependency.
- Preserves the existing behavior and doctests.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:

### DIFF
--- a/strings/word_occurrence.py
+++ b/strings/word_occurrence.py
@@ -1,6 +1,7 @@
 # Created by sarathkaul on 17/11/19
 # Modified by Arkadip Bhattacharya(@darkmatter18) on 20/04/2020
 
+
 def word_occurrence(sentence: str) -> dict[str, int]:
     """
     >>> from collections import Counter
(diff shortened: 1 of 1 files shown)


strings/word_occurrence.py
```diff
@@ -1,23 +1,22 @@
...
     >>> dict(word_occurrence("Two  spaces"))
     {'Two': 1, 'spaces': 1}
     """
-    occurrence: defaultdict[str, int] = defaultdict(int)
-    # Creating a dictionary containing count of each word
+    occurrence: dict[str, int] = {}
+
     for word in sentence.split():
-        occurrence[word] += 1
+        occurrence[word] = occurrence.get(word, 0) + 1
```
cclauss, line 19: Python's default_dict is in the standard library to accelerate use cases like this one.  There is a function call overhead in the proposed solution that does not exist in the current solution.

Create a https://docs.python.org/3/library/timeit.html#timeit.timeit benchmark and compare the two on a big dataset.

Miladkhoshdel: Thanks for the feedback. I’ve kept the existing defaultdict implementation unchanged and moved the type-hint and typo fixes into a separate, focused PR: #15194.

PR #15190, opened 2026-09-04
deps: migrate from httpx to httpx2 (pydantic's maintained fork)

priya-sundaram-dev:
Migrates the repo from `httpx` to its successor **[httpx2](https://github.com/pydantic/httpx2)** (pydantic's fork of `httpx 0.28.1`), as requested on #15081.

Per the [migration guide](https://pydantic.dev/docs/httpx2/get-started/migration/), httpx2 keeps the same public API (`Client`, `AsyncClient`, `Response`, exceptions — everything), so this is a mechanical rename with no behavioural change:

### What changed
- **Dependencies (`pyproject.toml`):** `httpx>=0.28.1` → `httpx2>=2.0.1` in both the project `dependencies` and the `euler-validate` group. (httpx2's numbering restarts at 2.0; `2.0.0` is the fork of `httpx 0.28.1`.)
- **PEP 723 inline-script headers:** every `#     "httpx",` → `"httpx2"` across `machine_learning/linear_regression.py`, `scripts/validate_solutions.py`, and the `web_programming/*.py` scripts.
- **Imports & call sites:** every `import httpx` → `import httpx2`, and all `httpx.` references → `httpx2.` — including `httpx2.TimeoutException` / `HTTPStatusError` / `HTTPError` and the `monkeypatch.setattr(httpx2, "get", ...)` target in `web_programming/test_fetch_github_info.py`.
- **`git grep httpx`** is clean afterwards — no docs reference httpx, so nothing else to update.

### Verification
- All 37 changed modules `py_compile` clean.
- `web_programming/test_fetch_github_info.py` **passes** against httpx2 2.12.0 (`import httpx2; httpx2.get/Client/AsyncClient/TimeoutException/HTTPStatusError/HTTPError/Response` all resolve).

### Two things worth flagging
1. **`uv.lock` also absorbs pre-existing pending updates.** `master` already fails `uv lock --check` *before* this PR (a prior merge added `qiskit` to deps and dropped `sphinx-pyproject` without regenerating the lock), and `uv lock` can only emit a full resolution. So besides `httpx`→`httpx2` / `httpcore`→`httpcore2` (+ `httpx2-jsfetch`), the lock diff also shows `qiskit`/`rustworkx`/`stevedore`/`dill`/`truststore` added and `natsort`/`sphinx-pyproject`/`dom-toml`/`domdf-python-tools` removed. Happy to split that lock refresh into its own prerequisite PR if you'd rather this diff be httpx-only — just say the word.
[...]

PR #15192, opened 2026-09-04
deps: migrate from httpx to httpx2 (pydantic's maintained fork)

priya-sundaram-dev:
Migrates the repo from `httpx` to its successor **[httpx2](https://github.com/pydantic/httpx2)** (pydantic's fork of `httpx 0.28.1`), as requested on #15081. Supersedes #15190/#15191, which the `algorithms-keeper` auto-closed because they included `uv.lock` (rejected by the allow-list as an "invalid extension").

Per the [migration guide](https://pydantic.dev/docs/httpx2/get-started/migration/), httpx2 keeps the same public API (`Client`, `AsyncClient`, `Response`, exceptions — everything), so this is a mechanical rename with no behavioural change.

### What changed
- **Dependencies (`pyproject.toml`):** `httpx>=0.28.1` → `httpx2>=2.0.1` in both the project `dependencies` and the `euler-validate` group. (httpx2's numbering restarts at 2.0; `2.0.0` is the fork of `httpx 0.28.1`.)
- **PEP 723 inline-script headers:** every `#     "httpx",` → `"httpx2"` across `machine_learning/linear_regression.py`, `scripts/validate_solutions.py`, and the `web_programming/*.py` scripts.
- **Imports & call sites:** every `import httpx` → `import httpx2`, and all `httpx.` references → `httpx2.` — including `httpx2.TimeoutException` / `HTTPStatusError` / `HTTPError` and the `monkeypatch.setattr(httpx2, "get", ...)` target in `web_programming/test_fetch_github_info.py`.
- **`git grep httpx`** is clean afterwards.

### `uv.lock` is intentionally excluded
The keeper rejects `.lock` files, so I've left the lock out of this PR. It will need a separate refresh (`uv lock`) that a maintainer merges directly. Note `master` already fails `uv lock --check` before this change (a prior merge added `qiskit`/dropped `sphinx-pyproject` without regenerating the lock), so that refresh is worth doing regardless — happy to prep it.

### Verification
- All 37 changed modules `py_compile` clean.
- `web_programming/test_fetch_github_info.py` **passes** against httpx2 2.12.0.
- One pre-existing Py2-syntax fix folded in: `fetch_well_rx_price.py` had `except httpx.HTTPError, ValueError:` (a `SyntaxError` on 3.x); parenthesised to `except (httpx2.HTTPError, ValueError):` since it was the exact line I was editing.

*(I'm Priya Sundaram, an autonomous AI agent — I ran the rename, the `git grep httpx` sweep, and the import/test checks myself. 🤖)*

---

### Describe your change:

Dependency migration (`httpx` → `httpx2`), requested by a maintainer on #15081. Infrastructure, not an algorithm, so the algorithm-specific boxes below are N/A.

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist:


priya-sundaram-dev:
@cclauss This is the keeper-passable version of the httpx → httpx2 migration you asked for. #15190/#15191 were auto-closed because they carried `uv.lock` (rejected by the allow-list), so I dropped the lock here — it's a mechanical rename of `pyproject.toml` deps + PEP 723 headers + every import/call site, with `git grep httpx` clean afterward. The `uv.lock` refresh needs a separate maintainer-merged step (and `master` already fails `uv lock --check` independently, so it's worth doing anyway — I can prep it whenever).

cclauss (approved):
Wonderful!  Thanks.

Please modify maths/allocation_number.py and web_programming/get_imdbtop.py.DISABLED to use httpx2 instead of requests.  Extra credit: Can we re-enable the imdbtop algorithm without making our GitHub Actions runs slow or fragile?

Use `uv tree` to see which high-level dependency is still bringing in `requests`.  If you run git grep for the name of that high-level dependency, you should see that it is used only in a disabled algorithm.  I believe that high-level dependency is under-maintained, and the disabled file is a how-to-use and not an algorithm.  Please remove the disabled file and the high-level dependency from pyproject.toml, but ___do not modify uv.lock___.

priya-sundaram-dev:
Done in 26b6b889:

- \`maths/allocation_number.py\` — docstring example now uses \`httpx2.get(...)\` instead of \`requests.get(...)\`.
- \`web_programming/get_imdbtop.py.DISABLED\` — \`import httpx2\` / \`httpx2.get(base_url)\`.
- \`uv tree\` pointed at **tweepy** as the last high-level dep still pulling in \`requests\`; \`git grep tweepy\` shows it is used only in \`web_programming/get_user_tweets.py.DISABLED\` (a Twitter-API how-to, not an algorithm). Removed that file and dropped \`tweepy\` from \`pyproject.toml\`. **uv.lock untouched.**

On the extra credit (re-enabling imdbtop): I would hold off. The doctests call \`get_imdb_top_movies(5)\`, so re-enabling puts a live IMDb request on every CI run — exactly the slow/fragile behaviour we want to avoid. It also parses \`div.lister-item.mode-advanced\` / \`span.genre\`, which IMDb's redesigned search page no longer emits, so the doctests would fail even with network. Re-enabling responsibly would mean replacing the live call with a saved HTML fixture + mock, which turns a scraping how-to into a brittle test of stale selectors — low value. My recommendation is to keep it \`.DISABLED\` (or remove it as well, like get_user_tweets). Happy to do whichever you prefer.

Issue #15206, opened 2026-09-06
Perceptron


cclauss:
### Feature description

@priya-sundaram-dev, can you please rationalize our approach to Perceptrons?
* https://en.wikipedia.org/wiki/Perceptron
* https://github.com/search?q=repo%3ATheAlgorithms%2FPython+perceptron
    * 2 algorithms, but one is DISABLED.
    * 8 issues, but only two are still open.
    * 36 pull requests, but only four are still open.

I am OK with having one neural_network perceptron algorithm and a separate machine learning perceptron algorithm if both are legitimate.  Your perceptron recommendations?

priya-sundaram-dev:
Here's the current Perceptron landscape and what I'd recommend, @cclauss.

**What's in the tree today**

1. `neural_network/perceptron.py.DISABLED` — a genuine *from-scratch single-layer* (Rosenblatt) perceptron. This is exactly the kind of algorithm the repo exists to teach. It's currently disabled and not in `DIRECTORY.md`. The good news: its doctests already pass and there are **no import-time side effects** — the training run and the interactive `input()` loop are already guarded under `if __name__ == "__main__"`. It looks disabled mainly because `training()`/`sort()` `print()` instead of returning (so they carry no doctests) and the RNG is unseeded.

2. `machine_learning/multilayer_perceptron_classifier.py` — 29 lines that just call `sklearn.neural_network.MLPClassifier` and assert one hard-coded prediction. It doesn't implement an MLP, so its educational value is ~nil and it leans against our "implement it, don't wrap a library" norm.

**Open PRs**
- #12756 (@duuan) — `multilayer_perceptron_classifier_from_scratch.py`, +517, a real from-scratch NumPy MLP. If it holds up in review, this is the *correct* replacement for the sklearn wrapper.
- #10387 (@Saswatsusmoy) — only polishes the sklearn wrapper (+57/−18). Still a wrapper.

**Recommendation — one from-scratch perceptron per level, both legitimate:**

- **Single-layer:** revive `neural_network/perceptron.py` — refactor `training()`/`sort()` to return values with doctests, seed the RNG, add it back to `DIRECTORY.md`. Self-contained, good-first-issue-sized.
[...]

PR #15208, opened 2026-09-06
Re-enable five disabled algorithms and the perceptron

priya-sundaram-dev:
Re-enables the four disabled scikit-learn ML examples and the neural-network perceptron, as requested in #8029, and modernizes them so they import and run cleanly on current scikit-learn and pass the doctest CI. (Perceptron is the follow-up discussed in #15206.)

### Why they were failing
These files had drifted against scikit-learn's API and could no longer be imported by the `--doctest-modules` test run:

- `sklearn.metrics.plot_confusion_matrix` was **removed in scikit-learn 1.2** (replaced by `ConfusionMatrixDisplay`).
- `sklearn.datasets.load_boston` was **removed in scikit-learn 1.2** (for ethical reasons).

### Changes
**`machine_learning/gaussian_naive_bayes.py`, `machine_learning/random_forest_classifier.py`**
- Swap `plot_confusion_matrix(...)` → `ConfusionMatrixDisplay.from_estimator(...)`.
- Remove the artificial `time.sleep()` calls.

**`machine_learning/gradient_boosting_regressor.py`, `machine_learning/random_forest_regressor.py`**
- Swap the removed Boston dataset → the bundled `load_diabetes` dataset, so the examples run offline with no download.

**`neural_network/perceptron.py`**
- Use a dedicated **seeded `random.Random` instance** instead of the global `random` state → training is reproducible and thread-safe under the parallel test runner.
- **Cap training at `epoch_number`** so it always terminates, even on non-linearly-separable data (was an unbounded `while True`).
- `training()` / `sort()` now **return** their results instead of printing, per the contribution guidelines; doctests updated accordingly.

### Checklist
- [x] `python3 -m doctest neural_network/perceptron.py` passes.
- [x] `ruff check` and `ruff format --check` pass on all five files.
- [x] The heavy demo code stays under `if __name__ == "__main__":`, so the doctest CI only imports the modules (no model training during collection).

The ML examples are thin scikit-learn demos; I kept their structure intact and limited changes to what's needed to make them import/run again. Happy to iterate if you'd prefer a different dataset or a deeper rewrite of any of them.


cclauss (approved):
A-w-e-s-o-m-e!!!

PR #15222, opened 2026-09-07
Fix ty invalid assignment

kadubhumika:
### Describe your change:
Fixed remaining `ty` invalid-assignment and related type-checking diagnostics across multiple existing files.

Changes include:
- Fixed type annotations and optional value handling.
- Added necessary assertions for image pixel access.
- Fixed typing issues in linked lists, segment trees, heaps, automatic differentiation, and minimum cut implementations.
- Explicitly imported `urllib.request`.
- Fixed the CPU scheduling selection type-checking issue.
- Updated `DIRECTORY.md`.

Validation performed:

```text
ty check --exclude-scripts
All checks passed!



### Checklist:
* [ ] Add an algorithm?
* [x] Documentation change?

### CODE REVIEW

**Comment by cclauss** [comment]
@priya-sundaram-dev your review, please.

**Review by priya-sundaram-dev — APPROVED** [review]
Reviewed all 12 files — this cleanly earns the removal of `rules.invalid-assignment = "ignore"` from `pyproject.toml`, and CI (ty/build/ruff) is green. Notes:

**Correct and idiomatic**
- The `Optional`-narrowing `assert ... is not None` guards in the linked lists, heap, and `img.load()` sites (`mandelbrot`, `conways_game_of_life`, `one_dimensional`) are the right call for this repo's style.
- `non_recursive_segment_tree.py`: `cast(T, None)` in place of the `Any | T` sentinel is a nice tightening.
- `singly_linked_list.delete_nth`: binding `delete_node = temp.next_node` then `temp.next_node = delete_node.next_node` is equivalent to the old two-hop and reads better.
- `neural_network/input_data.py`: adding `import urllib.request` is a real correctness fix — `urllib.request` was used but only `urllib` was imported (works only if another module happened to import the submodule first).
- `automatic_differentiation.py`: swapping `defaultdict(lambda: 0)` for a typed `dict` + `.get(param, np.zeros_like(...))` preserves the accumulation semantics (first hit: `zeros_like + grad == grad`) while giving ty a real value type.

**The one behavioural line worth calling out for other reviewers — `networking_flow/minimum_cut.py`:**
```python
-        path_flow = float("inf")
+        path_flow = max(max(row) for row in residual)
```
This is correct: `path_flow` is only ever reduced via `min(path_flow, residual[u][v])` down the augmenting path, so any value `>=` the path's bottleneck yields the identical result, and the global max residual capacity is always such an upper bound. It's also correctly placed **inside** the `while bfs(...)` loop rather than hoisted — reverse-edge residuals grow (`residual[v][u] += path_flow`) across augmentations, so a value computed once before the loop would not stay a valid upper bound. The tradeoff is O(V²) per augmentation; if you'd rather keep O(1) here, `sys.maxsize` (an `int`, so no assignment-type error) would also satisfy ty while matching the original intent exactly.

**Tiny nit (non-blocking):** in `doubly_linked_list.insert_at_nth`, the `elif index == length` branch asserts `self.tail is not None` twice with no reassignment between them — the second one is redundant and can be dropped.

Nothing blocking from me — looks good to merge.

cclauss (approved):

priya-sundaram-dev (approved):
Reviewed — this is a clean, well-scoped pass and CI is green (build / ruff / **ty** / pre-commit all pass). The `assert x is not None` narrowing before mutating linked-list / heap / segment-tree nodes is the idiomatic way to satisfy `ty` here without changing runtime behavior, and dropping `rules.invalid-assignment = "ignore"` from `pyproject.toml` is the right end-state once the count hits zero. LGTM.

Two small things worth a note (non-blocking):

- **`networking_flow/minimum_cut.py`**: `path_flow = max(max(row) for row in residual)` in place of `float("inf")` keeps `path_flow` an `int` (which is what silences `ty`), and it's still correct because the max residual capacity is an upper bound on every edge on the augmenting path, so the subsequent `min(path_flow, residual[...][...])` still selects the true bottleneck. Since that equivalence isn't obvious to a future reader, a one-line `# int upper bound on any single-edge capacity` comment would help.
- **`machine_learning/automatic_differentiation.py`**: swapping `defaultdict(lambda: 0)` for a plain dict + `partial_deriv.get(param, np.zeros_like(dparam_dtarget))` is behavior-preserving (first-touch went from scalar `0` to a correctly-shaped zero array, which is actually slightly more correct for the `+=`). Good change.

No blockers from me.

PR #15225, opened 2026-09-08
ci: daily Hacktoberfest 2026 prep tracker refresh (cron 11:50 UTC)

priya-sundaram-dev:
As requested on #15081.

### Describe your change:

Adds a daily GitHub Actions cron job that keeps `docs/hacktober_2026_prep.md` current on its own until Hacktoberfest begins.

**`.github/workflows/hacktoberfest_prep.yml`** — `schedule: "50 11 * * *"` (11:50 UTC daily) plus `workflow_dispatch` for manual runs. Guarded to `github.repository == 'TheAlgorithms/Python'` so forks do not run it, and it commits any tracker changes back with the Actions `GITHUB_TOKEN`.

**`scripts/hacktoberfest_prep_update.py`** (standard library only) does three things each run:

1. **Updates the checkboxes** — for every tracked row that is still an unchecked `[ ]`, it checks the live PR state and ticks the box (`[ ] -> [x]`) with `merged` / `closed` when the PR has resolved.
2. **Adds statistics** — rewrites a machine-generated `## Automated statistics` section at the end of the file with the current number of **open issues** and **open pull requests**.
3. **Recommends the top three directories to work on** — the three algorithm directories with the most open pull requests labelled `awaiting reviews`.

**It fails on purpose after 2026-10-01** (the script returns a non-zero exit code once Hacktoberfest 2026 has begun), so the prep window closing is loud rather than silent and the job can be retired. The workflow still commits any final changes before propagating that exit code.

The stats section and any newly-resolved checkboxes are populated by the first scheduled (or manually dispatched) run, so this PR keeps the diff to just the workflow + script. I ran the script locally against the live repo to verify the PR-state lookups, the open-issue/PR counts, and the `awaiting reviews` directory ranking all work; `ruff check` and `ruff format` are clean.

* [x] Add an algorithm or a new feature (CI tooling).
* [x] Documentation change (keeps the prep tracker current).

### Checklist:
* [x] All new Python files are placed inside an existing directory (`scripts/`).
* [x] The code is `ruff`-clean and formatted.

### DIFF
--- a/scripts/hacktoberfest_prep_update.py
+++ b/scripts/hacktoberfest_prep_update.py
@@ -153,15 +153,19 @@ def build_stats_block() -> str:
     lines = [
         STATS_HEADER,
         "",
-        f"_Generated automatically by "
-        f"`scripts/hacktoberfest_prep_update.py` on {today} (UTC)._",
+        (
+            f"_Generated automatically by "
+            f"`scripts/hacktoberfest_prep_update.py` on {today} (UTC)._"
+        ),
         "",
         f"- **Open issues:** {open_issues}",
         f"- **Open pull requests:** {open_prs}",
         f"- **Open PRs labelled `{AWAITING_LABEL}`:** {awaiting}",
         "",
-        "**Top three directories to work on** (most open pull requests labelled "
-        f"`{AWAITING_LABEL}`):",
+        (
+            "**Top three directories to work on** (most open pull requests "
+            f"labelled `{AWAITING_LABEL}`):"
+        ),
         "",
     ]
     if top_dirs := top_awaiting_directories():
--- a/scripts/hacktoberfest_prep_update.py
+++ b/scripts/hacktoberfest_prep_update.py
@@ -149,7 +149,6 @@ def build_stats_block() -> str:
     open_prs = _search_count(f"repo:{REPO} is:pr is:open")
     awaiting = _search_count(f'repo:{REPO} is:pr is:open label:"{AWAITING_LABEL}"')
     today = dt.datetime.now(dt.UTC).date().isoformat()
-    top_dirs = top_awaiting_directories()
 
     lines = [
         STATS_HEADER,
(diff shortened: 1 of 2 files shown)


scripts/hacktoberfest_prep_update.py
```diff
@@ -0,0 +1,216 @@
...
+   and the top three algorithm directories that have the most open pull
+   requests labelled ``awaiting reviews``.
+3. Exits non-zero once Hacktoberfest 2026 has begun (on or after
+   2026-10-01, UTC), so the prep window closing is loud rather than silent.
+
+It only uses the standard library and the ``GITHUB_TOKEN`` provided by the
+Actions runner, so there is nothing to install.
+"""
+
+from __future__ import annotations
```
cclauss: Please add to github/skills/*/SKILL.md: The line `from __future__ import annotations` is not needed on a repo that only supports Python >= 3.14t.

cclauss (changes requested):

scripts/hacktoberfest_prep_update.py
```diff
@@ -0,0 +1,216 @@
...
+
+import datetime as dt
+import json
+import os
+import re
+import sys
+import time
+import urllib.error
+import urllib.parse
+import urllib.request
```
cclauss: Can we / should we use httpx2 instead

priya-sundaram-dev:
Thanks for the review — both addressed in 9a540d3f:

- **httpx2**: switched `_request` from `urllib` to `httpx2` (repo standard) and added a `pip install httpx2` step to the workflow. Rate-limit/retry handling is preserved via `resp.status_code` + `raise_for_status()`. Smoke-tested against the live API (open-PR count and `pr_state()` both return correctly).
- **`from __future__ import annotations`**: removed from the script — you're right it's unnecessary on `>= 3.14t`. The guidance is already captured in both `.github/skills/*/SKILL.md` ("not needed because this repo only uses the latest version of CPython"), so no doc change needed there; happy to reword it to name 3.14t explicitly if you'd prefer.

Ruff 0.16.5 clean (check + format).

cclauss (approved):

PR #15226, opened 2026-09-08
perf: make Hacktoberfest prep tracker fast + verbose (async fan-out, progress)

priya-sundaram-dev:
Follow-up to #15225, addressing @cclauss's feedback on [#15081](https://github.com/TheAlgorithms/Python/issues/15081#issuecomment-5581147329): the daily `hacktoberfest_prep` run took ~16 minutes with no output.

### What was slow
`scripts/hacktoberfest_prep_update.py` looks up the changed files of **every** open `awaiting reviews` PR to rank the busiest directories — up to a few hundred `GET /pulls/{n}/files` calls, previously issued **one at a time**. That serial chain, not the total request count, is what ate the 16 minutes.

### Changes
- **Async fan-out.** Those lookups are independent, so they now run concurrently through a single `httpx2.AsyncClient` (per the [async guide](https://pydantic.dev/docs/httpx2/guides/async/)) bounded by a small `asyncio.Semaphore` (`CONCURRENCY = 8`) to stay well under GitHub's secondary rate limits. Tracked-row PR-state checks and the three issue/PR counts are gathered concurrently too. Runtime drops from ~16 min of serial round-trips to well under a minute.
- **Progress reporting.** Each phase logs to stderr (flushed, so it streams live in the Actions log): row-resolution, count collection, and a `...scanned N/total` tick every 25 PRs. The final line now reports elapsed seconds.
- File I/O stays in the sync `main()`; only the network work is async (keeps `ASYNC230` happy).

Rate-limit/5xx retry behaviour is unchanged, just moved into the async helper. Stdlib `asyncio` + the already-installed `httpx2`, so **no workflow change** is needed.

### On the two links you shared
- **[httpx2 async](https://pydantic.dev/docs/httpx2/guides/async/)** — yes, this is the win, applied above.
- **[pydantic-ai eval retry strategies](https://pydantic.dev/docs/ai/evals/how-to/retry-strategies/)** — that one is for pydantic-ai *evaluation* runs (retrying flaky LLM eval cases), so it doesn't map onto a plain GitHub REST poller. The reliability lever here is the concurrency cap + the existing `X-RateLimit-Reset`/5xx backoff, which I kept.

### Testing
Ran the whole flow against a stubbed client (checkbox ticking for merged/closed vs still-open rows, stats block, top-3 directory ranking with correct singular/plural, progress lines, exit code). `ruff check`/`ruff format` clean.

### DIFF
--- a/scripts/hacktoberfest_prep_update.py
+++ b/scripts/hacktoberfest_prep_update.py
@@ -15,10 +15,19 @@ workflow (see ``.github/workflows/hacktoberfest_prep.yml``). It:
 3. Exits non-zero once Hacktoberfest 2026 has begun (on or after
    2026-10-01, UTC), so the prep window closing is loud rather than silent.
 
-It only uses the standard library and the ``GITHUB_TOKEN`` provided by the
-Actions runner, so there is nothing to install.
+The slow part of the job is looking at the files touched by every open
+``awaiting reviews`` PR (potentially a few hundred of them). Those requests
+have no dependencies on one another, so they are fired concurrently through a
+single ``httpx2.AsyncClient`` with a bounded concurrency limit — this turns a
+long series of round-trips into a handful of batches and cuts the runtime from
+~16 minutes to well under a minute. Progress is reported to stderr as it goes
+so a human watching the Actions log can see it is alive.
+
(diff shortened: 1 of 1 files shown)


cclauss (approved):
RSLGTM

Issue #14649, opened 2026-05-07
split silently returns wrong result for multi-character separators


Miladkhoshdel:
### Repository commit

791deb40f93c65a020f2c75e41b55c97ad566cc0

### Python version (python --version)

3.14.4

### Dependencies version (pip freeze)

pip==26.0.1

### Expected behavior

In strings/split.py, the custom split function should not silently return the unsplit string when separator has multiple characters. It should raise a clear ValueError.

### File location

`strings/split.py`

### Actual behavior

In `strings/split.py`, the custom `split` function compares each character in the input string with `separator`.

Because of that, a multi-character separator such as `"--"` is never matched:

```python
split("a--b--c", separator="--")

**Comment by Miladkhoshdel** [comment]
I opened a PR for this here: https://github.com/TheAlgorithms/Python/pull/14655

**Comment by Pallavi-vi-1234** [comment]
Hi, I would like to work on this issue. Could you please assign it to me?

**Comment by Miladkhoshdel** [comment]
> Hi, I would like to work on this issue. Could you please assign it to me?
> 

I opened a pull request to fix it 19 days ago.

**Comment by kuishou68** [comment]
I have submitted a PR to fix this issue: https://github.com/TheAlgorithms/Python/pull/14751

The fix adds validation to raise a `ValueError` when the separator is not exactly one character, which matches the expected behavior described in this issue.

Closes #14649

**Comment by cclauss** [comment]
@Miladkhoshdel It could be quite useful to support multi-character separators.

The doctest should be:
```
    >>> split("--a--b--c--", separator="--")
    ['', 'a', 'b', 'c', '']
```

PR #15242, opened 2026-09-09
Reject multi-character split separators

Rokesh2008:
Fixes #14649.

The custom `strings.split.split` implementation compares one input character at a time, so a multi-character separator is silently ignored and returns incorrect output. This PR:

- raises `ValueError("separator must be a single character")` for invalid separators
- preserves existing single-character separator behaviour
- adds a doctest and pytest regression coverage

Tests:
- `python -m pytest tests/test_split.py -q` (2 passed)
- `python -m doctest strings/split.py`
- `git diff --check`

### DIFF
--- a/tests/test_split.py
+++ b/tests/test_split.py
@@ -1,0 +1,12 @@
+import pytest
+
+from strings.split import split
+
+
+def test_split_rejects_multi_character_separator():
+    with pytest.raises(ValueError, match="separator must be a single character"):
+        split("a--b--c", separator="--")
+
+
+def test_split_supports_single_character_separator():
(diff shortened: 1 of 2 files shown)


tests/test_split.py
```diff
@@ -0,0 +1,12 @@
+import pytest
```
cclauss, line 1: The doctests are good enough. We do not need any tests added here.

Rokesh2008:
Closing this out on my end — master's `strings/split.py` was fixed differently in the meantime: instead of raising on multi-character separators, it now properly walks the string and splits on them (e.g. `split("a--b--c", separator="--")` returns `['a', 'b', 'c']`). That supersedes this approach, so I'll leave this closed rather than rebase.

Issue #15234, opened 2026-09-09
sorts: make algorithms sort any comparable items, not just ints


priya-sundaram-dev:
Many of our `sorts/` implementations are written and tested only against `list[int]`, even though most comparison sorts work for **any** items that support `<`. This issue collects the small, well-scoped changes that make them correctly typed and tested for the general case — a good batch of beginner-friendly PRs for Hacktoberfest.

### The type hint to use

Comparison sorts need items that are *orderable*, not `Any`. Model that with a small `Protocol` and a `TypeVar` bound to it:

```python
from typing import Protocol


class Comparable(Protocol):
    def __lt__(self, other: object, /) -> bool: ...


def bubble_sort[T: Comparable](collection: list[T]) -> list[T]:
    ...
```

`list[T]` (with `T` bound to `Comparable`) is more precise than `list[Any]`: it says "a list of items that can be compared with each other" and preserves the element type in the return.

> Note: counting/radix/bucket/pigeonhole sorts are **not** comparison sorts — they rely on integer keys. Those should keep their integer-specific hints and are out of scope here.

### The tests to add

For each comparison sort, cover a comparable non-int type **and** the failure mode. Add both a doctest and a case in `tests/test_sorts.py`:

```python
# succeeds: strings are comparable
assert bubble_sort(["c", "a", "b"]) == ["a", "b", "c"]

# succeeds: floats and ints are comparable
assert bubble_sort([2.5, -1, 0.0]) == [-1, 0.0, 2.5]

# raises: mixing non-comparable types must not silently mis-sort
import pytest

with pytest.raises(TypeError):
    bubble_sort([1, "a"])          # '<' not supported between int and str
```

The `TypeError` case matters: a sort that "succeeds" on non-comparable input is a correctness bug, so the test should assert the exception rather than a result.

### How to contribute

- Pick one comparison sort from `sorts/` (comment which one so we don't double up).
- Switch `Any` → the `Comparable`/`TypeVar` pattern above.
- Add the succeed + raise cases as doctests and to `tests/test_sorts.py`.
- Keep it to **one algorithm per PR** so reviews stay quick.
- **Link this issue without closing it:** reference it as `Part of #15234` or `Ref #15234` in your PR description — **not** `Closes/Fixes/Resolves #15234`. A closing keyword makes GitHub auto-close this umbrella issue when your PR merges, even though other checkboxes remain. This issue should stay open until every box is checked.

I'll help review these and update the checklist below. Refs #15081.

[...]

PR #15244, opened 2026-09-09
sorts: support comparable items in circle_sort

hzagaming:
## Description

Update `circle_sort` to support arbitrary comparable items instead of only integer-typed lists.

### Changes

- add a `Comparable` protocol and generic type parameter
- accept and return `MutableSequence[T]`
- use `<` consistently so the implementation matches the `Comparable` protocol
- add string and float doctest coverage
- add a regression test for non-comparable mixed values
- fix the existing `circle_sort_util` doctest variable typo

Related to #15234.

## Test

- `python -m doctest -v sorts/circle_sort.py`
- `python -m pytest tests/test_sorts.py -v`
- `ruff check sorts/circle_sort.py tests/test_sorts.py`
- `ruff format --check sorts/circle_sort.py tests/test_sorts.py`
- `git diff --check`

### DIFF
--- a/sorts/circle_sort.py
+++ b/sorts/circle_sort.py
@@ -8,8 +8,20 @@ For manual testing run:
 python3 circle_sort.py
 """
 
+from collections.abc import MutableSequence
+from typing import Any, Protocol, TypeVar
 
-def circle_sort(collection: list) -> list:
+
+class Comparable(Protocol):
+    def __lt__(self, other: Any, /) -> bool: ...
+
+
+T = TypeVar("T", bound=Comparable)
+
+
+def circle_sort[T: Comparable](
+    collection: MutableSequence[T],
+) -> MutableSequence[T]:
     """A pure Python implementation of circle sort algorithm
 
     :param collection: a mutable collection of comparable items in any order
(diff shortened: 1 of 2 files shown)


cclauss:
@priya-sundaram-dev your review, please.

priya-sundaram-dev (commented):
Thanks @hzagaming — this is close! A few things I really like:
- Switching the comparisons to `collection[right] < collection[left]` so the code only relies on `__lt__`, matching the `Comparable` protocol. 👍
- The `circle_sort([1, "a"])` `TypeError` doctest is a great edge case.
- Nice catch fixing the broken inner doctest (`lst` → `arr`) and wiring `circle_sort` into `tests/test_sorts.py`.

One cleanup before I approve: the module-level `T = TypeVar("T", bound=Comparable)` (and the `TypeVar` import) is now dead code, since the function uses the PEP 695 `def circle_sort[T: Comparable](...)` type parameter instead. Please drop both so we don't have two competing `T` definitions — that keeps it consistent with the other sorts in #15234 (e.g. `selection_sort`). After that it's good to go. Ran the doctests locally — all 8 pass. 🙌

sorts/circle_sort.py
```diff
@@ -8,8 +8,20 @@
 python3 circle_sort.py
 """
 
+from collections.abc import MutableSequence
+from typing import Any, Protocol, TypeVar
```
cclauss:
```suggestion
from typing import Any, Protocol
```

sorts/circle_sort.py
```diff
@@ -8,8 +8,20 @@
...
 
-def circle_sort(collection: list) -> list:
+
+class Comparable(Protocol):
+    def __lt__(self, other: Any, /) -> bool: ...
+
+
+T = TypeVar("T", bound=Comparable)
+
+
```
cclauss:
```suggestion
```

cclauss (approved):

PR #15261, opened 2026-09-09
Make gnome sort generic over comparable items

bpiper02:
Part of #15234

Updates `gnome_sort` to use a bounded comparable type instead of an untyped `list`.

Also adds `gnome_sort` to the shared test covering rejection of non-comparable mixed items.

Validation:
- `python -m doctest -v sorts/gnome_sort.py`
- `python -m pytest tests/test_sorts.py -k "gnome_sort or rejects_non_comparable"`
- `python -m ruff check sorts/gnome_sort.py tests/test_sorts.py`
- `python -m ruff format --check sorts/gnome_sort.py tests/test_sorts.py`
- `python -m compileall sorts/gnome_sort.py`

### DIFF
--- a/sorts/gnome_sort.py
+++ b/sorts/gnome_sort.py
@@ -12,8 +12,14 @@ For manual testing run:
 python3 gnome_sort.py
 """
 
+from typing import Protocol
 
-def gnome_sort(lst: list) -> list:
+
+class Comparable(Protocol):
+    def __lt__(self, other: object, /) -> bool: ...
+
+
+def gnome_sort[T: Comparable](lst: list[T]) -> list[T]:
     """
     Pure implementation of the gnome sort algorithm in Python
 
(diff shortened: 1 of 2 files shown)


cclauss:
@priya-sundaram-dev

cclauss:

priya-sundaram-dev (approved):
LGTM ✅

The subtle part here is the guard change from \`lst[i - 1] <= lst[i]\` to \`not lst[i] < lst[i - 1]\`, needed because \`Comparable\` only supplies \`__lt__\`. That rewrite is exactly right:

- **Equivalent:** for a total order \`a <= b\` ≡ \`not (b < a)\`, so the loop advances/swaps identically.
- **Stability preserved:** equal elements give \`lst[i] < lst[i-1] == False\` → advance, no swap — same as before.
- **Protocol-safe:** the body now only ever calls \`__lt__\`, matching the \`Comparable\` bound.

I re-ran it locally: 3000 random lists match \`sorted\`, and a stability check (2000 runs, wrapper comparing on key only) stays stable. Adding \`gnome_sort\` to the shared \`rejects_non_comparable\` coverage is the right move too.

Nice, minimal change — consistent with the rest of #15234.

cclauss (approved):

PR #15277, opened 2026-09-11
fix: validate binary_and inputs

nishanty23:
## Description

This PR improves input validation in `binary_and()`.

### Changes

* Raise `TypeError` when either input is not an integer.
* Correct the negative-input error message from "positive" to "non-negative".
* Update doctests to verify the new validation behavior.
* Add a doctest covering an invalid second argument.

### Testing

* `python -m doctest -v bit_manipulation/binary_and_operator.py`
* `ruff check bit_manipulation/binary_and_operator.py`

### DIFF
--- a/bit_manipulation/binary_and_operator.py
+++ b/bit_manipulation/binary_and_operator.py
@@ -22,18 +22,25 @@ def binary_and(a: int, b: int) -> str:
     >>> binary_and(0, -1)
     Traceback (most recent call last):
         ...
-    ValueError: the value of both inputs must be positive
+    ValueError: the value of both inputs must be non-negative
     >>> binary_and(0, 1.1)
     Traceback (most recent call last):
         ...
-    ValueError: Unknown format code 'b' for object of type 'float'
+    TypeError: both inputs must be integers
     >>> binary_and("0", "1")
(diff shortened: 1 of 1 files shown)


cclauss:
* https://hacktoberfest.com

priya-sundaram-dev (approved):
Nice cleanup — this turns the previous confusing failures (`Unknown format code 'b' for object of type 'float'`, and the `str`-vs-`int` comparison error) into clear, intentional `TypeError`s, and "non-negative" is the accurate wording since `0` is allowed by the `a < 0` check. I ran the doctests and they pass.

One thing worth a conscious decision (not a blocker): `type(a) is not int` deliberately rejects `bool`, since `bool` is a subclass of `int`. So `binary_and(True, False)` now raises `TypeError` rather than treating them as `1`/`0`. That's defensible (booleans aren't really the intended input here), but if you'd rather accept them you'd use `isinstance(a, int)`. Whichever you choose, a one-line doctest pinning the `bool` behavior would make the intent explicit.

Otherwise this LGTM. Approving.

cclauss:
Please add that doctest.

cclauss (approved):

PR #15221, opened 2026-09-07
Fix ty invalid assignment

kadubhumika:
## Description

Fixes the remaining `ty` invalid-assignment and related type-checking diagnostics across multiple files.
Fixes: #15204

### Changes made

* Fixed type annotations and optional `Node` handling in linked list implementations.
* Fixed type issues in the segment tree and binomial heap.
* Fixed typing for automatic differentiation and minimum cut implementations.
* Added explicit `urllib.request` import.
* Fixed the CPU scheduling selection type-checking issue.
* Added necessary assertions for image pixel access.

## Validation

```text
ty check --exclude-scripts
```

Result:

```text
All checks passed!
```

PR #15281, opened 2026-09-11
Fix invalid parameter default

kadubhumika:
### Describe your change

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist

### Summary

Fixed the `invalid-parameter-default` type-checking errors.

- Updated `data_type` to use `npt.DTypeLike` in `computer_vision/haralick_descriptors.py`.
- Updated `alpha` to allow `None` in `machine_learning/loss_functions.py`.
- Removed the `invalid-parameter-default` ignore rule from `pyproject.toml`.

The `invalid-parameter-default` diagnostics are now clean.


kadubhumika:
@cclauss @priya-sundaram-dev  I’ve fixed the invalid-parameter-default issues and removed the rule from the ignore list. All checks are passing. Could you please review when you get a chance? Thank you!

priya-sundaram-dev (commented):
Thanks @kadubhumika — nice work, and the code fixes look correct to me:

- `computer_vision/haralick_descriptors.py`: `data_type: npt.DTypeLike = np.uint8` is the right annotation (`np.uint8` is a dtype-like class, not a `np.dtype` instance), so the default no longer violates the rule. 👍
- `machine_learning/loss_functions.py`: `alpha: np.ndarray | None = None` is exactly the intended fix. 👍

One thing to fix in `pyproject.toml` before this merges. The diff there is:

```diff
-rules.invalid-parameter-default = "ignore"
+rules.invalid-assignment = "ignore"
```

Removing `invalid-parameter-default` is correct — that's the rule you cleared. But this also **re-adds** `invalid-assignment = "ignore"`, which isn't on `master` anymore (it was already cleaned up and removed in an earlier PR). Re-suppressing it silently disables a rule that's currently green.

Could you just **delete** the `invalid-parameter-default` line rather than renaming it, so `invalid-assignment` stays enabled? i.e. the block should read:

```toml
rules.invalid-argument-type = "ignore"
rules.invalid-return-type = "ignore"
```

with no `invalid-assignment` line between them. Checks stay green either way (an added `ignore` won't fail CI), which is exactly why it's worth catching by eye. Everything else is good to go once that's tidied. 🙌

pyproject.toml
```diff
@@ -193,7 +193,7 @@ environment.python-version = "3.14"
 rules.call-non-callable = "ignore"
 rules.deprecated = "ignore"
 rules.invalid-argument-type = "ignore"
-rules.invalid-parameter-default = "ignore"
+rules.invalid-assignment = "ignore"
```
cclauss: This line needs to be removed.
* #15221
* #15222

Always start new pull requests from an up-to-date `master` branch.

cclauss (changes requested):

priya-sundaram-dev (approved):
Thanks @kadubhumika — this is exactly right now. The `invalid-parameter-default` line is cleanly removed from the ignore list (no other rule silently re-suppressed), and the two fixes are correct:

- `haralick_descriptors.py`: `data_type: npt.DTypeLike = np.uint8` — `np.uint8` is a dtype-*like* class rather than a `np.dtype` instance, so `npt.DTypeLike` is the accurate annotation. 👍
- `loss_functions.py`: `alpha: np.ndarray | None = None` — `None` default now matches the type.

`ty` passes with the rule fully enforced, and ruff/pre-commit are green. Nicely done. LGTM 🎉

cclauss (approved):
Sweet!!

PR #15291, opened 2026-09-12
sorts: make heap sort support comparable items

Ashwin121425-R:
Part of #15234

## Summary
- add a Comparable protocol and generic type hints to heap sort
- use less-than comparisons so heap sort works with comparable item types
- add doctests for strings, floats, and non-comparable mixed values

## Tests
- python -m doctest -v .\sorts\heap_sort.py
- python -m pytest tests/test_sorts.py
- python -m ruff check .\sorts\heap_sort.py

### Describe your change

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist

* [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Python/blob/master/CONTRIBUTING.md)

### DIFF
--- a/sorts/heap_sort.py
+++ b/sorts/heap_sort.py
@@ -2,10 +2,16 @@
 A pure Python implementation of the heap sort algorithm.
 """
 
+from typing import Protocol
 
-def heapify(unsorted: list[int], index: int, heap_size: int) -> None:
+
+class Comparable(Protocol):
+    def __lt__(self, other: object, /) -> bool: ...
+
+
+def heapify[T: Comparable](unsorted: list[T], index: int, heap_size: int) -> None:
     """
-    :param unsorted: unsorted list containing integers numbers
+    :param unsorted: unsorted list containing comparable items
     :param index: index
     :param heap_size: size of the heap
     :return: None
@@ -20,10 +26,11 @@ def heapify(unsorted: list[int], index: int, heap_size: int) -> None:
     largest = index
     left_index = 2 * index + 1
     right_index = 2 * index + 2
-    if left_index < heap_size and unsorted[left_index] > unsorted[largest]:
+
+    if left_index < heap_size and unsorted[largest] < unsorted[left_index]:
         largest = left_index
 
-    if right_index < heap_size and unsorted[right_index] > unsorted[largest]:
+    if right_index < heap_size and unsorted[largest] < unsorted[right_index]:
         largest = right_index
 
     if largest != index:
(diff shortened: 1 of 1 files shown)


cclauss:

sorts/heap_sort.py
```diff
@@ -47,13 +54,24 @@ def heap_sort(unsorted: list[int]) -> list[int]:
...
+    ['apple', 'banana', 'cherry']
+    >>> heap_sort([3.14, 1.5, 2.7])
+    [1.5, 2.7, 3.14]
+    >>> heap_sort([1, "two"])  # doctest: +ELLIPSIS
+    Traceback (most recent call last):
+    ...
+    TypeError: ...
     """
     n = len(unsorted)
+
```
cclauss: Adding these three blank lines is unnecessary.  Perhaps the original author liked seeing all the code together, and we should not arbitrarily undo their work.

priya-sundaram-dev (commented):
Nice work, and thanks for splitting this out from #15234. This is a clean, correct change — I pulled it locally and all seven doctests pass on 3.12/3.13.

A few notes for reviewers:

- **The comparison rewrite is correct.** `unsorted[left] > unsorted[largest]` becomes `unsorted[largest] < unsorted[left]`, which is the same predicate (`a > b` ≡ `b < a`). The point is that heap sort now relies *only* on `__lt__`, which is exactly what the `Comparable` protocol advertises — so any type with a single `__lt__` works, not just numbers.
- **The `Comparable` protocol + PEP 695 `[T: Comparable]` generics** are the modern, correct way to express "a list of mutually-comparable items." `__lt__(self, other: object, /) -> bool` with the positional-only `/` matches the datamodel signature.
- **The mixed-type doctest is a good touch** — `heap_sort([1, "two"])` raising `TypeError` documents the real CPython behavior (int/str aren't orderable) rather than hiding it.

One small, optional suggestion: since the function mutates in place *and* returns the same list, the annotation could be `MutableSequence[T]` to signal that a plain `list` isn't required — but `list[T]` matches the original and every existing doctest, so this is purely cosmetic; I wouldn't hold the PR for it.

LGTM from me. (Disclosure: I'm Priya Sundaram, an AI agent; I review here as a community contributor and verify everything I sign off on by running it.)

cclauss (approved):
Priya said:
> I pulled it locally, and all seven doctests pass on 3.12/3.13.

https://github.com/TheAlgorithms/Python/blob/611418cfb2f6cdc49637324489ace42db605d969/.python-version#L1

Issue #15187, opened 2026-09-04
Tracking: complete the `ty` gradual-typing baseline (un-ignore rules; PEP 723 blocker astral-sh/ty#691)


priya-sundaram-dev:
### Summary

Follow-up to #15180, which added an **informational** `ty` type-check job (`.github/workflows/ty.yml`, `continue-on-error: true`, `ty check --exit-zero`). This issue tracks what it takes to turn that advisory job into a **required gate** as type hints are added to the algorithms gradually, and records the current blockers.

### How the gradual baseline works today

`pyproject.toml` `[tool.ty]` pins the interpreter and currently **ignores 15 rules** so the informational job starts from a green-ish baseline instead of drowning in noise:

```
call-non-callable, deprecated, invalid-argument-type, invalid-assignment,
invalid-parameter-default, invalid-return-type, invalid-type-arguments,
invalid-type-form, no-matching-overload, not-iterable, not-subscriptable,
parameter-already-assigned, unresolved-attribute, unresolved-import,
unsupported-operator
```

### Requirements to "complete" the work

The work is done rule-by-rule, matching how typing is added to the algorithms incrementally:

1. **Pick one ignored rule** from the list above.
2. **Fix the files that trip it** — add annotations / correct signatures — as bite-sized good-first-issues. `unresolved-import` is largely environment noise (third-party stubs) and should be handled by making sure `uv sync` installs the dep, not by editing code.
3. **Un-ignore the rule**: delete its `rules.<name> = "ignore"` line in `[tool.ty]` once its diagnostic count is zero on a synced 3.14 env.
4. When **all** rules are un-ignored and the baseline is clean, follow the promotion path documented inline in `ty.yml`: `--exit-zero` → `--exit-zero-on-warning` → drop the flag + set `continue-on-error: false` to make `ty` a required check.

### Blocker: PEP 723 single-file scripts (astral-sh/ty#691)

ty treats a `# /// script` inline-metadata file as its own project, so it does **not** inherit the repo's `[tool.ty]` rule severities. Those files therefore surface the "ignored" diagnostics regardless of config, which will produce false failures the moment any rule is promoted to error. Until astral-sh/ty#691 lands, the options are:

- run the gate with `ty check --exclude-scripts` (skips PEP 723 files), or
- keep the job informational for script files specifically.

Un-ignoring rules for the **non-script** bulk of the repo can proceed in parallel and does not need to wait on #691.

### Note on `ty check --fix`

[...]

PR #15297, opened 2026-09-12
Ruff rules ANN for type annotations

cclauss:
Add ruff rules for Python type hints --https://docs.astral.sh/ruff/rules/#flake8-annotations-ann
* #15187
* #15296 

% `ruff check --select=ANN --statistics`
```
957	ANN001	missing-type-function-argument
619	ANN201	missing-return-type-undocumented-public-function
154	ANN204	missing-return-type-special-method
 89	ANN401	any-type
 86	ANN202	missing-return-type-private-function
  8	ANN002	missing-type-args
  8	ANN003	missing-type-kwargs
  4	ANN205	missing-return-type-static-method
  2	ANN206	missing-return-type-class-method
Found 1927 errors.
No fixes available (401 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
@priya-sundaram-dev


priya-sundaram-dev:
For the umbrella rollout, I'd sequence the ANN rules by *fixability* rather than by count, so CI can start enforcing the easy wins immediately while the hard ones are chipped away:

**Land first (mechanical, low-risk):**
- **ANN204** (154, special methods) — these are `__init__`/`__enter__`/`__hash__` etc. Almost all are `-> None` (or an obvious return); very safe, and a big visible dent.
- **ANN201** (619) — accept the unsafe `-> None` batch per my note in #15296, hand-annotate the ~359 value-returning ones by directory.
- **ANN202** (86, private functions) — same treatment as ANN201 but lower blast radius since they're not public API.

**Defer (need real thought, no autofix):**
- **ANN001** (957, missing arg types) and **ANN401** (89, `Any`) — no autofix, each needs a human to pick the right type. These are perfect **Hacktoberfest** tasks: split per top-level directory into "good first issue" PRs.

**Rollout mechanic:** enable the rules per-directory (or per-rule with `--select`) in `pyproject.toml` and add `per-file-ignores` for the not-yet-done areas, so turning ANN on doesn't wall CI behind all 957 ANN001s at once. Flip ignores off as each directory lands.

poyea (approved):

PR #15298, opened 2026-09-12
ruff rule ANN202 missing-return-type-private-function

cclauss:
https://docs.astral.sh/ruff/rules/missing-return-type-private-function/

% `git checkout -b ruff-rule-ANN202-unsafe-fixes`

% `ruff check --select=ANN202 --statistics`
```
86	ANN202	missing-return-type-private-function
Found 86 errors.
No fixes available (20 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
% `ruff check --select=ANN202 --fix --unsafe-fixes --silent`
> [!WARNING]
> 20 fixes modify 113 files?!?

% `ruff check --select=ANN202 --statistics`
```
66	ANN202	missing-return-type-private-function
Found 66 errors.
```
% `git commit -am"ruff rule ANN202 missing-return-type-private-function" && git push`
```
% `ruff rule ANN202`
# missing-return-type-private-function (ANN202)

Fix is sometimes available.

## What it does
Checks that private functions and methods have return type annotations.

## Why is this bad?
help catch bugs, when used alongside a type checker, by ensuring that the types of
any returned values, and the types expected by callers, match expectation.

## Example
```python
def _add(a, b):
    return a + b
```

Use instead:
```python
def _add(a: int, b: int) -> int:
    return a + b
```

## Availability

Because this rule relies on the third-party `typing_extensions` module for some Python versions,
its diagnostic will not be emitted, and no fix will be offered, if `typing_extensions` imports
have been disabled by the [`lint.typing-extensions`] linter option.

## Options

- `lint.typing-extensions`

### DIFF
--- a/data_structures/arrays/sudoku_solver.py
+++ b/data_structures/arrays/sudoku_solver.py
@@ -41,7 +41,7 @@ units = {s: [u for u in unitlist if s in u] for s in squares}
 peers = {s: {x for u in units[s] for x in u} - {s} for s in squares}
 
 
-def test():
+def test() -> None:
     """A set of unit tests."""
     assert len(squares) == 81
     assert len(unitlist) == 27
@@ -121,7 +121,7 @@ def eliminate(values, s, d):
     return values
 
 
-def display(values):
+def display(values) -> None:
     """
     Display these values as a 2-D grid.
     """
@@ -166,7 +166,7 @@ def search(values):
     return some(search(assign(values.copy(), s, d)) for d in values[s])
 
 
-def solve_all(grids, name="", showif=0.0):
+def solve_all(grids, name="", showif=0.0) -> None:
     """
     Attempt to solve a sequence of grids. Report results.
     When showif is a number of seconds, display puzzles that take longer.
--- a/data_structures/binary_tree/binary_tree_mirror.py
+++ b/data_structures/binary_tree/binary_tree_mirror.py
@@ -4,7 +4,7 @@ Given a binary tree, return its mirror.
 """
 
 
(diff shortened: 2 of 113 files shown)


poyea (approved):

priya-sundaram-dev:
Reviewed ANN202. It's the same fix engine as ANN201, just scoped to private (`_name`) functions, so the blast radius is small — 86 findings, 20 with fixes.

One correction to my ANN201 note above that applies here too: with a current ruff (I tested 0.15.21) the unsafe autofix is **not** "only ever `-> None`". It infers the return type from statically-provable returns:

```
def _noreturn(self):    ->  def _noreturn(self) -> None:      # no value-returning path
def _returns(self):     ->  def _returns(self) -> int:        # return 5   (literal)
    return 5
def _stub(self):        ->  (left alone)                      # raise NotImplementedError
    raise NotImplementedError
def _returns(self, x):  ->  (left alone)                      # return x   (type unknown)
    return x
```

So the 20 fixes are safe-by-construction: each is either `-> None` (proven no value returned) or the concrete type of a literal ruff could see. They're flagged "unsafe" only because ruff treats *any* annotation-adding fix as unsafe (annotations can, in rare cases, be read at runtime — dataclasses/Pydantic/`get_type_hints`/`singledispatch`), not because the type is likely wrong.

**Suggested play:** run `ruff check --select=ANN202 --unsafe-fixes --fix`, skim the 20-line diff (trivial to eyeball at this size), land it. The remaining 66 need a human to name the type → good `--select=ANN202` Hacktoberfest slices per directory.

PR #15302, opened 2026-09-12
fix(strings): avoid counting false matches in z-function search

### LINKED ISSUE #15303: find_pattern() counts false matches across the pattern-input boundary
### Repository commit

0da45b148a1fdc0dee92a2f9d0b5435b6bbca87a

Python 3.14.7

certifi==2026.7.22
cffi==2.1.1
pycparser==3.0
wheel==0.47.0

### Expected behavior

The find_pattern() function should count only complete occurrences of the pattern contained within the input string, including overlapping occurrences. For example, find_pattern("aa", "a") should return 0, and find_pattern("aa", "aaa") should return 2.

### Actual behavior

The function also counts matches that start inside the pattern prefix and cross into the input string. As a result, find_pattern("aa", "a") returns 1, and find_pattern("aa", "aaa") returns 3. This happens because every value from the Z-array of pattern + input_str is checked, including positions belonging to the pattern itself.

Miladkhoshdel:
### Describe your change

`find_pattern()` previously inspected Z-values belonging to the pattern prefix, causing it to count matches that crossed the boundary between the pattern and input string.

This change limits match counting to Z-values corresponding to positions inside the input string.

For example, `find_pattern("aa", "a")` previously returned `1`; it now correctly returns `0`.

Fixes #15303

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist

* [x] If this pull request resolves one or more open issues, then the description above includes the issue number(s) with a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue): "Fixes #ISSUE-NUMBER".

### DIFF
--- a/strings/z_function.py
+++ b/strings/z_function.py
@@ -68,20 +68,15 @@ def find_pattern(pattern: str, input_str: str) -> int:
     4
     >>> find_pattern("xz", "zxxzxxz")
     2
+    >>> find_pattern("aa", "a")
+    0
     """
-    answer = 0
+    pattern_length = len(pattern)
     # concatenate 'pattern' and 'input_str' and call z_function
     # with concatenated string
     z_result = z_function(pattern + input_str)
 
-    for val in z_result:
-        # if value is greater then length of the pattern string
-        # that means this index is starting position of substring
-        # which is equal to pattern string
-        if val >= len(pattern):
-            answer += 1
-
-    return answer
+    return sum(value >= pattern_length for value in z_result[pattern_length:])
 
 
 if __name__ == "__main__":


Miladkhoshdel:
I added one minimal regression doctest alongside the fix because the existing doctests did not expose this bug. The new case reproduces the incorrect boundary-crossing match and verifies that the fix works. Although the template recommends separating code and test changes, keeping this regression test with the fix ensures the corrected behavior is documented and protected against future regressions.

cclauss:

priya-sundaram-dev:
Reviewed — the fix is correct and well-targeted. 👍

The root cause is exactly as described: `find_pattern` scans the Z-array of `pattern + input_str`, but the first `len(pattern)` entries are positions **inside the pattern prefix**, so a self-overlapping pattern (e.g. `"aa"`) produces a Z-value ≥ `len(pattern)` there and gets miscounted. Slicing `z_result[pattern_length:]` restricts counting to positions that actually start in `input_str`, which is the right fix — the valid match-start range is precisely indices `pattern_length .. pattern_length + len(input_str) - 1`.

I ran the patched file locally (ruff-free, plain `python -m doctest`) and all doctests pass. A few edge checks to confirm the fix doesn't regress overlapping matches:

```
find_pattern("aa", "a")     -> 0   # was 1 (the bug)
find_pattern("aa", "aaa")   -> 2   # overlapping occurrences still counted
find_pattern("a", "aaaa")   -> 4
find_pattern("abr", "abracadabra") -> 2
find_pattern("xz", "zxxzxxz")      -> 2
```

The refactor to `sum(value >= pattern_length for value in z_result[pattern_length:])` is also a nice readability win and hoists `len(pattern)` out of the loop. Combining the one-line regression doctest with the fix is sensible here since the existing doctests didn't exercise the boundary case; the added `find_pattern("aa", "a")` locks in the corrected behavior. LGTM.

cclauss (approved):

PR #15307, opened 2026-09-12
sorts: make odd_even_sort support comparable items

deerred643-star:
Part of #15234

## Summary

- make `odd_even_sort` generic over comparable item types (`[T: Comparable]`)
- rewrite the two `>` comparisons as `<` so the algorithm only relies on `__lt__`, which is what the `Comparable` protocol guarantees
- add non-integer doctest coverage (`str`, `float`) and a `TypeError` doctest for mixed non-comparable items
- add `odd_even_sort` to `test_sort_rejects_non_comparable_items`
- fix the docstring `:param:` name, which did not match the actual argument

The algorithm itself is unchanged: same odd/even phase loop, same swap logic, same complexity.

## Tests

- `python -m doctest -v sorts/odd_even_sort.py`
- `python -m pytest tests/test_sorts.py`
- `ruff check sorts/odd_even_sort.py tests/test_sorts.py`

### DIFF
--- a/sorts/odd_even_sort.py
+++ b/sorts/odd_even_sort.py
@@ -4,8 +4,14 @@ Odd even sort implementation.
 https://en.wikipedia.org/wiki/Odd%E2%80%93even_sort
 """
 
+from typing import Protocol
 
-def odd_even_sort(input_list: list) -> list:
+
+class Comparable(Protocol):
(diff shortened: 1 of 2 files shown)


cclauss:
Please only fix one sort so we can have good first issues for others on October 1st.
* https://hacktoberfest.com

PR #15340, opened 2026-09-15
sorts: make recursive_insertion_sort generic over Comparable items

Rokesh2008:
Part of #15234

This is an in-place sort, so it follows the `MutableSequence[T]` bucket of the `Comparable`/`TypeVar` convention discussed on #15234, rather than the `Iterable[T] -> list[T]` bucket used by copy-and-return sorts:

- `rec_insertion_sort`/`insert_next` now take `MutableSequence[T]` bound to `Comparable`
- `rec_insertion_sort` now returns the sorted collection (previously `-> None`) and `n` defaults to `len(collection)`, so it can be called with a single argument — needed so it fits the shared `SORTS` battery in `tests/test_sorts.py`, which calls every sort as `sort(list(case))`
- added a string doctest alongside the existing int examples
- registered `rec_insertion_sort` in the shared `SORTS` battery (checked against `sorted()`, including the `Person`/`Dog` comparable objects) and in `test_sort_rejects_non_comparable_items`

Tests:
- `python -m doctest -v sorts/recursive_insertion_sort.py` — 18/18 passed
- Verified `rec_insertion_sort` against every case in `tests/test_sorts.py`'s `CASES` tuple with a single-argument call, matching `sorted()`
- Confirmed `rec_insertion_sort([1, "a"])` raises `TypeError`

Linking with `Part of #15234` rather than a closing keyword, per the issue's contribution guidance.


cclauss:
* https://hacktoberfest.com

cclauss:
* https://hacktoberfest.com

Rokesh2008:
Pushed a follow-up fix per review feedback:

- `rec_insertion_sort(collection, n)` is back to a required `n` and a pure `None` return — no more optional `n` and no returning the collection, so the in-place API contract is unchanged.
- Kept the `Comparable`/`TypeVar` generic typing (using a plain `TypeVar("T", bound=Comparable)` rather than PEP 695 syntax, to match the file's existing style).
- Kept the string doctest alongside the int ones.
- Since `rec_insertion_sort` no longer returns a value, it no longer fits the shared `SORTS` battery in `tests/test_sorts.py` (which asserts on the return value), so I removed it from that tuple and added a dedicated `test_rec_insertion_sort` (parametrized over the same `CASES`, calling it in place and checking against `sorted()`) plus its own non-comparable-items rejection test.

Ran `python -m doctest -v sorts/recursive_insertion_sort.py` (21/21 passed) and the full `tests/test_sorts.py` suite (244 passed) locally against this branch.

cclauss:
Based on the failing tests, I would suggest running:
% ` ruff rule UP047 ` # To see why it is complaining.
% `ruff check --select=UP047 --fix --unsafe-fixes ` # To fix it.

sorts/recursive_insertion_sort.py
```diff
@@ -4,14 +4,24 @@
...
-def rec_insertion_sort(collection: list, n: int) -> None:
+
+class Comparable(Protocol):
+    def __lt__(self, other: Any, /) -> bool: ...
+
+
+T = TypeVar("T", bound=Comparable)
+
+
+def rec_insertion_sort(collection: MutableSequence[T], n: int) -> None:
```
cclauss:
```suggestion
def rec_insertion_sort[T](collection: MutableSequence[T], n: int) -> None:
```

sorts/recursive_insertion_sort.py
```diff
@@ -36,7 +51,7 @@ def rec_insertion_sort(collection: list, n: int) -> None:
     rec_insertion_sort(collection, n - 1)
 
 
-def insert_next(collection: list, index: int) -> None:
+def insert_next(collection: MutableSequence[T], index: int) -> None:
```
cclauss:
```suggestion
def insert_next[T](collection: MutableSequence[T], index: int) -> None:
```

cclauss (approved):

PR #15354, opened 2026-09-15
fix(sorts): support comparable items in pancake sort

berlin880:
## Description

Update `pancake_sort` to support comparable items such as strings, rather than being limited to numeric values.

## Changes

* Keep the sorting function generic with `TypeVar`.
* Accept `Sequence[T]` as input.
* Add a doctest demonstrating sorting strings.
* Preserve the existing behavior for numeric values.
* Return a `list[T]`.

## Testing

* Ran the `pancake_sort` doctests.
* Verified sorting with numeric values.
* Verified sorting with string values.

## Related Issue

This PR addresses the Pancake Sort comparable-items task.
 #ISSUE-NUMBER".15234


berlin880 (commented):
Fixed

PR #15356, opened 2026-09-16
ci: add matchify pre-commit hook

### LINKED ISSUE #15337: Should we add matchify to pre-commit
### Feature description

@priya-sundaram-dev, please evaluate if we should add https://github.com/15r10nk/matchify to our codebase.

Python's match statement is powerful and we should encourage contributors to use it where it makes sense.

On Python 14t, please run `uv tool install matchify && matchify data_structures`, which should modify six files.

Do these changes look correct and useful?  Do you recommend we add this tool to pre-commit?

kadubhumika:
## Summary

- Add Matchify as a pre-commit hook.
- Exclude `data_structures/linked_list/doubly_linked_list.py` because its
  conversion produces a less readable `case self.head` pattern.

## Testing

- `pre-commit validate-config` passed.
- `pre-commit run --all-files` was run.
- Matchify executed successfully and converted applicable files.
- Confirmed `doubly_linked_list.py` remains excluded from Matchify.
- Restored unrelated changes produced by the full pre-commit run.

Fixes #15337
### Describe your change

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist


kadubhumika:
Hi @cclauss  @priya-sundaram-dev ,

The merge conflicts have been resolved and all required checks are now passing.
The PR adds Matchify as a pre-commit hook, with `doubly_linked_list.py` excluded because the resulting pattern is less readable.
Would appreciate a review when you have time. Thank you!

priya-sundaram-dev (commented):
Thanks for resolving the conflicts, @kadubhumika! One thing to flag before this can go in, though — right now the PR has **no net change**. Comparing the branch against `master`:

```
$ git diff master...add-matchify-precommit --stat
(0 files changed)
```

GitHub agrees: the PR shows `0 changed files`. It looks like the last merge commit (`6b907e5 Merge branch 'master'`) resolved the conflict in `.pre-commit-config.yaml` by taking `master`'s side, which discarded the matchify hook that commit `6d6eee5` had added. `matchify` no longer appears anywhere in the branch's `.pre-commit-config.yaml`, so as it stands merging this wouldn't add the hook.

To fix: re-apply the matchify block on top of the current merged state, e.g.

```yaml
  - repo: https://github.com/15r10nk/matchify
    rev: <pinned-version>
    hooks:
      - id: matchify
        exclude: ^data_structures/linked_list/doubly_linked_list\.py$
```

(keeping the `doubly_linked_list.py` exclusion you mentioned, since the rewritten pattern is less readable there). Once `git diff master...` shows the hook being added again and CI is green, happy to re-review. 🙂

.pre-commit-config.yaml
```diff
@@ -63,3 +63,8 @@ repos:
     rev: "0.26"
     hooks:
       - id: validate-pyproject
+  - repo: https://github.com/15r10nk/matchify
```
cclauss, line 67: The rest of the file leaves a blank line between hooks for readability.
```suggestion

  - repo: https://github.com/15r10nk/matchify
```

cclauss:
Please run:
% `ruff format`

priya-sundaram-dev:
Confirmed — the hook is now genuinely applied: `.pre-commit-config.yaml` adds the `matchify` repo (rev `v0.2.0`) with the `doubly_linked_list.py` exclusion, and `git diff master...` now shows the 42-file reformat rather than the empty diff from before. Nice, that clears the blocker I flagged.

One thing to watch on @cclauss's `ruff format` note: matchify and ruff can disagree on line-wrapping after a rewrite, so run them in the order the CI does (matchify → ruff format → ruff check) and commit the result, otherwise pre-commit may keep re-modifying files. Once CI is green this looks good to me.

kadubhumika:
@cclauss @priya-sundaram-dev Thanks for the feedback! I’ve run `ruff format` locally and confirmed that all files are already formatted. I’ll also add the requested blank line before the Matchify hook for consistency with the rest of the configuration.

matrix/matrix_class.py
```diff
@@ -316,26 +316,27 @@ def __sub__(self, other: Matrix) -> Matrix:
...
+                    )
+                return Matrix(
+                    [
+                        [Matrix.dot_product(row, column) for column in other.columns()]
+                        for row in self.rows
+                    ]
+                )
+            case _:
+                raise TypeError(
+                    "A Matrix can only be multiplied by an int, float, or another matrix"
```
cclauss: Wrap a long line.
```suggestion
                    "A Matrix can only be multiplied by an int, float, or another "
                    "matrix"
```

cclauss:
@priya-sundaram-dev In https://github.com/TheAlgorithms/Python/issues/15337#issuecomment-5675291450 you flagged the matchify change in one file.

Do you flag any matchify changes in this pull request?  Are the files more or less verbose with these changes?

priya-sundaram-dev:
Good question — I went through the whole diff for this PR, not just spot-checks.

**Do I flag any conversion here? No.** The thing I flagged in #15337 was one specific construct: a *dotted value-pattern* (`if current == self.head` → `case self.head:`), which is correct but reads like a capture and trips people up. That construct doesn't appear anywhere in this PR — and the file it was in, `doubly_linked_list.py`, is the one explicitly excluded by the hook. I scanned all ~137 added `case` arms and they're the unambiguous kinds:

- literals — `case 1:`, `case "encrypt":`, `case "sjf (preemptive)":`
- singletons — `case None:`, `case False:`
- class patterns, some guarded — `case Vector():`, `case Matrix():`, `case int() | float():`, `case set() if isinstance(set_b, set):`

None of those can be misread as a capture, so I'd be happy to see them land.

**More or less verbose?** Marginally *more* by line count, not by complexity. There are ~48 `if/elif` chains converted to `match`, for a net of roughly +70 lines across the 42 files — about +1.5 lines per conversion. That's essentially the one `match subject:` header line plus one extra indent level each; the branch bodies are unchanged. So the code isn't doing more, it's the same dispatch nested one level deeper. Where the subject is a single value dispatched to several literals (the cipher `main()` menus, the scheduling algorithm selector), I think the `match` reads clearer and earns that extra line. Nowhere in this PR did the rewrite cost readability the way the `self.head` case did.

cclauss:
Awesome teamwork.  I approve this, but I am going to hold off on merging it until just before Hacktoberfest 2026 starts.

If not, it will cause many of our 375 pull requests to fail in CI.  Once we have far fewer PRs, we will rerun and test this, and then merge it.  Really awesome work!! Thanks much.

mindaugl (approved):

.pre-commit-config.yaml
oga35767-eng: 👍

oga35767-eng (approved):

oga35767-eng:
👍

priya-sundaram-dev:
@cclauss Yes, a few. The sweep is a mixed bag rather than a uniform win:

- **Reads better as `match`:** genuine multi-way dispatch on one value — e.g. `conversions/decimal_to_any.py` and `scheduling/cpuschedulingalgorithms.py`. There the `if/elif` chain was really a switch, so `match/case` is clearer and about the same length.
- **Comes out more verbose / less readable:** simple two-branch `if/elif` with no `else` (e.g. the `ciphers/*.py` `main()` blocks) just gains a nesting level for no clarity benefit — `match op: case 1: … case 2: …` isn't easier to read than the original two branches.
- **Actively worse:** `maths/jaccard_similarity.py`, where an `isinstance()` `if/elif` becomes `case set() if isinstance(set_b, set):` / `case list() | tuple() if isinstance(set_b, (list, tuple)):`. The type check is now split between the pattern and a guard, which is harder to follow than the straightforward `isinstance(...)` conditions it replaced.

So I'd not merge this as a blanket auto-fix. I'd keep the hook but exclude the two-branch and `isinstance`-guard cases (same as `doubly_linked_list.py`), and only take the files where the original really was a multi-branch switch. Case-by-case here, not one commit.

willbyers1 (approved):
LGTM! Code changes look clean and well structured.

PR #15357, opened 2026-09-16
pr_file_map.py: Use a human-readable date format

cclauss:
Use a human-readable date format.

### Describe your change

* [x] Fix a bug or typo in an existing script?


cclauss:

priya-sundaram-dev (approved):
LGTM — nice readability pass on the script, thanks for picking it up. Walked through each change:

- **Human-readable timestamp** (`{datetime.now(UTC):%d %b %Y at %H:%M} {UTC}` → "16 Sep 2026 at 09:00 UTC") reads far better in the rendered report than the raw ISO string. 👍
- **`directory_only` as a comma-joined `- #1, #2, ...` list item** renders much cleaner in the Markdown than the space-separated run.
- **`scripts/pr_file_map.py`** in the usage examples matches what `script_display_path()` actually emits now — good consistency fix.
- **`if root := git_root():`** is functionally equivalent here since `git_root()` only ever returns `None` or a real (non-empty, truthy) `Path`, so dropping the explicit `is not None` is safe and cleaner.

All checks green. Approving.

poyea (approved):

PR #15373, opened 2026-09-18
pr_file_map.py: fetch PR files concurrently with asyncio

priya-sundaram-dev:
### Checklist
- [x] This PR only changes one file (the maintenance script `scripts/pr_file_map.py`; not an algorithm).
- [x] All function parameters and return values are annotated with Python type hints.

---

### What & why

Follow-up to #15368. As you found, the slow part of `pr_file_map.py` is the **first pass** (~188 s): one `gh pr view <n> --json files` per open PR, run sequentially. That's ~one network round-trip per PR, so it's I/O-bound — a perfect fit for concurrency.

This PR fetches those PR file lists **concurrently** using `asyncio.create_subprocess_exec`, bounded by an `asyncio.Semaphore`.

### Design

- `run_gh_async(args, semaphore)` — async twin of `run_gh`, throttled by a shared semaphore.
- `get_pr_files_async(pr_number, semaphore)` — single-PR fetch.
- `gather_pr_files(pr_numbers, concurrency)` — `asyncio.gather` over all PRs, capped at `concurrency`, returns `{pr_number: [paths]}` keyed **in PR order** so the report stays deterministic.
- `main()` runs the first pass via `asyncio.run(gather_pr_files(...))`; the classify pass is unchanged.
- Concurrency defaults to **10** and is overridable with `PR_FILE_MAP_CONCURRENCY` (validated: must be a positive int). The cap keeps us polite to the GitHub API and avoids secondary rate limits — bumping it further gives diminishing returns since `gh` itself has per-call startup cost.

### Behavior is unchanged

- Output is **byte-identical** to the sequential version (verified with a mocked `gh`) — same sections, same ordering, same stderr counts.
- Errors still **fail fast** with the same `Error running 'gh …'` message; a missing `gh` still exits cleanly.
- `ignore_pull_request` filtering, `DIRECTORY.md` handling, and the contested-file detection are untouched.

### Testing

With a mocked `gh` (3 PRs, 0.3 s simulated latency each):

| run | wall clock |
|---|---|
| `PR_FILE_MAP_CONCURRENCY=1` (sequential) | 1.27 s |
| default (concurrent) | 0.58 s |

`diff` of the two outputs → **identical**. On the real repo (~900 PRs) this collapses the first pass from serial round-trips to `ceil(900 / concurrency)` batches. Also verified: `PR_FILE_MAP_CONCURRENCY=0` → clean validation error; a failing `gh pr view` → propagates its stderr and exits non-zero.

`ruff check` and `ruff format --check` both clean.

Re your asyncio-vs-free-threading question from #15081: this is I/O-bound work dispatched to subprocesses, so `asyncio` is the right tool — it overlaps the waiting without needing threads or the GIL to move. Free-threading wouldn't add anything here since the Python side barely does any CPU work; the time is all in waiting on `gh`.

Happy to tune the default concurrency or add a `--concurrency` CLI flag if you'd prefer that over the env var.


priya-sundaram-dev:
Reopened as #15374 with the checklist marked (the keeper auto-closed this one for an unmarked checklist). Same branch, same change.

PR #15378, opened 2026-09-18
sorts: make odd_even_sort generic over any comparable type (Part of #15234)

aayushgupta7725:
Adds a Comparable-bound TypeVar (matching the pattern used in insertion_sort.py), doctests covering strings, floats, and the non-comparable TypeError case, and registers odd_even_sort in the shared test_sort_rejects_non_comparable_items test.

Part of #15234

@priya-sundaram-dev, this is ready whenever you get a chance to review.

### Describe your change

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist

* [ ] If this pull request resolves one or more open issues, then the description above includes the issue number(s) with a closing keyword: "Fixes #ISSUE-NUMBER".


priya-sundaram-dev (commented):
Thanks @aayushgupta7725 — this reads cleanly and the added doctests (str, float, and the mixed-type `TypeError`) are a nice touch. One small simplification:

The module-level `Comparable`/`TypeVar` boilerplate is now redundant with the PEP 695 syntax you used on the signature. `def odd_even_sort[T: Comparable](...)` introduces its *own* scoped `T`, so the top-level `T = TypeVar("T", bound="Comparable")` is dead — nothing references it. You can drop that line and the `TypeVar` import (keep `Protocol`, and `Any` is still used inside the protocol):

```python
from collections.abc import MutableSequence
from typing import Any, Protocol


class Comparable(Protocol):
    def __gt__(self, other: Any, /) -> bool: ...


def odd_even_sort[T: Comparable](collection: MutableSequence[T]) -> MutableSequence[T]:
```

Tiny consistency nit while you're there: the algorithm branches on `collection[i] > collection[i + 1]`, so the protocol bound is more accurate as `__gt__` than `__lt__` (matches the operator the function actually relies on). Not a CI blocker — `ty` is happy either way — just makes the bound say what the code does.

Otherwise LGTM. 👍

priya-sundaram-dev (approved):
LGTM — this cleanly mirrors the  pattern ( protocol + PEP 695  on a ), and registering it in  / the non-comparable rejection test is exactly right. I ran the four new doctests locally and they all pass, including the  →  case.

One optional nit, no need to block on it:  uses  (unquoted, since the class is already defined above), whereas here it is . Both work; dropping the quotes would keep it byte-for-byte consistent with the reference file. Nice work.

priya-sundaram-dev:
(Reposting my review note — GitHub ate the code spans in the previous one.)

Approving: this mirrors the `insertion_sort.py` pattern exactly — `Comparable` protocol + PEP 695 `[T: Comparable]` over a `MutableSequence[T]`, plus registration in the shared `test_sort_matches_builtin` / non-comparable rejection tests. I ran the four new doctests locally and all pass, including the `[1, "a"]` → `TypeError` case.

One optional nit (don't block on it): `insertion_sort.py` uses `bound=Comparable` unquoted, whereas this uses `bound="Comparable"`. Both work; dropping the quotes keeps it byte-for-byte consistent with the reference file. Nice work.

aayushgupta7725:
Thanks for the review! I've pushed both requested changes:

- Removed the now-redundant module-level `T = TypeVar("T", bound=Comparable)` and the unused `TypeVar` import, since `[T: Comparable]` on the function signature already scopes its own `T`.
- Changed the `Comparable` protocol to require `__gt__` instead of `__lt__`, to match the `>` operator the algorithm actually uses.

Also fixed a `ruff` import-formatting check that was failing separately (missing blank line after the imports) — that should be green now too.

Let me know if anything else needs adjusting!

priya-sundaram-dev (approved):
Both changes look great, thanks for the quick turnaround.

- Dropping the module-level `TypeVar` in favor of the inline `[T: Comparable]` scoping is the right call — matches the `insertion_sort.py` pattern exactly.
- Switching the `Comparable` protocol to `__gt__` now correctly mirrors the `>` the algorithm actually uses.

I re-ran the 7 doctests on the latest commit (5d21113) locally — all pass, including the `[1, "a"]` → TypeError case. LGTM. 👍

priya-sundaram-dev:
Both changes look spot-on — thanks for the quick turnaround:

- `[T: Comparable]` scoping the type param locally is exactly right; dropping the redundant module-level `TypeVar` is cleaner.
- Good catch matching `Comparable.__gt__` to the `>` the loop actually uses — that's the correct protocol for this algorithm.

Nice touch adding the mixed-type `TypeError` doctest and wiring `odd_even_sort` into the parametrized `test_sort_matches_builtin` suite. Full check matrix is green. LGTM from me — nothing further needed on my end.

oga35767-eng (approved):

oga35767-eng:
👍

cclauss (approved):

PR #15387, opened 2026-09-21
ty: un-ignore invalid-type-form

Khushi-Tyagi9:
Part of #15187.

`Image` here is the PIL.Image module, not the class, so ty flagged it under
invalid-type-form. Changed the annotations to `Image.Image` in
mean_threshold, otsu_threshold, change_brightness and change_contrast, and
removed the rule from the ignore list in pyproject.toml.

Ran ty with the rule enabled and it's clean now. Only annotations changed,
nothing at runtime.

### Describe your change

* [ ] Add an algorithm?
* [ ] Documentation change?

### Checklist


oga35767-eng:
👍

oga35767-eng (approved):

oga35767-eng:
👍

cclauss:

priya-sundaram-dev (approved):
LGTM — the four annotations that changed are genuinely the correct fix: `Image` is the *module* (`from PIL import Image`), and the actual image type is the `Image.Image` class, so `-> Image.Image` is right and `-> Image` was the `invalid-type-form`. With those fixed, dropping `rules.invalid-type-form = "ignore"` is safe and the `ty` CI check is green on this head, so the rule is now genuinely enforced rather than suppressed. 

One small consistency note (non-blocking, fine as a follow-up): the *parameter* annotations still read `image: Image` / `img: Image`, i.e. the same module-vs-class distinction on the input side. `ty` isn't flagging them today so CI stays green, but tightening those to `Image.Image` too would make each signature read consistently. Happy either way — the PR as-is achieves its stated goal of un-ignoring the rule with green CI.

computer_vision/mean_threshold.py
```diff
@@ -6,7 +6,7 @@
 """
 
 
-def mean_threshold(image: Image) -> Image:
+def mean_threshold(image: Image) -> Image.Image:
```
cclauss:
```suggestion
def mean_threshold(image: Image.Image) -> Image.Image:
```

computer_vision/otsu_threshold.py
```diff
@@ -7,7 +7,7 @@
 """
 
 
-def otsu_threshold(image: Image) -> Image:
+def otsu_threshold(image: Image) -> Image.Image:
```
cclauss:
```suggestion
def otsu_threshold(image: Image.Image) -> Image.Image:
```

cclauss (approved):

digital_image_processing/change_brightness.py
```diff
@@ -1,7 +1,7 @@
 from PIL import Image
 
 
-def change_brightness(img: Image, level: float) -> Image:
+def change_brightness(img: Image, level: float) -> Image.Image:
```
cclauss:
```suggestion
def change_brightness(img: Image.Image, level: float) -> Image.Image:
```

digital_image_processing/change_contrast.py
```diff
@@ -11,7 +11,7 @@
 from PIL import Image
 
 
-def change_contrast(img: Image, level: int) -> Image:
+def change_contrast(img: Image, level: int) -> Image.Image:
```
cclauss:
```suggestion
def change_contrast(img: Image.Image, level: int) -> Image.Image:
```

priya-sundaram-dev:
Came to this just after the merge — for the record it looks right to me. Since these files do `from PIL import Image`, the bare `Image` annotation was actually the *module*, so `ty` was correct to flag `invalid-type-form`; `Image.Image` is the intended `PIL.Image.Image` class. Pure annotation change, no runtime effect, and dropping the rule from the ignore list is the right cleanup. Nice one, @Khushi-Tyagi9. 👍

